PR #27, opened 2014-06-13
Typo in the @deprecated documentation

extesy:
It looks like CreateResponse>YourKey< but it should be CreateResponse<YourKey>


maximelamure:
Thanks for the bug @extesy . It has been updated and pushed.

extesy:
4 years, nice...

PR #38, opened 2015-03-23
Fixes spelling on EntityResponse#hasError

sbilstein:


CrendKing:
Hi Siegfried. Thank you very much for the PR. Unfortunately since this method is public, it is part of interface, so we can’t remove it directly. Instead we have to deprecate it first and remove in the next major release. We will take care of this problem.

PR #44, opened 2015-06-08
Update build.gradle to use siging.signPom instead of just signPom

jpbetz:
Seems to be required to publish. This may have changed in a recent version of Gradle.


jpbetz:
This is obsolete.

PR #45, opened 2015-06-26
Upgrade to apache commons IO 2.4

jpbetz:
The current 1.4 version is very old.

When using 2.4 in source code and depending on rest.li which depends on 1.4, build system conflict resolution gets unhappy, mainly because of how build systems handle version conflict resolution for major versions.


CrendKing:
Hi Joe. Goksel will take care of this soon. Thanks for the report!

CrendKing:
We updated commons IO to 2.4.

PR #48, opened 2015-07-25
Add validation mode for disallowing unrecognized fields.

jpbetz:
This gives developers the option run the validator in a mode that disallows unrecognized fields.  For most use cases, ignoring unrecognized fields is the best approach for backward compatibility reasons.  But there are a couple use cases at system boundaries where a validation mode that disallows unrecognized fields might be generally useful:

- When accepting API requests from the public web of via 3rd party API endpoints.
- Before persisting data at the storage layer.

We're hoping to get this into rest.li because we need to perform this type of validation and would like to leverage the existing validator to do so.

### DIFF
--- a/data/src/main/java/com/linkedin/data/schema/validation/UnrecognizedFieldMode.java
+++ b/data/src/main/java/com/linkedin/data/schema/validation/UnrecognizedFieldMode.java
@@ -1,0 +1,17 @@
+package com.linkedin.data.schema.validation;
+
+/**
+ * Enum for how unrecognized fields should be handled during validation.
+ */
+public enum UnrecognizedFieldMode {
+
+  /**
+   * Ignore unrecognized fields during validation.
+   */
+  IGNORE,
+
+  /**
+   * If an unrecognized is present, then validation fails.
+   */
+  DISALLOW
+}
--- a/data/src/main/java/com/linkedin/data/schema/validation/ValidateDataAgainstSchema.java
+++ b/data/src/main/java/com/linkedin/data/schema/validation/ValidateDataAgainstSchema.java
@@ -163,6 +163,10 @@ public final class ValidateDataAgainstSchema
         {
           validate(nextElement, nextElementSchema, nextElement.getValue());
         }
+        else if (_options.getUnrecognizedFieldMode() == UnrecognizedFieldMode.DISALLOW)
+        {
+          addMessage(element, "unrecognized field found but not allowed");
+        }
       }
     }
 
@@ -682,7 +686,7 @@ public final class ValidateDataAgainstSchema
         return object;
       }
     }
-    
+
     protected void addMessage(DataElement element, String format, Object... args)
     {
       _messages.add(new Message(element.path(), format, args));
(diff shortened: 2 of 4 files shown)


jpbetz:
An alternative way to provide this functionality would be to, somehow, allow the validator to visit all data elements, including data elements that are unrecognized (have not schema).  Then a developer could write define a custom `Validator` that they pass into `ValidateDataAgainstSchema` that could detect when `DataElement.getSchema()` returns null and produce it's own validation errors.

This approach seemed slightly simpler to me, but I'm certainly open to feedback.

el378:
Hi Joe,

This looks like a concise approach. The caveat, as you alluded to, is that removing optional fields from the schema will now result in an effectively backwards incompatible change. Given the use case is front facing APIs, clients probably cannot be migrated at will. Can you elaborate on how you think people will use this, e.g. as part of a rest.li filter, within resource method before writing to cold storage, etc?

An alternative to consider is to provide a third mode/option where extraneous fields are removed from DataTemplate objects. If you want to take this path and use the validator approach, you can probably change the way RecordTemplates are handled and return a fixup that removes the fields not present in the schema. This is a more involved change than the current diff, but I suspect it will cause fewer problems in the long run in terms of backwards compatibility.

For reference, when we approached a similar problem previously, we decided to use a custom rest.li filter to trim the fields not in the schema. For our use case, we required a way to trim extraneous data while also supporting an override mechanism (e.g. preserve specific items not in the schema by describing a positive mask tree), and therefore did not end up using the validator for this purpose.

-Erluo

jpbetz:
Hi Erluo,

I'm well aware of the trade-offs when handling unrecognized fields via pass-through, filtering and disallowing them during validation.  **If** we are able to get Pegasus adoption going.  Most of our codebase we will be using pass-through, and for a number of major system boundaries we will be doing filtering.

Inconveniently (and as you mentioned), Pegasus does not currently support a Validation mode that filters unrecognized fields as part of the traversal.  So we currently do a separate pass, which basically looks like:

```
Builder.create(data, schema, IterationOrder.PRE_ORDER).filterBy(AbsentSchemaPredicate).remove();
```

We will most likely need to change how this is done. If Linkedin is onboard with it being done via an optional Fixup, that would be great.

For a few remaining use cases, developers would like to disallow unrecognized fields via validation.  It is possible that some of those use cases will transition to filtering in the future, but some may stay that way indefinitely. In some these cases data models are very stable and developers would currently prefer to fail fast, at least initially.  While I agree in principal that, generally, filtering is preferred.  Their preference to disallow the data in these particular cases is reasonable.

jpbetz:
I should mention, we are not using rest.li, only Pegasus.  The main place we are considering disallowing unrecognized fields is at the storage layer, if we ever disallowed unrecognized fields at the API layer, it would most likely be a 3rd API endpoint, but we haven't discussed that in detail, it is also possible that we would just filter there.

el378:
Hi Joe,

Thanks for clarifying your use case -- I was under the assumption that this was used for the API layer (which was our use case), which is why I was pointing out the backwards compatibility issue (given that unrecognized field and removed fields are somewhat indistinguishable from the perspective of the validator).

I certainly agree with you that pass/filter/fail options are tradeoffs. My point is that there seems to be enough use cases (such as the ones you mentioned) to consider offering filtering in parallel to this pull request. We are onboard with the fixup approach -- I'll follow up in the other thread.

-Erluo

jpbetz:
Erluo,

Thanks for getting back to me. Looking forward to your feedback on the Filter issue.

To close the loop on this pull-request, are you planning to accept this? or are there any remaining concerns?

-Joe

jpbetz:
I'll combine this with the filtering change and send a single pull request with both.  Closing.

PR #50, opened 2015-07-30
Add validation mode for trimming and disallowing unrecognized fields

jpbetz:
This gives developers the option run the validator in a mode that disallows or filters out unrecognized fields when needed.  For most use cases ignoring unrecognized fields for backward compatibility reasons.  But there are a couple system boundaries where these approaches have been found useful.

Note that I've included a specialized version of `com.linkedin.data.it.Remover.ToRemove` called `ToTrim`.  Alternatively, could modified and then exposed some of the currently private functionality from `Remover`, but that would have resulted in a public API where developers could directly accumulate `Remover.ToRemove` objects, which is risky because if the order of those objects is not correct, the wrong elements can be removed from arrays. Also, `ToTrim` needs to track the paths of the elements it visits for validation message purposes.

Questions for reviewers:
* Does the naming look reasonable for the new Enum values? "TRIM" could alternatively have been named "FILTER".
* Did I handle fixup correctly here?

### DIFF
--- a/data/src/main/java/com/linkedin/data/schema/validation/ValidateDataAgainstSchema.java
+++ b/data/src/main/java/com/linkedin/data/schema/validation/ValidateDataAgainstSchema.java
@@ -153,24 +153,23 @@ public final class ValidateDataAgainstSchema
      * This is a specialized version of {@link com.linkedin.data.it.Remover.ToRemove} which is and
      * should remain private, so could not be used here.
      */
-    private class FieldToTrim
+    private static class FieldToTrim
     {
       private FieldToTrim(DataMap parent, String fieldName)
       {
+        assert(!parent.isReadOnly());
         _parent = parent;
         _fieldName = fieldName;
       }
 
       private void trim()
       {
-        if (_parent != null) {
-          if (_parent.isReadOnly())
-          {
-            throw new ConcurrentModificationException("Map marked as read-only during validation.");
-          }
-          else {
-            _parent.remove(_fieldName);
-          }
+        if (_parent.isReadOnly())
+        {
+          throw new ConcurrentModificationException("Map marked as read-only during validation.");
+        }
+        else {
+          _parent.remove(_fieldName);
         }
       }
 
--- a/data/src/main/java/com/linkedin/data/schema/validation/ValidateDataAgainstSchema.java
+++ b/data/src/main/java/com/linkedin/data/schema/validation/ValidateDataAgainstSchema.java
@@ -190,7 +190,7 @@ public final class ValidateDataAgainstSchema
 
           if (map.isReadOnly()) {
             _hasFixupReadOnlyError = true;
-            addMessage(element.path(), "unrecognized field cannot be trimmed because DataMap backing it is read-only");
+            addMessage(element, "unrecognized field cannot be trimmed because DataMap backing it is read-only");
           } else {
             _toTrim.add(new FieldToTrim(map, fieldName));
           }
@@ -759,12 +759,6 @@ public final class ValidateDataAgainstSchema
     protected void addMessage(DataElement element, String format, Object... args)
     {
       _messages.add(new Message(element.path(), format, args));
-      _valid = false;
-    }
-
-    protected void addMessage(Object[] path, String format, Object... args)
-    {
-      _messages.add(new Message(path, format, args));
       _valid = false;
     }
 
--- a/data/src/main/java/com/linkedin/data/schema/validation/ValidateDataAgainstSchema.java
+++ b/data/src/main/java/com/linkedin/data/schema/validation/ValidateDataAgainstSchema.java
@@ -21,6 +21,7 @@ import java.math.BigDecimal;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
@@ -124,6 +125,7 @@ public final class ValidateDataAgainstSchema
     private Object _fixed = null;
     private boolean _valid = true;
     private final Context _context;
+    private List<ToTrim> _toTrim = new ArrayList<ToTrim>(0);
 
     private State(ValidationOptions options, Validator validator)
     {
@@ -142,6 +144,53 @@ public final class ValidateDataAgainstSchema
       {
         validateIterative(element);
       }
+
+      if (_toTrim.size() > 0) {
+        for (ToTrim toTrim: _toTrim) {
+          toTrim.trim();
+        }
+      }
+    }
+
+    /**
+     * Holds a data elements for later removal.
+     *
+     * This is a specialized version of {@link com.linkedin.data.it.Remover.ToRemove} which is and
+     * should remain private, so could not be used here.
+     */
+    private class ToTrim
+    {
+      private ToTrim(DataElement element)
+      {
+        DataElement parentElement = element.getParent();
+        _path = element.path();
+        _parent = parentElement == null ? null : parentElement.getValue();
+        _name = element.getName();
+      }
+
+      private void trim()
+      {
+        if (_parent != null) {
+          Class<?> nameClass = _name.getClass();
+          Class<?> parentClass = _parent.getClass();
+          if (nameClass == String.class) {
+            assert (parentClass == DataMap.class);
+            DataMap map = (DataMap) _parent;
+            if (map.isReadOnly())
+            {
+              _hasFixupReadOnlyError = true;
+              addMessage(_path, "unrecognized field cannot be trimmed because DataMap backing it is read-only");
+            }
+            else {
+              map.remove(_name);
+            }
+          }
+        }
+      }
+
+      private Object[] _path;
+      private Object _parent;
+      private Object _name;
     }
 
     protected void validateRecursive(DataElement element)
@@ -154,6 +203,7 @@ public final class ValidateDataAgainstSchema
     {
       _recursive = false;
       _fixed = element.getValue();
+      UnrecognizedFieldMode unrecognizedFieldMode = _options.getUnrecognizedFieldMode();
       ObjectIterator it = new ObjectIterator(element, IterationOrder.POST_ORDER);
       DataElement nextElement;
       while ((nextElement = it.next()) != null)
@@ -162,6 +212,18 @@ public final class ValidateDataAgainstSchema
         if (nextElementSchema != null)
         {
           validate(nextElement, nextElementSchema, nextElement.getValue());
+        }
+        else if (unrecognizedFieldMode != UnrecognizedFieldMode.IGNORE)
+        {
+          switch (unrecognizedFieldMode)
+          {
+            case DISALLOW:
+              addMessage(nextElement, "unrecognized field found but not allowed");
+              break;
+            case TRIM:
+              _toTrim.add(new ToTrim(nextElement));
+              break;
+          }
         }
       }
     }
@@ -682,10 +744,16 @@ public final class ValidateDataAgainstSchema
         return object;
       }
     }
-    
+
     protected void addMessage(DataElement element, String format, Object... args)
     {
       _messages.add(new Message(element.path(), format, args));
+      _valid = false;
+    }
+
+    protected void addMessage(Object[] path, String format, Object... args)
+    {
+      _messages.add(new Message(path, format, args));
       _valid = false;
     }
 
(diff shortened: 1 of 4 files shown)


data/src/main/java/com/linkedin/data/schema/validation/ValidateDataAgainstSchema.java
```diff
@@ -142,6 +144,53 @@ protected void validate(DataElement element)
...
+     *
+     * This is a specialized version of {@link com.linkedin.data.it.Remover.ToRemove} which is and
+     * should remain private, so could not be used here.
+     */
+    private class ToTrim
+    {
+      private ToTrim(DataElement element)
+      {
+        DataElement parentElement = element.getParent();
+        _path = element.path();
```
jpbetz: Just like done in `ToRemove`, we cannot hold a reference to DataElement, so we instead capture the fields we need from it.

jpbetz:
I'm fine with someone downloading the diff and running it through the usual review/commit workflow.  If it's possible to retain attribution, that would be kinda nice, but not a huge deal.

kvidhani:
Hi Joe,

Thanks for the fix, it looks like what we discussed on and agreed earlier.

We'll work on getting it through the normal review process and getting it
committed soon.

I'll notify you once it's in.

Thanks!

data/src/main/java/com/linkedin/data/schema/validation/ValidateDataAgainstSchema.java
```diff
@@ -142,6 +144,53 @@ protected void validate(DataElement element)
...
+              addMessage(_path, "unrecognized field cannot be trimmed because DataMap backing it is read-only");
+            }
+            else {
+              map.remove(_name);
+            }
+          }
+        }
+      }
+
+      private Object[] _path;
```
jpbetz: Thinking about this diff more, I should remove this field, it's expensive to compute.

jpbetz:
Karim,

I thought about this a bit more and realized I could be quite a bit more efficient.

I've updated the pull request to not trim fields of a DataMap that itself is being trimmed (because there is no need). I've also made sure to only calculate the `path()` (a somewhat expensive operation) when needed.

jpbetz:
I should also mention:

I've only considered record fields in this diff.  My understanding is that a unrecognized union member or enum symbol fail validation so do not need to be addressed by this change, but please double check me on that.

kvidhani:
Hi Joe,

Thanks for the improvement.

Also, I just checked and your comment about union/enum is indeed true.

We are working in pushing this into pegasus as we speak.

Stay tuned.

- Karim

kvidhani:
Hi Joe

It seems on your diff that your fix won't work with recursive validation.

This is obviously only the case if avro union mode is enabled.

Is this intentional?

- Karim

kvidhani:
Also, this line:

Object parent = parentElement == null ? null : parentElement.getValue();

Wouldn't parent always be non-null at this point?

Also, what happens if the parent is a DataList?

- Karim

jpbetz:
Karim,

You can include comments one lines of code pull-request.  It's easier to maintain threaded conversations about code changes that way.

In response to above:

1.) I didn't consider avro union mode in this change. It looks like it should be trivial to add the check. Do you need me to submit an updated pull request?

2.) You may be correct that parent can never be null at:
```
Object parent = parentElement == null ? null : parentElement.getValue();
```
Would you like the defensive check removed?

3.) The intend here is that we only consider parents of an unrecognized field which should always be record. It certainly may be possible that there is an error in how we identifier unrecognized fields that I'm missing, if so, would you provide an example of how this might be a `DataList`?

jpbetz:
Okay, I wrote the code to address (1) from above and everything was going well until I tried to add a test case for the fix, and then I ran into a major issue: #53, which basically means `AvroUnionMode` completely broken for any schemas that contain unions.

So I gotta ask: What is `AvroUnionMode` validation used for?

If you're not using `AvroUnionMode` for anything, I recommend removing it.  If you are, I'm curious how it is that you've not run into this issue.

I'm happy to provide my fix for (1), but it can't provide a decent test case for it given the state of things.

Let me know.

jpbetz:
Okay, I've gone ahead and fixed `AvroUnionMode` to do what I believe is intended. I'm still not convinced   `AvroUnionMode` is really needed, since the erroneous code comes from 959bb2f6d3a18bd7e299cb9a36bac4fbc52847b6, which is "the beginning of time" as far as this source repo is concerned.

I'll leave it up to the rest.li team to decide what to do here, but, given the overall lack of test coverage for `AvroUnionMode`, I wouldn't trust it with anything important. I recommend either removing or adding substantially more tests.

jpbetz:
Okay, I've followed up on Swee's guidance from #53. Let's try this again!

data/src/main/java/com/linkedin/data/schema/validation/ValidateDataAgainstSchema.java
```diff
@@ -278,13 +292,82 @@ protected void recurseRecord(DataElement element, RecordDataSchema schema, DataM
...
+        assert(!parent.isReadOnly());
+        _parent = parent;
+        _fieldName = fieldName;
+      }
+
+      private void trim()
+      {
+        if (_parent.isReadOnly())
+        {
+          throw new ConcurrentModificationException("Map marked as read-only during validation.");
```
kvidhani: This should be unsupported operation exception. This is what Java uses for such scenarios (i.e unmodifiable list/map). What do you think?

data/src/main/java/com/linkedin/data/schema/validation/ValidateDataAgainstSchema.java
```diff
@@ -124,6 +120,7 @@ public static ValidationResult validate(DataElement element, ValidationOptions o
     private Object _fixed = null;
     private boolean _valid = true;
     private final Context _context;
+    private List<FieldToTrim> _toTrim = new ArrayList<FieldToTrim>(0);
```
kvidhani: Why is this set at zero during construction?

data/src/main/java/com/linkedin/data/schema/validation/ValidateDataAgainstSchema.java
jpbetz: That seems reasonable.

data/src/main/java/com/linkedin/data/schema/validation/ValidateDataAgainstSchema.java
jpbetz: Premature optimization :)

It seemed plausible to me that the most common number of fields to trim would be 0.

kvidhani: Sounds reasonable.

data/src/main/java/com/linkedin/data/schema/validation/ValidateDataAgainstSchema.java
```diff
@@ -278,13 +292,82 @@ protected void recurseRecord(DataElement element, RecordDataSchema schema, DataM
...
         RecordDataSchema.Field field = schema.getField(key);
+        Object value = entry.getValue();
         if (field != null)
         {
-          Object value = entry.getValue();
           DataSchema childSchema = field.getType();
           childElement.setValueNameSchema(value, key, childSchema);
           validate(childElement, childSchema, value);
         }
+        else
```
kvidhani: Are you sure that this is the only area (recurseRecord()) where you have to add this?

data/src/main/java/com/linkedin/data/schema/validation/ValidateDataAgainstSchema.java
```diff
@@ -278,13 +292,82 @@ protected void recurseRecord(DataElement element, RecordDataSchema schema, DataM
...
+        {
+          childElement.setValueNameSchema(value, key, null);
```
kvidhani: I assume parent type is always recognized at this point?

data/src/main/java/com/linkedin/data/schema/validation/ValidateDataAgainstSchema.java
```diff
@@ -278,13 +292,82 @@ protected void recurseRecord(DataElement element, RecordDataSchema schema, DataM
...
+          handleUnrecognizedField(childElement);
```
kvidhani: Also what about checking the unrecognized field mode at this point in time like we do for iteratively?

data/src/test/java/com/linkedin/data/schema/validation/TestValidation.java
```diff
@@ -2081,4 +2098,161 @@ public void testNonRootStartDataElement() throws IOException
...
+    Object goodObjects[] =
+        {
+            "a valid string"
+        };
+
+    Object badObjects[] =
+        {
+        };
+
+    // There is no coercion for this type.
```
kvidhani: Are these comments copy paste leftovers?

data/src/test/java/com/linkedin/data/schema/validation/TestValidation.java
```diff
@@ -2081,4 +2098,161 @@ public void testNonRootStartDataElement() throws IOException
...
+    // Test with all coercion modes, result should be the same for all cases.
+    testCoercionValidation(schemaText, "bar", goodObjects, badObjects, disallowUnrecognizedFieldOption());
+    testCoercionValidation(schemaText, "bar", goodObjects, badObjects, trimUnrecognizedFieldOption());
```
kvidhani: This test, as far as I can tell, will assert that the result of the validation is false (meaning it failed). Why would such validation fail? If we trim, doesn't that lead to successful validation?

data/src/test/java/com/linkedin/data/schema/validation/TestValidation.java
```diff
@@ -2081,4 +2098,161 @@ public void testNonRootStartDataElement() throws IOException
...
+    }
+  }
+
+  @Test
+  public void testUnrecognizedFieldsWithAvroUnionDefault() throws IOException
+  {
+    ValidationOptions options = new ValidationOptions(
+        RequiredMode.CAN_BE_ABSENT_IF_HAS_DEFAULT,
+        CoercionMode.NORMAL,
+        UnrecognizedFieldMode.TRIM);
```
kvidhani: Can you write a test with disallow as well? It should be similar except the result would not be valid

data/src/main/java/com/linkedin/data/schema/validation/ValidateDataAgainstSchema.java
jpbetz: Yes

data/src/main/java/com/linkedin/data/schema/validation/ValidateDataAgainstSchema.java
jpbetz: Yes, since we only recurse for fields with recognized types.

data/src/main/java/com/linkedin/data/schema/validation/ValidateDataAgainstSchema.java
jpbetz: Already done. See the implementation of `handleUnrecognizedField`.  I moved that logic into that function.

data/src/test/java/com/linkedin/data/schema/validation/TestValidation.java
jpbetz: Good catch, yes.  Removing.

data/src/test/java/com/linkedin/data/schema/validation/TestValidation.java
jpbetz: Added.

jpbetz:
I'd addressed all feedback I'm aware of. Any additional concerns with this pull-request?

kvidhani:
No this is all for now. We're in the process of committing it soon.

Thanks again!

data/src/main/java/com/linkedin/data/schema/validation/ValidationOptions.java
```diff
@@ -249,6 +300,7 @@ public String toString()
 
   private CoercionMode _coercionMode;
   private RequiredMode _requiredMode;
+  private UnrecognizedFieldMode _unrecognizedFieldMode;
```
soojungha: equals, hashCode, and toString methods need to be changed.

jpbetz:
Good catch @soojungha.  I've added them.

jpbetz:
Any update on this?

el378:
I just committed your diff along with some other refactoring.

Thanks for your input!

PR #60, opened 2015-12-02
Improve support for alternate schema parsers and generator target languages.

jpbetz:
This primarily cleans up DataSchemaParser to better support alternate generator target languages as well as allow for alternate schema parsers.

This is primarily for Courier (coursera.github.io/courier) but is general enough to be useful to anyone extending pegasus.

Courier supports multiple target languages (Swift, Scala, Java for Android) and also has it's own custom file format for authoring pegasus schemas, and would benefit significantly from having these extension points.

Separate from this pull-request I'd like to discuss merging the custom schema language used by Courier into Pegasus, but regardless what we decide on that topic, this seems generally useful.

Testing:

```
./gradlew test
...
SUCCESS
```


data/src/main/java/com/linkedin/data/schema/AbstractSchemaParser.java
```diff
@@ -0,0 +1,153 @@
...
+import java.util.Set;
+import java.util.TreeMap;
+
+/**
+ * Common base class for parsers that parse Schema objects.
+ *
+ * @author slim
+ * @author jbetz
+ */
+abstract public class AbstractSchemaParser extends AbstractDataParser
```
jpbetz: This code is has simply been refactored up from `SchemaParser` and is otherwise unchanged (except the constructor and `addTopLevelSchema()` method).

generator/src/main/java/com/linkedin/pegasus/generator/DataSchemaParser.java
```diff
@@ -52,14 +53,15 @@
 {
   private final String _resolverPath;
   private final DataSchemaResolver _schemaResolver;
+  private final SchemaParserFactory _schemaParserFactory;
+  private final String _fileExtension;
```
jpbetz: These are the two main extension points added to `DataSchemaParser`.

mchen07:
IMHO, custom java binding currently in our PDSC ("java" meta property) should not be a part of data model spec, since it is language dependent, while PDSC is designed to describe data model that should be language agnostic. So this change is just a workaround for this limitation to support other language binding. An ideal solution to me should be: Template spec layer should not be directly tied to the custom java class specified in current PDSC, instead it should just store a CustomInfoSpec to not break existing Pegasus java binding. For other language bindings, we should get custom binding information from some configuration outside PDSC, and generate the correct data model based on Template spec with the configuration information provided in code-gen phase. Just two cents for your reference.

jpbetz:
I'm supportive of Pegasus being extended to allow language specific
properties being provided from an external source (I.e. Not pdsc files).
Let's keep in mind though, that unless we want to break backward
compatibility, we would need to continue to support language specific
properties written in pdsc files.

With that in mind, are there any concerns with accepting this change so
that we can have reasonable integration with template spec?

jpbetz:
Any interest in this pull request?

jpbetz:
This is high priority for us.

data/src/main/java/com/linkedin/data/schema/DataSchemaParser.java
```diff
@@ -0,0 +1,114 @@
+package com.linkedin.data.schema;
+
+import java.io.InputStream;
+import java.io.Reader;
+import java.util.List;
+
+/**
+ * A schema parser is used to parse data schemas for a particular source representation.
+ */
+public interface DataSchemaParser {
```
mchen07: I would prefer to choose a different name for this interface, since in generator module, we have another class named as DataSchemaParser as well.

generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
```diff
@@ -418,47 +425,46 @@ else if (schema instanceof PrimitiveDataSchema)
...
-        {
-          throw new IllegalArgumentException(schema + " has \"java\" property with \"coercerClass\" that is not a string");
+        final DataMap map = (DataMap) java;
+        final Object custom = map.get(CLASS_PROPERTY);
+        if (custom != null) {
+          if (custom.getClass() != String.class) {
+            throw new IllegalArgumentException(schema + " has \"" + customTypeLanguage + "\" property with \"class\" that is not a string");
+          }
+          // a custom class specification has been found
+          customClasses = new CustomClasses();
```
mchen07: CustomClasses information may not be sufficient to support different language, for example, Swift does not have namespace, only classname is not enough to do codegen later. Since we want to change Pegasus meta data layer to support alternate language binding, we need to take this into consideration.

data/src/main/java/com/linkedin/data/schema/DataSchemaParser.java
mchen07: Would you mind sharing the use case where you need to provide another implementation of DataSchemaParser that is different from current SchemaParser implementation?

mchen07: Overall, I don't have objections to pull this interface generalization into Pegasus to facilitate different implementation of data schema parser, except some naming suggestions.

generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
mchen07: This is my major concern for this pull request. Basically current TemplateSpecGeneration implementation introduced such an issue, that is, introduced this language-coupling interpretation of CustomInfoSpec. Instead, at template spec layer, we should only present all the meta information of CustomInfo specified in the PDSC instead of directly interpreting that as Java Class, and as for how to interpret such CustomInfoSpec in code generation, it should leave to individual language code generator to decide. We are in the process of fixing this issue, and should have a RB to commit in the next month. If we make this pull request into master branch, when we later on committed the correct approach of handling custom information in TemplateSpecGenerator, we will have to remove your newly added constructor and bring in a backwards-incompatible change that may break your current usage. If you are ok with this (your code may be broken later), we can pull this in for now.

data/src/main/java/com/linkedin/data/schema/DataSchemaParser.java
jpbetz: Okay, cool. I believe `SchemaParser` is also taken.  How about `PegasusSchemaParser` ?

generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
jpbetz: This all makes sense. I'm okay with `getCustomClasses(DataSchema schema, String customTypeLanguage)` being a temporary work around that get's removed once we have a method that returns all meta information.

data/src/main/java/com/linkedin/data/schema/DataSchemaParser.java
mchen07: Sounds good.

data/src/main/java/com/linkedin/data/schema/AbstractSchemaParser.java
```diff
@@ -0,0 +1,252 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Common base class for schema parsers.
+ *
+ * @author slim
+ * @author jbetz
+ */
+abstract public class AbstractSchemaParser extends AbstractDataParser implements PegasusSchemaParser
```
mchen07: With this new AbstractSchemaParser abstract class, why do we still need AbstractDataParser?

jpbetz: Good catch.  We really don't.  I had left `AbstractDataParser` alone, but I can certainly consolidate these classes together.  I'll update the pull request.

mchen07: Thanks for update. I will download the diff and run some sanity check before committing to master, will let you know if I run into any issues.

mchen07:
Your patch cannot be applied to the master:

Applying: Add support for alternate schema parsers.
Applying: Move common schema parser code from SchemaParser into new AbstractSchemaParser abstract class.
Applying: Remove unused imports
Applying: Introduce DataSchemaParser and DataSchemaParserFactory interfaces.
error: patch failed: restli-docgen/src/main/java/com/linkedin/restli/docgen/RestLiResourceRelationship.java:36
error: restli-docgen/src/main/java/com/linkedin/restli/docgen/RestLiResourceRelationship.java: patch does not apply
Patch failed at 0004 Introduce DataSchemaParser and DataSchemaParserFactory interfaces.
The copy of the patch that failed is found in:
   /Users/mnchen/dev/pegasus_trunk/pegasus/.git/rebase-apply/patch
When you have resolved this problem, run "git am --continue".
If you prefer to skip this patch, run "git am --skip" instead.
To restore the original branch and stop patching, run "git am --abort".

jpbetz:
I've rebased.

mchen07:
This has been pulled into master, you should be able to see this reflected in github soon.

jpbetz:
Thanks @mchen07 !

PR #61, opened 2015-12-02
Add typed map key support to DataSchemas and TemplateSpecs.

jpbetz:
This adds the ability to indicate the type of a map key in pegasus schemas. E.g.:

```
{ "type": "map", "keys": "int", "values": "org.example.Fortune" }
```

The wire protocol stays exactly as-is. Keys are still encoded as strings in JSON objects.

The benefit of this change is that data binding generators may provide improved type-safety.  For example, a data binding generator might bind to `Map<Int, Fortune>` for the above schema example, and could also validate that the map keys are all valid integers.

This is only the groundwork for typed map keys.

If this is accepted there are two potential follow up changes:
- Improve the existing pegasus data schema generator for Java to generate typesafe bindings for typed map keys.
- Introduce a string representation for complex types OR deliberately limit keys to primitive types and typerefs to primitive types.  In Courier at Courser we use [InlineStringCodec](https://github.com/coursera/courier/blob/master/runtime/src/main/scala/org/coursera/courier/codecs/InlineStringCodec.scala), so we could potentially add a Java implementation of such a codec to Pegasus to be the standard way complex types are encoded as map keys.  Or we could just add the primitive key type support, which would enable developers to clearly indicate things like that a key must be an integer or must be a URN (which is a typeref to string).

We have been using typed map keys in Courier in production for quite some time and are really happy with them.

I realize that there should be unit tests,  if this approach looks reasonable, let me know and I'll add the unit tests to this pull-request.

### DIFF
--- a/data/src/main/java/com/linkedin/data/schema/DataSchemaConstants.java
+++ b/data/src/main/java/com/linkedin/data/schema/DataSchemaConstants.java
@@ -40,6 +40,7 @@ public class DataSchemaConstants
   public static final String SYMBOLS_KEY = "symbols";
   public static final String SYMBOL_DOCS_KEY = "symbolDocs";
   public static final String TYPE_KEY = "type";
+  public static final String KEYS_KEY = "keys";
   public static final String VALUES_KEY = "values";
 
   public static final String NULL_TYPE = "null";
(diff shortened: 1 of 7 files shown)


jpbetz:
Just a reminder, this PR is not yet ready to be landed.  I'd like to get a response on if Pegasus would like to have typed map key support merge in. If so, I can do the needed polish work to get this ready for full code review and merge.

jpbetz:
Any interest in this pull request?

jpbetz:
This is lower priority for us, we can continue to use what we have if needed.  Mostly proposing the change to see if there is interest in introducing direct map key support into Pegasus or not.

jpbetz:
@kvidhani This is one of the underlying improvements that is still outstanding to add courier support for pegasus.

xma000:
I am for this proposal. It can be a valuable addition to the data model binding support. A couple of things to think of are
 - We probably should reuse Rest.li protocol 2.0 complex type serialization. We already use this key serialization in responses for a few batch methods.
 - iOS and Android codegen should also support key type.
But the first two steps can be done without much question:
 1. Formally support key type for map.
 2. Improve Pegasus data template codegen to support typed key.

jpbetz:
Great, I'll rebase and flesh out the unit test coverage.

PR #68, opened 2016-05-03
Handle 'comment' key in client configuration as a special cases, fixes #67.

### LINKED ISSUE #67: d2-quickstart example is broken
when running the examples/d2-quickstart client code, the client only runs once then fails.

looks like the "comment" key needs to be handled better:

``` json
    "trafficProportion" : {
        "comment" : "this represent the number of request per second for a particular service",
        "newsArticle": 3,
        "jobRecommendation": 2,
        "articleRecommendation" : 1
    },
```

"this represent the number of request per second for a particular service" is not a number...

beejeebus:
fixes issue #67.


FreCap:
Thank you for signaling, it will be solved in the next release

PR #85, opened 2017-11-07
This fixes a NullPointerException issue where the underlying ErrorRes…

mannylavery:
This fixes a NullPointerException issue where the underlying ErrorResponse is null but no null check is possible.

### DIFF
--- a/restli-client/src/main/java/com/linkedin/restli/client/RestLiResponseException.java
+++ b/restli-client/src/main/java/com/linkedin/restli/client/RestLiResponseException.java
@@ -28,6 +28,8 @@ import com.linkedin.r2.message.rest.RestResponseBuilder;
 import com.linkedin.restli.common.ErrorResponse;
 import com.linkedin.restli.internal.common.HeaderUtil;
 
+import java.util.Optional;
+
 /**
  * @author Josh Walker
  * @version $Revision: $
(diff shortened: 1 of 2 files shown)


restli-client/src/main/java/com/linkedin/restli/client/RestLiResponseException.java
```diff
@@ -64,74 +66,84 @@ public RestLiResponseException(RestResponse rawResponse,
                                  Throwable cause)
   {
     super(rawResponse, cause);
-    _status = rawResponse.getStatus();
-    _errorResponse = errorResponse;
-    _decodedResponse = decodedResponse;
+    _status = Optional.of(rawResponse.getStatus());
+    _errorResponse = Optional.ofNullable(errorResponse);
```
mchen07: From pegasus codebase, we have guaranteed that ErrorResponse passed in cannot be null, which can be seen from ExceptionUtil.getErrorResponse() method. I am curious in what case you encountered a NULL passed in for _errorResponse?

mlavery:
@mchen07 not sure if that has been fixed I encountered this problem in August of last year. A lower level service was throwing a ServiceException which was if I recall getting converted to a ResponseException in the framework and passed back.

The issue I had when testing API was that when trying to access the ServiceErrorCode and ServiceErrorMessage the underlying _errorResponse was null (some of the time).

Its been almost a year so the details are fuzzy.

ChrisCarini:

PR #97, opened 2018-10-03
[docs] Edit documentation for BATCH_PARTIAL_UPDATE return entity

evanw555:
Forgot to add an entry for this feature in the request builders functionality table in the client user guide.


maximelamure:
LGTM!

PR #99, opened 2018-10-26
Fix the format issue of all tables and code blocks

bolobobo:
The format of the tables and code blocks are disordered. Standardized the their formats.


maximelamure:
Thanks for your contribution @bolobobo !

PR #118, opened 2019-02-17
Add search support

maximelamure:


evanw555:
Looks like the search functionality includes script files in its search. Use "count" or "jsonify" as a search query and you'll see the top entry called `null` which links to `lunr-feed.js`. Maybe it's being included in the {% for post in site.pages %}` loop?

maximelamure:
Good catch @evanw555 , I updated the PR to index MD files only. Let me know if you have additional comments.

PR #125, opened 2019-03-06
Add documentation for BATCH_FINDER

bolobobo:
* Add documentation for BATCH_FINDER
* Fixed dead links, typo and errors


maximelamure:
Thanks, @bolobobo for this PR. Can you add a page in the "API reference" section where it will provide more details on the BatchFinder: The protocol, pagination support, the behavior, how the response is generated (order maintained), error management, IDL/JSON, client code. Each time you can provide an example. It would also be great to add more context on what is not supported (primitive type for a criteria)

batch_finder_resource_method.md
```diff
@@ -0,0 +1,418 @@
...
+Eg.
+
+    curl "d2://userSearchResults?bq=findUsers&batchCriteria=List((firstName:pauline),(lastName:iglou))&age=21" -X GET
+
+#### Pagination support
+##### 1) Common pagination for all search criteria  
+The developer can pass additional parameters to specify a common pagination. It will be more efficient than adding a pagination context inside each criteria object.  
+Eg.
+
+    curl "d2://userSearchResults?q=findUsers&batchCriteria=List((firstName:pauline, age:12),(lastName:iglou))&firstName=max&start=10&count=10" -X GET
```
maximelamure: Can we also have the examples with the protocol V2?

batch_finder_resource_method.md
```diff
@@ -0,0 +1,418 @@
...
+public class BatchFinderResult<QK,V extends RecordTemplate,MD extends RecordTemplate>
+{
+   private final Map<QK,CollectionResult<V,R>> _elements;
+   private final Map<QK,RestLiServiceException> _errors;
+   ...
+}
+```
+
+## Future consideration
+### Support primitive type(not ready)
```
maximelamure: I would say not supported. For now, we don't want to support this feature that will prevent any evolutions.

batch_finder_resource_method.md
```diff
@@ -0,0 +1,418 @@
...
+        -   [Reponse Body](#batchcollectionresponse)
+        -   [Error Handling](#error-handling)
+-   [Resource API](#resource-api)
+    -   [Method Annotation and Parameters](#method-annotation-and-parameters)
+    -   [Criteria Filter](#criteria-filter)
+    -   [BatchFinderResult](#batchfinderresult)
+-   [Future consideration](#future-consideration) 
+    
+## Overview   
+The BATCH_FINDER allows combining multiple requests to one FINDER within a Rest.li resource into a single call.
```
maximelamure: I know I am the author of this sentence :) but by reading it again, I feel like the reader might understand that batch-finder will allow batching finder methods, like a multiplexer. In fact, we support a new resource method that accepts a list of filters set.  Instead of callings multiple finders with different filter values, we can call 1 batch-finder method with a list of filters.

batch_finder_resource_method.md
```diff
@@ -0,0 +1,418 @@
...
+
+`BatchFinderCriteriaResult` fields:
+
+- (optional) "elements" : JSON serialized list of entity types (in success case)
+- (optional) "metadata":
+- (optional) "paging" : JSON serialized CollectionMetadata object
+- (optional) "error" : it's an ErrorResponse which fail to get a list of entities to corresponding search criteria(in failure)
+- "isError" : which indicates whether the result is a successful case or not
+
+E.g.
```
maximelamure: Can we have the curl request at the origin of this response?

batch_finder_resource_method.md
```diff
@@ -0,0 +1,418 @@
...
...n\tat org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:141)\n\tat org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:134)\n\tat org.eclipse.jetty.server.Server.handle(Server.java:534)\n\tat org.eclipse.jetty.server.HttpChannel.handle(HttpChannel.java:333)\n\tat org.eclipse.jetty.server.HttpConnection.onFillable(HttpConnection.java:251)\n\tat org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:283)\n\tat org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:108)\n\tat org.eclipse.jetty.io.SelectChannelEndPoint$2.run(SelectChannelEndPoint.java:93)\n\tat org.eclipse.jetty.util.thread.strategy.ExecuteProduceConsume.executeProduceConsume(ExecuteProduceConsume.java:303)\n\tat org.eclipse.jetty.util.thread.strategy.ExecuteProduceConsume.produceConsume(ExecuteProduceConsume.java:148)\n\tat org.eclipse.jetty.util.thread.strategy.ExecuteProduceConsume.run(ExecuteProduceConsume.java:136)\n\tat org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:671)\n\tat org.eclipse.jetty.util.thread.QueuedThreadPool$2.run(QueuedThreadPool.java:589)\n\tat java.lang.Thread.run(Thread.java:745)\n",
```
maximelamure: Don't provide the full stack trace.

batch_finder_resource_method.md
```diff
@@ -0,0 +1,418 @@
...
+At least, 2 query parameters will have to be set for a batch finder:
+
+- The "bq" query parameter is reserved for passing the batch finder method name
+- A second query parameter will be used to pass a set of different search criteria. The name of this query parameter is set in the [BatchFinder method annotation](/rest.li/batch_finder_resource_method#method-annotation-and-parameters).
+For example, with @BatchFinder(value="findUsers", batchParam="batchCriteria"), the batch query parameter name is "batchCriteria". 
+The type of this query parameter is a List.
+
+Eg.
+
+    curl "d2://userSearchResults?bq=findUsers&batchCriteria=List((firstName:pauline, age:12),(lastName:iglou))" -X GET
```
maximelamure: if it's "curl", "d2" won't work

batch_finder_resource_method.md
```diff
@@ -0,0 +1,418 @@
...
+Pagination default to start=0 and count=10. Clients may set both of these parameters to any desired value.
+### Method Annotation and Parameters
+The @`BatchFinder` annotation takes 2 required parameter:
+- `value` : which indicates the BATCH_FINDER method name
+- `batchParam` : which indicates the name of the batch criteria parameter, each BATCH_FINDER method must have and can only have one batch parameter
+
+For example: 
+```
+  @BatchFinder(value = "searchGreetings", batchParam = "criteria")
+  public BatchFinderResult<GreetingCriteria, Greeting, EmptyRecord> searchGreetings(@PagingContextParam PagingContext context,
```
maximelamure: Can we update all the examples to be async ?

batch_finder_resource_method.md
bolobobo: This is the example with Protocol V2 and I will add a V1.

maximelamure: The "curl" request won't work with d2. For protocol V2, you need to set the Header.
For example
`curl --header "X-RestLi-Protocol-Version: 2.0.0"  "http://localhost:10546/userSearchResults?q=findUsers&batchCriteria=List((firstName:pauline, age:12),(lastName:iglou))&firstName=max&start=10&count=10"`

batch_finder_resource_method.md
bolobobo: Got it.

batch_finder_resource_method.md
```diff
@@ -0,0 +1,418 @@
...
+For example, it should be safe to call whenever the client wishes.
+However, this is not something enforced by the framework, and it is up to the application developer that there are no side effects.
+
+It is important to note that:
+- the operations may execute on the server out of order
+- the response objects are expected to be returned in the same order and position as the respective input search criteria.
+- The BATCH_FINDER will require implementing a resource method to handle a BATCH_FINDER requests. It won't behave like a multiplexer that will call automatically existing finders
+
+## Protocol
+### Request
```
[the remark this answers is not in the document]
bolobobo: For this section, they are actually in different part in protocol document(Actually, that doc is not very readable). And I add more details here.

protocol.md
```diff
@@ -448,6 +457,66 @@ The URI templates below assume variables with types as follows:
...
+- A second query parameter will be used to pass a set of different search criteria. The name of this query parameter is set in the [BatchFinder method annotation](/rest.li/batch_finder_resource_method#method-annotation-and-parameters).
+For example, with @BatchFinder(value="findUsers", batchParam="batchCriteria"), the batch query parameter name is "batchCriteria". 
+The type of this query parameter is a List.
+
+Different data type has different representation in Rest.li protocol 1.0 and 2.0. See more details in [Rest.li Protocol](/rest.li/spec/protocol).
+
+Eg.
+In Rest.li protocol 1.0
+
+    curl "http://localhost:8080/userSearchResults?bq=findUsers&batchCriteria[0].firstName=pauline&batchCriteria[0].age=12&batchCriteria[1].lastName=iglou" --globoff
```
maximelamure: Does `--globoff` required?

protocol.md
```diff
@@ -448,6 +457,66 @@ The URI templates below assume variables with types as follows:
...
+The other query parameters will be applied as common filters across all batch requests.
+
+Here is an example batch request with two individual finders using the following criteria:
+
+- filter by first name and age
+- filter by last name and age
+
+Eg.
+
+    curl "http://localhost:8080/userSearchResults?bq=findUsers&batchCriteria=List((firstName:pauline),(lastName:iglou))&age=21" -X GET
```
maximelamure: Looks like we use protocol V2 here. In this case, can you add the HTTP header `--header "X-RestLi-Protocol-Version: 2.0.0"`? The goal is to provide a working example if they copy/paste.

protocol.md
```diff
@@ -956,6 +1026,65 @@ or complex data type, e.g.:
       }
     }
     
+### Batch Collection Response
```
maximelamure: Won't it be `BatchFinder collection response`?

batch_finder_resource_method.md
```diff
@@ -0,0 +1,310 @@
...
+## Resource API
+BATCH_FINDER is supported on Collection and Association Resources only(See more details about [Collection Resource](/rest.li/spec/protocol#collection-resources) and [Association Resource](/rest.li/spec/protocol#association-resources)).
+Resources may provide zero or more BATCH_FINDER resource methods. Each BATCH_FINDER method must be annotated with the @`BatchFinder` annotation.
+
+Pagination default to start=0 and count=10. Clients may set both of these parameters to any desired value.
+### Method Annotation and Parameters
+The @`BatchFinder` annotation takes 2 required parameter:
+- `value` : which indicates the BATCH_FINDER method name
+- `batchParam` : which indicates the name of the batch criteria parameter, each BATCH_FINDER method must have and can only have one batch parameter
+
```
maximelamure: It might be useful to have an example of how to call this resource method with curl to highlight the mapping between the query parameter `bq` and the `value` parameter in the `BatchFinder` annotation.

batch_finder_resource_method.md
```diff
@@ -0,0 +1,310 @@
...
+- Can only be Arrays of Record template type, if have to use some other data types like Pegasus Enum, etc as the array item,
+ need to wrap it into a Record Template (`.pdsc` schema)
+ 
+### Criteria Filter
+To implement a batch finder, the resource owner has to define a `RecordTemplate` to define a criteria filter parameter.
+The batch finder method will have to accept a array of this criteria filter.
+
+Example:
+
+The "GreetingCriteria" class represent a criteria filter to filter by "id" or by "tone".
```
maximelamure: Can we swap and have the PDSC first, then the code generated from this PDSC?

batch_finder_resource_method.md
```diff
@@ -0,0 +1,310 @@
...
+
+-   [Overview](#overview)
+-   [Protocol](#protocol)
+    -   [Request](#request)
+    -   [Response](#response)
+-   [Client](#client)
+    -   [Java Request Builders](#java-request-builders)    
+-   [Resource API](#resource-api)
+    -   [Method Annotation and Parameters](#method-annotation-and-parameters)
+    -   [Criteria Filter](#criteria-filter)
```
maximelamure: What do you think about this order?
 - CriteriaFilter
 - Method annotation
 - BatchFinder results
 - Error Handling
It will follow the steps the developer has to do to build a batch finder.

protocol.md
bolobobo: Yes, since we use [] in our curli call.

protocol.md
bolobobo: The class name is BatchCollectionResponse.

maximelamure: Do you think we should rename it?

batch_finder_resource_method.md
maximelamure: I suggested moving the PDSC example before the java class since the code is generated from the PDSC.

batch_finder_resource_method.md
```diff
@@ -0,0 +1,314 @@
...
+    public GreetingsCreateRequestBuilder create()
+    public GreetingsGetRequestBuilder get()
+    ...
+    public GreetingsFindBySearchRequestBuilder findBySearch()
+    // the BATCH_FINDER resource method named "SearchGreetings"
+    public GreetingsBatchFindBySearchGreetingsRequestBuilder batchFindBySearchGreetings() 
+}
+```
+The request builders factory class provides a method to generate the corresponding BATCH_FINDER request builder.
+More details about generated class and method declaration in [BATCH FINDER Request Builder](/rest.li/user_guide/restli_client#batch-finder-request-builder).
```
maximelamure: This document should provide all the details. The user guide should provide a simple use case. Having a link to the user-guide from this doc means the details should me moved here.

guide_restli_server.md
```diff
@@ -875,29 +867,140 @@ Keys, query parameters, or action parameters.
...
 @QueryParam(value="o", typeref=CustomObjectRef.class) CustomObject o
 
-@QueryParam(value="oArray", typeref=CustomObjectRef.class)
-CustomObject[] oArray
+@QueryParam(value="oArray", typeref=CustomObjectRef.class) CustomObject[] oArray
 ```
+<a id="BATCHFINDER"></a>
+
+#### BATCH FINDER
+The BATCH_FINDER allows combining multiple requests to one FINDER within a Rest.li resource into a single call.
```
maximelamure: This whole section seems a duplicate to the new batch finder document. The idea was to explain only the basic concept here and to provide all the details in the new document. In the end, we should provide a link to this new doc to guide the reader if she/he wants more details. Same comment in the "client" document.

guide_restli_server.md
```diff
@@ -875,30 +867,78 @@ Keys, query parameters, or action parameters.
...
+        // on error: to construct error response for test
+        batchFinderResult.putError(currentCriteria, new RestLiServiceException(HttpStatus.S_404_NOT_FOUND, "Fail to find Greeting!"));
+      }
+    }
+
+    return batchFinderResult;
+  }
 ```
 
+
```
maximelamure: We should add a link to redirect the reader to the resource method doc if they want more details.

guide_restli_client.md
```diff
@@ -286,6 +287,37 @@ a method to set each of them, of the form:
...
+
+The value **must** be non-null. For the batch query parameter, it also uses the form above
+like the other regular parameters.
+
+If the batchFinder specifies `AssocKey` parameters, the builder will contain
+a method to set each of them, of the form:
+```java
+    public <BuilderType> <assocKeyName>Key(<AssocKeyType> value);
+```
+
```
maximelamure: Same here, we should add a link to the resource method doc if the reader wants more details.

PR #128, opened 2019-03-09
Add ByteString to valid query parameter.

mchen07:


maximelamure:
LGTM! Thanks @mchen07

PR #135, opened 2019-05-15
Add documentation for namespace override in Avro translation

evanw555:
Adding documentation for how to use the namespace override option in Avro translation, since apparently there's no documentation at all about this. Also updated the part that describes the command-line arguments to use in order to add the system properties for this as well as for the optional default option.


maximelamure:
LGTM! Thanks, @evanw555

PR #140, opened 2019-07-19
WIP: Add documentation for service errors

evanw555:
Not done yet, would just like some preliminary review.

Added:

- Page for detailing service errors in detail. So far just has ErrorResponse info.
- Page for describing how to set up service errors step-by-step.
- Added information for success responses.
- Added examples of IDLs with service error and success response info.
- Added data file containing base URL for GitHub repo for easy usage when making links.

### DIFF
--- a/_data/api_reference_navigation.yml
+++ b/_data/api_reference_navigation.yml
@@ -17,6 +17,8 @@ toc:
         url: /rest.li/Rest_li-2_0-response-API
   - title: Rest.li Server
     subfolderitems:
+      - page: Service Errors
+        url: /rest.li/spec/service_errors
       - page: Return Entity
         url: /rest.li/spec/return_entity
   - title: Rest.li Resource Method
--- a/_data/urls.yml
+++ b/_data/urls.yml
@@ -1,0 +1,1 @@
+repo: https://github.com/linkedin/rest.li/blob/master
--- a/_data/user_guide_navigation.yml
+++ b/_data/user_guide_navigation.yml
@@ -61,6 +61,8 @@ toc:
         url: /rest.li/testsuite_how_to
       - page: Test Suite - Add New Language
         url: /rest.li/howto_add_new_language
+      - page: Configure Service Errors in Java
+        url: /rest.li/user_guide/service_errors_java
   - title: FAQ
     subfolderitems:    
       - page: Rest.li - FAQ
--- a/guide_restli_server.md
+++ b/guide_restli_server.md
@@ -2046,8 +2046,8 @@ By default, Rest.li returns an extensive HTTP error response that
 includes:
 
 -   HTTP Status Code (manditory)
--   X-LinkedIn-Error-Response header (this will be renamed to
-    X-RestLi-Error-Response shortly)
+-   `X-LinkedIn-Error-Response` header (this will be renamed to
+    `X-RestLi-Error-Response` shortly)
 -   A response body containing:
     -   A full stack trace
     -   A service error code (optional)
@@ -2067,10 +2067,13 @@ the HTTP response. However if any other Java exception is thrown,
 Rest.li automatically provides a default error message of "Error in
 application code" in the error response. This default error message may
 be customized via RestLiConfig as well, as shown in this example:
-java
-```
+
+```java
 restLiConfig.setInternalErrorMessage("Internal error, please try again later.");
 ```
+
+See [this page](/rest.li/spec/service_errors#error-responses) for more information on
+error responses in Rest.li.
 
 <a id="wiki-Projections"></a>
 
--- a/guide_service_errors_java.md
+++ b/guide_service_errors_java.md
@@ -1,0 +1,254 @@
+---
+layout: guide
+title: Configuring Service Errors in Java
+permalink: /user_guide/service_errors_java
+index: 2
+excerpt: This page describes how to configure service errors for a resource in Java.
+---
+
+# Configuring Service Errors in Java
+
+{{ page.excerpt }}
+
+See [Service Errors](/rest.li/spec/service_errors) for an in-depth reference of service errors in Rest.li.
+
+## Contents
+
+-   [Background](#background)
+-   [Step 1: Define your Service Errors](#step-1-define-your-service-errors)
+-   [Step 2: Write Your Service Error Definition](#step-2-write-your-service-error-definition)
+-   [Step 3: Apply the Service Errors to your Resource](#step-3-apply-the-service-errors-to-your-resource)
+-   [Step 4: Returning Service Errors](#step-4-returning-service-errors)
+-   [Step 5: Enabling Service Error Validation](#step-5-enabling-service-error-validation)
+
+## Background
+
+It's important to communicate to the consumers of a service what types of errors may be returned by a particular API.
+Fortunately, Rest.li allows application developers to configure service errors for their resources. This makes APIs
+clearer by documenting in the IDL what sorts of errors should be expected, allowing clients to effectively handle and
+process returned errors. Also, enabling service error validation allows services to guarantee which errors a client
+should expect to encounter.
+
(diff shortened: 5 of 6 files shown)


guide_service_errors.md
```diff
@@ -0,0 +1,234 @@
...
+to guarantee which errors a client should expect to encounter.
+
+## Step 1: Define your Service Errors
+
+First, you need to determine which service error codes your service may return, and what
+information will be contained in error responses corresponding with these codes. You'll
+need to define four things:
+
+- The application-specific string code associated with some service error
+- The HTTP status it'll return
```
maximelamure: HTTP status code?

guide_service_errors.md
```diff
@@ -0,0 +1,234 @@
...
+and process returned errors. Also, enabling service error validation allows services
+to guarantee which errors a client should expect to encounter.
+
+## Step 1: Define your Service Errors
+
+First, you need to determine which service error codes your service may return, and what
+information will be contained in error responses corresponding with these codes. You'll
+need to define four things:
+
+- The application-specific string code associated with some service error
```
maximelamure: HTTP status text?

guide_service_errors.md
```diff
@@ -0,0 +1,234 @@
...
+- The HTTP status it'll return
+- The string message it'll return
+- The schema defining the format of the error details
+
+For example, say we want to standardize the way errors are returned as a result of violating
```
maximelamure: To guide the developer to find the right HTTP status, can we suggest this list: https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml.

guide_service_errors.md
```diff
@@ -0,0 +1,234 @@
...
+the service's rate-limiting constraint. We would need to pick a string error code such as
+`QUOTA_EXCEEDED`, an HTTP status such as `429`, a string message describing the failure, and
+some schema describing the custom error details sent back to the client.
+
+## Step 2: Write Your Service Error Definition
+
+Write a Java enum implementing the `ServiceError` interface, which will contain a set of
```
maximelamure: Let's not be java specific in the explanation, but we can provide an example in java to illustrate it.

maximelamure: just a side note, how will it work in Python?

maximelamure: Does it have to be an enum, or any java class that will implement the interface?

guide_service_errors.md
```diff
@@ -0,0 +1,234 @@
...
+
+  private final HttpStatus _status;
+  private final String _message;
+  private final Class<? extends RecordTemplate> _errorDetailType;
+
+  /*
+   * Suggested: Define string constants for each of the error codes for easy access.
+   * These codes should match the enum value names.
+   */
+  public class Codes
```
maximelamure: Does Java or Rest.li provide a list of Status code/text the developer can re-use? If they want custom status, they can still define them as illustrated here.

guide_service_errors.md
```diff
@@ -0,0 +1,234 @@
...
+  {
+    return _errorDetailType;
+  }
+}
+```
+
+## Step 3: Apply the Service Errors to your Resource
+
+Now that you have defined your service errors, you can use them to specify which resources
+or particular resource methods may return such errors. You can do this by using a few
```
maximelamure: Can we say that in java, you can do this by using annotations as described below?

guide_service_errors.md
```diff
@@ -0,0 +1,234 @@
...
+to a resource) or as a method-level annotation (to refer to a resource method).
+This is used during build-time when generating IDLs and also during runtime when
+validating service errors returned by your resource.
+In the following example,
+all methods in the resource may return a `QUOTA_EXCEEDED` error, but only `getAll` can return
+an `INVALID_PERMISSIONS` error:
+
+```java
+@ServiceErrorDef(SampleServiceError.class)
+@ServiceErrors(SampleServiceError.Codes.QUOTA_EXCEEDED)
```
maximelamure: Won't it be easier for the developer to write something like
`@ServiceErrors(SampleServiceError.QUOTA_EXCEEDED)`
`QUOTA_EXCEEDED` would be a `ServiceError` type. At the build time, the framework can use it to generate the IDL and at the execution to validate. 
The benefit is to have to set only 1 annotation, not 2.

guide_service_errors.md
evanw555: I'll clarify as "HTTP status code"

guide_service_errors.md
evanw555: No, this refers to something like `"BAD_REQUEST"`, I can add this as an example.

guide_service_errors.md
evanw555: Good resource, I'll add the link. Thanks

guide_service_errors.md
evanw555: I thought the same thing, but this whole documentation is very Java specific, especially the guides [(example)](https://linkedin.github.io/rest.li/user_guide/restli_server#sub-resources).

This feature is not supported in Python. I can specify that.

It must be an enum, that is a hard requirement.

guide_service_errors.md
evanw555: There are no provided default error codes, but there are examples for reference such as this. Do you think we should add provided default codes?

guide_service_errors.md
evanw555: Not sure what you mean by this. Do you think it would be better if I renamed this page to "Configuring Service Errors in Java"?

guide_service_errors.md
evanw555: This is something Gevorg, Karthik, and I discussed, but it's not possible. Java annotations only allow certain return types in their methods: https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html#jls-9.6.1

evanw555: Java does allow enums in this case, but we can't use interfaces that enums implement, nor can we use a generic enum class to capture all possible enums.

guide_service_errors.md
maximelamure: `BAD_REQUEST` seems the HTTP status text. HTTP status has a code and a text. For example, `404` -> `NOT_FOUND` ....

guide_service_errors.md
maximelamure: I think we can explain the concept without being language-specific, then adding a section on how it works in java. By doing this, it opens the door to an example in a different language.

guide_service_errors.md
maximelamure: I remember we had this discussion in the past. In `go` for example, they provide this `https://golang.org/src/net/http/status.go` You can create your own, but the standard ones are already provided.

guide_service_errors.md
maximelamure: Oh, it's related to my previous comment related to `java`. My comment was to suggest to replace `You can do this by using a few...` by `In java, you can do this by using annotations as described below`

guide_service_errors.md
maximelamure: Thanks for the explanation. I remember that now.

guide_service_errors.md
evanw555: That was a bad example because it is also an HTTP status text. I mean something application-specified like `INVALID_ALBUM_ID` or `QUOTA_EXCEEDED`

guide_service_errors.md
evanw555: I renamed the guide to be Java-specific, and I changed the reference page to include mentions of "in Java..."

guide_service_errors.md
evanw555: I will ask Karthik and Gevorg about this.

maximelamure:
LGTM!

maximelamure:
LGTM! Thanks, @evanw555 for your contribution.

PR #143, opened 2019-08-13
Edit modeling & asynchronous documentation to recommend Async.

sarahjohanknecht:
**Modeling docs:** 
 -Add information about async templates available for use.

**Async docs:**
 -Refactor documentation to recommend Task/Async templates, since Promise template is deprecated.

**Testing:**
 -jekyll serve to check formatting & links.

### DIFF
--- a/Asynchronous-Servers-and-Clients-in-Rest_li.md
+++ b/Asynchronous-Servers-and-Clients-in-Rest_li.md
@@ -36,7 +36,7 @@ There are two main options available to write async server implementations in Re
 
 ### Async Server Templates
 
-Rest.li also includes templates to make writing async resources slightly easier. There are templates that use `Callback`s, `Promise`s, and `Task`s for each type of Rest.li resource. For example, for collection resources with primitive keys we have `com.linkedin.restli.server.resources.CollectionResourceAsyncTemplate`, `com.linkedin.restli.server.resources.CollectionResourceTaskTemplate`, and `com.linkedin.restli.server.resources.CollectionResourcePromiseTemplate`. There are similar templates for complex key resources, association resources, and simple resources.
+Rest.li also includes templates to make writing async resources slightly easier. There are templates that use `Callback`s and `Task`s for each type of Rest.li resource. For example, for collection resources with primitive keys we have `com.linkedin.restli.server.resources.CollectionResourceAsyncTemplate` and `com.linkedin.restli.server.resources.CollectionResourceTaskTemplate`. There are similar templates for complex key resources, association resources, and simple resources.
 
 ### Server Configuration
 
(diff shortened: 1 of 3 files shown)


Asynchronous-Servers-and-Clients-in-Rest_li.md
```diff
@@ -36,7 +36,7 @@ There are two main options available to write async server implementations in Re
 
 ### Async Server Templates
 
-Rest.li also includes templates to make writing async resources slightly easier. There are templates that use `Callback`s, `Promise`s, and `Task`s for each type of Rest.li resource. For example, for collection resources with primitive keys we have `com.linkedin.restli.server.resources.CollectionResourceAsyncTemplate`, `com.linkedin.restli.server.resources.CollectionResourceTaskTemplate`, and `com.linkedin.restli.server.resources.CollectionResourcePromiseTemplate`. There are similar templates for complex key resources, association resources, and simple resources.
+Rest.li also includes templates to make writing async resources slightly easier. There are templates that use `Callback`s and `Task`s for each type of Rest.li resource. For example, for collection resources with primitive keys we have `com.linkedin.restli.server.resources.CollectionResourceAsyncTemplate` and `com.linkedin.restli.server.resources.CollectionResourceTaskTemplate`. There are similar templates for complex key resources, association resources, and simple resources.
```
maximelamure: I might be wrong, but I think we recommended to use the `task` one.  Can you check? if yes, we can explicitly mention that.

sarahjohanknecht: The Promise template is deprecated, so I am decreasing the mentions of that in this document. From the documentation on the templates, it doesn't seem that we specifically recommend task over async. The main difference is whether the user wants to use Parseq and Tasks or a callback. I can change the documentation to specifically recommend Task if thats what we would like to do going forward.

modeling.md
```diff
@@ -171,18 +175,28 @@ keys are “foreign” keys from the referenced entities. Instead of using
...
+public class MyRelationResource extends AssociationResourceTaskTemplate<Relation> { … }  
 ```
 
 Defines a child resource with a URI of the form:
 
 ```  
 /myRelations/key1={longKey1}\&key2={longKey2}  
 ```
+#### Asynchronous resources
+Asynchronous Rest.li resources should be used when the downstream implementations are non-blocking. There are asynchronous resource templates for Collection, Simple, Association, and ComplexKey resources.
```
evanw555: What do you mean "when the downstream implementations are non-blocking"? Like if a service downstream from your resource uses an async template? May need a little more clarification

modeling.md
```diff
@@ -171,18 +175,28 @@ keys are “foreign” keys from the referenced entities. Instead of using
 Create, new association relationships are made by using Update, which
 allows the client to provide the resource identifier.
 
+Asynchronous association resources are declared by creating a class that extends AssociationResourceAsyncTemplate or AssociationResourceTaskTemplate.
```
evanw555: Class names and other inline code snippets should be wrapped with graves (``) to make it look nice, (e.g. DataSchema -> `DataSchema`). GitHub's highlighting logic might be a little complicated, so idk if you've already done this. I don't see the graves but it's highlighted.

modeling.md
sarahjohanknecht: I added a little more clarification. The page I linked at the bottom of these section goes more into detail about asynchronous, as well.

modeling.md
sarahjohanknecht: Fixed! This file wasn't using that formatting for other class names of the templates, so I changed all of them to be formatted with graves.

sarahjohanknecht:
I think changing ServiceErrors was outside of the scope.

PR #144, opened 2019-08-15
Adding parseq support to spring bridge

keremsahin1:


restli-spring-bridge/src/main/java/com/linkedin/restli/server/spring/RestliHttpRequestHandler.java
```diff
@@ -82,12 +86,29 @@ public RestliHttpRequestHandler(RestLiConfig config,
...
+  }
+
+  public RestliHttpRequestHandler(RestLiConfig config,
+                                  SpringInjectResourceFactory injectResourceFactory,
+                                  FilterChain filterChain,
+                                  int parseqThreadPoolSize)
+  {
+    Engine parseqEngine = null;
+    if (parseqThreadPoolSize > 0) {
+      final ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(parseqThreadPoolSize);
```
karthikbalasub: You should make the scheduler also a constructor parameter so it can be configured using spring. A service might want to customize the executor service used with the server.

keremsahin1: Thanks for the review, updated.

PR #149, opened 2019-09-16
Fix the layout issue in example respones on Rest.li Protocol page

nickibi:
This PR is to fix a layout issue in example responses on Rest.li Protocol page.


maximelamure:
LGTM! Thanks @nickibi  for your contribution

PR #153, opened 2019-11-07
Fix X-Restli-Method header documentation 

klee97:
Protocol incorrectly said X-RestLi-Method would be set to an uppercase string. It will actually be set to a lowercase string


maximelamure:
LGTM!

PR #154, opened 2019-11-16
Fix gradle vesrsion for wrapper

mdubinko:
I noticed that running through the quickstart produced an error. It was grabbing an ancient version 2.12 of  gradlew. This should fix it. Tested to make sure ./gradlew publishRestliIdl and ./gradlew build still work in this directory.


evanw555:
Thanks for the pointer, looks like this was already fixed by our team. We went ahead and made the same changes for the other examples, too.

PR #156, opened 2019-12-19
Remove references to ".pdsc" and use PDL in examples

evanw555:


.gitignore
```diff
@@ -22,4 +22,5 @@ ligradle/
 .DS_Store
 _site/
 .sass-cache/
+.jekyll-cache/
```
nickibi: what's the purpose to add this one?

evanw555: I'm assuming this shouldn't be included in the source code. It only popped up now because it's new Jekyll functionality, I guess.

PR #165, opened 2020-02-05
Add "Migrating from PDSC to PDL" page

evanw555:
This page shows how to migrate from PDSC to PDL, as well as a comparison
showing notable differences between the two formats.

### DIFF
--- a/pdl_migration.md
+++ b/pdl_migration.md
@@ -1,0 +1,255 @@
+---
+layout: guide
+title: Migrating from PDSC to PDL
+permalink: /pdl_migration
+excerpt: Guide for migrating from PDSC to PDL with a side-by-side comparison showing notable differences.
+---
+
+# Migrating from PDSC to PDL
+
+{{ page.excerpt }}
+
+## Contents
+- [Why Use PDL?](#why-use-pdl)
+- [How to Convert Schemas to PDL](#how-to-convert-schemas-to-pdl)
+- [Notable Differences Between PDSC and PDL](#notable-differences-between-pdsc-and-pdl)
+    - [PDL is More Java-Like](#pdl-is-more-java-like)
+    - [Import Statements](#import-statements)
+    - [Shorthand for Custom Properties](#shorthand-for-custom-properties)
+
+## Why Use PDL?
+
+PDL as a schema definition language was designed to be read and written by humans, making it a much more enjoyable experience for developers
+to define their service's models. PDSC, despite its syntax being simply a subset of JSON, cannot boast this human-readability factor.
+
+In addition to its inherently more readable syntax, PDL also features some extra shorthand which developers can leverage to write less and
+make their schemas more readable for others. See [below](#notable-differences-between-pdsc-and-pdl) for more information on this.
+
+## How to Convert Schemas to PDL
+
+If you have an existing project with PDSC schemas, converting everything to PDL is pretty straightforward.
+Rest.li's Gradle plugin provides a task `convert<sourceSet>ToPdl` which will automatically convert your schemas for you.
+It should be noted that this task verifies the converted schemas against the original schemas.
+If this verification is failed, then the whole conversion will be aborted.
+
+The following command will convert every PDSC schema in your project to PDL.
+
+```
+gradle convertToPdl
+```
+
+You can selectively migrate only one particular module of your project with the following:
+
+```
+gradle :<moduleName>:convertToPdl
+```
+
+This task also takes in a few options:
+
+```
+gradle :<moduleName>:convertToPdl \
+    [-PconvertToPdl.reverse=(true|false)] \
+    [-PconvertToPdl.keepOriginal=(true|false)] \
+    [-PconvertToPdl.preserveSrcCmd]
+```
+
+| Property | Type | Description
+|----------|------|------------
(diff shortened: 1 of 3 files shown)


pdl_migration.md
```diff
@@ -0,0 +1,255 @@
...
+
+If you have an existing project with PDSC schemas, converting everything to PDL is pretty straightforward.
+Rest.li's Gradle plugin provides a task `convert<sourceSet>ToPdl` which will automatically convert your schemas for you.
+It should be noted that this task verifies the converted schemas against the original schemas.
+If this verification is failed, then the whole conversion will be aborted.
+
+The following command will convert every PDSC schema in your project to PDL.
+
+```
+gradle convertToPdl
```
nickibi: Shall we mention rest.li version for this convert tool?

pdl_migration.md
```diff
@@ -0,0 +1,255 @@
...
+    [-PconvertToPdl.reverse=(true|false)] \
+    [-PconvertToPdl.keepOriginal=(true|false)] \
+    [-PconvertToPdl.preserveSrcCmd]
+```
+
+| Property | Type | Description
+|----------|------|------------
+| `convertToPdl.reverse` | boolean | If true, converts PDL schemas back to PDSC (and vice versa if false).
+| `convertToPdl.keepOriginal` | boolean | If true, keeps the source schemas (the source schemas are deleted by default).
+| `convertToPdl.preserveSrcCmd` | string | Command which is run for each file, useful for running special VCS logic. The command should be a template string containing `$src` and `$dst` as references to the source and destination filename, respectively.
```
nickibi: I think we should provide an example here, like - gradle  convertToPdl -PconvertTodl.preserveSrcCmd "/usr/bin/svn mv \$src \$dst"

pdl_syntax.md
```diff
@@ -494,7 +494,7 @@ record InlineWithinUnion {
 ```
 
 ### Union with aliases
-***Note:*** _Union with aliases is a recent feature in the pegasus schema language and it might not be fully supported in non-java languages. Please check the [support level]({% link testsuite_compatibility_matrix.md %}) on all languages you intend to use before using aliases_
+***Note:*** _Union with aliases is a recent feature in the pegasus schema language and it might not be fully supported in non-java languages. Please check the [support level](/rest.li/multi_language_compatibility_matrix) on all languages you intend to use before using aliases_
```
nickibi: Thanks for fixing the link.

pdl_migration.md
```diff
@@ -0,0 +1,255 @@
...
+
+/**
+ * Foo is a record.
+ */
+record Foo {
+    x: int = 1
+}
+        </pre></td>
+    </tr>
+</table>
```
nickibi: Shall we put a comprehensive example here?

pdl_migration.md
evanw555: Nope, it's been around since version 10, so it's pretty old.

nickibi: But we updated the cmd tool, did we?

evanw555: You're right, I'll address this.

pdl_migration.md
```diff
@@ -0,0 +1,412 @@
...
+    [-PconvertToPdl.reverse=(true|false)] \
+    [-PconvertToPdl.keepOriginal=(true|false)] \
+    [-PconvertToPdl.preserveSrcCmd]
+```
+
+| Property | Type | Since | Description
+|----------|------|-------|-------------
+| `convertToPdl.reverse` | boolean | `28.0.11` | If true, converts PDL schemas back to PDSC (and vice versa if false).
+| `convertToPdl.keepOriginal` | boolean | `28.0.11` | If true, keeps the source schemas (the source schemas are deleted by default).
+| `convertToPdl.preserveSrcCmd` | string | `28.1.1` | Command which is run for each file, useful for running special VCS logic. The command should be a template string containing `$src` and `$dst` as references to the source and destination filename, respectively (e.g. `gradle convertToPdl -PconvertToPdl.preserveSrcCmd "/usr/bin/svn mv $src $dst"`).
```
nickibi: We did the whitespace fix in 28.1.3. We probably should suggest customer use 28.1.3+ to avoid schema validation failure. WDYT?

evanw555: By this logic, we may as well just say that any version below 28.1.3 isn't guaranteed to work correctly for this command.

nickibi: Umm, why can't we just recommend a good version?

evanw555: Same thing.

pdl_migration.md
```diff
@@ -0,0 +1,414 @@
...
+```
+gradle :<moduleName>:convertToPdl \
+    [-PconvertToPdl.reverse=(true|false)] \
+    [-PconvertToPdl.keepOriginal=(true|false)] \
+    [-PconvertToPdl.preserveSrcCmd]
+```
+
+| Property | Type | Description
+|----------|------|-------------
+| `convertToPdl.reverse` | boolean | If true, converts PDL schemas back to PDSC (and vice versa if false).
```
karthikbalasub: Include skip-verification flag

evanw555: I thought that too at first when I wrote this, but the `convertToPdl` Gradle task is separate from the `SchemaFormatTranslator` class which supports this flag.

PR #166, opened 2020-02-11
Change pdl property encoding logic to not flatten if nested branches > 1

aman1309:
Change pdl property encoding logic to not flatten if nested branches > 1

### DIFF
--- a/data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
+++ b/data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
@@ -129,8 +129,6 @@ abstract class PdlBuilder
   /**
    * Writes a set of schema properties that share a common prefix to .pdl.
    *
-   * TODO: Support configuration for preferring collapsed "map" style over expanded "path" style.
-   *
    * @param prefix provides the common prefix of all the properties.
    * @param properties provides the properties to write.
    */
@@ -149,14 +147,14 @@ abstract class PdlBuilder
       {
         // Favor @x.y.z = "value" property encoding style over @x = { "y": { "z": "value" } }
         DataMap dm = (DataMap) value;
-        if (!dm.isEmpty() && dm.size() == 1)
+        if (dm.size() == 1)
         {
           // encode non-empty value property like @x.y.z = "value"
           writeProperties(pathParts, dm);
         }
-        else if (!pathParts.isEmpty())
+        else
         {
-          // encode empty value property like @x.y = {}
+          // encode value property like @x.y = {jsonValue}
           writeProperty(pathParts, dm);
         }
       }
--- a/data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
+++ b/data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
@@ -149,7 +149,7 @@ abstract class PdlBuilder
       {
         // Favor @x.y.z = "value" property encoding style over @x = { "y": { "z": "value" } }
         DataMap dm = (DataMap) value;
-        if (!dm.isEmpty())
+        if (!dm.isEmpty() && dm.size() == 1)
         {
           // encode non-empty value property like @x.y.z = "value"
           writeProperties(pathParts, dm);
--- a/data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
+++ b/data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
@@ -145,16 +145,16 @@ abstract class PdlBuilder
 
       if (value instanceof DataMap)
       {
-        // Favor @x.y.z = "value" property encoding style over @x = { "y": { "z": "value" } }
         DataMap dm = (DataMap) value;
+        // Decide encoding style based on map branches/size
         if (dm.size() == 1)
         {
-          // encode non-empty value property like @x.y.z = "value"
+          // encode value property like @x.y.z = "value"
           writeProperties(pathParts, dm);
         }
         else
         {
-          // encode value property like @x.y = {jsonValue}
+          // encode value property like @x = { "y": { "z": "value" } }
           writeProperty(pathParts, dm);
         }
       }
--- a/data/src/test/java/com/linkedin/data/schema/TestPdlBuilder.java
+++ b/data/src/test/java/com/linkedin/data/schema/TestPdlBuilder.java
@@ -1,0 +1,41 @@
+package com.linkedin.data.schema;
+
+import com.linkedin.data.DataMap;
+import java.io.IOException;
+import java.io.StringWriter;
+import java.util.Collections;
+import java.util.Map;
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+
+public class TestPdlBuilder {
+
+  @DataProvider
+  private static Object[][] propertiesMapProvider()
+  {
+    DataMap properties1 = new DataMap();
+    properties1.put("empty", new DataMap());
+    DataMap properties2 = new DataMap();
+    properties2.put("validate", properties1);
+    DataMap properties3 = new DataMap();
+    DataMap nestedMap = new DataMap(properties2);
+    nestedMap.putAll(properties1);
+    properties3.put("nested", nestedMap);
+    return new Object[][]{{properties1, "@empty = {}\n"},
+        {properties2, "@validate.empty = {}\n"}
+        //TODO Add test case for multiple properties in a map level once iteration logic is fixed to be deterministic
+    };
+  }
+
+  @Test(dataProvider = "propertiesMapProvider")
+  public void testPropertiesWriter(Map<String, Object> properties, String pdlString) throws IOException
+  {
+    StringWriter writer = new StringWriter();
+    PdlBuilder pdlBuilder = (new IndentedPdlBuilder.Provider()).newInstance(writer);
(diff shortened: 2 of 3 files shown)


data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
```diff
@@ -149,7 +149,7 @@ PdlBuilder writeProperties(List<String> prefix, Map<String, Object> properties)
       {
         // Favor @x.y.z = "value" property encoding style over @x = { "y": { "z": "value" } }
```
evanw555: Most of the comments in this method are now invalid. You should update them.

> Favor @x.y.z = "value" property encoding style over @x = { "y": { "z": "value" } }
> encode non-empty value property like @x.y.z = "value"
> encode empty value property like @x.y = {}

I'd also suggest cleaning this up.

> Use shorthand for boolean true.  Instead of writing "@deprecated = true",
> write "@deprecated".

Also, remove this from the method javadoc.

> TODO: Support configuration for preferring collapsed "map" style over expanded "path" style.

data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
```diff
@@ -149,7 +149,7 @@ PdlBuilder writeProperties(List<String> prefix, Map<String, Object> properties)
...
         DataMap dm = (DataMap) value;
-        if (!dm.isEmpty())
+        if (!dm.isEmpty() && dm.size() == 1)
```
evanw555: Redundant. Just `dm.size() == 1` suffices.

data/src/test/java/com/linkedin/data/schema/TestSchemaToPdlEncoder.java
```diff
@@ -22,19 +22,38 @@
...
+    nestedMap.putAll(properties1);
+    properties3.put("nested", nestedMap);
+    return new Object[][]{{properties1, "namespace com.linkedin.test\n" + "\n" + "@empty = {}\nrecord RecordDataSchema {}"},
+        {properties2, "namespace com.linkedin.test\n" + "\n" + "@validate.empty = {}\n" + "@empty = {}\nrecord RecordDataSchema {}"},
+        {properties3, "namespace com.linkedin.test\n" + "\n" + "@nested = {\"validate\":{\"empty\":{}},\"empty\":{}}\n"
+            + "@empty = {}\n" + "record RecordDataSchema {}"}};
+  }
+
+  @Test(dataProvider = "recordSchemaProvider")
+  public void testEncodeRecordWithEmptyDataMapInProperty(Map<String, Object> properties, String pdlString) throws IOException
```
evanw555: By adding more test cases that don't apply to the original intent of the test method, this becomes more confusing. I'd suggest making this more generalized of a method which just tests custom property encoding.

However, this brings up another interesting point. This test assumes the ordering of properties, which is currently NOT deterministic. I'd suggest keeping the tests right now limited to only one property per test case, and adding a TODO to be addressed by whoever implements deterministic ordering. _In addition_, this sort of testing should be done by `PdlEncoderTest`, which makes testing the encoder much easier. You can avoid defining the schema as a string literal.

data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
aman1309: done. Initially thought dm could be null also which is not the case

data/src/test/java/com/linkedin/data/schema/TestSchemaToPdlEncoder.java
```diff
@@ -22,19 +22,38 @@
...
   {
     RecordDataSchema source =
         new RecordDataSchema(new Name("com.linkedin.test.RecordDataSchema"), RecordDataSchema.RecordType.RECORD);
-    Map<String, Object> properties = new HashMap<>();
     properties.put("empty", new DataMap());
```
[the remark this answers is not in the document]
aman1309: done

data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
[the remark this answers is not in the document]
aman1309: done

data/src/test/java/com/linkedin/data/schema/TestSchemaToPdlEncoder.java
aman1309: I thought of writing it in PdlEncoderTest but then its in a different module, and also uses SchemaToPdlEncoder.
Now, I have added a new test class for TestPdlEncoder in same package.

Regarding the deterministic ordering I have removed the last test for now.

data/src/test/java/com/linkedin/data/schema/TestPdlBuilder.java
```diff
@@ -0,0 +1,41 @@
+package com.linkedin.data.schema;
```
evanw555: License

data/src/test/java/com/linkedin/data/schema/TestPdlBuilder.java
```diff
@@ -0,0 +1,41 @@
...
+import com.linkedin.data.DataMap;
+import java.io.IOException;
+import java.io.StringWriter;
+import java.util.Collections;
+import java.util.Map;
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+
```
evanw555: Add javadoc, my personal convention for test javadocs is: `Tests for {@link PdlBuilder}.`

data/src/test/java/com/linkedin/data/schema/TestPdlBuilder.java
```diff
@@ -0,0 +1,41 @@
...
+    properties1.put("empty", new DataMap());
+    DataMap properties2 = new DataMap();
+    properties2.put("validate", properties1);
+    DataMap properties3 = new DataMap();
+    DataMap nestedMap = new DataMap(properties2);
+    nestedMap.putAll(properties1);
+    properties3.put("nested", nestedMap);
+    return new Object[][]{{properties1, "@empty = {}\n"},
+        {properties2, "@validate.empty = {}\n"}
+        //TODO Add test case for multiple properties in a map level once iteration logic is fixed to be deterministic
```
evanw555: It's fine that each test (for now) has one root-level property, but none of them branch. Thus, these cases don't actually test your changes. You'll need to make this a little more comprehensive.

data/src/test/java/com/linkedin/data/schema/TestPdlBuilder.java
```diff
@@ -0,0 +1,41 @@
...
+    };
+  }
+
+  @Test(dataProvider = "propertiesMapProvider")
+  public void testPropertiesWriter(Map<String, Object> properties, String pdlString) throws IOException
```
evanw555: nit: I'd suggest mapping the test method name directly to the method being tested; testing `writeProperties` -> `testWriteProperties`

data/src/test/java/com/linkedin/data/schema/TestPdlBuilder.java
```diff
@@ -0,0 +1,41 @@
...
+import java.io.IOException;
+import java.io.StringWriter;
+import java.util.Collections;
+import java.util.Map;
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+
+public class TestPdlBuilder {
```
evanw555: Please use the Rest.li brace style convention (braces on new lines). This goes especially for your data provider, the return value should be expanded neatly.

data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
evanw555: Your changes still don't address most of this.

data/src/test/java/com/linkedin/data/schema/TestPdlBuilder.java
aman1309: done

data/src/test/java/com/linkedin/data/schema/TestPdlBuilder.java
aman1309: done

data/src/test/java/com/linkedin/data/schema/TestPdlBuilder.java
aman1309: done

data/src/test/java/com/linkedin/data/schema/TestPdlBuilder.java
aman1309: done

data/src/test/java/com/linkedin/data/schema/TestPdlBuilder.java
aman1309: as discussed this is not possible right now, as ordering is not deterministic.

data/src/test/java/com/linkedin/data/schema/TestPdlBuilder.java
```diff
@@ -0,0 +1,67 @@
+/*
+   Copyright (c) 2015 LinkedIn Corp.
```
evanw555: Looks good overall, just fix this to 2020.

PR #169, opened 2020-02-13
Fix bug that would cause pdl encoder to import types conflicting with inline types.

karthikbalasub:
Also improves how imports for includes are handled.

### DIFF
--- a/data/src/main/java/com/linkedin/data/schema/SchemaToPdlEncoder.java
+++ b/data/src/main/java/com/linkedin/data/schema/SchemaToPdlEncoder.java
@@ -718,6 +718,7 @@ public class SchemaToPdlEncoder extends AbstractSchemaEncoder
    * The following rules are used to determine whether a type should be imported:
    * (1) The type is outside the root namespace of the document.
    * (2) The type is declared outside the document (i.e. not inlined in this document).
+   * (3) The type's name does not conflict with name of an Inlined type.
    *
    * When multiple referenced types with the same unqualified name may be imported, the type with the alphabetically
    * first namespace is chosen. (e.g. "com.a.b.c.Foo" is chosen over "com.x.y.z.Foo")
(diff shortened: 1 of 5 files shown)


generator-test/src/test/pegasus/com/linkedin/pegasus/generator/test/idl/enums/EscapedSymbols.pdl
```diff
@@ -5,7 +5,6 @@ namespace com.linkedin.pegasus.generator.test.idl.enums
  */
 enum EscapedSymbols {
   ENUM,
-  `enum`,
```
nickibi: why do we delete this one?

karthikbalasub: This was an issue introduced in previous commit. enum cannot be used as it is reserved keyword in java. The template generation step fails for this.

evanw555: Oops

data/src/main/java/com/linkedin/data/schema/SchemaToPdlEncoder.java
```diff
@@ -782,7 +782,7 @@ private void gatherTypes(DataSchema schema, boolean isDeclaredInline, Set<Name>
         }
         for (NamedDataSchema include : recordSchema.getInclude())
         {
-          gatherTypes(include, true, encounteredTypes, inlinedTypes);
+          gatherTypes(include, recordSchema.isIncludeDeclaredInline(include), encounteredTypes, inlinedTypes);
```
evanw555: Nice catch.

data/src/main/java/com/linkedin/data/schema/SchemaToPdlEncoder.java
```diff
@@ -756,7 +756,7 @@ private void writeDocAndProperties(RecordDataSchema.Field field) throws IOExcept
    * @param inlinedTypes cumulative set of all inlined types in this schema (and its descendents).
```
evanw555: We should clarify that this set is a collection of simple names, not full names.

generator-test/src/test/pegasus/com/linkedin/pegasus/generator/test/idl/imports/InlineTypeConflict.pdl
```diff
@@ -0,0 +1,28 @@
...
+
+  /**
+   * Inline type can also conflict with types used for other fields.
+   */
+  inlineType2: record Note {
+    a: int
+  }
+
+  /**
+   * This field reference a type (Note) that conflicts with Note defined inline, and thus must use fully qualified name.
```
evanw555: nit: `reference` -> `references`

PR #173, opened 2020-02-14
Provide a fast path to just modify the URI of a request

karthikrg:
This is a hot path in many services that just want to rewrite URIs.

### DIFF
--- a/r2-core/src/main/java/com/linkedin/r2/message/BaseMessage.java
+++ b/r2-core/src/main/java/com/linkedin/r2/message/BaseMessage.java
@@ -41,12 +41,25 @@ public abstract class BaseMessage implements MessageHeaders
 
   protected BaseMessage(Map<String, String> headers, List<String> cookies)
   {
-    ArgumentUtil.notNull(headers, "headers");
-    ArgumentUtil.notNull(cookies, "cookies");
-    TreeMap<String, String> tmpHeaders = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
-    tmpHeaders.putAll(headers);
-    _headers = Collections.unmodifiableSortedMap(tmpHeaders);
-    _cookies = Collections.unmodifiableList(new ArrayList<>(cookies));
+    this(headers, cookies, true);
+  }
+
+  protected BaseMessage(Map<String, String> headers, List<String> cookies, boolean validateArgs)
+  {
+    if (validateArgs)
+    {
+      ArgumentUtil.notNull(headers, "headers");
+      ArgumentUtil.notNull(cookies, "cookies");
+      TreeMap<String, String> tmpHeaders = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+      tmpHeaders.putAll(headers);
+      _headers = Collections.unmodifiableSortedMap(tmpHeaders);
+      _cookies = Collections.unmodifiableList(new ArrayList<>(cookies));
+    }
+    else
+    {
+      _headers = headers;
+      _cookies = cookies;
+    }
   }
 
   @Override
--- a/r2-core/src/main/java/com/linkedin/r2/message/BaseRequest.java
+++ b/r2-core/src/main/java/com/linkedin/r2/message/BaseRequest.java
@@ -18,11 +18,24 @@ public abstract class BaseRequest extends BaseMessage implements Request
 
   protected BaseRequest(Map<String, String> headers, List<String> cookies, URI uri, String method)
   {
-    super(headers, cookies);
-    ArgumentUtil.notNull(uri, "uri");
-    ArgumentUtil.notNull(method, "method");
-    _uri = uri;
-    _method = method;
+    this(headers, cookies, uri, method, true);
+  }
+
+  protected BaseRequest(Map<String, String> headers, List<String> cookies, URI uri, String method, boolean validateArgs)
+  {
+    super(headers, cookies, validateArgs);
+    if (validateArgs)
+    {
+      ArgumentUtil.notNull(uri, "uri");
+      ArgumentUtil.notNull(method, "method");
+      _uri = uri;
+      _method = method;
+    }
+    else
+    {
+      _uri = uri;
+      _method = method;
+    }
   }
 
   @Override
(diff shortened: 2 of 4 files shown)


karthikbalasub (commented):
This is in r2-core, so please include @ChaoLinkedIn  and @FreCap as reviewers

karthikrg:
> This is in r2-core, so please include @ChaoLinkedIn and @FreCap as reviewers

Can't seem to add them?

r2-core/src/main/java/com/linkedin/r2/message/rest/RestRequestImpl.java
```diff
@@ -33,10 +33,22 @@
...
+
+  private RestRequestImpl(ByteString entity, Map<String, String> headers,
+      List<String> cookies, URI uri, String method, boolean validateArgs)
+  {
+    super(headers, cookies, uri, method, validateArgs);
     _entity = entity;
   }
 
+  @Override
+  public RestRequest copyWithURI(URI uri)
```
mchen07: This feels weird to me. Given a `RestRequest`, you want to provide a convenient method to get another RestRequest with only URI changed, this can be done completely outside pegasus, because `RestRequest` already provide a `builder` for it. So client should be very easy to do this by

```
   RestRequest origRequest = ....
   RestRequest updatedRequest = origRequest.builder().setURI(uri).build();
```
If this is the only purpose of this PR, I don't think that we need this here. Because  by analogy, I may question why we don't add any other convenient methods to update some part of an existing RestRequest. Personally feel current way to expose a `builder` API is cleaner.

r2-core/src/main/java/com/linkedin/r2/message/BaseMessage.java
```diff
@@ -41,12 +41,25 @@
...
+      ArgumentUtil.notNull(headers, "headers");
+      ArgumentUtil.notNull(cookies, "cookies");
+      TreeMap<String, String> tmpHeaders = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+      tmpHeaders.putAll(headers);
+      _headers = Collections.unmodifiableSortedMap(tmpHeaders);
+      _cookies = Collections.unmodifiableList(new ArrayList<>(cookies));
+    }
+    else
+    {
+      _headers = headers;
```
mchen07: This changed header order in previous code, may break existing code.

r2-core/src/main/java/com/linkedin/r2/message/BaseRequest.java
```diff
@@ -18,11 +18,24 @@
...
-    _method = method;
+    this(headers, cookies, uri, method, true);
+  }
+
+  protected BaseRequest(Map<String, String> headers, List<String> cookies, URI uri, String method, boolean validateArgs)
+  {
+    super(headers, cookies, validateArgs);
+    if (validateArgs)
+    {
+      ArgumentUtil.notNull(uri, "uri");
```
mchen07: Is this really costly? Not sure if it is worth to use flag to control such minor CPU.

r2-core/src/main/java/com/linkedin/r2/message/rest/RestRequestImpl.java
karthikrg: That is exactly what the default method in the interface does. However, calling this method can result in onerous performance costs, especially if it is in the hot path, since creating a builder and then building a request from it, involves some header copying and name validation.

I sent you a flamegraph offline showing the CPU impact of this on one of our internal services.

r2-core/src/main/java/com/linkedin/r2/message/BaseMessage.java
karthikrg: Sorry I didn't get what you were alluding to. The existing method signature and behavior is unchanged.

r2-core/src/main/java/com/linkedin/r2/message/BaseRequest.java
karthikrg: Refer to previous comment about why it's expensive

r2-core/src/main/java/com/linkedin/r2/message/BaseMessage.java
mchen07: I meant if `validateArgs` is true, `_headers` is sorted, but it is not if `validateArgs` is false. This may cause issues for code that handles the R2 request if the code has assumption on the headers passed.

r2-core/src/main/java/com/linkedin/r2/message/BaseRequest.java
mchen07: In this particular routine, you are just saving two Null checks? I didn't see them stand out in your flamegraph, or maybe I overlooked something in the flamegraph.

evanw555:
@li-kramgopa You can't add them because they haven't been added to the SI GitHub team yet (i.e. they don't have write access). We'll work to add them. In the meantime, you should at least link them to this PR so they can look it over.

r2-core/src/main/java/com/linkedin/r2/message/BaseRequest.java
karthikrg: I don't think we care about map sorting at all! What we care about is case-insensitivity to follow the http header spec. TreeMap just happens to be a sorted map that can be configured as a case -insensitive map.

The entire point of validateArgs is this: We will not validate/copy anything and trust whatever the subclass sent us if this is false. This is supposed to be carefully used by subclasses who know what they are doing (like us creating a url rewritten request copy here).

The default value of validateArgs is true, and in this mode all the existing validation and copying checks run.

r2-core/src/main/java/com/linkedin/r2/message/BaseMessage.java
karthikrg: I don't think we care about map sorting at all! What we care about is case-insensitivity to follow the http header spec. TreeMap just happens to be a sorted map that can be configured as a case -insensitive map.

The entire point of validateArgs is this: We will not validate/copy anything and trust whatever the subclass sent us if this is false. This is supposed to be carefully used by subclasses who know what they are doing (like us creating a url rewritten request copy here).

The default value of validateArgs is true, and in this mode all the existing validation and copying checks run.

r2-core/src/main/java/com/linkedin/r2/message/BaseRequest.java
karthikrg: Most of the use here is to proxy the validateArgs to the superclass (BaseMessage) where all the expensive header copy and field name validation happens. Thought while we are it might as well also save on the null checks.

li-bsoetarman (commented):
We need to be able to quantify these performance improvements.
For these kind of logic only changes, please also add the jmh tests, and before and after jmh results.
In addition, also please post the before and after flame graph; how much time are we actually saving?
Thanks.

ChrisCarini:

PR #175, opened 2020-02-17
Update AVRO compat layer in pegasus to use adapters from avro-util

mchen07:
With avro-util 0.1.8 released and ELRed, this PR is ready to merge

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -99,7 +99,7 @@ project.ext.externalDependency = [
 
   'jsr305': 'com.google.code.findbugs:jsr305:3.0.0',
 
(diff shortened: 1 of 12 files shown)


build.gradle
```diff
@@ -224,6 +227,9 @@ subprojects {
     repositories {
       mavenLocal()
       mavenCentral()
+      maven {
```
evanw555: Can you not just use `jcenter()`?

radai-rosenblatt: we dont publish avro-util to jcenter or central (yet?)

evanw555: Nvm, I forgot about the distinction between "central" and our own org repo on Bintray

PR #178, opened 2020-02-20
Fix for DataTranslator where PegasusToAvroDefaultFieldTranslation mode throws unexpcted exception

junchuanwang:
PR migrated from RB: https://rb.corp.linkedin.com/r/196355

This is a bug fix when using the DataTranslator with the translation mode set to DO_NOT_TRANSLATE, the translator throws an exception for schemas without defaults

Related jira ticket: [SI-14208](https://jira01.corp.linkedin.com:8443/browse/SI-14208)

Tests:
(0) unit-tests
(1) mint build
(2) container wc-tests


data-avro/src/test/java/com/linkedin/data/avro/TestDataTranslator.java
```diff
@@ -1154,48 +1158,211 @@ private void testDataTranslation(String schemaText, String[][] row) throws IOExc
 
   @DataProvider
   public Object[][] defaultToAvroOptionalTranslationProvider() {
-    // This test simulates the case where the DataMap for update is empty
-    // Each object array contains four element,
+    // These tests test DataMap translation under different PegasusToAvroDefaultFieldTranslationMode modes.
+    // it will test whether the Avro map generated from expected AvroJsonString is as expected.
+    // Each object array contains eight element,
```
nickibi: nit: element**s**

data-avro/src/test/java/com/linkedin/data/avro/TestDataTranslator.java
```diff
@@ -1154,48 +1158,211 @@ private void testDataTranslation(String schemaText, String[][] row) throws IOExc
...
     // 1. first element is the schema Text,
-    // 2. second element is the String representation of the data map
-    // 3. third element is the dataTranslationMode for the default field and
-    // 4. fourth element is is the expected AvroJsonString after translation
-    // The test will test whether the Avro map generated from expected AvroJsonString is as expected from the program
+    // 2. second element is the schema translation mode
+    // 3. third element is the expected schema translation outcome
+    // 4. fourth element is the String representation of the data map
+    // 5. fifth element is the dataTranslationMode for the default field and
```
nickibi: nit: remove "and"

data-avro/src/test/java/com/linkedin/data/avro/TestDataTranslator.java
```diff
@@ -1154,48 +1158,211 @@ private void testDataTranslation(String schemaText, String[][] row) throws IOExc
...
+    // 6. sixth element is is the expected AvroJsonString after translation
```
nickibi: nit: remove extra "is"

data-avro/src/test/java/com/linkedin/data/avro/TestDataTranslator.java
```diff
@@ -1154,48 +1158,211 @@ private void testDataTranslation(String schemaText, String[][] row) throws IOExc
 
   @DataProvider
   public Object[][] defaultToAvroOptionalTranslationProvider() {
-    // This test simulates the case where the DataMap for update is empty
-    // Each object array contains four element,
+    // These tests test DataMap translation under different PegasusToAvroDefaultFieldTranslationMode modes.
+    // it will test whether the Avro map generated from expected AvroJsonString is as expected.
```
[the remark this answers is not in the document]
evanw555: If he adds this extra "is", then he has to remove the "is" at the end of the sentence.

data-avro/src/main/java/com/linkedin/data/avro/DataTranslator.java
```diff
@@ -642,10 +641,20 @@ private Object translate(Object value, DataSchema dataSchema, Schema avroSchema)
             }
             else if (fieldValue == null)
             {
+              // if no fieldValue present, either NULL value or defaultValue will be assigned
               Object defaultValue = field.getDefault();
               if (defaultValue != null)
               {
-                fieldValue = defaultValue;
+                if (_dataTranslationOptions == null || ((DataMapToAvroRecordTranslationOptions) _dataTranslationOptions). getDefaultFieldDataTranslationMode()
```
evanw555: nit: remove space after the dot

data-avro/src/test/java/com/linkedin/data/avro/TestDataTranslator.java
```diff
@@ -50,6 +52,8 @@
 
 public class TestDataTranslator
 {
+  private static final Logger LOG = LoggerFactory.getLogger(TestDataTranslator.class);
```
evanw555: You've said this is only for debug, but it's not actually used in the code you intend to commit. You should delete it.

data-avro/src/main/java/com/linkedin/data/avro/DataTranslator.java
```diff
@@ -609,18 +609,17 @@ private Object translate(Object value, DataSchema dataSchema, Schema avroSchema)
             Schema fieldAvroSchema = avroField.schema();
             Object fieldValue = map.get(fieldName);
             boolean isOptional = field.getOptional();
-            if (isOptional  || (
-                _dataTranslationOptions != null &&
-                ((DataMapToAvroRecordTranslationOptions) _dataTranslationOptions). getDefaultFieldDataTranslationMode()
-                    == PegasusToAvroDefaultFieldTranslationMode.DO_NOT_TRANSLATE))
+            if (isOptional)
```
evanw555: I originally left this comment, which I'm still curious about.

> Based on my understanding, this if-statement is just used to determine if the field schema was converted to a union. This doesn't seem to be consistent with the logic that actually determines whether to convert the schema field to a union.

evanw555: After arguing for a few hours, I've come to the conclusion that the way you're doing it here is valid.

data-avro/src/main/java/com/linkedin/data/avro/DataTranslator.java
```diff
@@ -609,18 +609,17 @@ private Object translate(Object value, DataSchema dataSchema, Schema avroSchema)
...
             {
+              // process optional fields
```
evanw555: Generally LGTM, though the documentation here could be drastically improved. This would help to avoid long conversations in the future to figure out what's going on here. It looks like you already added a few, but here are my suggested edits:

> // process optional fields

This comment is unnecessary, since the `isOptional` condition above is clear.

Add a comment like this:

```
if (fieldDataSchema.getDereferencedType() != DataSchema.Type.UNION)
{
  // Translating from a non-union pegasus type to an embedded union type in Avro
  if (fieldValue == null)
```

> // optional field will be represented as union in Avro

This doesn't really explain what's happening here. I'd suggest:
```
// Extract the Avro type corresponding to the pegasus type from the Avro union
```

> // already a union

Should be something like:
```
// Translating from a pegasus union to an Avro union
```

> // if no fieldValue present, either NULL value or defaultValue will be assigned

Should be a little more descriptive:

```
// Required field is missing, assign default value if present
```

Then add another comment in the subcondition saying:

```
// Translate default value as null, depending on specified options
```

PR #179, opened 2020-02-20
fix template exception while request build init for custom response types

aman1309:
Test: 
mint build
wrote an action test to test changes manually.


restli-common/src/main/java/com/linkedin/restli/common/TypeSpec.java
```diff
@@ -67,7 +68,11 @@ public TypeSpec(Class<T> type)
   private static DataSchema backfillSchemaIfPossible(Class<?> type)
   {
     // These are all the classes used for type specs that are "schema-less".
-    if(type == CompoundKey.class || type == ComplexResourceKey.class || type == Void.class) return null;
+    if (type == CompoundKey.class || type == ComplexResourceKey.class || type == Void.class
+        || (!DataSchemaUtil.isPrimitiveClass(type) && DataTemplateUtil.hasCoercer(type)))
```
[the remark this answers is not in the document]
aman1309: there are data schemas available for primitives and coercers also registers primitives

restli-client/src/main/java/com/linkedin/restli/client/ActionRequestBuilder.java
```diff
@@ -193,7 +195,10 @@ public ActionRequestBuilder(String baseUriTemplate, TypeSpec<V> elementType, Res
     if (_resourceSpec.getRequestMetadata(_name) == null) // old builder code in use
     {
       requestDataSchema = DynamicRecordMetadata.buildSchema(_name, _actionParams.keySet());
-
+      if (_elementType == null)
+      {
+        _elementType = new TypeSpec<>(_elementClass);
```
karthikbalasub: This is not loading from resourespec?

aman1309: this is old request builder code where spec doesn't have that info. see the else part

karthikbalasub: Ah, didn't notice that.

restli-common/src/main/java/com/linkedin/restli/common/TypeSpec.java
evanw555: Why don't you just keep track of registered custom classes in `Custom.java`? I think that would be simpler

aman1309: we could do that but didn't see much value for adding another map when this can be achieved using existing ones. Also, since coercers are only for customs, so felt like doing it this way. Let me know if you see any issue with this.

evanw555: true, though I feel like this can be more abstracted. What if you created a new helper method (e.g. `DataTemplateUtil#isCustomType(Class<?>)` which itself checks the coercers and does the logic to exclude primitives. It doesn't make sense to me that this consumer method needs to know that primitives also have coercers, it just wants to know if a type is a custom typeref/custom type/whatever.

evanw555: Another alternative is to create the helper method as `Custom#isCustomType(Class<?>)`, since it may not be appropriate for `DataTemplateUtil` to have a sense of what is "custom" and what isn't.

aman1309: done

data/src/main/java/com/linkedin/data/template/Custom.java
```diff
@@ -149,4 +151,16 @@ public static void initializeCustomClass(Class<?> customClass)
...
+  /**
+   * Return whether the input class is a Custom type or not
+   * uses coercer map to identify as each custom type must register a coercer excluding the primitive types.
+   *
+   * @param clazz is the class to check.
+   * @return true if an object of the input class is a Custom type.
+   */
+  public static boolean isCustomType(Class<?> clazz)
+  {
+    return !DataSchemaUtil.isPrimitiveClass(clazz) && DataTemplateUtil.hasCoercer(clazz);
```
evanw555: Thinking about this now. How will this behave for enums? You should test this. It looks like the `Enum` type has a coercer, yet it's not in the "primitives" mapping. Would this method return true?

data/src/main/java/com/linkedin/data/template/Custom.java
```diff
@@ -149,4 +151,16 @@ public static void initializeCustomClass(Class<?> customClass)
...
   }
+
+  /**
+   * Return whether the input class is a Custom type or not
+   * uses coercer map to identify as each custom type must register a coercer excluding the primitive types.
+   *
+   * @param clazz is the class to check.
+   * @return true if an object of the input class is a Custom type.
+   */
+  public static boolean isCustomType(Class<?> clazz)
```
evanw555: I suggest renaming to `isCustomClass` to be consistent with the other methods in this class. Update the Javadoc to explain that this method checks if a given class is registered as a custom Java binding.

aman1309: removed the changes

data/src/main/java/com/linkedin/data/template/Custom.java
aman1309: removed the changes from typespec as its anyway legacy code and not needed for fix as it handled in action request builder change.

karthikbalasub: Did you check this wouldn't be needed for other places that use TypeSpec? I think it would still be good to have the check in TypeSpec as fallback. It doesn't hurt to be extra safe?

aman1309: will take it as a separate PR since adding multiple checks like primitive and enum didn't seem right.
Also, BatchRequests are build in a same way which would affect the performance but other places the class is constant.

evanw555:
I noticed that you totally scrapped your old solution. Could you please provide some context on what the context is here?

aman1309:
> I noticed that you totally scrapped your old solution. Could you please provide some context on what the context is here?

For the action request builder fix, now schema is taken from ResourceSpec instead of reflection so the code which is throwing the exception is no longer used.
TypeSpec fix done before had complications like checking for enum and primary types which would complicate things in adding coercers for new types. We can look at this later if required since the constructor being used is legacy code which anyway would not be used. if used its a constant class where a schema is available and the exception is not thrown except BatchRequest which I will open a separate PR.

restli-client/src/main/java/com/linkedin/restli/client/ActionRequestBuilder.java
evanw555: Why are we lazily loading this class-level field in this method? Wouldn't it make more sense to just do so in the constructor?

aman1309: yeah, we have to since this is required for legacy flow only when there were no resourcespec generated. Also, It should be fine since this is builder class not the actual one.

evanw555: Ok, then why isn't it a local variable? It's not clean to have a class-level field which is only initialized and read in one particular method.

aman1309: because with another constructor we accept typespec directly so that we don't have to extract the schema from the class using reflection which is a costly thing

evanw555: Nevermind

PR #185, opened 2020-02-24
AsyncPool Improvements and Fixes

nizarm:
When a object creation request to the pool is faster than the actual creation itself - the rate limiter queues all the incoming requests and creates the object without considering the fact that the checked out objects would come back to the pool.

This change is to make sure that the rate limiter task always check if there is a waiter in the queue or the pool is under the minimum level.

### DIFF
--- a/r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
+++ b/r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
@@ -21,6 +21,7 @@
 package com.linkedin.r2.transport.http.client;
 
 import com.linkedin.common.stats.LongTracker;
+import com.linkedin.r2.util.SingleTimeout;
 import com.linkedin.util.ArgumentUtil;
 import com.linkedin.util.clock.Clock;
 import com.linkedin.util.clock.SystemClock;
@@ -30,13 +31,11 @@ import java.util.Collections;
 import java.util.Deque;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.Queue;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 
-import com.linkedin.common.stats.LongStats;
 import com.linkedin.common.stats.LongTracking;
 import com.linkedin.r2.SizeLimitExceededException;
 import org.slf4j.Logger;
@@ -64,10 +63,14 @@ public class AsyncPoolImpl<T> implements AsyncPool<T>
   private final int _maxSize;
   private final int _maxWaiters;
   private final long _idleTimeout;
+  private final long _waiterTimeout;
   private final ScheduledExecutorService _timeoutExecutor;
   private final int _minSize;
   private volatile ScheduledFuture<?> _objectTimeoutFuture;
   private final RateLimiter _rateLimiter;
+
+  public static final int MIN_WAITER_TIMEOUT = 300;
+  public static final int MAX_WAITER_TIMEOUT = 3000;
 
   private enum State { NOT_YET_STARTED, RUNNING, SHUTTING_DOWN, STOPPED }
 
@@ -85,11 +88,12 @@ public class AsyncPoolImpl<T> implements AsyncPool<T>
   private final Deque<TimedObject<T>> _idle = new LinkedList<TimedObject<T>>();
   // When no unused objects are available, callbacks live here while they wait
   // for a new object (either returned by another user, or newly created)
-  private final LinkedDeque<Callback<T>> _waiters = new LinkedDeque<Callback<T>>();
+  private final LinkedDeque<Callback<T>> _waiters = new LinkedDeque<>();
   private Throwable _lastCreateError = null;
   private State _state = State.NOT_YET_STARTED;
   private Callback<None> _shutdownCallback = null;
   private final AsyncPoolStatsTracker _statsTracker;
+  private final Clock _clock;
 
   /**
    * Constructs an AsyncPool with maxWaiters equals to ({@code Integer.MAX_VALUE}).
@@ -156,6 +160,23 @@ public class AsyncPoolImpl<T> implements AsyncPool<T>
         maxWaiters, strategy, minSize, rateLimiter, SystemClock.instance(), new LongTracking());
   }
 
+  @Deprecated
+  public AsyncPoolImpl(String name,
+      Lifecycle<T> lifecycle,
+      int maxSize,
+      long idleTimeout,
+      ScheduledExecutorService timeoutExecutor,
+      int maxWaiters,
+      Strategy strategy,
+      int minSize,
+      RateLimiter rateLimiter,
+      Clock clock,
+      LongTracker waitTimeTracker)
+  {
+    this(name, lifecycle, maxSize, idleTimeout, Integer.MAX_VALUE, timeoutExecutor, maxWaiters, strategy, minSize,
+        rateLimiter, clock, waitTimeTracker);
+  }
+
   /**
    * Creates an AsyncPoolImpl with a specified strategy of
    * returning pool objects and a minimum pool size.
@@ -192,6 +213,7 @@ public class AsyncPoolImpl<T> implements AsyncPool<T>
       Lifecycle<T> lifecycle,
       int maxSize,
       long idleTimeout,
+      long waiterTimeout,
       ScheduledExecutorService timeoutExecutor,
       int maxWaiters,
       Strategy strategy,
@@ -209,11 +231,13 @@ public class AsyncPoolImpl<T> implements AsyncPool<T>
     _lifecycle = lifecycle;
     _maxSize = maxSize;
     _idleTimeout = idleTimeout;
+    _waiterTimeout = waiterTimeout;
     _timeoutExecutor = timeoutExecutor;
     _maxWaiters = maxWaiters;
     _strategy = strategy;
     _minSize = minSize;
     _rateLimiter = rateLimiter;
+    _clock = clock;
     _statsTracker = new AsyncPoolStatsTracker(
         () -> _lifecycle.getStats(),
         () -> _maxSize,
@@ -324,7 +348,7 @@ public class AsyncPoolImpl<T> implements AsyncPool<T>
     boolean create = false;
     boolean reject = false;
     final LinkedDeque.Node<Callback<T>> node;
-    final Callback<T> callbackWithTracking = new TimeTrackingCallback<T>(callback);
+    final TimeTrackingCallback<T> callbackWithTracking = new TimeTrackingCallback<T>(callback);
     for (;;)
     {
       TimedObject<T> obj = null;
@@ -347,7 +371,7 @@ public class AsyncPoolImpl<T> implements AsyncPool<T>
             if (_waiters.size() < _maxWaiters)
             {
               // No objects available and the waiter list is not full; add to waiter list and break out of loop
-              node = _waiters.addLastNode(callbackWithTracking);
+              node = _waiters.addLastNode(new WaiterTimeoutCallback(callbackWithTracking));
               create = shouldCreate();
             }
             else
@@ -601,6 +625,27 @@ public class AsyncPoolImpl<T> implements AsyncPool<T>
       @Override
       public void run(final SimpleCallback callback)
       {
+        boolean shouldIgnore;
+        synchronized (_lock) {
+          // Ignore the object creation if no one is waiting for the object and the pool already has _minSize objects
+          int totalObjects = _checkedOut + _idle.size();
+          shouldIgnore = _waiters.size() == 0 && totalObjects >= _minSize;
+          if (shouldIgnore) {
+            _statsTracker.incrementIgnoredCreation();
+            if (_poolSize >= 1)
+            {
+              // _poolSize also include the count of creation requests pending. So we have to make sure the pool size
+              // count is updated when we ignore the creation request.
+              _poolSize--;
+            }
+          }
+        }
+
+        if (shouldIgnore) {
+          callback.onDone();
+          return;
+        }
+
         _lifecycle.create(new Callback<T>()
         {
           @Override
@@ -668,33 +713,33 @@ public class AsyncPoolImpl<T> implements AsyncPool<T>
 
   private void timeoutObjects()
   {
-    Collection<T> idle = reap(_idle, _idleTimeout);
-    if (idle.size() > 0)
+    Collection<T> expiredObjects = getExpiredObjects();
+    if (expiredObjects.size() > 0)
     {
-      LOG.debug("{}: disposing {} objects due to idle timeout", _poolName, idle.size());
-      for (T obj : idle)
+      LOG.debug("{}: disposing {} objects due to idle timeout", _poolName, expiredObjects.size());
+      for (T obj : expiredObjects)
       {
         destroy(obj, false);
       }
     }
   }
 
-  private <U> Collection<U> reap(Queue<TimedObject<U>> queue, long timeout)
+  private Collection<T> getExpiredObjects()
   {
-    List<U> toReap = new ArrayList<U>();
-    long now = System.currentTimeMillis();
-    long target = now - timeout;
+    List<T> expiredObjects = new ArrayList<T>();
+    long now = _clock.currentTimeMillis();
 
     synchronized (_lock)
     {
+      long deadline = now - _idleTimeout;
       int excess = _poolSize - _minSize;
-      for (TimedObject<U> p; (p = queue.peek()) != null && p.getTime() < target && excess > 0; excess--)
+      for (TimedObject<T> p; (p = _idle.peek()) != null && p.getTime() < deadline && excess > 0; excess--)
       {
-        toReap.add(queue.poll().get());
+        expiredObjects.add(_idle.poll().get());
         _statsTracker.incrementTimedOut();
       }
     }
-    return toReap;
+    return expiredObjects;
   }
 
   private void shutdownIfNeeded()
@@ -749,7 +794,7 @@ public class AsyncPoolImpl<T> implements AsyncPool<T>
     shutdown.onSuccess(None.none());
   }
 
-  private static class TimedObject<T>
+  private class TimedObject<T>
   {
     private final T _obj;
     private final long _time;
@@ -757,7 +802,7 @@ public class AsyncPoolImpl<T> implements AsyncPool<T>
     public TimedObject(T obj)
     {
       _obj = obj;
-      _time = System.currentTimeMillis();
+      _time = _clock.currentTimeMillis();
     }
 
     public T get()
@@ -771,6 +816,47 @@ public class AsyncPoolImpl<T> implements AsyncPool<T>
     }
   }
 
+  private class WaiterTimeoutCallback implements Callback<T>
+  {
+    private final SingleTimeout<Callback<T>> _timeout;
+
+    private WaiterTimeoutCallback(final Callback<T> callback)
+    {
+      _timeout = new SingleTimeout<>(_timeoutExecutor, _waiterTimeout, TimeUnit.MILLISECONDS, callback, () -> {
+
+        synchronized (_lock)
+        {
+          _waiters.remove(this);
+          _statsTracker.incrementWaiterTimedOut();
+        }
+        LOG.debug("{}: failing waiter due to waiter timeout", _poolName);
+        callback.onError(
+            new WaiterTimeoutException(
+                "Exceeded waiter timeout of " + _waiterTimeout + "ms: in Pool: "+ _poolName));
+      });
+    }
+
+    @Override
+    public void onError(Throwable e)
+    {
+      Callback<T> callback = _timeout.getItem();
+      if (callback != null)
+      {
+        callback.onError(e);
+      }
+    }
+
+    @Override
+    public void onSuccess(T result)
+    {
+      Callback<T> callback = _timeout.getItem();
+      if (callback != null)
+      {
+        callback.onSuccess(result);
+      }
+    }
+  }
+
   private class TimeTrackingCallback<T> implements Callback<T>
   {
     private final long _startTime;
@@ -779,13 +865,13 @@ public class AsyncPoolImpl<T> implements AsyncPool<T>
     public TimeTrackingCallback(Callback<T> callback)
     {
       _callback = callback;
-      _startTime = System.currentTimeMillis();
+      _startTime = _clock.currentTimeMillis();
     }
 
     @Override
     public void onError(Throwable e)
     {
-      long waitTime = System.currentTimeMillis() - _startTime;
+      long waitTime = _clock.currentTimeMillis() - _startTime;
       synchronized (_lock)
       {
         _statsTracker.trackWaitTime(waitTime);
@@ -797,13 +883,18 @@ public class AsyncPoolImpl<T> implements AsyncPool<T>
     @Override
     public void onSuccess(T result)
     {
-      long waitTime = System.currentTimeMillis() - _startTime;
+      long waitTime = _clock.currentTimeMillis() - _startTime;
       synchronized (_lock)
       {
         _statsTracker.trackWaitTime(waitTime);
         _statsTracker.sampleMaxWaitTime(waitTime);
       }
       _callback.onSuccess(result);
+    }
+
+    public long getTime()
+    {
+      return _startTime;
     }
   }
 
--- a/r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolStats.java
+++ b/r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolStats.java
@@ -36,6 +36,8 @@ public class AsyncPoolStats implements PoolStats
   private final int _totalDestroyErrors;
   private final int _totalBadDestroyed;
   private final int _totalTimedOut;
+  private final int _totalWaiterTimedOut;
+  private final int _totalCreationIgnored;
 
   private final int _checkedOut;
   private final int _maxPoolSize;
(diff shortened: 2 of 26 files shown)


r2-netty/src/main/java/com/linkedin/r2/transport/http/client/common/ChannelPoolLifecycle.java
```diff
@@ -86,33 +92,56 @@ public ChannelPoolLifecycle(SocketAddress address, Bootstrap bootstrap, ChannelG
...
-        {
-          c.config().setOption(ChannelOption.TCP_NODELAY, true);
-        }
-        _channelGroup.add(c);
         channelCallback.onSuccess(c);
+        return;
       }
-      else
+
+      c.pipeline().addAfter(SslHandlerUtil.PIPELINE_SSL_HANDLER, CHANNELPOOL_CALLBACK_HANDLER, new ChannelDuplexHandler()
```
FreCap: CHANNELPOOL_SSL_CALLBACK_HANDLER?

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
```diff
@@ -255,13 +275,15 @@ public void start()
         throw new IllegalStateException(_poolName + " is " + _state);
       }
       _state = State.RUNNING;
-      if (_idleTimeout > 0)
+      long timeOut = Math.min(_idleTimeout, _waitTimeout);
+      if (timeOut > 0)
       {
-        long freq = Math.min(_idleTimeout / 10, 1000);
+        long freq = Math.min(timeOut / 10, 1000);
```
FreCap: waittimeout can be very low, let's add a min of 20-50 ms

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
```diff
@@ -601,6 +623,25 @@ private void create()
...
+        boolean shouldIgnore;
+        synchronized (_lock) {
+          // Ignore the object creation if no one is waiting for the object and the pool already has _minSize objects
+          int totalObjects = _checkedOut + _idle.size();
+          shouldIgnore = _waiters.size() == 0 && totalObjects >= _minSize;
+          if (shouldIgnore) {
+            _statsTracker.incrementIgnoredCreation();
+            if (_poolSize >= 1)
+            {
+              _poolSize--;
```
FreCap: The correctness of this line is not immediately clear and we might want to clarify that it is tightly tied to the shouldCreate methd

FreCap: also I'm not super clear why this would go below 1, at first sight it seem that it would be kind of a bug if ti went below.
I'd be curious if it ever happens

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
```diff
@@ -697,6 +751,23 @@ private void timeoutObjects()
...
+    List<TimeTrackingCallback<U>> toReap = new ArrayList<>();
+    long now = _clock.currentTimeMillis();
+    long target = now - timeout;
+
+    synchronized (_lock)
+    {
+      for (TimeTrackingCallback<U> p; (p = queue.peek()) != null && p.getTime() < target;)
+      {
+        toReap.add(queue.poll());
+        _statsTracker.incrementWaiterTimedOut();
```
FreCap: We can move the increment on line 731.
`        obj.onError(new WaiterTimeoutException("Waiter timeout after " + _waitTimeout + "ms"));`

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
```diff
@@ -797,14 +868,24 @@ public void onError(Throwable e)
...
         _statsTracker.trackWaitTime(waitTime);
         _statsTracker.sampleMaxWaitTime(waitTime);
       }
       _callback.onSuccess(result);
     }
+
+    public Callback<T> get()
+    {
+      return _callback;
+    }
```
FreCap: Is this method unused?

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolStats.java
```diff
@@ -64,6 +66,8 @@ public AsyncPoolStats(
       int totalDestroyErrors,
       int totalBadDestroyed,
       int totalTimedOut,
+      int totalWaiterTimedOut,
+      int totalCreationIgnored,
```
FreCap: totalCreationsIgnored
or to be even more explicit
totalCreationsCancelledNoWaiter

r2-core/src/main/java/com/linkedin/r2/transport/http/client/RateLimiter.java
```diff
@@ -57,6 +57,11 @@
    */
   Collection<Task> cancelPendingTasks();
 
+  default int numberOfPendingTasks()
+  {
+    return -1;
+  }
+
```
FreCap: There is not explicit need for having this other than testing. Our usual stance is to add methods to interfaces only if really needed. Adding it in the interface means extra support in the future for any ratelimiter.

I'd rather remove this and make the method @VisibleForTesting

r2-core/src/main/java/com/linkedin/r2/transport/http/client/ExponentialBackOffRateLimiter.java
```diff
@@ -173,24 +173,10 @@ public void incrementPeriod()
...
-        _pending.add(t);
-        schedule();
-      }
-    }
-
-    if (runNow)
-    {
-      t.run(_doneCallback);
+      _pending.add(t);
+      schedule();
```
FreCap: Is this change actually part of your RB?

r2-core/src/main/java/com/linkedin/r2/transport/http/client/ExponentialBackOffRateLimiter.java
```diff
@@ -208,6 +194,12 @@ public void submit(Task t)
     }
   }
 
+  @Override
+  public int numberOfPendingTasks()
```
FreCap: @VisibleForTesting

r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
```diff
@@ -534,7 +544,7 @@ public void testGetStatsWithErrors() throws Exception
     Assert.assertEquals(stats.getCheckedOut(), GET - PUT_BAD - DISPOSE);
     // When the each create fails, it will retry and cancel the waiter,
     // resulting in a second create error.
```
FreCap: This comment is not updated anymore

r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
```diff
@@ -570,6 +580,389 @@ public void testWaitTimeStats() throws Exception
...
+        Integer.MAX_VALUE,
+        _executor,
+        Integer.MAX_VALUE,
+        AsyncPoolImpl.Strategy.MRU,
+        0, rateLimiter
+    );
+
+    pool.start();
+
+    // Phase A : Checking out object 'numberOfCheckout' times !
```
FreCap: just formatting, no need of space before ! or :

r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
```diff
@@ -570,6 +580,389 @@ public void testWaitTimeStats() throws Exception
...
+    // Phase D : All the object creation in phase B gets unblocked now
+    blockableObjectCreator.unblockCreation();
+    try
+    {
+      // Wait for all object creation to be unblocked
+      future.get(5, TimeUnit.SECONDS);
+    }
+    catch (Exception e)
+    {
+      Assert.fail("Unexpected interruption", e);
```
FreCap: Could we add in what way it is unexpected?

r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
```diff
@@ -570,6 +580,389 @@ public void testWaitTimeStats() throws Exception
...
+    }
+
+    // Making sure the rate limiter pending tasks are submitted to the executor
+    while (rateLimiter.numberOfPendingTasks() > 0)
+    {
+      try
+      {
+        Thread.sleep(1);
+      }
+      catch (Exception e)
```
FreCap: Here no need of catching all exceptions, we can just specify the ones that Thread.sleep returns

r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
```diff
@@ -570,6 +580,389 @@ public void testWaitTimeStats() throws Exception
...
+      int poolSize, int concurrency, int waiterTimeout)
+  {
+    CreationBlockableSynchronousLifecycle blockableObjectCreator =
+        new CreationBlockableSynchronousLifecycle(numberOfCheckoutsInPhaseB, concurrency);
+    ScheduledExecutorService executor = Executors.newScheduledThreadPool(500);
+    RateLimiter rateLimiter = new ExponentialBackOffRateLimiter(0, 5000,
+        10, executor, concurrency);
+
+    ClockedExecutor clockedExecutor = new ClockedExecutor();
+    SettableClock settableClock = new SettableClock();
```
FreCap: ClockedExecutor should extend Clock, so no need to have two

r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
```diff
@@ -570,6 +580,389 @@ public void testWaitTimeStats() throws Exception
...
+    settableClock.addDuration(waiterTimeout+1);
+    clockedExecutor.runFor(waiterTimeout+2);
+
+    // Phase D : All the object creation in phase B gets unblocked now
+    blockableObjectCreator.unblockCreation();
+    try
+    {
+      future.get(5, TimeUnit.SECONDS);
+    }
+    catch (Exception e)
```
FreCap: Expand exceptions everywhere :)

r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
```diff
@@ -570,6 +580,389 @@ public void testWaitTimeStats() throws Exception
...
+    {
+      try
+      {
+        Thread.sleep(1);
+      }
+      catch (Exception e)
+      {
+        Assert.fail("Unexpected interruption", e);
+      }
+    }
```
FreCap: use AssertWithTimeout instead to avoid locking situations, here and in the other places with sleep

r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
```diff
@@ -570,6 +580,389 @@ public void testWaitTimeStats() throws Exception
...
+    Assert.assertEquals(stats.getTotalCreationIgnored(), numberOfCheckoutsInPhaseB - concurrency);
+    Assert.assertEquals(stats.getCheckedOut(), numberOfCheckoutsInPhaseA);
+    Assert.assertEquals(stats.getIdleCount(), concurrency);
+    Assert.assertEquals(stats.getTotalCreated(), numberOfCheckoutsInPhaseA + concurrency);
+    Assert.assertEquals(stats.getPoolSize(), numberOfCheckoutsInPhaseA + concurrency);
+    Assert.assertEquals(stats.getTotalWaiterTimedOut(), numberOfCheckoutsInPhaseB - numbOfObjectsToBeReturnedInPhaseC);
+  }
+
+  @DataProvider
+  public Object[][] dataProvider()
```
FreCap: You can specify the name of the dataProvider including what kind of data it'll return

r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
```diff
@@ -570,6 +580,389 @@ public void testWaitTimeStats() throws Exception
...
+  {
+    // 500 represent a good sample for the randomized data.
+    // This has been verified against 100K test cases in local
+    int numberOfTestCases = 500;
+    Random randomNumberGenerator = new Random();
```
FreCap: ThreadLocalRandom.current()

r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
```diff
@@ -570,6 +580,389 @@ public void testWaitTimeStats() throws Exception
...
+        try
+        {
+          Object checkedOutObject = cb.get();
+          synchronized (sync)
+          {
+            checkedOutObjects.add(checkedOutObject);
+          }
+
+          latch.countDown();
+
```
FreCap: extra lines

r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
```diff
@@ -622,6 +1015,60 @@ public int getLive()
...
+    public void waitUntilAllBlocked()
+    {
+      try
+      {
+        _blockersDoneLatch.await();
+      }
+      catch (Exception ex)
+      {
+
```
FreCap: missing exception. either log, rethrow runtime or rename exception ignored. 
Expand Exception

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
FreCap: This tightly coupled condition is very dangerous being so distant from the origin of the update

r2-netty/src/main/java/com/linkedin/r2/netty/handler/common/SessionResumptionSslHandler.java
```diff
@@ -46,14 +50,19 @@
 public class SessionResumptionSslHandler extends ChannelOutboundHandlerAdapter
 {
   public static final String PIPELINE_SESSION_RESUMPTION_HANDLER = "SessionResumptionSslHandler";
+  public static final AttributeKey<SessionResumptionSslHandler> CHANNEL_SESSION_RESUMPTION_HANDLER =
+      AttributeKey.valueOf("sslSessionResumptionHandler");
 
   private final SslHandlerGenerator _hostPortToSslHandler;
+  private Future<Channel> _handshakeFuture;
```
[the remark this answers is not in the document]
nizarm: Good catch. I have removed other related changes, but missed removing these member variable. Removed it now.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
nizarm: Sure,  20-50 ms will be too aggressive IMHO. But yea - this matters only when someone override the default.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
nizarm: Added a comment to make it clear.  If you closely watch the creation onError code path - this is exactly done there as well.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
nizarm: The reason that is not done is to avoid acquiring the lock again just for updating the status (similar to timeoutObjects method). Since any slight increase in wait in this class is in the critical path of the request - we should avoid that. Let me know.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
nizarm: Good catch

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolStats.java
nizarm: I will vote for totalCreationsIgnored

r2-core/src/main/java/com/linkedin/r2/transport/http/client/RateLimiter.java
nizarm: Sure, we do not have VisibleForTesting anywhere in pegasus. Need to have additional external dependency  for that. To make things simpler I will have this method part of ExponentialRateLimitor alone and not part of the interface.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/ExponentialBackOffRateLimiter.java
nizarm: Yes,  to make it more asynchronous and the task submission always happens in the executor thread.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/ExponentialBackOffRateLimiter.java
nizarm: See my comments above for this.

r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
nizarm: Good Catch. removed non relevant comments. Thanks

r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
nizarm: Done

r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
nizarm: The settableClock is needed here to simulate the time.  or Am I missing something ?

r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
nizarm: Thanks.

r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
nizarm: Done. Curios, What is the difference between Random() and ThreadLocalRandom.current() ?

r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
FreCap: clockedExecutor should do the same, no? if you replace all settableClock with the clockedExecutor, should work the same

r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
FreCap: I think here is not really important, but it was just to avoid using Random in general which if used in prod code brings slowness compared to ThreadLocal one. (still here we are not using it in prod, and both ways are good)

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
```diff
@@ -278,7 +279,7 @@ public void start()
       long timeOut = Math.min(_idleTimeout, _waitTimeout);
       if (timeOut > 0)
       {
-        long freq = Math.min(timeOut / 10, 1000);
+        long freq = Math.min(timeOut / 10, 50);
```
FreCap: What I meant here was to set an upper and lower bound like in   Math.max(50,Math.min(timeOut / 10, 1000));

r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
```diff
@@ -647,34 +648,35 @@ public void testObjectsAreNotCreatedWhenThereAreNoWaiters(int numberOfCheckouts,
...
-      {
-        Thread.sleep(1);
-      }
-      catch (Exception e)
+    AssertionMethods.assertWithTimeout(5000, () -> {
+      while (rateLimiter.numberOfPendingTasks() > 0)
       {
-        Assert.fail("Unexpected interruption", e);
-      }
-    }
```
FreCap: AssertionMethods.assertWithTimeout(5000, () ->       Assert.assertEquals(rateLimiter.numberOfPendingTasks(),0,"Number of tasks has to drop to 0")

FreCap: Same for others

nizarm: Neat !

r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
nizarm: Updated, had to make a small adjustment to the until time to consider the pool thread interval.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
```diff
@@ -209,11 +231,13 @@ public AsyncPoolImpl(String name,
     _lifecycle = lifecycle;
     _maxSize = maxSize;
     _idleTimeout = idleTimeout;
+    _waitTimeout = waiterTimeout;
```
ssheng: Please name the parameter and class variable consistently. _waitTimeout vs. waiterTimeout.

r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
```diff
@@ -647,34 +648,25 @@ public void testObjectsAreNotCreatedWhenThereAreNoWaiters(int numberOfCheckouts,
...
-      {
-        Thread.sleep(1);
-      }
-      catch (Exception e)
-      {
-        Assert.fail("Unexpected interruption", e);
-      }
-    }
+    AssertionMethods.assertWithTimeout(5000, () ->
+            Assert.assertEquals(rateLimiter.numberOfPendingTasks(),0,"Number of tasks has to drop to 0"));
```
FreCap: reformat

r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
```diff
@@ -755,8 +746,7 @@ public void testWaiterTimeout(int numberOfCheckoutsInPhaseA, int numberOfCheckou
       pool.put(checkedOutObjects.remove(0));
     }
 
-    settableClock.addDuration(waiterTimeout+1);
-    clockedExecutor.runFor(waiterTimeout+2);
+    clockedExecutor.runFor(waiterTimeout + AsyncPoolImpl.MAX_WAITER_TIMEOUT);
```
FreCap: I'd be ok also with copy pasting the formula ;)  Math.max(MIN_WAITER_TIMEOUT, Math.min(timeOut / 10, MAX_WAITER_TIMEOUT));

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
```diff
@@ -697,6 +757,23 @@ private void timeoutObjects()
     return toReap;
   }
 
+  private <U> Collection<TimeTrackingCallback<U>> getExpiredWaiters(Queue<TimeTrackingCallback<U>> queue, long timeout)
```
ssheng: The waiters queue is already a class variable and doesn't need to be passed around, correct?

nizarm:
> I think after next iteration I've no more comments.
> Thank you for the change! It'll be instrumental to backpressure connections in highly distributed systems in the https era ;)

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
nizarm: Based this method on existing reap. I have refactored that one as well.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
```diff
@@ -255,13 +278,15 @@ public void start()
         throw new IllegalStateException(_poolName + " is " + _state);
       }
       _state = State.RUNNING;
-      if (_idleTimeout > 0)
+      long timeOut = Math.min(_idleTimeout, _waiterTimeout);
+      if (timeOut > 0)
       {
-        long freq = Math.min(_idleTimeout / 10, 1000);
+        long freq = Math.max(MIN_WAITER_TIMEOUT, Math.min(timeOut / 10, MAX_WAITER_TIMEOUT));
```
FreCap: Thinking more about this, instead of modifying the min timeout (which could have an impact on servers with thousand of AsyncPools when set to a low value), we could run the expire logic per request.

This would guarantee us that we are doing the checking work only when actually needed and not keep checking every 100-300ms x 5000 Async Pools which can be a lot of load on the servers if we don't actually have any new connection.

r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
FreCap: I don't see the exception expanded to TimeoutException here

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
nizarm: Thanks for the offline discussion. This absolutely make sense. I have pushed the changes as discussed.

PR #186, opened 2020-02-24
Fix prepare-release script so that tag is deleted on failed push

evanw555:
When the `prepare-release` script fails to push the locally created
release tag, it should delete the tag locally so as not to confuse
the developer.

### DIFF
--- a/prepare-release
+++ b/prepare-release
@@ -44,3 +44,10 @@ fi
 # Push release tag
 echo "Pushing tag $TAG_NAME..."
 git push $REMOTE $TAG_NAME
+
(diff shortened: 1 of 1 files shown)


prepare-release
```diff
@@ -44,3 +44,10 @@ fi
 # Push release tag
 echo "Pushing tag $TAG_NAME..."
 git push $REMOTE $TAG_NAME
+
+if [ $? -ne 0 ]
+then
+  echo 'Push failed, tag will be locally deleted.'
```
karthikbalasub: tag will be locally deleted -> clearing tag from the local repo

PR #187, opened 2020-02-25
Fix issues when encoding/parsing custom properties in PDL.

karthikbalasub:
Allow PDL parser and encoder to support property keys with special characters.
Also sort properties by key before encoding them to ensure ordering consistency.

### DIFF
--- a/data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
+++ b/data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
@@ -29,6 +29,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import org.apache.commons.lang3.StringEscapeUtils;
 
@@ -51,6 +52,7 @@ abstract class PdlBuilder
       "record", "typeref", "union", "null", "true", "false"
   ));
   private static final char ESCAPE_CHAR = '`';
+  private static final Pattern NON_IDENTIFIER_CHARS = Pattern.compile(".*[^0-9a-zA-Z_-].*");
 
   /**
    * Each subclass must define a provider for creating new instances.
@@ -127,14 +129,17 @@ abstract class PdlBuilder
   abstract boolean writeDoc(String doc) throws IOException;
 
   /**
-   * Writes a set of schema properties that share a common prefix to .pdl.
+   * Writes a set of schema properties that share a common prefix to .pdl. Sorts the properties by key before writing.
    *
    * @param prefix provides the common prefix of all the properties.
    * @param properties provides the properties to write.
    */
   PdlBuilder writeProperties(List<String> prefix, Map<String, Object> properties) throws IOException
   {
-    for (Map.Entry<String, Object> entry : properties.entrySet())
+    for (Map.Entry<String, Object> entry :
+        properties.entrySet().stream()
+            .sorted(Map.Entry.comparingByKey())
+            .collect(Collectors.toList()))
     {
       String key = entry.getKey();
       Object value = entry.getValue();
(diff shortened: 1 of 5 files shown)


data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
```diff
@@ -127,14 +129,17 @@ PdlBuilder writeSpace() throws IOException
...
+   * Writes a set of schema properties that share a common prefix to .pdl. Sorts the properties by key before writing.
    *
    * @param prefix provides the common prefix of all the properties.
    * @param properties provides the properties to write.
    */
   PdlBuilder writeProperties(List<String> prefix, Map<String, Object> properties) throws IOException
   {
-    for (Map.Entry<String, Object> entry : properties.entrySet())
+    for (Map.Entry<String, Object> entry :
+        properties.entrySet().stream()
```
evanw555: I'd avoid packing this stream builder chain into the for-loop statement. Either construct the list on its own line, or use `Stream#forEach`

generator-test/src/test/pegasus/com/linkedin/pegasus/generator/test/idl/escaping/PropertyKeyEscaping.pdl
```diff
@@ -2,6 +2,7 @@ namespace com.linkedin.pegasus.generator.test.idl.escaping
 
 record PropertyKeyEscaping {
   @`namespace` = "foo.bar"
+  @path.`/*.*/`
```
evanw555: I'd suggest buffing this test case up a little bit to test the following things:
1. All types of comments (e.g. single-line, javadoc):
```
@`//`.`/**`.`*/`.`/*.*/`
```
2. Test the encoder pattern you just added (`Pattern.compile(".*[^0-9a-zA-Z_-].*")`):
```
@shouldNotBeEscaped.abc_XYZ-123
```

data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
```diff
@@ -51,6 +52,7 @@
       "record", "typeref", "union", "null", "true", "false"
   ));
   private static final char ESCAPE_CHAR = '`';
+  private static final Pattern NON_IDENTIFIER_CHARS = Pattern.compile(".*[^0-9a-zA-Z_-].*");
```
evanw555: Tbh this is kind of confusing, it took me a minute to realize why this also contains `.*`...

Perhaps it would make more sense to use the inverse:
```
IDENTIFIER_CHARS = Pattern.compile("[0-9a-zA-Z_-]+")
```
Then do...
```
!IDENTIFIER_CHARS.matcher(propertyKey).matches()
```

data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
karthikbalasub: fixed

generator-test/src/test/pegasus/com/linkedin/pegasus/generator/test/idl/escaping/PropertyKeyEscaping.pdl
karthikbalasub: Done, improved the grammar to handle non-identifier but should not be escaped property segments

data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
karthikbalasub: done

data/src/test/java/com/linkedin/data/schema/grammar/TestPdlSchemaParser.java
```diff
@@ -85,6 +87,14 @@ public void testParseNestedProperties() throws IOException
     expected.put("validate", validate);
     expected.put("com.linkedin.CustomValidator", "abc");
 
+    DataMap propertyWithPath = new DataMap();
+    DataMap propertyWithSpecialChars = new DataMap();
+    propertyWithPath.put("/*", propertyWithSpecialChars);
+    propertyWithSpecialChars.put("*/.$", false);
+    expected.put("pathProp", propertyWithPath);
+
```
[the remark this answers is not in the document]
karthikbalasub: The extra lines are to separate the different cases, for better reading

PR #192, opened 2020-02-26
Fix for PDL schema cyclic referencing detection in "includes" and "typeref" 

junchuanwang:
Right now PDL parser wouldn't detect cyclic referencing for "includes" and "typeref"

This change will fix this issue so we have feature parity between PDL and PDSC on cyclic referencing detection.

### DIFF
--- a/data/src/main/java/com/linkedin/data/schema/grammar/PdlSchemaParser.java
+++ b/data/src/main/java/com/linkedin/data/schema/grammar/PdlSchemaParser.java
@@ -578,6 +578,7 @@ public class PdlSchemaParser extends AbstractSchemaParser
       setDocAndProperties(context, schema);
       bindNameToSchema(name, schema.getAliases(), schema);
       DataSchema refSchema = toDataSchema(typeref.ref);
+      checkTyperefCycle(schema, refSchema);
       schema.setReferencedType(refSchema);
       schema.setRefDeclaredInline(isDeclaredInline(typeref.ref));
     }
@@ -1093,6 +1094,10 @@ public class PdlSchemaParser extends AbstractSchemaParser
     if (schema == null)
     {
       schema = getResolver().findDataSchema(fullName, errorMessageBuilder());
+      if (schema != null)
+      {
+        checkForCycleWithInclude(((NamedDataSchema) schema).getFullName());
+      }
     }
     return schema;
   }
--- a/data/src/main/java/com/linkedin/data/schema/grammar/PdlSchemaParser.java
+++ b/data/src/main/java/com/linkedin/data/schema/grammar/PdlSchemaParser.java
@@ -1082,26 +1082,6 @@ public class PdlSchemaParser extends AbstractSchemaParser
     }
   }
 
-  /**
-   * Look for {@link DataSchema} with the specified name.
-   *
-   * @param fullName to lookup.
-   * @return the {@link DataSchema} if lookup was successful else return null.
-   */
-  public DataSchema lookupName(String fullName)
-  {
-    DataSchema schema = DataSchemaUtil.typeStringToPrimitiveDataSchema(fullName);
-    if (schema == null)
-    {
-      schema = getResolver().findDataSchema(fullName, errorMessageBuilder());
-      if (schema != null)
-      {
-        checkForCycleWithInclude(((NamedDataSchema) schema).getFullName());
-      }
-    }
-    return schema;
-  }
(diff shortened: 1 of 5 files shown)


data/src/test/java/com/linkedin/data/schema/resolver/TestDataSchemaResolver.java
```diff
@@ -460,22 +697,36 @@ public void testMapDataSchemaResolver()
...
 
     for (String[] testLookupAndExpectedResult : testLookupAndExpectedResults)
     {
-      DataSchemaResolver resolver = new MapDataSchemaResolver(
-          SchemaParserFactory.instance(),
-          Arrays.asList(buildSystemIndependentPath("a1")),
-          ".pdsc", testSchemas);
-      lookup(resolver, new String[][] { testLookupAndExpectedResult}, File.separatorChar, debug);
+      DataSchemaResolver schemaResolver = new MapDataSchemaResolver(
+          extension.equals("pdsc") ? SchemaParserFactory.instance() : PdlSchemaParserFactory.instance(),
```
karthikbalasub: Can we use MultiFormatDataSchemaResolver instead?

data/src/test/java/com/linkedin/data/schema/resolver/TestDataSchemaResolver.java
```diff
@@ -236,6 +238,7 @@ public void testMapDataSchemaResolver()
     {
         {
           "Two records including each other",
+            "pdsc",
```
evanw555: Use `com.linkedin.data.schema.SchemaFormatType` here rather than a String. It's more type-safe.

This also gives you easy access to the parser factory instance.

data/src/main/java/com/linkedin/data/schema/grammar/PdlSchemaParser.java
```diff
@@ -1093,6 +1094,10 @@ public DataSchema lookupName(String fullName)
     if (schema == null)
     {
       schema = getResolver().findDataSchema(fullName, errorMessageBuilder());
+      if (schema != null)
```
evanw555: This method is now completely identical to `AbstractSchemaParser#lookupName`. Why don't we just delete this one and use that?

data/src/test/java/com/linkedin/data/schema/resolver/TestDataSchemaResolver.java
junchuanwang: MapDataSchemaResolver is written for this test cases so that is why I want to re-use them

data/src/main/java/com/linkedin/data/schema/grammar/PdlSchemaParser.java
junchuanwang: good idea

data/src/test/java/com/linkedin/data/schema/resolver/TestDataSchemaResolver.java
```diff
@@ -460,22 +700,36 @@ public void testMapDataSchemaResolver()
...
 
   @Test(dataProvider = "circularReferenceData")
-  public void testCircularReferences(String desc, Map<String, String> testSchemas, String[][] testLookupAndExpectedResults)
+  public void testCircularReferences(String desc, SchemaFormatType extension, Map<String, String> testSchemas, String[][] testLookupAndExpectedResults)
   {
     boolean debug = false;
 
     for (String[] testLookupAndExpectedResult : testLookupAndExpectedResults)
     {
-      DataSchemaResolver resolver = new MapDataSchemaResolver(
```
evanw555: As I mentioned in my previous comment, using the enum gives you easy access to the schema parser instance. You shouldn't be doing `extension.equals(PDSC) ? SchemaParserFactory.instance() : PdlSchemaParserFactory.instance()`. You should be doing `extension.getSchemaParserFactory()`

evanw555: You could potentially also avoid having to pass in the schema parser factory into the `MapDataSchemaResolver` itself, you could just pass in only the extension, from which you could derive the parser factory and the string extension.

junchuanwang: I don't see how you second suggestion is significantly better, but I did the change according to your suggestion.

PR #193, opened 2020-02-26
Use pathspec object as hash map key in ResolvedPropertiesReaderVisitor

junchuanwang:
Currently ResolvedPropertiesReaderVisitor Use PathSpec-generated string as key in hash map, it uses up a lot of memory. 

This change is an optimization to use the PathSpec object in the hasp so it will reduce memory footprint.

### DIFF
--- a/data/src/main/java/com/linkedin/data/schema/PathSpec.java
+++ b/data/src/main/java/com/linkedin/data/schema/PathSpec.java
@@ -20,6 +20,8 @@
 
 package com.linkedin.data.schema;
 
+import java.util.AbstractCollection;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -82,6 +84,17 @@ public class PathSpec
   public PathSpec(String... segments)
   {
     _path = new ArrayList<String>(Arrays.asList(segments));
+  }
+
+  /**
+   * Construct a new {@link PathSpec} from {@link AbstractCollection} type
+   *
+   * @param pathSpecCollection the collection that contains path segments.
+   */
+  public PathSpec(AbstractCollection<String> pathSpecCollection)
+  {
+    _path = new ArrayList<>();
+    _path.addAll(pathSpecCollection);
   }
 
   /**
--- a/data/src/main/java/com/linkedin/data/schema/PathSpec.java
+++ b/data/src/main/java/com/linkedin/data/schema/PathSpec.java
@@ -20,10 +20,9 @@
 
 package com.linkedin.data.schema;
 
-import java.util.AbstractCollection;
-import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -87,13 +86,13 @@ public class PathSpec
   }
 
   /**
-   * Construct a new {@link PathSpec} from {@link AbstractCollection} type
+   * Construct a new {@link PathSpec} from {@link java.util.Collection} type
    *
    * @param pathSpecCollection the collection that contains path segments.
    */
-  public PathSpec(AbstractCollection<String> pathSpecCollection)
+  public PathSpec(Collection<String> pathSpecCollection)
   {
-    _path = new ArrayList<>();
+    _path = new ArrayList<>(pathSpecCollection.size());
     _path.addAll(pathSpecCollection);
   }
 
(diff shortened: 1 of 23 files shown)


data/src/main/java/com/linkedin/data/schema/PathSpec.java
```diff
@@ -84,6 +86,17 @@ public PathSpec(String... segments)
     _path = new ArrayList<String>(Arrays.asList(segments));
   }
 
+  /**
+   * Construct a new {@link PathSpec} from {@link AbstractCollection} type
+   *
+   * @param pathSpecCollection the collection that contains path segments.
+   */
+  public PathSpec(AbstractCollection<String> pathSpecCollection)
```
karthikbalasub: Just use Collection

data/src/main/java/com/linkedin/data/schema/PathSpec.java
```diff
@@ -84,6 +86,17 @@ public PathSpec(String... segments)
...
+  {
+    _path = new ArrayList<>();
```
karthikbalasub: initialize size using Collection::size()

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/ValidateSchemaAnnotationTask.java
```diff
@@ -198,10 +198,16 @@ private void checkHandlerAnnotation(String name, List<String> foundClasses) thro
...
-      } catch (ClassNotFoundException e)
+      }
+      catch (ClassNotFoundException e)
       {
         throw new IOException("Failed to load class while scanning classes", e);
       }
+      catch (Exception | Error e)
+      {
+        getProject().getLogger()
+                    .info("Unexpected exceptions or errors found during scanning class, ok to skip", e);
```
karthikbalasub: Why is it okay to skip?

junchuanwang: Sorry I found I made a mistake regarding the content to publish. Let me create a separate view for above change.

I want this review to only focus on PathSpec key type change

data/src/test/java/com/linkedin/data/schema/annotation/TestSchemaAnnotationProcessor.java
```diff
@@ -344,7 +345,7 @@ public void testDenormalizedSchemaProcessing(String filePath, List<List<String>>
     ResolvedPropertiesReaderVisitor resolvedPropertiesReaderVisitor = new ResolvedPropertiesReaderVisitor();
     DataSchemaRichContextTraverser traverser = new DataSchemaRichContextTraverser(resolvedPropertiesReaderVisitor);
     traverser.traverse(result.getResultSchema());
-    Map<String, Map<String, Object>> pathSpecToResolvedPropertiesMap = resolvedPropertiesReaderVisitor.getLeafFieldsPathSpecToResolvedPropertiesMap();
+    Map<PathSpec, Map<String, Object>> pathSpecToResolvedPropertiesMap = resolvedPropertiesReaderVisitor.getLeafFieldsPathSpecToResolvedPropertiesMap();
```
mchen07: Only change in test?

evanw555:
> This has breaking changes. Are you going to do a major version bump? If yes, check with the team before doing it.

This is such a small change for a major version bump. I'd suggest just adding the old method back, deprecating it, and "reconstructing" the same return value (I'm assuming the map keyed by strings is deterministic given a map keyed by PathSpecs).

data/src/main/java/com/linkedin/data/schema/annotation/ResolvedPropertiesReaderVisitor.java
```diff
@@ -122,8 +122,26 @@ public void callbackOnContext(DataSchemaRichContextTraverser.TraverserContext co
...
   {
-    return _leafFieldsPathSpecToResolvedPropertiesMap;
+    return _leafFieldsPathSpecObjToResolvedPropertiesMap.entrySet()
+                                                        .stream()
+                                                        .collect(Collectors.toMap(e -> e.getKey().toString(),
+                                                                                  Map.Entry::getValue));
+
+  }
+
+  public Map<PathSpec, Map<String, Object>> getLeafFieldsPathSpecObjToResolvedPropertiesMap()
```
evanw555: I understand that you need to rename the method to avoid conflicts, but this is very wordy. I'd suggest cutting it down. How about `getLeafFieldToResolvedPropertiesMap`? You don't need to say "pathSpec" because it's implied based on the return type.

PR #196, opened 2020-02-27
Update documentation for PDL schema

karthikbalasub:


pdl_schema.md
```diff
@@ -0,0 +1,1105 @@
...
+```
+Deprecate a record:
+```pdl
+namespace com.example.models
+
+@deprecated = "Use Person type instead."
+record User {
+  firstName: string
+}
+```
```
nickibi: Should we provide an example of deprecation which doesn't have any string value?

pdl_schema.md
```diff
@@ -0,0 +1,1105 @@
...
+ * `firstName` is a primitive field of type `string`.
+ * `birthDay` references the type `com.example.time.Date` from the first example
+* PDL supports Java like `import` feature that allows you to define references
+ to external types and then use them in the schema using their simple name.
+
+### Record Field attributes
+Pegasus supports additional features for defining the behavior of fields in a
+record.
+#### Record fields can be defined as optional.
+In pegasus, a field is required unless the field is explicitly declared as
```
nickibi: Pegasus should start with capital letter? If yes, please fix other places as well.

evanw555: This is a good question. I can never decide which way it is tbh

karthikbalasub: I'm considering to update based on this rule:
 - Use "Pegasus" when the doc is talking about the Pegasus framework.
 - Use "pegasus" otherwise, eg, pegasus file, pegasus schema.

What do you think?

pdl_schema.md
karthikbalasub: I went back and forth on this: Should this doc provide examples for all supported features?.

I finally decided to skip some features that are not recommended and have better alternative.
This was one of them. Another example is specifying default for optional field.

pdl_schema.md
karthikbalasub: Decided to use "Pegasus" everywhere

pdl_schema.md
```diff
@@ -0,0 +1,1105 @@
...
+-   [Fixed Type](#fixed-type)
+-   [Import](#import)
+-   [Properties](#properties)
+-   [Package](#package)
+-   [Escaping](#escaping)
+
+## PDL Schema Definition
+
+Pegasus is a schema definition and serialization system developed as part of
+rest.li framework. It provides [multi-language
```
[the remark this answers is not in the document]
karthikbalasub: Done

pdl_schema.md
```diff
@@ -0,0 +1,1106 @@
...
+
+## Primitive Types
+The above examples already introduced `int`, `long`, `string` and `boolean`
+primitive types that are supported by Pegasus.
+The full list of supported Pegasus primitive types are: `int`, `long`, `float`,
+`double`, `boolean`, `string` and `bytes`.
+
+The actual types used for the primitives depends on the language specific
+binding implementation. For details on Java bindings for Pegasus primitives, see
+[Primitive Types] (https://linkedin.github.io/rest.li/java_binding#primitive-types)
```
evanw555: This link isn't rendering properly. I think it's the space between the two components.

pdl_schema.md
```diff
@@ -0,0 +1,1106 @@
...
+
+The Pegasus code generator implements a resolver that is similar to Java class
+loaders. If there is a reference to a named schema, the code generator will try
+to look for a file in the code generator’s resolver path. The resolver path is
+similar to a Java classpath. The fully qualified name of the named schema will
+be translated to a relative file name. The relative file name is computed by
+replacing dots (“.”) in the fully qualified name by the directory path separator
+(typically “/”) and appending a `.pdl` extension. This relative file name looked
+up using each location in the resolver path until it finds a file that contains
+the named schema.
```
evanw555: nit: I think it looks nicer if you escape references to characters. e.g. "/" -> `/`, dot (.) -> (`.`)

pdl_schema.md
```diff
@@ -0,0 +1,1106 @@
+---
+layout: guide
+title: PDL Schema
+permalink: /pdl_schema
+redirect_from: /pdl_syntax
+excerpt: Rest.li PDL Schema.
```
evanw555: This excerpt is used as the description text in the search functionality, so keep that in mind here. Something like "Documentation of pegasus schemas and specification of the PDL syntax."

pdl_schema.md
```diff
@@ -0,0 +1,1106 @@
...
+
+For example:
+```pdl
+namespace com.example.models
+
+fixed MD5 16
+```
+The example above defines a type called `MD5`, which has a size of `16 bytes`.
+
+## Import
```
evanw555: nit: Should `Import` be pluralized? It looks like you made it singular for a reason. I see it the same as "Properties". If you do update it, make sure to update the TOC link.

pdl_schema.md
```diff
@@ -0,0 +1,1106 @@
...
+
+import com.example.models.time.Time
+
+record AuditStamp {
+  createdAt: Time
+  updatedAt: Time
+}
+```
+
+**Note:**
```
evanw555: Maybe be worthwhile to mention that import statements take precedence in name resolution.

PR #202, opened 2020-03-02
Fix data translator for translating Pegasus non-union field to Avro union and vise versa

junchuanwang:
Due to introduction of PegasusToAvroDefaultFieldTranslationMode, some of the translation from (Pegasus required with default) to (Avro optional) or (Avro optional) to (Pegasus required with default) cannot be performed correctly.

This PR is for that fix. Also added test cases.

### DIFF
--- a/data-avro/src/main/java/com/linkedin/data/avro/DataTranslator.java
+++ b/data-avro/src/main/java/com/linkedin/data/avro/DataTranslator.java
@@ -16,6 +16,7 @@
 
 package com.linkedin.data.avro;
 
+import com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper;
 import com.linkedin.data.ByteString;
 import com.linkedin.data.Data;
 import com.linkedin.data.DataList;
@@ -31,7 +32,6 @@ import com.linkedin.data.schema.FixedDataSchema;
 import com.linkedin.data.schema.MapDataSchema;
 import com.linkedin.data.schema.RecordDataSchema;
 import com.linkedin.data.schema.UnionDataSchema;
-import com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper;
 import java.nio.ByteBuffer;
 import java.util.AbstractMap;
 import java.util.ArrayDeque;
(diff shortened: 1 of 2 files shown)


data-avro/src/main/java/com/linkedin/data/avro/DataTranslator.java
```diff
@@ -666,6 +645,19 @@ else if (fieldValue == null)
...
+              Map.Entry<String, Schema> fieldAvroEntry = findUnionMember(fieldDataSchema, fieldAvroSchema);
+              if (fieldAvroEntry == null)
+              {
+                _path.removeLast();
+                continue;
+              }
+              fieldAvroSchema = fieldAvroEntry.getValue();
+            }
+
             // Translate default value as null, depending on specified options
```
evanw555: Realizing that this comment makes no sense. I think you accidentally copied one of the comments from above in your last PR. You should fix this.
`// Translate default value as null, depending on specified options` -> `// Translate the pegasus field data to Avro data`

data-avro/src/main/java/com/linkedin/data/avro/DataTranslator.java
```diff
@@ -666,6 +645,19 @@ else if (fieldValue == null)
                 continue;
               }
             }
+
+            if ((fieldDataSchema.getDereferencedType() != DataSchema.Type.UNION) && fieldAvroSchema.getType() == Schema.Type.UNION) {
```
evanw555: nit: remove unnecessary parentheses

data-avro/src/main/java/com/linkedin/data/avro/DataTranslator.java
```diff
@@ -666,6 +645,19 @@ else if (fieldValue == null)
...
+              // Since optional field will be represented as union in Avro
```
evanw555: This comment is misleading. This case is no longer is unique to optional fields (after all, this revelation is the reason for this PR). I'd suggest just removing this line comment and keeping the comment in the line below.

data-avro/src/main/java/com/linkedin/data/avro/DataTranslator.java
```diff
@@ -355,7 +355,9 @@ private Object translate(Object value, DataSchema dataSchema, Schema avroSchema)
             boolean isOptional = field.getOptional();
             DataSchema fieldDataSchema = field.getType();
             Schema fieldAvroSchema = avroSchema.getField(fieldName).schema();
-            if (isOptional && (fieldDataSchema.getDereferencedType() != DataSchema.Type.UNION))
+            if (isOptional && (fieldDataSchema.getDereferencedType() != DataSchema.Type.UNION) ||
```
evanw555: There are a few things wrong with this if-statement:
1. `isOptional` is no longer necessary, since we can no longer assume that this scenario is unique to optional fields.
2. It would be a lot simpler if you just made this consistent with the other if-statement in this diff; this condition should simply see if the pegasus type is *not* a union yet the Avro type *is* a union.
3. It has a lot of unnecessary parentheses, though if you follow my suggestions above, this should solve itself.

data-avro/src/main/java/com/linkedin/data/avro/DataTranslator.java
junchuanwang: I just think it hard to read. I split this line into two so I can remove unnecessary parentheses but still make it look nice.

PR #203, opened 2020-03-02
Adding a PathSpec constructor that takes in a List of String path com…

jadesym:
…ponents.


evanw555:
How does this provide anything that `PathSpec(Collection<String>)` doesn't already provide?

jadesym:
Oh, looks like this was added in the last week :). Didn't see the Collection option. Nice! Closing pull request since it's not necessary.

PR #204, opened 2020-03-03
Make TraverseContext and some of its method in DataSchemaRichContextTraverser public so custmoized SchemaVisitor can access

junchuanwang:
Allow customized implementation of DataSchemaRichContextTraverser.SchemaVisitor to access DataSchemaRichContextTraverser.TraverserContext and some of its method so that they can implement the customized logic in `callbackOnContext`


data/src/main/java/com/linkedin/data/schema/annotation/DataSchemaRichContextTraverser.java
```diff
@@ -505,6 +508,16 @@ TraverserContext getNextContext(String nextTraversePathComponent, String nextSch
       return nextContext;
     }
 
+    public DataSchema getOriginalTopLevelSchema()
+    {
+      return _originalTopLevelSchema;
+    }
+
+    void setOriginalTopLevelSchema(DataSchema originalTopLevelSchema)
```
mchen07: You want customized visitor to modify context?

junchuanwang: (1) some of the methods still have access control, they can't set all the field of the context.
(2) but yes, the visitor implementation could have access "currentSchema" and "topLevelSchema" now. I weighted between "hide implementation for transparency" vs "expose implementation for optimization", and feel that visitor implementation should have access to the details.

junchuanwang: > You want customized visitor to modify context?

(1) some of the methods still have access control, they can't set all the field of the context.
(2) but yes, the visitor implementation could have access "currentSchema" and "topLevelSchema" now. I weighted between "hide implementation for transparency" vs "expose implementation for optimization", and feel that visitor implementation should have access to the details.

data/src/main/java/com/linkedin/data/schema/annotation/DataSchemaRichContextTraverser.java
```diff
@@ -411,7 +412,7 @@ public String formatToErrorMessage()
    * A new {@link TraverserContext} object will be created before entering child from parent.
    * In this way, we simulate {@link TraverserContext} as elements inside stack during recursive traversal.
    */
-  static class TraverserContext
+  public static class TraverserContext
```
karthikbalasub: I suggest making this a top-level class now that its scope has increased

data/src/main/java/com/linkedin/data/schema/annotation/DataSchemaRichContextTraverser.java
```diff
@@ -420,6 +421,7 @@ public String formatToErrorMessage()
     Boolean _shouldContinue = null;
```
karthikbalasub: You should make all these variables private. Package-private is still loose a constraint.

And also create a final constructor so this cannot be extended.

data/src/main/java/com/linkedin/data/schema/annotation/DataSchemaRichContextTraverser.java
karthikbalasub: Make all the setters private and then create a new method for getFirstContext? That way you don't have to expose the setters outside of this class.

Infact, you should consider defining an interface for the context and then making this implementation class private.

mchen07: As @karthikbalasub commented, I may prefer that we make all setters private.

data/src/main/java/com/linkedin/data/schema/annotation/DataSchemaRichContextTraverser.java
junchuanwang: I don't think you can make a constructor final? What is the purpose?

karthikbalasub: I mean make the class final, the goal is to avoid extending this class

junchuanwang: I made "TraverserContext" an interface, and made a "TraverserContextImpl" private implementation class now so it shouldn't be extended.

PR #207, opened 2020-03-05
Enforce schema properties order in JsonBuilder so Pegasus schema and Avro schema properties would have deterministic order

junchuanwang:
Currently when translating to Avro or Pegasus schema, nested properties wouldn't have deterministic and based on its implementation (DataMap type), the order is given from HashMap, which is indeterministic.

This change would affect [pegasus to avro] translation AND [avro to peagsus] translation when outputting string (only for properties encoding). It wouldn't affect memory representation.

wc-test:
https://crt.prod.linkedin.com/#/testing/executions/7037b41c-8501-4db9-9237-1f54c5964c68/execution

### DIFF
--- a/data-avro/src/test/java/com/linkedin/data/avro/TestSchemaTranslator.java
+++ b/data-avro/src/test/java/com/linkedin/data/avro/TestSchemaTranslator.java
@@ -1238,7 +1238,7 @@ public class TestSchemaTranslator
                   "]" +
                   "}",
               allModes,
-              "{ \"type\" : \"record\", \"name\" : \"foo\", \"fields\" : [ { \"name\" : \"result\", \"type\" : { \"type\" : \"record\", \"name\" : \"fooResult\", \"fields\" : [ { \"name\" : \"success\", \"type\" : [ \"string\", \"null\" ], \"doc\" : \"Success message\", \"default\" : \"Union with aliases.\" }, { \"name\" : \"failure\", \"type\" : [ \"null\", \"string\" ], \"doc\" : \"Failure message\", \"default\" : null }, { \"name\" : \"fieldDiscriminator\", \"type\" : { \"type\" : \"enum\", \"name\" : \"fooResultDiscriminator\", \"symbols\" : [ \"success\", \"failure\" ] }, \"doc\" : \"Contains the name of the field that has its value set.\" } ] }, \"default\" : { \"fieldDiscriminator\" : \"success\", \"success\" : \"Union with aliases.\", \"failure\" : null } } ] }",
+              "{ \"type\" : \"record\", \"name\" : \"foo\", \"fields\" : [ { \"name\" : \"result\", \"type\" : { \"type\" : \"record\", \"name\" : \"fooResult\", \"fields\" : [ { \"name\" : \"success\", \"type\" : [ \"string\", \"null\" ], \"doc\" : \"Success message\", \"default\" : \"Union with aliases.\" }, { \"name\" : \"failure\", \"type\" : [ \"null\", \"string\" ], \"doc\" : \"Failure message\", \"default\" : null }, { \"name\" : \"fieldDiscriminator\", \"type\" : { \"type\" : \"enum\", \"name\" : \"fooResultDiscriminator\", \"symbols\" : [ \"success\", \"failure\" ] }, \"doc\" : \"Contains the name of the field that has its value set.\" } ] }, \"default\" : { \"failure\" : null, \"fieldDiscriminator\" : \"success\", \"success\" : \"Union with aliases.\" } } ] }",
               emptyFooSchema,
               emptyFooValue,
               "{\"result\": {\"success\": \"Union with aliases.\", \"failure\": null, \"fieldDiscriminator\": ##Q_STARTsuccess##Q_END}}"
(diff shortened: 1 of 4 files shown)


data/src/main/java/com/linkedin/data/codec/AbstractJacksonDataCodec.java
```diff
@@ -296,12 +296,20 @@ public void startMap(DataMap map) throws IOException
       _generator.writeStartObject();
     }
 
+    @Override
+    public Iterable<Map.Entry<String, Object>> orderMap(DataMap map)
```
nickibi: I am curious about how this method is being called?

data-avro/src/test/java/com/linkedin/data/avro/TestSchemaTranslator.java
```diff
@@ -1660,6 +1660,46 @@ public void testToAvroSchema_testTypeRefAnnotationPropagation(String schemaBefor
...
+          {
+              " { " +
+              "   \"type\" : \"record\", " +
+              "   \"name\" : \"Foo\", " +
+              "   \"fields\" : [ { " +
+              "     \"name\" : \"field1\", " +
+              "     \"type\" : \"int\", " +
+              "     \"b_customerAnnotation\" : \"f1\", " +
+              "     \"c_customerAnnotation\" : \"f1\", " +
+              "     \"a_customerAnnotation\" : \"f1\" " +
```
evanw555: nit: I think you mean to say `x_customAnnotation`?

data/src/main/java/com/linkedin/data/schema/JsonBuilder.java
```diff
@@ -282,7 +280,10 @@ public void writeProperties(Map<String, ?> value) throws IOException
   {
     if (value.isEmpty() == false)
     {
-      for (Map.Entry<String, ?> entry : value.entrySet())
+      List<Map.Entry<String, ?>> orderedProperties =  value.entrySet().stream()
```
evanw555: The unit tests seems to imply this, however I'd like to confirm; this method is only for writing custom properties, right? Since this is a generic `JsonBuilder`, does it have any sense of what "properties" are? Or is this just for writing any generic subtree of the JSON object?

data/src/main/java/com/linkedin/data/codec/AbstractJacksonDataCodec.java
```diff
@@ -296,6 +296,12 @@ public void startMap(DataMap map) throws IOException
...
```
karthikbalasub: This is not a good idea. The jackson codec is used for serializing data on the servers and making this change here will affect performance of services.

data/src/main/java/com/linkedin/data/schema/JsonBuilder.java
junchuanwang: this is just for "writing  generic subtree of the JSON object", any concerns?

data-avro/src/test/java/com/linkedin/data/avro/TestSchemaTranslator.java
junchuanwang: yes sir.

data/src/main/java/com/linkedin/data/codec/AbstractJacksonDataCodec.java
junchuanwang: Update code ,can you please check? Now I limit the scope of the change so it only applies when writing properties.

data/src/main/java/com/linkedin/data/codec/AbstractJacksonDataCodec.java
```diff
@@ -213,12 +224,35 @@ public DataList readList(InputStream in) throws IOException
     return parse(_factory.createParser(in), mesg, locationMap);
   }
 
-  public void objectToJsonGenerator(Object object, JsonGenerator generator) throws IOException
```
karthikbalasub: Do not change public APIs. Keep this one and add the second API

data/src/main/java/com/linkedin/data/codec/AbstractJacksonDataCodec.java
```diff
@@ -133,7 +133,18 @@ protected void writeObject(Object object, JsonGenerator generator) throws IOExce
 
   protected Data.TraverseCallback createTraverseCallback(JsonGenerator generator)
   {
-    return new JacksonTraverseCallback(generator);
+    return createTraverseCallback(generator, false);
+  }
+
+  /**
+   * Create {@link JacksonTraverseCallback} instance for Data object traverse
```
karthikbalasub: JacksonTraverseCallback -> Traverse.Callback
Avoid implementation details in the java doc

data/src/main/java/com/linkedin/data/codec/AbstractJacksonDataCodec.java
```diff
@@ -213,12 +224,35 @@ public DataList readList(InputStream in) throws IOException
     return parse(_factory.createParser(in), mesg, locationMap);
   }
 
-  public void objectToJsonGenerator(Object object, JsonGenerator generator) throws IOException
+  /**
+   * Convert an object to Json format representation.
+   * @param object the object that needs to be converted
+   * @param generator the generator that could generate Json output based on the object
+   * @param mapKeyOrdered if the object contains map element (which could be nested map), can set this flag to order the
```
karthikbalasub: If true, then map elements in the object (can be the object itself or its nested elements) will be sorted by Key in the generated Json.

data/src/main/java/com/linkedin/data/codec/AbstractJacksonDataCodec.java
```diff
@@ -213,12 +224,35 @@ public DataList readList(InputStream in) throws IOException
...
+   * @throws IOException
+   */
+  public void objectToJsonGenerator(Object object, JsonGenerator generator, boolean mapKeyOrdered) throws IOException
   {
-    Data.TraverseCallback callback = createTraverseCallback(generator);
+    Data.TraverseCallback callback = createTraverseCallback(generator, mapKeyOrdered);
     Data.traverse(object, callback);
   }
 
+  public static class JacksonTraverseCallbackMapKeyOrdered extends JacksonTraverseCallback
```
karthikbalasub: Why do you need this class? Can't you add a flag in JacksonTraverseCallback and a second constructor to set it?

data/src/main/java/com/linkedin/data/schema/JsonBuilder.java
```diff
@@ -269,11 +267,24 @@ public void writeMapField(String fieldName, Map<String, ?> value, boolean requir
    */
   public void writeData(Object object) throws IOException
   {
-    _jacksonDataCodec.objectToJsonGenerator(object, _jsonGenerator);
+    _jacksonDataCodec.objectToJsonGenerator(object, _jsonGenerator, false);
   }
 
+  /**
+   * Write Data object. But if the Data Object contains DataMap, the result would output the map entries using sorted key order.
```
karthikbalasub: document that this applies to object also.

data/src/main/java/com/linkedin/data/codec/AbstractJacksonDataCodec.java
junchuanwang: Good catch. Fail to notice it was public

data/src/main/java/com/linkedin/data/codec/AbstractJacksonDataCodec.java
junchuanwang: I wouldn't say I totally agree with you in this one. But I made the change according to your suggestion/

The reason is that this method take in "generator" as input parameter in order to return Data.TraverseCallback, this is already very implementation specific and very detailed implementation description. Something unavoidable.

data/src/main/java/com/linkedin/data/codec/AbstractJacksonDataCodec.java
junchuanwang: Because I need to override the 
`orderMap` method in the new class. 

I don't think I got you. How do I set the override method in the constructor?

junchuanwang: after off line discussion, adopted the suggestion

data/src/main/java/com/linkedin/data/schema/JsonBuilder.java
evanw555: Would this affect the order in which all schema attributes are written, even non-custom properties? e.g. [namespace, name, type, fields] vs [type, fields, name, namespace]. Or are those written individually in the schema encoder? I'm concerned that this would affect the ordering of things that aren't custom properties

data/src/main/java/com/linkedin/data/template/JacksonDataTemplateCodec.java
```diff
@@ -87,7 +87,7 @@ protected void dataTemplateToJsonGenerator(Object data,
     }
     else
     {
-      objectToJsonGenerator(data, generator);
+      objectToJsonGenerator(data, generator, false);
```
karthikbalasub: You can revert this change now.

data/src/main/java/com/linkedin/data/schema/JsonBuilder.java
junchuanwang: it shouldn't. non-customer properties are written using other code. This code only manage the properties, so all properties written as a trunk. shouldn't affect others

data/src/main/java/com/linkedin/data/template/JacksonDataTemplateCodec.java
nickibi: No need for this change.

PR #212, opened 2020-03-07
Add Streaming Timeout to TransportClient Properties

nizarm:
Add Streaming Timeout to Transport Client Properties Converter so that these properties get serialized and deserialized properly.

### DIFF
--- a/d2-schemas/src/main/pegasus/com/linkedin/d2/D2TransportClientProperties.pdsc
+++ b/d2-schemas/src/main/pegasus/com/linkedin/d2/D2TransportClientProperties.pdsc
@@ -17,6 +17,12 @@
       "optional": true
     },
     {
+      "name": "streamingTimeout",
+      "type": "long",
+      "doc": "Streaming Timeout in ms for this transport client. Disabled by default (-1)",
+      "optional": true
+    },
+    {
       "name": "maxResponseSize",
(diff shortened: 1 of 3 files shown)


d2-schemas/src/main/pegasus/com/linkedin/d2/D2TransportClientProperties.pdsc
```diff
@@ -16,6 +16,12 @@
       "doc": "Timeout in ms for this transport client. Defaults to 10s.",
       "optional": true
     },
+    {
+      "name": "streamingTimeout",
+      "type": "long",
+      "doc": "Streaming Timeout in ms for this transport client. Disabled by default (-1)",
```
ChaoLinkedIn: Why -1? 0 should disable it as well, right?

nizarm: The streaming implementation expect -1, 0 will make the request immediately timeout.

ChaoLinkedIn: We probably do not have any use case to immediately timeout a request. Prefer to have 0 to disable it as other configs. But that probably belong to another changes.

PR #214, opened 2020-03-07
Fix Classpath resource schema resolver to handle pdl resources.

karthikbalasub:


karthikbalasub:
The github diff is confusing for this PR, it might be easier to review it by checking out the branch locally.

data/src/main/java/com/linkedin/data/schema/resolver/ClasspathResourceDataSchemaResolver.java
```diff
@@ -33,61 +34,106 @@
...
+    void setExtension(String extension)
     {
-      errorMessageBuilder.append(String.format("Failed to read/close data schema file \"%s\" in classpath: \"%s\"", schemaResourcePath, e.getMessage()));
+      this._extension = extension;
     }
-    return schema;
-  }
 
-  private final ClassLoader _classLoader;
+    private String _extension = SchemaParser.FILE_EXTENSION;
```
evanw555: I suggest you move this field and its corresponding setter near the top of the inner class. It would be easier to read.

data/src/main/java/com/linkedin/data/schema/resolver/ClasspathResourceDataSchemaResolver.java
```diff
@@ -33,61 +34,106 @@
  *
  * @author Min Chen
  */
-public class ClasspathResourceDataSchemaResolver extends DefaultDataSchemaResolver
```
evanw555: This class no longer extends `DefaultDataSchemaResolver`. Are you sure you're not removing any default functionality it had before?

data/src/main/java/com/linkedin/data/schema/resolver/AbstractMultiFormatDataSchemaResolver.java
```diff
@@ -0,0 +1,179 @@
...
+
+/**
+ * Combines multiple file format specific resolvers (and respective file format specific parsers)
+ * into a single resolver.
+ *
+ * E.g. a resolver for the ".pdsc" file format and the ".pdl" file format, each with their
+ * own file format specific parsers, can be combined into a single resolver able to look up
+ * schemas of either file format.
+ */
+public class AbstractMultiFormatDataSchemaResolver implements DataSchemaResolver
```
mchen07: How is this different from current `MultiFormatDataSchemaResolver` (https://github.com/linkedin/rest.li/blob/master/data/src/main/java/com/linkedin/data/schema/resolver/MultiFormatDataSchemaResolver.java)?

evanw555: Based on my understanding, he wanted to add similar logic to `ClasspathResourceDataSchemaResolver`, so it makes more sense to add the shared logic in a common abstract class.

karthikbalasub: Correct, refactored the existing resolver to extract the common logic to abstract class. Forgot to make it abstract, will do it.

The existing multi-format resolver is the file based one.

data/src/main/java/com/linkedin/data/schema/resolver/ClasspathResourceDataSchemaResolver.java
karthikbalasub: Since this delegates to the SingleFormat resolver (inner class) which extends the Default resolver, there shouldn't be any regression.

data/src/main/java/com/linkedin/data/schema/resolver/AbstractMultiFormatDataSchemaResolver.java
mchen07: Never mind, I found out when I read through to the end.

data/src/main/java/com/linkedin/data/schema/resolver/ClasspathResourceDataSchemaResolver.java
evanw555: Ah, that's right. It's interesting how the relationship of the multi-format resolvers with the default resolver logic is more "has-a" than "is-a". Reminds me of CS class...

data/src/main/java/com/linkedin/data/schema/resolver/ClasspathResourceDataSchemaResolver.java
```diff
@@ -33,61 +34,106 @@
...
    * @param classLoader provides the {@link ClassLoader}.
    */
-  public ClasspathResourceDataSchemaResolver(SchemaParserFactory parserFactory, ClassLoader classLoader)
+  public ClasspathResourceDataSchemaResolver(ClassLoader classLoader)
   {
-    super(parserFactory);
+    for (DataSchemaParserFactory parserForFormat: BUILTIN_FORMAT_PARSER_FACTORIES)
+    {
+      SingleFormatClasspathSchemaResolver resolver = new SingleFormatClasspathSchemaResolver(parserForFormat);
+      resolver.setExtension("." + parserForFormat.getLanguageExtension());
```
mchen07: why do we need this setter exposed? Should constructor handle this automatically?

data/src/main/java/com/linkedin/data/schema/resolver/ClasspathResourceDataSchemaResolver.java
mchen07: I may prefer that we make this `final`, and initialized in constructor and hide `setter` for this.

data/src/main/java/com/linkedin/data/schema/resolver/ClasspathResourceDataSchemaResolver.java
```diff
@@ -33,61 +34,107 @@
...
-  @Override
-  protected NamedDataSchema locateDataSchema(String schemaName, StringBuilder errorMessageBuilder)
+  private class SingleFormatClasspathSchemaResolver extends DefaultDataSchemaResolver
   {
-    NamedDataSchema schema = null;
-    final String schemaResourcePath = getDataSchemaResourcePath(schemaName);
-    try (InputStream stream = _classLoader.getResourceAsStream(schemaResourcePath))
+    private String _extension = SchemaParser.FILE_EXTENSION;
+
+    void setExtension(String extension)
```
evanw555: one last nit: move to below constructor

data/src/main/java/com/linkedin/data/schema/resolver/ClasspathResourceDataSchemaResolver.java
evanw555: +1

karthikbalasub: Fixed, this was remnant of an alternate implementation I started with. Cleaned it now.

PR #216, opened 2020-03-09
pretty print json in pdl builder

aman1309:
Manually tested JSON indent, can not add those test cases till we have the deterministic ordering of properties.

### DIFF
--- a/data/src/main/java/com/linkedin/data/schema/IndentedPdlBuilder.java
+++ b/data/src/main/java/com/linkedin/data/schema/IndentedPdlBuilder.java
@@ -18,6 +18,7 @@ package com.linkedin.data.schema;
 
 import com.fasterxml.jackson.core.util.DefaultIndenter;
 import com.fasterxml.jackson.core.util.DefaultPrettyPrinter;
+import com.linkedin.data.codec.JacksonDataCodec;
 import java.io.IOException;
 import java.io.Writer;
 import org.apache.commons.lang3.StringUtils;
@@ -123,11 +124,13 @@ class IndentedPdlBuilder extends PdlBuilder
   @Override
   PdlBuilder writeJson(Object value) throws IOException
   {
+    JacksonDataCodec jsonCodec = new JacksonDataCodec();
     DefaultPrettyPrinter prettyPrinter = new DefaultPrettyPrinter();
     prettyPrinter.indentObjectsWith(
         new DefaultIndenter(getIndentSpaces(1), DefaultIndenter.SYS_LF + getIndentSpaces(_indentDepth)));
-    _jsonCodec.setPrettyPrinter(prettyPrinter);
-    return super.writeJson(value);
+    jsonCodec.setPrettyPrinter(prettyPrinter);
+    write(toJson(value, jsonCodec));
+    return this;
   }
 
   /**
--- a/data/src/main/java/com/linkedin/data/schema/IndentedPdlBuilder.java
+++ b/data/src/main/java/com/linkedin/data/schema/IndentedPdlBuilder.java
@@ -16,6 +16,8 @@
 
 package com.linkedin.data.schema;
 
+import com.fasterxml.jackson.core.util.DefaultIndenter;
+import com.fasterxml.jackson.core.util.DefaultPrettyPrinter;
 import java.io.IOException;
 import java.io.Writer;
 import org.apache.commons.lang3.StringUtils;
@@ -71,13 +73,7 @@ class IndentedPdlBuilder extends PdlBuilder
   @Override
   PdlBuilder indent() throws IOException
   {
-    final int numSpaces = _indentDepth * DEFAULT_INDENT_WIDTH;
-    final StringBuilder sb = new StringBuilder(numSpaces);
-    for (int i = 0; i < numSpaces; i++)
-    {
-      sb.append(" ");
-    }
-    write(sb.toString());
+    write(getIndentSpaces());
     return this;
   }
 
@@ -124,6 +120,15 @@ class IndentedPdlBuilder extends PdlBuilder
     return false;
   }
 
+  @Override
+  PdlBuilder writeJson(Object value) throws IOException
+  {
+    DefaultPrettyPrinter prettyPrinter = new DefaultPrettyPrinter();
+    prettyPrinter.indentObjectsWith(new DefaultIndenter(" ", DefaultIndenter.SYS_LF + getIndentSpaces()));
+    _jsonCodec.setPrettyPrinter(prettyPrinter);
+    return super.writeJson(value);
+  }
+
   /**
    * Write an intended line of .pdl code.
    * The code will be prefixed by the current indentation and suffixed with a newline.
@@ -132,5 +137,15 @@ class IndentedPdlBuilder extends PdlBuilder
   private void writeLine(String code) throws IOException
   {
     indent().write(code).newline();
+  }
+
+  private String getIndentSpaces() {
+    final int numSpaces = _indentDepth * DEFAULT_INDENT_WIDTH;
+    final StringBuilder sb = new StringBuilder(numSpaces);
+    for (int i = 0; i < numSpaces; i++)
+    {
+      sb.append(" ");
+    }
+    return sb.toString();
   }
 }
--- a/data/src/main/java/com/linkedin/data/schema/IndentedPdlBuilder.java
+++ b/data/src/main/java/com/linkedin/data/schema/IndentedPdlBuilder.java
@@ -73,7 +73,7 @@ class IndentedPdlBuilder extends PdlBuilder
   @Override
   PdlBuilder indent() throws IOException
   {
-    write(getIndentSpaces());
+    write(getIndentSpaces(_indentDepth));
     return this;
   }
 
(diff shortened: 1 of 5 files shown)


aman1309:
> 1. This doesn't look like it accounts for the PDL builder's stateful indent property.
          Fixed. Thanks

> 2. It's not true that you can't add testing for this because of the deterministic ordering problem. Consider a property like `@a = { "b": { "c": 123 } }` or `@z = [ 1, 2, 3 ]`
         updated test for array but 1 test cases will be compacted to @a.b.c = 123



Also manually tested for below PDL
namespace com.linkedin.pegasus.generator.test.idl.escaping

record PropertyKeyEscaping {
  @grammarChars.`foo[type=a.b.c].bar`.`{:=}` = "grammarChars"
  **@`namespace` = {
    "bar": "foo",
    "foo": "bar"
  }**
  @path.`//`.`/*.*/`.`/** foo */`
  @path2.`/**`.`*/`
  @shouldNotBeEscaped.ABC.1-2.a_b.ab124.000.ABC-123
  @`test.path` = 1
  @validate.`com.linkedin.CustomValidator` = "foo"
  aField: string
}

data-avro/src/test/java/com/linkedin/data/avro/TestSchemaTranslator.java
```diff
@@ -1660,6 +1660,46 @@ public void testToAvroSchema_testTypeRefAnnotationPropagation(String schemaBefor
               null,
               null,
               null
+          },
```
karthikbalasub: these are not related to your change.. please check

aman1309: yes updated the branch

data/src/main/java/com/linkedin/data/schema/IndentedPdlBuilder.java
```diff
@@ -124,6 +120,15 @@ boolean writeDoc(String doc) throws IOException
     return false;
   }
 
+  @Override
+  PdlBuilder writeJson(Object value) throws IOException
+  {
+    DefaultPrettyPrinter prettyPrinter = new DefaultPrettyPrinter();
+    prettyPrinter.indentObjectsWith(new DefaultIndenter(" ", DefaultIndenter.SYS_LF + getIndentSpaces()));
```
karthikbalasub: Should this be using two spaces? SPACE * INDENT_WIDTH?

data/src/main/java/com/linkedin/data/schema/IndentedPdlBuilder.java
```diff
@@ -124,6 +120,15 @@ boolean writeDoc(String doc) throws IOException
...
+    _jsonCodec.setPrettyPrinter(prettyPrinter);
```
karthikbalasub: Suggest creating a local codec for this. 
This way of changing the state of a member variable for each function call is not clean.

data/src/test/java/com/linkedin/data/schema/TestPdlBuilder.java
```diff
@@ -37,31 +39,39 @@
...
+              "@empty = [ 1, 2, 3 ]\n",
+              "@empty=[1,2,3]"
             },
             {
               properties2,
-              "@validate.empty = {}\n"
+              "@validate.empty = [ 1, 2, 3 ]\n",
+              "@validate.empty=[1,2,3]"
             }
         //TODO Add test case for multiple properties in a map level once iteration logic is fixed to be deterministic
```
karthikbalasub: Can you add the testcases you used for testing, but comment them out for now. It would be easier to just enable those tests later.

data/src/main/java/com/linkedin/data/schema/IndentedPdlBuilder.java
aman1309: that is why I changed static to property, this is need to add pdl stateful indentation at the beginning of each line in json

data/src/main/java/com/linkedin/data/schema/IndentedPdlBuilder.java
aman1309: fixed

data/src/test/java/com/linkedin/data/schema/TestPdlBuilder.java
aman1309: done

data/src/test/java/com/linkedin/data/schema/TestPdlBuilder.java
```diff
@@ -36,32 +38,57 @@
...
+              arrayValueProperty,
+              "@`array` = [ 1, 2, 3 ]\n",
+              "@`array`=[1,2,3]"
+            },
+            {
+              flattenProperty,
+              "@flatten.`array` = [ 1, 2, 3 ]\n",
+              "@flatten.`array`=[1,2,3]"
+            },
+            /* TODO Add test case for multiple properties in a map level once iteration logic is fixed to be deterministic
```
evanw555: I'd suggest adding a map test case with only 1 property per level, along with a TODO saying to add more properties at each level.

data/src/main/java/com/linkedin/data/schema/IndentedPdlBuilder.java
```diff
@@ -124,6 +120,16 @@ boolean writeDoc(String doc) throws IOException
     return false;
   }
 
+  @Override
+  PdlBuilder writeJson(Object value) throws IOException
```
evanw555: I don't think this method is necessary. You can avoid adding this logic as an overridden method which calls the super method (a little confusing) by just updating the codec's pretty printer on indent change (i.e. `#increaseIndent` and `#decreaseIndent`). It seems more intuitive IMO to compute the indent string and update the pretty printer on indent change rather than on each use.

data/src/main/java/com/linkedin/data/schema/IndentedPdlBuilder.java
```diff
@@ -133,4 +139,15 @@ private void writeLine(String code) throws IOException
   {
     indent().write(code).newline();
   }
+
+  private String getIndentSpaces(int indentDepth)
```
evanw555: I would make this method static, since it's just a helper method to compute indent text. Whether to use static methods when possible or to avoid static methods seems to be a contentious and subjective issue, so it's up to you tbh.

data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
```diff
@@ -63,6 +61,7 @@
   }
 
   private final Writer _writer;
+  final JacksonDataCodec _jsonCodec = new JacksonDataCodec();
```
evanw555: Should be protected.

data/src/test/java/com/linkedin/data/schema/TestPdlBuilder.java
aman1309: flatten array property map is that use case and we need it to test compact logic.
that is why added separate test for json case.

data/src/main/java/com/linkedin/data/schema/IndentedPdlBuilder.java
aman1309: It shouldn't be confusing at Indenter is kind of decorating the vanilla builder so it should be good to have child class calling super method.

Regarding having this logic in increase and decrease indent I would avoid it as to why build it when it's not needed.

data/src/main/java/com/linkedin/data/schema/IndentedPdlBuilder.java
aman1309: lets keep it like this for now, as scope of method is for this class only

data/src/main/java/com/linkedin/data/schema/IndentedPdlBuilder.java
karthikbalasub: The state is not for the entire PdlBuilder, but only when writeJson is called (ie at a function scope). After this function is completed, the state of _jsonBuilder is invalid. So using a member variable to set the state for a particular function is odd. 

What if someone uses the _jsonCodec on some other code path without updating the prettyPrinter? 
   It is not explicit that pretty printer should be updated before the code should be used.
What happens if PdlBuilder stops using _jsonCodec in toJson? 
  It is not explict that PdlBuilder is expecting json to be formatted by custom builders.

data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
```diff
@@ -275,11 +274,11 @@ private String toJson(Object value) throws IOException
   {
     if (value instanceof DataMap)
     {
-      return JSON_CODEC.mapToString((DataMap) value);
+      return _jsonCodec.mapToString((DataMap) value);
```
karthikbalasub: Cleaner option is to create two protected abstrac methods toJson(DataMap) toJson(DataList) and let concrete implementations decide how the json is generated.

aman1309: why not have only 1 ie toJson(Object) and writeJson is kind of fulfilling that but if needed we can separate out toJson

PR #176, opened 2020-02-19
Add resource key validation while parsing

aman1309:
Rest of the unit cases will add later if there are no major comments for refactoring.

PR #217, opened 2020-03-11
Add resource key validation while parsing

aman1309:
I had deleted the fork, raising the PR again. 
Refer https://github.com/linkedin/rest.li/pull/176 for initial comments.

### DIFF
--- a/restli-common/src/main/java/com/linkedin/restli/common/ComplexResourceKey.java
+++ b/restli-common/src/main/java/com/linkedin/restli/common/ComplexResourceKey.java
@@ -23,6 +23,8 @@ import com.linkedin.data.schema.validation.CoercionMode;
 import com.linkedin.data.schema.validation.RequiredMode;
 import com.linkedin.data.schema.validation.ValidateDataAgainstSchema;
 import com.linkedin.data.schema.validation.ValidationOptions;
+import com.linkedin.data.schema.validation.ValidationResult;
+import com.linkedin.data.schema.validator.DataSchemaAnnotationValidator;
 import com.linkedin.data.template.DataTemplateUtil;
 import com.linkedin.data.template.RecordTemplate;
 import com.linkedin.jersey.api.uri.UriComponent;
@@ -214,11 +216,10 @@ public final class ComplexResourceKey<K extends RecordTemplate, P extends Record
    * @return {@link ComplexResourceKey} initialized with id and param values specified in
    *         the input DataMap
    */
-  public static ComplexResourceKey<RecordTemplate, RecordTemplate> buildFromDataMap(DataMap keyDataMap,
-                                                                                    Class<? extends RecordTemplate> keyKeyClass,
-                                                                                    Class<? extends RecordTemplate> keyParamsClass)
+  public static ComplexResourceKey<RecordTemplate, RecordTemplate> buildFromDataMap(DataMap keyDataMap, Class<? extends RecordTemplate> keyKeyClass,
+      Class<? extends RecordTemplate> keyParamsClass)
   {
-    return buildFromDataMap(keyDataMap, ComplexKeySpec.forClassesMaybeNull(keyKeyClass, keyParamsClass));
+    return buildFromDataMap(keyDataMap, ComplexKeySpec.forClassesMaybeNull(keyKeyClass, keyParamsClass), false);
   }
 
   /**
@@ -233,6 +234,23 @@ public final class ComplexResourceKey<K extends RecordTemplate, P extends Record
   public static ComplexResourceKey<RecordTemplate, RecordTemplate> buildFromDataMap(DataMap keyDataMap,
                                                                                     ComplexKeySpec<?, ?> complexKeyType)
   {
+    return buildFromDataMap(keyDataMap, complexKeyType, false);
+  }
+
+  /**
+   * Build complex key instance from an untyped datamap representing a complex key as
+   * defined in {@link QueryParamsDataMap}
+   *
+   * @param keyDataMap untyped DataMap - all primitive values are represented as strings.
+   * @param complexKeyType type of {@link ComplexResourceKey}
+   * @param enforceValidation if set true throws IllegalArgumentException on validation failure
+   * @return {@link ComplexResourceKey} initialized with id and param values specified in
+   *         the input DataMap
+   */
+  public static ComplexResourceKey<RecordTemplate, RecordTemplate> buildFromDataMap(DataMap keyDataMap,
+                                                                                    ComplexKeySpec<?, ?> complexKeyType,
+                                                                                    boolean enforceValidation)
+  {
     // Copy in case the original is immutable
     keyDataMap = new DataMap(keyDataMap);
 
@@ -242,8 +260,8 @@ public final class ComplexResourceKey<K extends RecordTemplate, P extends Record
     {
       paramsDataMap = new DataMap();
     }
-    RecordTemplate key = validateDataMap(keyDataMap, complexKeyType.getKeyType());
-    RecordTemplate params = validateDataMap(paramsDataMap, complexKeyType.getParamsType());
+    RecordTemplate key = validateDataMap(keyDataMap, complexKeyType.getKeyType(), enforceValidation);
+    RecordTemplate params = validateDataMap(paramsDataMap, complexKeyType.getParamsType(), enforceValidation);
 
     return new ComplexResourceKey<RecordTemplate, RecordTemplate>(key, params);
   }
@@ -270,7 +288,7 @@ public final class ComplexResourceKey<K extends RecordTemplate, P extends Record
     Map<String, List<String>> queryParameters =
         UriComponent.decodeQuery(URI.create("?" + currentPathSegment), true);
     DataMap allParametersDataMap = QueryParamsDataMap.parseDataMapKeys(queryParameters);
-    return buildFromDataMap(allParametersDataMap, complexKeyType);
+    return buildFromDataMap(allParametersDataMap, complexKeyType, false);
   }
 
   /**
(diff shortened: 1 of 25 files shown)


restli-int-test-api/src/main/snapshot/com.linkedin.restli.examples.greetings.client.typerefCustomDoubleAssociationKeyResource.snapshot.json
```diff
@@ -26,6 +26,11 @@
     "ref" : "double",
     "java" : {
       "class" : "com.linkedin.restli.examples.custom.types.CustomDouble"
+    },
+    "validate" : {
+      "regex" : {
+        "regex" : "[0-9]*\\.[0-9]"
```
junchuanwang: Question. Why is not not "[0-9]*\\.[0-9]*"   how many digits are allowed after the "."?

aman1309: yup that is only 1

restli-server/src/main/java/com/linkedin/restli/server/config/ResourceMethodConfigImpl.java
```diff
@@ -10,15 +10,17 @@
  *
  * @author mnchen
  */
-class ResourceMethodConfigImpl implements ResourceMethodConfig
+public class ResourceMethodConfigImpl implements ResourceMethodConfig
```
junchuanwang: My last question before giving a ship:

Why is this public? Is this used anywhere?

aman1309: this is need for RoutingResult default initialization.

junchuanwang: sorry I still didn't quite get why package private is not enough.
Aren't they in the same package? RoutingResult  is in the parent package of ResourceMethodConfigImpl ?

aman1309: routingresult is not in same package

restli-server/src/main/java/com/linkedin/restli/server/config/ResourceMethodConfig.java
```diff
@@ -21,4 +21,9 @@
    * Config for whether this method will need to validate query parameters.
    */
   boolean shouldValidateQueryParams();
+
+  /**
+   * Config for whether this method will need to validate path/resource keys.
+   */
+  boolean shouldValidateResourceKeyParams();
```
evanw555: Replying to your comment on the last PR:

> Resource keys are identifiers for entity on which the resource takes action on.
> there are two configs here one to validate keys in the path, 2nd in method level config which validates keys in batch requests or resource method argument/params

So are you saying that the service-level config only determines whether to validate path keys? And the method-level config determines whether to validate the key (either from the path or query param) passed into the resource method invocation? If so, then this naming is confusing and doesn't seem to align with this logic. If my assessment is correct, then a name like `RestLiConfig#shouldValidatePathKeys` and `RestLiMethodConfig#shouldValidateResourceKeys` are more intuitive.

Also, since these are interfaces meant for public consumption, this should be thoroughly clarified via javadocs on the getters and setters of both configs.

aman1309: shouldValidateResourceKeys at all places. updated docs

restli-server/src/main/java/com/linkedin/restli/internal/server/util/ArgumentUtils.java
```diff
@@ -507,16 +559,25 @@ public static Object parseSimplePathKey(final String value,
     }
   }
 
+  public static Object convertSimpleValue(final String value,
+      final DataSchema schema,
```
junchuanwang: in pegasus we align method parameters in code style, can you fix this?

junchuanwang:
A general comment: 

I am ok with setting this config to RestliConfig level, but still want to ask, do you have any considerations between setting this on "RestLiConfig" level versus "RestliMethodConfig" level for finer granularity?

restli-server/src/main/java/com/linkedin/restli/server/config/ResourceMethodConfig.java
```diff
@@ -21,4 +21,9 @@
    * Config for whether this method will need to validate query parameters.
    */
   boolean shouldValidateQueryParams();
+
+  /**
+   * Config for whether this method will need to validate path/resource keys.
+   */
+  boolean shouldValidateResourceKey();
```
evanw555: nit: this should be pluralized, right? Based on my understanding, there can be multiple keys per method call (e.g. batch method)

restli-server/src/main/java/com/linkedin/restli/server/config/RestLiMethodConfig.java
```diff
@@ -19,4 +18,9 @@
    * Gets whether query parameter validation against its parameter data template is enabled
    */
   boolean shouldValidateQueryParams();
+
+  /**
+   * Gets whether resource/path keys validation against its parameter data template is enabled
+   */
+  boolean shouldValidateResourceKey();
```
evanw555: nit: same as other comment.

evanw555:
> A general comment:
> 
> I am ok with setting this config to RestliConfig level, but still want to ask, do you have any considerations between setting this on "RestLiConfig" level versus "RestliMethodConfig" level for finer granularity?

@junchuanwang This is a good idea because it would allow us to avoid hackily transferring the `RestLiConfig` setting into the `RestLiMethodConfig` setting as a private field. There may be technical limitations due to the fact that this setting is needed when the path is parsed (may be earlier than the actual routing). Aman will need to check this.

aman1309:
> > A general comment:
> > I am ok with setting this config to RestliConfig level, but still want to ask, do you have any considerations between setting this on "RestLiConfig" level versus "RestliMethodConfig" level for finer granularity?
> 
> @junchuanwang This is a good idea because it would allow us to avoid hackily transferring the `RestLiConfig` setting into the `RestLiMethodConfig` setting as a private field. There may be technical limitations due to the fact that this setting is needed when the path is parsed (maybe earlier than the actual routing). Aman will need to check this.

ArgumentBuilder methods would also explode again if we try to use RestliConfig there. We would require similar refactoring of methods like in RestliRouter.
Also, there are 2 cases here validating resource keys in paths and validating keys at method specific location Ex for batch its in requests. We may also want to keep it distinguished for clients too in future.
Both the configs are used at respective place but method config is derived from server config, thats why I would say lets keep it like this for now.

evanw555:
> ArgumentBuilder methods would also explode again if we try to use RestliConfig there. We would require similar refactoring of methods like in RestliRouter.

@aman1309 There seems to be some miscommunication here. I think Junchuan is suggesting that you remove the config from RestLiConfig and control it solely in ResourceMethodConfig. Is this a possibility?

aman1309:
> > ArgumentBuilder methods would also explode again if we try to use RestliConfig there. We would require similar refactoring of methods like in RestliRouter.
> 
> @aman1309 There seems to be some miscommunication here. I think Junchuan is suggesting that you remove the config from RestLiConfig and control it solely in ResourceMethodConfig. Is this a possibility?

Oh my bad, that is not possible since MethodConfig is formed after building the keys and method resolution. we can do the other way around but that would need refactoring which can be done later.

restli-server/src/main/java/com/linkedin/restli/server/config/ResourceMethodConfigImpl.java
```diff
@@ -31,26 +33,46 @@ public boolean shouldValidateQueryParams() {
...
+
   @Override
   public String toString()
   {
     return "ResourceMethodConfigImpl{" +
-            "_timeoutMs=" + _timeoutMs +
-            '}';
+          "_timeoutMs=" + _timeoutMs +
+          ", _validateQueryParams=" + _validateQueryParams +
+          ", _validateResourceKeyParams=" + _validateResourceKeys +
```
evanw555: nit: update this string

restli-server/src/main/java/com/linkedin/restli/server/config/RestLiMethodConfigImpl.java
```diff
@@ -6,11 +6,13 @@
 {
   private final Map<String, Long> _timeoutMsConfig;
   private boolean _validateQueryParams;
+  private boolean _validateResourceKeyParams;
```
evanw555: nit: update this variable

restli-common/src/main/java/com/linkedin/restli/common/ComplexResourceKey.java
```diff
@@ -319,16 +356,24 @@ public void makeReadOnly()
...
   {
     RecordTemplate recordTemplate = wrapWithSchema(dataMap, spec);
     // Validate against the class schema with FixupMode.STRING_TO_PRIMITIVE to parse the
     // strings into the
     // corresponding primitive types.
-    ValidateDataAgainstSchema.validate(recordTemplate.data(),
-                                       recordTemplate.schema(),
-                                       new ValidationOptions(RequiredMode.CAN_BE_ABSENT_IF_HAS_DEFAULT,
-                                                             CoercionMode.STRING_TO_PRIMITIVE));
+    DataSchemaAnnotationValidator validator = new DataSchemaAnnotationValidator(recordTemplate.schema());
```
aman1309: @karthikbalasub  this is code which introduced the bug

PR #218, opened 2020-03-12
Change ValidateSchemaAnnotationTask to ignore exception or errors durng initializaing class by classLoader

junchuanwang:
Current in ValidateSchemaAnnotationTask, we try to find annotation handler by searching a class with certain java annotation ("RestLiSchemaAnnotationHandler") inside the dependencies. Therefore we need to use "Class.forName(<className>)}" methods.

However there might be exception/errors raised during this period of time, such as java.lang.NoClassDefFoundError, java.lang.IllegalAccessError, java.lang.IllegalAccessError

This is mostly likely due to the fact the the user didn't config the grade dependencies properly. For those classes that failed to be instantiated, they will be ignored.

### DIFF
--- a/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/ValidateSchemaAnnotationTask.java
+++ b/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/ValidateSchemaAnnotationTask.java
@@ -173,13 +173,12 @@ public class ValidateSchemaAnnotationTask extends DefaultTask
    * This method will check the java class annotation on the class instantiated by a className;
    * It will search if that class has an annotation matching {@link SCHEMA_HANDLER_JAVA_ANNOTATION},
    * if found, this class name will be added to "foundClasses" list.
+   * if exceptions or error detected during instantiation of the class, this method will return without doing anything.
    *
    * @param name the name of class to be search annotation from.
    * @param foundClasses a list of class names of the classes that contains {@link SCHEMA_HANDLER_JAVA_ANNOTATION}
-   * @return whether this
-   * @throws IOException
    */
-  private void checkHandlerAnnotation(String name, List<String> foundClasses) throws IOException
+  private void checkHandlerAnnotation(String name, List<String> foundClasses)
   {
     if (name.endsWith(CLASS_SUFFIX))
     {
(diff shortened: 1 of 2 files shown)


gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/ValidateSchemaAnnotationTask.java
```diff
@@ -173,34 +173,36 @@ private String toNativePath(final String path)
...
+      Class<?> clazz = null;
       try
       {
-        Class<?> clazz = classForName(clazzName);
-        for (Annotation a : clazz.getAnnotations())
+        clazz = classForName(clazzName);
+      } catch (Exception | Error e)
+      {
+        getProject().getLogger().info("Unexpected exceptions or errors [{}] found during instantiating the class, will skip this class: [{}]", e.getClass(), clazzName);
+        getProject().getLogger().debug("Unexpected exceptions or errors found during instantiating the class [{}], detailed error: ", clazzName, e);
```
nickibi: Why do we need both info and debug logs? One is not enough?

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/ValidateSchemaAnnotationTask.java
```diff
@@ -173,34 +173,36 @@ private String toNativePath(final String path)
    * This method will check the java class annotation on the class instantiated by a className;
    * It will search if that class has an annotation matching {@link SCHEMA_HANDLER_JAVA_ANNOTATION},
    * if found, this class name will be added to "foundClasses" list.
+   * if exceptions or error detected during instantiation of the class, this method will return without doing anything.
```
nickibi: nit: errors

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/ValidateSchemaAnnotationTask.java
junchuanwang: usually people don't open "--debug" flag, so I don't want them to be spammed by so many exception details. If they want to find out what are the concrete exceptions, they can check the debug information.

PR #219, opened 2020-03-12
Introduce an object creation timeout to completely de-couple channel creation from channelPool

nizarm:
Introduce an object creation timeout to completely de-couple channel creation from channelPooll
Remove retry during failure
Fix ssl handshake timeout isssue

### DIFF
--- a/r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
+++ b/r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
@@ -64,6 +64,7 @@ public class AsyncPoolImpl<T> implements AsyncPool<T>
   private final int _maxWaiters;
   private final long _idleTimeout;
   private final long _waiterTimeout;
+  private final long _creationTimeout;
   private final ScheduledExecutorService _timeoutExecutor;
   private final int _minSize;
   private volatile ScheduledFuture<?> _objectTimeoutFuture;
@@ -71,6 +72,8 @@ public class AsyncPoolImpl<T> implements AsyncPool<T>
 
   public static final int MIN_WAITER_TIMEOUT = 300;
   public static final int MAX_WAITER_TIMEOUT = 3000;
+  public static final int DEFAULT_OBJECT_CREATION_TIMEOUT = 2000;
+
 
   private enum State { NOT_YET_STARTED, RUNNING, SHUTTING_DOWN, STOPPED }
 
@@ -232,6 +235,7 @@ public class AsyncPoolImpl<T> implements AsyncPool<T>
     _maxSize = maxSize;
     _idleTimeout = idleTimeout;
     _waiterTimeout = waiterTimeout;
+    _creationTimeout = DEFAULT_OBJECT_CREATION_TIMEOUT; // TODO: currently hard coded to 2 seconds. Expose this as a parameter !
     _timeoutExecutor = timeoutExecutor;
     _maxWaiters = maxWaiters;
     _strategy = strategy;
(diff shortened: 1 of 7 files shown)


r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
```diff
@@ -816,6 +831,40 @@ public long getTime()
     }
   }
 
+  private class ObjectCreationTimeoutCallback implements Callback<T>
```
ChaoLinkedIn: This looks like a duplicate of TimeoutCallback. Can you just reuse that?

nizarm: I wanted to throw a retry exception since it is connection related. I have refactored TimeoutCallback to accommodate this need.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
```diff
@@ -698,17 +709,24 @@ public void onError(final Throwable e)
                 LOG.error("Encountered error while invoking error waiter callback", ex);
               }
             }
+
+            // Now after cancelling all the pending tasks, lets make sure to back off on the creation
+            _rateLimiter.incrementPeriod();
```
ssheng: Why is this moved down?

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
```diff
@@ -646,8 +648,10 @@ public void run(final SimpleCallback callback)
           return;
         }
 
-        _lifecycle.create(new Callback<T>()
-        {
+        // Lets not trust the _lifecycle to timely return a response here.
+        // Embedding the callback inside a timeout callback (ObjectCreationTimeoutCallback)
+        // to force a response withing creationTimeout deadline to reclaim the object slot in the pool
```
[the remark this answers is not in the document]
nizarm: This is taken care in the recent push

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
nizarm: call to this method '_rateLimiter.incrementPeriod()' will schedule the next pending task for object creation execution before the current thread get a chance to execute cancelPendingTasks. So by moving this down after cancelPendingTasks - we make sure that the rate limiting tasks only submit new requests.

r2-netty/src/main/java/com/linkedin/r2/transport/http/client/common/ChannelPoolLifecycle.java
```diff
@@ -118,19 +118,24 @@ public void create(final Callback<Channel> channelCallback)
...
-          if(evt == SslHandshakeCompletionEvent.SUCCESS){
+          if(evt == SslHandshakeCompletionEvent.SUCCESS)
+          {
             channelCallback.onSuccess(c);
             c.pipeline().remove(CHANNELPOOL_SSL_CALLBACK_HANDLER);
           }
+          else if (evt instanceof SslHandshakeCompletionEvent)
+          {
+            Throwable sslException = ((SslHandshakeCompletionEvent) evt).cause();
+            onError(channelCallback, sslException);
```
ssheng: Do we have guarantee that cause() is not null?

nizarm: Yes if it is not instance of SslHandshakeCompletionEvent.SUCCESS - it is guaranteed to have cause() in it.

https://github.com/netty/netty/blob/00afb19d7a37de21b35ce4f6cb3fa7f74809f2ab/handler/src/main/java/io/netty/handler/ssl/SslHandshakeCompletionEvent.java

But yea - it make sense to not depend on this implementation detail and I will add a null check.

FreCap: Let's add a comment here then

FreCap: Shouldn't this also be a retriable exception?

nizarm: Thank you. good point. I will address this in the next release as I already merged the pull request and Identity do not leverage retry exception.

nizarm: @FreCap  - please let me know if you have any other concerns !

PR #223, opened 2020-03-18
Raise exception in ValidateSchemaAnnotationTask if any handler configured not found.

junchuanwang:
In PR: https://github.com/linkedin/rest.li/pull/218

I made a change to ignore errors if the class failed to be instantiated during search for handlers.
The problem with this is that the handlers under the search, might also encounter failure in instantiation and we would accidentally skip any validation to be done by this handler.

This PR is to add an exception when we fail to see all configured handlers initialized and found. By doing this, we can ensure that all configured handlers need to be found before we run validation, otherwise the user will see error message.

### DIFF
--- a/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/ValidateSchemaAnnotationTask.java
+++ b/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/ValidateSchemaAnnotationTask.java
@@ -123,7 +123,7 @@ public class ValidateSchemaAnnotationTask extends DefaultTask
     // if the number of found handlers doesn't match number of configured modules, will throw exception
     if (foundClassNames.size() != expectedHandlersNumber)
     {
-      String errorMsg = String.format("Encountered errors when searching for annotation handlers: total %s configured, but %s handlers found: [%s].",
+      String errorMsg = String.format("Encountered errors when searching for annotation handlers: total %s dependencies configured, but %s handlers found: [%s].",
                                       expectedHandlersNumber,
                                       foundClassNames.size(),
                                       String.join(",", foundClassNames));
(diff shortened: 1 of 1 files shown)


gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/ValidateSchemaAnnotationTask.java
```diff
@@ -89,6 +91,15 @@ public void validateSchemaAnnotation() throws IOException
 
     getProject().getLogger().info("started schema annotation validation");
 
+    int expectedHandlersNumber = ((DefaultConfiguration) _handlerJarPath).getAllDependencies().size();
```
karthikbalasub: Is this assuming one handler per jar?

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/ValidateSchemaAnnotationTask.java
```diff
@@ -108,12 +119,16 @@ public void validateSchemaAnnotation() throws IOException
       scanHandlersInClassPathJar(f, foundClassNames, scannedClass);
     }
 
-    // skip if no handlers configured or found
-    if (handlerJarPathUrls.size() == 0 || foundClassNames.size() == 0)
+    // For now, every schema annotation handler should be in its own module
+    // if the number of found handlers doesn't match number of configured modules, will throw exception
+    if (foundClassNames.size() != expectedHandlersNumber)
```
karthikbalasub: This wouldn't work if one Jar has more than one handle and another fails to load. A better check would be to ensure atleast one handler per jar dependency.

Ideally, we shouldn't use these heuristics. We should provide a way for users to explicity say which class or package to load and we should error if it cannot be loaded.

junchuanwang: I thought about this, and I know you are referring to something like 

foundClassNames.size() < expectedHandlersNumber

That Is why I mentioned:
"every schema annotation handler should be in its own module". Can we enforce this rule?

karthikbalasub: It is hard to enforce. I think this change is okay for now, but create a backlog ticket to fix this cleanly. we could discuss during ideas meeting

PR #225, opened 2020-03-18
Update the default connection creation timeout to 10s from 2s

nizarm:
Update the default connection creation timeout to 10s from 2s
This will be made configurable in a subsequent release

### DIFF
--- a/r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
+++ b/r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
@@ -822,11 +822,11 @@ public class TestAsyncPool
 
       // run for the duration of default creation timeout
       // TODO: parameterize the creation duration when the default creation gets parameterized
-      clockedExecutor.runFor(10000);
+      clockedExecutor.runFor(AsyncPoolImpl.DEFAULT_OBJECT_CREATION_TIMEOUT);
     }
 
     // drain all the pending tasks
-    clockedExecutor.runFor(10000);
+    clockedExecutor.runFor(AsyncPoolImpl.DEFAULT_OBJECT_CREATION_TIMEOUT);
 
     // Make sure that all the creations are failed with CreationTimeout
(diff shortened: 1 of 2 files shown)


r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
```diff
@@ -822,11 +822,11 @@ public void testCreationTimeout(int poolSize, int concurrency) throws Exception
 
       // run for the duration of default creation timeout
       // TODO: parameterize the creation duration when the default creation gets parameterized
-      clockedExecutor.runFor(2000);
+      clockedExecutor.runFor(10000);
```
ChaoLinkedIn: use DEFAULT_OBJECT_CREATION_TIMEOUT

r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
```diff
@@ -822,11 +822,11 @@ public void testCreationTimeout(int poolSize, int concurrency) throws Exception
...
     }
 
     // drain all the pending tasks
-    clockedExecutor.runFor(2000);
+    clockedExecutor.runFor(10000);
```
ChaoLinkedIn: ditto

nizarm:
> You need to change it again next time when you change this back to 2000, right? How about define a constant and use it everywhere?

Thanks - Updated with common constant.

r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
nizarm: Fixed

r2-core/src/test/java/test/r2/transport/http/client/TestAsyncPool.java
nizarm: Fixed

PR #230, opened 2020-03-20
Use manifest-only JAR for TranslateSchemasTask

nickibi:
Use PathingJarUtil to prepare manifest-only JAR for invocations of TranslateSchemasTask

Avoids "error=7, Argument list too long" error

### DIFF
--- a/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/TranslateSchemasTask.java
+++ b/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/TranslateSchemasTask.java
@@ -1,8 +1,11 @@
 package com.linkedin.pegasus.gradle.tasks;
 
+import com.linkedin.pegasus.gradle.PathingJarUtil;
 import com.linkedin.pegasus.gradle.SchemaFileType;
 import java.io.File;
+import java.io.IOException;
 import org.gradle.api.DefaultTask;
+import org.gradle.api.GradleException;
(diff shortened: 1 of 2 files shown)


evanw555:
Now that this logic is duplicated here and in the rest client task (I think), do you think it would be worthwhile to refactor it into its own method in `PathingJarUtil`?

nickibi:
> Now that this logic is duplicated here and in the rest client task (I think), do you think it would be worthwhile to refactor it into its own method in `PathingJarUtil`?
> Edit: It's a pretty small change, so it's up in the air for me.

I don't quite understand what do you mean by " refactor it into own method in 'PathingJarUntil'. I am using PathingJarUtil#generatePathingJar interface to generate pathing jar. If anyone wants to generate pathing jar, they'd better use this interface. The differences are the parameters passed into the interface.

evanw555:
> > Now that this logic is duplicated here and in the rest client task (I think), do you think it would be worthwhile to refactor it into its own method in `PathingJarUtil`?
> > Edit: It's a pretty small change, so it's up in the air for me.
> 
> I don't quite understand what do you mean by " refactor it into own method in 'PathingJarUntil'. I am using PathingJarUtil#generatePathingJar interface to generate pathing jar. If anyone wants to generate pathing jar, they'd better use this interface. The differences are the parameters passed into the interface.

I was referring to the way it's called with a try-catch wrapped around it, but I'm realizing that that's already a very basic usage. I changed my mind, I'll approve this. Thanks for the fix.

PR #232, opened 2020-03-21
Configure Bintray release task

evanw555:
Removes the old MavenDeployer setup, adds a new OSS
publishing mechanism using the Bintray Gradle plugin.
New tasks: generatePomFileForReleasePublication, bintrayUpload.


build_script/bintrayRelease.gradle
```diff
@@ -0,0 +1,89 @@
...
+    vcsUrl = 'https://github.com/linkedin/rest.li.git'
+    githubRepo = 'linkedin/rest.li'
+    licenses = ['Apache-2.0']
+    version {
+      name = project.version
+      released = new Date()
+      vcsTag = "v${project.version}"
+      gpg {
+        sign = true
+      }
```
mockitoguy: I suspect those may be the defaults so you don't have to declare it.

build_script/maven.gradle
```diff
@@ -1,4 +1,5 @@
 //Configures maven plugin in a standard way
+// TODO: this isn't currently being used, delete when we're sure that we don't need it
 project.plugins.withType(MavenPlugin) {
```
mockitoguy: You no longer apply "maven" plugin, so you're right, this is not needed :-)

gradle.properties
```diff
@@ -1,6 +1,5 @@
 version=28.1.33
-sonatypeUsername=please_set_in_home_dir_if_uploading_to_maven_central
```
mockitoguy: nice!

build_script/bintrayRelease.gradle
evanw555: Looks like the plugin automatically picks up `project.version` as the `name` property, and the `released` property is redundant. However, the default is to _not_ sign the artifacts, so I'll need to keep that. I'll also keep the `vcsTag` property.

PR #234, opened 2020-03-25
Add local-release script

evanw555:
Publishes Maven artifacts to `~/local-repo`

How to use: `./local-release`

Tested that internal projects can build against these locally released artifacts, and they can.

### DIFF
--- a/local-release
+++ b/local-release
@@ -1,0 +1,22 @@
+#!/bin/sh
+# The purpose of this script is to publish artifacts to ~/local-repo
+
+LOCAL_REPO="${HOME}/local-repo"
+
+if [ ! -d $LOCAL_REPO ]; then
+  # If this error is encountered, then (1) $HOME is not set correctly or (2) ~/local-repo doesn't exist
+  echo "Cannot perform local release, '$LOCAL_REPO' is not a valid directory"
+  exit 1
+fi
+
+VERSION=$(awk 'BEGIN { FS = "=" }; $1 == "version" { print $2 }' gradle.properties)
+echo "Publishing pegasus $VERSION to $LOCAL_REPO"
+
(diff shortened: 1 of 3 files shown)


gradle.properties
```diff
@@ -4,3 +4,4 @@ org.gradle.configureondemand=true
 org.gradle.parallel=true
 org.gradle.jvmargs=-Xmx4096M
 aggregateFailures=false
+localRepoPath=~/local-repo
```
junchuanwang: Have you tested? did you verify if this expands?

DPUkyle:
lgtm

local-release
```diff
@@ -0,0 +1,22 @@
...
+
+LOCAL_REPO="${HOME}/local-repo"
+
+if [ ! -d $LOCAL_REPO ]; then
+  # If this error is encountered, then (1) $HOME is not set correctly or (2) ~/local-repo doesn't exist
+  echo "Cannot perform local release, '$LOCAL_REPO' is not a valid directory"
+  exit 1
+fi
+
+VERSION=$(awk 'BEGIN { FS = "=" }; $1 == "version" { print $2 }' gradle.properties)
```
karthikbalasub, line 12: Will this handle trailing whitespaces?

local-release
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+# The purpose of this script is to publish artifacts to ~/local-repo
+
+LOCAL_REPO="${HOME}/local-repo"
+
+if [ ! -d $LOCAL_REPO ]; then
```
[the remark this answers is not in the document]
evanw555: Good point. Updated the PR

local-release
evanw555: It won't, fixed that in the latest diff and fixed it in the other files as well. What's interesting is that other Gradle tasks break if there's tailing whitespace, so it's a moot point.

PR #237, opened 2020-03-26
Fix the issue introduced when trim whitespace started clobbering newlines between elements.

karthikbalasub:
Also adds a gradle task for re-formatting PDL, so MP owners can get improvements we make to PDL formatting.

### DIFF
--- a/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
+++ b/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
@@ -1447,6 +1447,7 @@ public class PegasusPlugin implements Plugin<Project>
       }));
     });
 
+    // Helper task for reformatting existing PDL schemas by generating them again.
     project.getTasks().create(sourceSet.getTaskName("reformat", "Pdl"), TranslateSchemasTask.class, task ->
     {
       task.setInputDir(dataSchemaDir);
--- a/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
+++ b/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
@@ -1446,6 +1446,21 @@ public class PegasusPlugin implements Plugin<Project>
         project.getLogger().lifecycle("You can use '-PconvertToPdl.reverse=true|false' to change the direction of conversion.");
       }));
     });
+
+    project.getTasks().create(sourceSet.getTaskName("reformat", "Pdl"), TranslateSchemasTask.class, task ->
+    {
+      task.setInputDir(dataSchemaDir);
+      task.setDestinationDir(dataSchemaDir);
+      task.setResolverPath(getDataModelConfig(project, sourceSet));
+      task.setCodegenClasspath(project.getConfigurations().getByName("pegasusPlugin"));
+      task.setSourceFormat(SchemaFileType.PDL);
+      task.setDestinationFormat(SchemaFileType.PDL);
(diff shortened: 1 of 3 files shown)


gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1446,6 +1446,21 @@ protected void configureConversionUtilities(Project project, SourceSet sourceSet
         project.getLogger().lifecycle("You can use '-PconvertToPdl.reverse=true|false' to change the direction of conversion.");
       }));
     });
+
+    project.getTasks().create(sourceSet.getTaskName("reformat", "Pdl"), TranslateSchemasTask.class, task ->
```
aman1309: do we plan to run this task for migration already done? does it need to run for new migrations?

aman1309: should we make this optional to run? or we run this default whenever we make a change to the format if this is the case are you planning to remove it later?

karthikbalasub: This task is already optional, it is not linked to any build task.
It can be run by running `ligradle reformatPdl`, we don't have to remove it later. One way this can be used is to auto format PDL files. MP owners can use this to keep the formatting consistent.

restli-tools/src/main/java/com/linkedin/restli/tools/data/SchemaFormatTranslator.java
```diff
@@ -60,7 +60,7 @@
 {
   private static final Logger LOGGER = LoggerFactory.getLogger(SchemaFormatTranslator.class);
 
-  private static final Pattern LINE_END_SPACES = Pattern.compile("\\s+$", Pattern.MULTILINE);
+  private static final Pattern LINE_END_SPACES = Pattern.compile(" +$", Pattern.MULTILINE);
```
junchuanwang: I am sorry that my last suggestion introduced an error. I misunderstood "MULTILINE" mode. This mode actually match "$" to "any arbitrary" line's ending. (not current or each line)

Under MULTILINE mode, the correct regular expression should be `\s+?$` so that '\s' is not greedily matching consecutive empty line.
see example: https://regex101.com/r/TfmGhW/1

karthikbalasub: Junchuan, np, and thanks for digging into the issue. 
In this case, I prefer to keep it simple and not change it. Since we are generating the PDL output, we control the whitespaces that needs trimming. And since we use only spaces, this is good.

gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=28.1.34
+version=28.1.35
```
nickibi: Master is already 28.1.35, please rebase master.

karthikbalasub: thanks, fixed.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1446,6 +1446,21 @@ protected void configureConversionUtilities(Project project, SourceSet sourceSet
         project.getLogger().lifecycle("You can use '-PconvertToPdl.reverse=true|false' to change the direction of conversion.");
       }));
     });
+
```
evanw555: Maybe add a comment here explaining the purpose of this task, similar to the comment corresponding to the above task. Other than that, LGTM

PR #239, opened 2020-03-27
Add resource key validation while parsing

aman1309:
Reopened #217 with null schema fix.

### DIFF
--- a/restli-common/src/test/java/com/linkedin/restli/common/TestComplexResourceKey.java
+++ b/restli-common/src/test/java/com/linkedin/restli/common/TestComplexResourceKey.java
@@ -22,6 +22,7 @@ import com.linkedin.data.template.DataTemplateUtil;
 import com.linkedin.data.template.RecordTemplate;
 
 import org.testng.Assert;
+import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 
 public class TestComplexResourceKey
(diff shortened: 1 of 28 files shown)


restli-common/src/test/java/com/linkedin/restli/common/TestComplexResourceKey.java
```diff
@@ -126,30 +127,59 @@ public void testReadOnlyWithNullParams()
...
-            (RecordDataSchema) DataTemplateUtil.parseSchema("{ \"type\" : \"record\", \"name\" : \"omni\", \"fields\" : [ { \"name\" : \"int\", \"type\" : \"int\" } ] }");
+            (RecordDataSchema) DataTemplateUtil.parseSchema("{ \"type\" : \"record\", \"name\" : \"omni\", \"fields\" : [ { \"name\" : \"int\", \"type\" : \"int\", \"validate\": { \"regex\": { \"regex\": \"[0-9][0-9]\" } } } ] }");
 
     public OmniRecord(DataMap map)
     {
       super(map, schema);
     }
   }
+
+  public static class NullSchemaRecord extends RecordTemplate {
```
karthikbalasub, line 179: Also add a test where the key type is NullSchemaRecord

restli-int-test/src/test/java/com/linkedin/restli/examples/TestTyperefCustomDoubleAssociationKeyResource.java
```diff
@@ -69,13 +68,15 @@
   @BeforeClass
   public void initClass() throws Exception
   {
+    System.setProperty(RestLiIntTestServer.VALIDATE_KEYS_PROPERTY_NAME, Boolean.TRUE.toString());
```
karthikbalasub, line 71: System properties are shared. Setting it this way might cause flakiness if tests are running in parallel.

You should ideally pass these as parameter to the init method. (There are similar methods already). To make this generic, you can add a init method that takes in RestLiConfig directly.

PR #240, opened 2020-03-27
Feature/zk batch with jitter

nizarm:
One of the major problems identified during Zookeeper Capacity investigation is the read storm created during deployments.  During deployment, nodes in the cluster de-announce and announce to zookeeper and this can create thousands of watch notifications based on the number of nodes in the deploying cluster. This can create read storms in Zookeeper if all the upstream nodes started reading for every watch update. This read storm is more visible for big clusters with large numbers of upstreams like identity-mt.

This pull request is to batch or delay setting the watch by the configured threshold duration along with applying jitter to randomize the point in which the actual watch is set.

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/D2ClientConfig.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/D2ClientConfig.java
@@ -28,6 +28,7 @@ import com.linkedin.d2.balancer.zkfs.ZKFSTogglingLoadBalancerFactoryImpl;
 import com.linkedin.d2.balancer.zkfs.ZKFSTogglingLoadBalancerFactoryImpl.ComponentFactory;
 import com.linkedin.d2.discovery.stores.zk.ZKPersistentConnection;
 import com.linkedin.d2.discovery.stores.zk.ZooKeeper;
+import com.linkedin.d2.discovery.stores.zk.ZooKeeperStore;
 import com.linkedin.d2.jmx.JmxManager;
 import com.linkedin.d2.jmx.NoOpJmxManager;
 import com.linkedin.r2.transport.common.TransportClientFactory;
@@ -71,6 +72,7 @@ public class D2ClientConfig
   int retryLimit = DEAULT_RETRY_LIMIT;
   boolean warmUp = false;
   int warmUpTimeoutSeconds = WarmUpLoadBalancer.DEFAULT_SEND_REQUESTS_TIMEOUT_SECONDS;
+  int readWindowMs = ZooKeeperStore.DEFAULT_READ_WINDOW_MS;
   int warmUpConcurrentRequests = WarmUpLoadBalancer.DEFAULT_CONCURRENT_REQUESTS;
   DownstreamServicesFetcher downstreamServicesFetcher = null;
   boolean backupRequestsEnabled = true;
@@ -136,7 +138,8 @@ public class D2ClientConfig
                  ZKPersistentConnection zkConnection,
                  ScheduledExecutorService startUpExecutorService,
                  JmxManager jmxManager,
-                 String d2JmxManagerPrefix)
+                 String d2JmxManagerPrefix,
+                 int readWindowMs)
   {
     this.zkHosts = zkHosts;
     this.zkSessionTimeoutInMs = zkSessionTimeoutInMs;
@@ -181,5 +184,6 @@ public class D2ClientConfig
     this.startUpExecutorService = startUpExecutorService;
     this.jmxManager = jmxManager;
     this.d2JmxManagerPrefix = d2JmxManagerPrefix;
+    this.readWindowMs = readWindowMs;
   }
 }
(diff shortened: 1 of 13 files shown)


d2/src/main/java/com/linkedin/d2/discovery/stores/zk/ZooKeeperEphemeralStore.java
```diff
@@ -610,7 +638,19 @@ public void processResult(int rc, String path, Object ctx, Stat stat)
     protected void processWatch(String propertyName, WatchedEvent event)
     {
       // Reset the watch
-      _zk.getChildren(getPath(propertyName), this, this, false);
+      if (_readWindowMs > 0 && _executorService != null)
+      {
+        // Delay setting the watch based on configured _readWindowMs
+        int midPoint = _readWindowMs / 2;
+        int delay = midPoint + ThreadLocalRandom.current().nextInt(midPoint);
```
ChaoLinkedIn: So the delay is in the range of [midPoint, readWindowMs] -- why not [0, readWindowMs]? There is no need to explicitly add latency for the calls.

d2/src/main/java/com/linkedin/d2/balancer/D2ClientConfig.java
```diff
@@ -71,6 +72,7 @@
   int retryLimit = DEAULT_RETRY_LIMIT;
   boolean warmUp = false;
   int warmUpTimeoutSeconds = WarmUpLoadBalancer.DEFAULT_SEND_REQUESTS_TIMEOUT_SECONDS;
+  int readWindowMs = ZooKeeperStore.DEFAULT_READ_WINDOW_MS;
```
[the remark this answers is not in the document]
nizarm: Thanks. Will rename this with the suggested name

d2/src/main/java/com/linkedin/d2/discovery/stores/zk/ZooKeeperEphemeralStore.java
nizarm: The purpose of adding the delay is to reduce the number of reads as discussed in the design document. This will help us to reduce a lot of unwanted watch updates by delaying within the acceptable threshold. Please check the design document and let me know you still have concerns with this approach.

d2/src/main/java/com/linkedin/d2/balancer/D2ClientConfig.java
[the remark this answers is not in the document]
nizarm: @ChaoLinkedIn  - Refactored all the variable names to zookeeperReadWindowMs and updated the pull request branch with the commit

d2/src/main/java/com/linkedin/d2/discovery/stores/zk/ZooKeeperEphemeralStore.java
ChaoLinkedIn: Thanks for the clarification. Reviewed the doc and agree there should be a delay.

d2/src/main/java/com/linkedin/d2/discovery/stores/zk/ZooKeeperStore.java
```diff
@@ -42,8 +42,9 @@
     PropertyEventPublisher<T>,
     PropertyStore<T>
 {
-  private static final Logger           _log =
-                                                 LoggerFactory.getLogger(ZooKeeperStore.class);
+  private static final Logger           _log = LoggerFactory.getLogger(ZooKeeperStore.class);
+
+  public static final int DEFAULT_READ_WINDOW_MS = -1; //disabled by default
```
bohhyang, line 47: any reason not using Timespan?

d2/src/main/java/com/linkedin/d2/balancer/LastSeenBalancerWithFacilitiesFactory.java
```diff
@@ -73,15 +75,21 @@ public LoadBalancerWithFacilities create(D2ClientConfig config)
     }
 
     // init all the stores
-    LastSeenZKStore<ClusterProperties> lsClusterStore = getClusterPropertiesLastSeenZKStore(config, zkPersistentConnection, d2ClientJmxManager);
+    LastSeenZKStore<ClusterProperties> lsClusterStore =
+      getClusterPropertiesLastSeenZKStore(config, zkPersistentConnection, d2ClientJmxManager,
+                                          config._executorService, config.zookeeperReadWindowMs);
```
bohhyang, line 80: just for my understanding, which config is this executor service coming from? where is it defined?

d2/src/main/java/com/linkedin/d2/discovery/stores/zk/ZooKeeperStore.java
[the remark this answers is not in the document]
nizarm: If you see the use of the read window in processWatch function - we are using this to randomize setting the watch. So we may not be able to leverage the TimeSpan here and it may introduce unwanted conversion back to int in processWatch to randomize.

d2/src/main/java/com/linkedin/d2/balancer/LastSeenBalancerWithFacilitiesFactory.java
nizarm: This is coming from R2SchedulerFactory in container

d2/src/main/java/com/linkedin/d2/discovery/stores/zk/ZooKeeperStore.java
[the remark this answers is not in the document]
bohhyang: could the function "getDurationInMilliseconds" in Timespan work for getting a long in milliseconds for the calculation?

PR #246, opened 2020-03-31
Fix #245: Refactor RestRequestBuilderGenerator to accept arg file wit…

### LINKED ISSUE #245: Gradle task generateMainDataTemplate can fail with "error=7, Argument list too long"
Task `generateMainDataTemplate` can fail when asked to process a large amount of schema files, as those files are passed by `` directly to the CLI as an array.

I will refactor `RestRequestBuilderGenerator` and `GenerateRestClientTask` to allow for this possibility.

DPUkyle:
…h '@' syntax

 - Also add test coverage for psvm method

Gradle task refactor is still pending, but this gives @evanw555 a chance to start reviewing.

### DIFF
--- a/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/internal/ArgumentFileGenerator.java
+++ b/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/internal/ArgumentFileGenerator.java
@@ -1,0 +1,67 @@
+/*
+   Copyright (c) 2020 LinkedIn Corp.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package com.linkedin.pegasus.gradle.internal;
+
+import org.gradle.api.GradleException;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.List;
+
+public class ArgumentFileGenerator
+{
+
+  private ArgumentFileGenerator()
+  {
+    // prevent instantiation
+  }
+
+  /**
+   *  Helper method to generate an argument file, containing one argument per line.
+   *  This should be later prefixed with '@' and passed to the final argument of RestRequestBuilderGenerator.
+   *  <br>
+   *  If invoked from a Gradle task, it is recommended pass org.gradle.api.Task#getTemporaryDir()
+   *  as the value for tempDir.
+   *
+   * @param prefix unique value to prepend to the arg file name
+   * @param args the iterable of arguments to write to an arg file, one line per entry
+   * @param tempDir the directory to hold the arg file
+   * @return argument file which can be passed to a CLI or other JavaExec task
+   */
+  public static File createArgFile(String prefix, List<String> args, File tempDir)
+  {
+    File argFile;
+    try {
+      argFile = File.createTempFile(prefix, "arguments", tempDir);
+      Files.write(argFile.toPath(), args);
+    } catch (IOException e) {
+      throw new GradleException("Unable to create argFile", e);
+    }
+    return argFile;
+  }
+
+  /**
+   * Prefixes a File's absolute path with '@', indicating that is in an arg file
+   * @param argFile the argFile to be passed to a CLI
+   * @return absolute path of the file, prefixed with '@'
+   */
+  public static String getArgFileSyntax(File argFile)
+  {
+    return "@" + argFile.getAbsolutePath();
+  }
+}
--- a/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/GenerateRestClientTask.java
+++ b/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/GenerateRestClientTask.java
@@ -7,6 +7,7 @@ import com.linkedin.pegasus.gradle.SharedFileUtils;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -15,6 +16,7 @@ import org.gradle.api.GradleException;
 import org.gradle.api.file.FileCollection;
 import org.gradle.api.tasks.CacheableTask;
 import org.gradle.api.tasks.Classpath;
+import org.gradle.api.tasks.Input;
 import org.gradle.api.tasks.InputDirectory;
 import org.gradle.api.tasks.Internal;
 import org.gradle.api.tasks.OutputDirectory;
@@ -22,6 +24,9 @@ import org.gradle.api.tasks.PathSensitive;
 import org.gradle.api.tasks.PathSensitivity;
 import org.gradle.api.tasks.SkipWhenEmpty;
 import org.gradle.api.tasks.TaskAction;
+
+import static com.linkedin.pegasus.gradle.internal.ArgumentFileGenerator.createArgFile;
+import static com.linkedin.pegasus.gradle.internal.ArgumentFileGenerator.getArgFileSyntax;
 
 
 /**
@@ -49,6 +54,7 @@ public class GenerateRestClientTask extends DefaultTask
   private boolean _restli1FormatSuppressed;
   private boolean _restli2FormatSuppressed;
   private boolean _restli1BuildersDeprecated = true;
+  private boolean _enableArgFile;
 
   @TaskAction
   public void generate()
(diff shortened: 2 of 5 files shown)


restli-tools/src/test/java/com/linkedin/restli/tools/clientgen/TestRestRequestBuilderGeneratorEntryPoint.java
```diff
@@ -0,0 +1,141 @@
+package com.linkedin.restli.tools.clientgen;
```
evanw555: Add license

restli-tools/src/test/java/com/linkedin/restli/tools/clientgen/TestRestRequestBuilderGeneratorEntryPoint.java
```diff
@@ -0,0 +1,141 @@
...
+import org.testng.annotations.Test;
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.Collections;
+
+public class TestRestRequestBuilderGeneratorEntryPoint {
```
evanw555: Is this basically just a copy of the existing test, but using PSVM instead? Would it be possible to keep this in the existing test class, or would that be too messy you think?

restli-tools/src/main/java/com/linkedin/restli/tools/clientgen/RestRequestBuilderGenerator.java
```diff
@@ -51,8 +51,8 @@
  */
 public class RestRequestBuilderGenerator
 {
-  private static final String GENERATOR_REST_GENERATE_DATATEMPLATES = "generator.rest.generate.datatemplates";
-  private static final String GENERATOR_REST_GENERATE_VERSION = "generator.rest.generate.version";
+  protected static final String GENERATOR_REST_GENERATE_DATATEMPLATES = "generator.rest.generate.datatemplates";
+  protected static final String GENERATOR_REST_GENERATE_VERSION = "generator.rest.generate.version";
```
evanw555: Can just be package-private, right?

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/GenerateRestClientTask.java
```diff
@@ -156,11 +168,26 @@ public void generate()
...
+        javaExecSpec.args(sources);
       }).assertNormalExitValue()
     );
   }
 
+  /**
+   * Helper method to generate an argument file, containing one input file per line.
+   * This should be prefixed with '@' and passed to the final argument of RestRequestBuilderGenerator.
+   */
+  private File createArgFile(String prefix, List<String> files) {
```
evanw555: Although this logic is quite simple, it would probably be worthwhile to refactor this into its own utility class within `:gradle-plugins`, something like `ArgumentFileBuilder` (I say "___Builder" to stress that static utilities are unmaintainable) which handles creating the temp file and returning its "ID" (e.g. `@/something/something`). I think this would be a simple refactoring and would allow us to easily reuse it later, plus it removes complexity from this class.

restli-tools/src/main/java/com/linkedin/restli/tools/clientgen/RestRequestBuilderGenerator.java
```diff
@@ -51,8 +51,8 @@
  */
```
evanw555: Can you add a `TODO` here which says that we want to refactor this tool into a nicer CLI app using Commons-CLI?

restli-tools/src/main/java/com/linkedin/restli/tools/clientgen/RestRequestBuilderGenerator.java
```diff
@@ -64,10 +64,19 @@
   public static void main(String[] args)
       throws IOException
   {
+    String[] sources = new String[0];
+
     if (args.length < 2)
     {
       _log.error("Usage: RestRequestBuilderGenerator targetDirectoryPath [sourceFile or sourceDirectory]+");
       System.exit(1);
+    } else if (args.length == 2 && args[1].startsWith("@")) {
```
evanw555: Similar comment as the one above, there should be a utility class within `:restli-tools` which handles the parsing side of things. Although this logic here isn't too complicated, this will proactively keep this sort of stuff from exploding if we need to extend or reuse it, plus it keeps this class clean.

restli-tools/src/test/java/com/linkedin/restli/tools/clientgen/TestRestRequestBuilderGeneratorEntryPoint.java
DPUkyle: Due to the messiness of the sysprop backup/restore, I chose to extract it to a separate class. I'm still worried about parallelism or unwanted side-effects of this, so we should kill this.

restli-tools/src/main/java/com/linkedin/restli/tools/clientgen/RestRequestBuilderGenerator.java
DPUkyle: Yes, I'll try with default access.

restli-tools/src/test/java/com/linkedin/restli/tools/clientgen/TestRestRequestBuilderGeneratorEntryPoint.java
DPUkyle: Will do.

restli-tools/src/main/java/com/linkedin/restli/tools/clientgen/RestRequestBuilderGenerator.java
DPUkyle: Default access works.

restli-tools/src/main/java/com/linkedin/restli/tools/clientgen/RestRequestBuilderGenerator.java
DPUkyle: Added class `ArgumentFileProcessor`

restli-tools/src/main/java/com/linkedin/restli/tools/clientgen/RestRequestBuilderGenerator.java
DPUkyle: Done.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/GenerateRestClientTask.java
DPUkyle: Added class `ArgumentFileGenerator`. Note that it can't live in `restli-tools` as `pegasus-plugins` has no dependency on it.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/internal/ArgumentFileGenerator.java
```diff
@@ -0,0 +1,64 @@
...
+    // prevent instantiation
+  }
+
+  /**
+   *  Helper method to generate an argument file, containing one input file per line.
+   *  This should be later prefixed with '@' and passed to the final argument of RestRequestBuilderGenerator.
+   *
+   * @param prefix unique value to prepend to the arg file name
+   * @param files the iterable to write to an arg file, one line per entry
+   * @param tempDir the directory to hold the file (If invoked from Gradle, use org.gradle.api.Task#getTemporaryDir()
```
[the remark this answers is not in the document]
DPUkyle: Yes I did it this way to use as few Gradle APIs as possible; if we use `Task#getTempoaryDir()` as default storage for the arg File, the caller would have to pass a reference to the `Task` object, which is kind of icky too. Let's leave it as-is.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/internal/ArgumentFileGenerator.java
```diff
@@ -0,0 +1,64 @@
...
+   * @return argument file which can be passed to a CLI or other JavaExec task
+   */
+  public static File createArgFile(String prefix, List<String> files, File tempDir)
```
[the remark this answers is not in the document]
DPUkyle: Fair enough - although I don't intend this to become the "restli standard". We need to stop passing arguments via sysprops on the CLI, and use a proper framework like jcommander. I'm OK with commons-cli too _if_ it can support the argfile syntax. Otherwise I strongly recommend jcommander.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/internal/ArgumentFileGenerator.java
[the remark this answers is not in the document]
evanw555: Now I get it, makes sense.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/internal/ArgumentFileGenerator.java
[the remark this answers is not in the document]
evanw555: Ah, I see. So for now, this will hold us over, and we'll potentially have to use it if we run into the same issue for another task in the near future. The long-term vision will be to migrate all our tools to C-CLI or JCommander.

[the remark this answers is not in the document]
DPUkyle: 👍

generator-test/build.gradle
```diff
@@ -12,7 +12,7 @@ project.idea.module.testSourceDirs.add('src/test/javaPegasus')
 apply from: "${buildScriptDirPath}/dataTemplate.gradle"
 apply from: "${buildScriptDirPath}/avroSchema.gradle"
 
-// generate pdsc files under "unionPegasus" directory with explicit ordering
+// generate pdsc args under "unionPegasus" directory with explicit ordering
```
DPUkyle, line 15: Oh no - IntelliJ refactor got a bit too creative. Let me undo this change.

DPUkyle: Fixed.

PR #249, opened 2020-04-03
Disable Request Waiter Timeout if not within fail fast threshold and Optimize SingleTimeout runnable to not capture callback to deal with the future.cancel behaviour in ScheduledExecutor

nizarm:
Disable Request Waiter Timeout if not within failfast threshold

### DIFF
--- a/r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
+++ b/r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
@@ -71,7 +71,7 @@ public class AsyncPoolImpl<T> implements AsyncPool<T>
   private final RateLimiter _rateLimiter;
 
   public static final int MIN_WAITER_TIMEOUT = 300;
-  public static final int MAX_WAITER_TIMEOUT = 3000;
+  public static final int MAX_WAITER_TIMEOUT = 1000;
   public static final int DEFAULT_OBJECT_CREATION_TIMEOUT = 10000;
 
 
@@ -352,7 +352,7 @@ public class AsyncPoolImpl<T> implements AsyncPool<T>
     boolean create = false;
     boolean reject = false;
     final LinkedDeque.Node<Callback<T>> node;
-    final TimeTrackingCallback<T> callbackWithTracking = new TimeTrackingCallback<T>(callback);
+    Callback<T> callbackWithTracking = new TimeTrackingCallback<T>(callback);
     for (;;)
     {
       TimedObject<T> obj = null;
@@ -374,8 +374,12 @@ public class AsyncPoolImpl<T> implements AsyncPool<T>
           {
             if (_waiters.size() < _maxWaiters)
             {
+              if (isWaiterTimeoutWithinFastFailThreshold())
+              {
+                callbackWithTracking = new WaiterTimeoutCallback(callbackWithTracking);
+              }
               // No objects available and the waiter list is not full; add to waiter list and break out of loop
-              node = _waiters.addLastNode(new WaiterTimeoutCallback(callbackWithTracking));
+              node = _waiters.addLastNode(callbackWithTracking);
               create = shouldCreate();
             }
             else
@@ -438,6 +442,11 @@ public class AsyncPoolImpl<T> implements AsyncPool<T>
         }
       }
     };
+  }
+
+  private boolean isWaiterTimeoutWithinFastFailThreshold()
+  {
+    return _waiterTimeout >= MIN_WAITER_TIMEOUT && _waiterTimeout <= MAX_WAITER_TIMEOUT;
   }
 
   @Override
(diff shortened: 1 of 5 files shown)


r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
```diff
@@ -440,6 +444,11 @@ public boolean cancel()
     };
   }
 
+  private boolean isWaiterTimeoutWithinFastFailThreshold()
```
FreCap, line 447: isWaiterTimeoutEnabled?

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
```diff
@@ -440,6 +444,11 @@ public boolean cancel()
...
+  {
+    return _waiterTimeout >= MIN_WAITER_TIMEOUT && _waiterTimeout <= MAX_WAITER_TIMEOUT;
+  }
+
   @Override
   public void put(T obj)
   {
```
FreCap, line 454: You can also fix the leak in the SingleTimeout

```
  public SingleTimeout(ScheduledExecutorService executor, long timeout, TimeUnit timeoutUnit, T item, Consumer<T> timeoutAction) 
  { 
    ArgumentUtil.ensureNotNull(item,"item"); 
    ArgumentUtil.ensureNotNull(timeoutAction,"timeoutAction"); 
 
    _item = new AtomicReference<>(item); 
    _future = executor.schedule(() -> { 
      T item1 = _item.getAndSet(null); 
      if (item1 != null) 
      { 
        try 
        { 
          timeoutAction.apply(item1); 
        } catch (Throwable e) 
        { 
          LOG.error("Failed to execute timeout action", e); 
        } 
      } 
    }, timeout, timeoutUnit); 
  } 
```

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
FreCap, line 447: isWaiterTimeoutEnabled?

FreCap: within fast fail threshold can be a impl comment

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
nizarm: Done

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
```diff
@@ -374,8 +374,12 @@ public Cancellable get(final Callback<T> callback)
           {
             if (_waiters.size() < _maxWaiters)
             {
+              if (isWaiterTimeoutWithinFastFailThreshold())
```
ssheng: This check should also exist within the constructor and offspring factory where it is passed in. We do address this in a separate PR if you'd like.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
ssheng: SingleTime not part of this PR.

nizarm: You mean - you dont want to include singletimeout? It is already present in this PR as a separate commit following this. I will squash these during merge time

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
nizarm: Why this needs to be there in other places ? checking here should take care of all the cases right?

ssheng: We should not silently ignore configs passed in. If we could fail early in HttpClientFactory.Builder or the Offspring factory user can detect much earlier.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
```diff
@@ -840,15 +850,15 @@ public long getTime()
...
+      _timeout = new SingleTimeout<>(_timeoutExecutor, _waiterTimeout, TimeUnit.MILLISECONDS, callback, (callback1) -> {
 
         synchronized (_lock)
         {
           _waiters.remove(this);
           _statsTracker.incrementWaiterTimedOut();
         }
         LOG.debug("{}: failing waiter due to waiter timeout", _poolName);
-        callback.onError(
+        callback1.onError(
```
FreCap, line 861: I would rather not put a number in the callback variable, something like: callbackIfTimeout

r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java
nizarm: Understood. I will create a separate PR for this.

PR #250, opened 2020-04-03
Add support for custom properties on anonymous complex schemas.

karthikbalasub:


data/src/main/java/com/linkedin/data/schema/grammar/PdlSchemaParser.java
```diff
@@ -756,6 +760,26 @@ private RecordDataSchema parseRecord(
...
   }
 
+  /**
+   * Sets properties on the provided {@link ComplexDataSchema} using data parsed from the provided
+   * {@link AnonymousTypeDeclarationContext}.
+   *
+   * @param source source to read properties from
+   * @param target target on which to set properties
+   */
+  private void setProperties(AnonymousTypeDeclarationContext source, ComplexDataSchema target)
```
junchuanwang, line 770: can `setDocAndProperties` reuse this?

data/src/test/java/com/linkedin/data/schema/grammar/TestPdlSchemaParser.java
```diff
@@ -237,6 +236,32 @@ public void testNamedDataSchemaWithAliases()
 
   }
 
+  @Test
+  public void testComplexTypeWithProperties()
```
evanw555, line 240: Might be overkill, but I'd test to make sure the fields themselves have no properties.

data/src/main/java/com/linkedin/data/schema/grammar/PdlSchemaParser.java
evanw555: No, since `AnonymousTypeDeclarationContext` and `NamedTypeDeclarationContext` are separate classes with no inheritance relationship. The only other option would be to pass in the parent `TypeDeclarationContext` object, but that would still require that the anonymous and named types have their properties extracted separately.

data/src/test/java/com/linkedin/data/schema/grammar/TestPdlSchemaParser.java
karthikbalasub: done

PR #255, opened 2020-04-10
Enable arg file for resolver path to avoid arg too long errors.

karthikbalasub:


generator/src/main/java/com/linkedin/pegasus/generator/PegasusDataTemplateGenerator.java
```diff
@@ -97,7 +98,15 @@ public static void main(String[] args)
 
     final String generateImportedProperty = System.getProperty(PegasusDataTemplateGenerator.GENERATOR_GENERATE_IMPORTED);
     final boolean generateImported = generateImportedProperty == null ? true : Boolean.parseBoolean(generateImportedProperty);
-    PegasusDataTemplateGenerator.run(System.getProperty(AbstractGenerator.GENERATOR_RESOLVER_PATH),
+    String resolverPath = System.getProperty(AbstractGenerator.GENERATOR_RESOLVER_PATH);
+    if (resolverPath != null && ArgumentFileProcessor.isArgFile(resolverPath))
```
DPUkyle, line 102: Nice. Thanks @evanw555 for convincing me to extract `isArgFile(String)` to a utility class.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1493,6 +1493,7 @@ protected void configureConversionUtilities(Project project, SourceSet sourceSet
       }
       task.setKeepOriginal(keepOriginal);
       task.setSkipVerification(skipVerification);
+      task.setEnableArgFile(true);
```
DPUkyle, line 1496: Are you sure you want to globally enable this feature? Evan and I reserved it as off by default, and allowing consumers with javaexec issues to opt-in in their build.gradle files.

evanw555: +1 on this. We've seen before that `gradle-plugins` and `restli-tools` are not always locked to the same version. Kyle and I think it's better to keep it opt-in for now, then make it enabled by default later (or if/when we create newer more resilient tools)

karthikbalasub: Done

evanw555: Actually, after thinking about this a little harder, I've realized that: `convertXXXToPdl` and `reformatXXXPdl` are not part of the normal build process, so enabling arg file by default may not be that risky. However, the same cannot be said for `generateXXXDataTemplate` or `XXXTranslateSchemas`; these should be opt-in IMO.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -527,6 +527,7 @@
   // by default it is disabled, because it triggers hot-reload (even if it results in a no-op)
   private static final String DESTROY_STALE_FILES_ENABLE = "enableDestroyStaleFiles";
   public static final Collection<String> DATA_TEMPLATE_FILE_SUFFIXES = new ArrayList<>();
+  private static final String DATA_TEMPLATE_ENABLE_ARGFILE = "generateDataTemplate.enableArgFile";
```
evanw555, line 530: Is this necessary? Task inputs can be configured directly as such:
```
generateMainDataTemplate {
  enableArgFile = true
}
```
This is how the fix for `generateXXXRestClient` is enabled.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -546,6 +547,7 @@
   private static final String CONVERT_TO_PDL_KEEP_ORIGINAL = "convertToPdl.keepOriginal";
   private static final String CONVERT_TO_PDL_SKIP_VERIFICATION = "convertToPdl.skipVerification";
   private static final String CONVERT_TO_PDL_PRESERVE_SOURCE_CMD = "convertToPdl.preserveSourceCmd";
+  private static final String TRANSLATE_SCHMEAS_ENABLE_ARGFILE = "translateSchemas.enableArgFile";
```
evanw555, line 550: nit: misspelled "schemas"

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
karthikbalasub: Remove the property based logic. We can enable this using task args directly.

PR #256, opened 2020-04-10
Bump spring dependencies to 3.2.18.RELEASE

evanw555:
To avoid security vulnerabilities in spring 3.2.0 - 3.2.14

This is a patch version bump, so there should theoretically be minimal side effects.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,7 +1,5 @@
 version=28.2.4
 group=com.linkedin.pegasus
-sonatypeUsername=please_set_in_home_dir_if_uploading_to_maven_central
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -1,7 +1,5 @@
 version=28.2.4
 group=com.linkedin.pegasus
-sonatypeUsername=please_set_in_home_dir_if_uploading_to_maven_central
-sonatypePassword=please_set_in_home_dir_if_uploading_to_maven_central
```
nickibi, line 4: I am curious about why do we remove this part along with this PR?

PR #258, opened 2020-04-11
Exposing projection information to AbstractVerifiers through the context

logancarmody:
This was a feature request (tracked internally by SI-8229), so that developers could write custom schema verifiers which were aware of the rest.li projections being applied to the model for the given request.


data/src/main/java/com/linkedin/data/schema/validation/ValidateDataAgainstSchema.java
```diff
@@ -241,6 +241,7 @@ else if (element instanceof MutableDataElement)
            validatorElement = new SimpleDataElement(fixed, element.getName(), schema, element.getParent());
         }
         _context._dataElement = validatorElement;
+        _context._dataSchema = schema;
```
karthikbalasub, line 244: This schema is already available as context::dataElement::schema. Why can't you use that?
(see line 227 above)

logancarmody:
The solution that was described by Karthik is adequate for our use case.

Issue #241, opened 2020-03-27
Upgrade gradle


DPUkyle:
Let's build rest.li with a more modern Gradle version. This shouldn't fundamentally affect the artifacts produced.

Even though Gradle 6.3 is out at the time of writing, I suggest we first bump to Gradle 5.2.1, then to Gradle 6.x once things stabilize.

PR #261, opened 2020-04-16
Fix Maven publication for extra artifacts

evanw555:
The existing Maven publication setup wasn't including extra non-standard artifacts
such as fat JARs and testing JARs. This change resolves that and also manually adds
test dependency information into the POM. Ideally, the test artifacts shouldn't be
published, but for now this change is necessary to make it all work.

**Note:**
* Depends on #241 for `maven-publish` plugin fixes added in Gradle 5


build_script/bintrayRelease.gradle
```diff
@@ -51,21 +47,84 @@ def pomConfig = {
...
+    if (!root.dependencies) {
+      root.appendNode('dependencies')
+    }
+
+    def dependencies = root.dependencies.first()
+    ['testCompile', 'testRuntime'].forEach { testConfiguration ->
+      project.configurations.getByName(testConfiguration).dependencies.forEach {
+        // Manually add dependency node
+        def dependency = dependencies.appendNode('dependency')
+        dependency.appendNode('groupId', it.group)
```
junchuanwang, line 81: where does this "it" come from?  I guess this is groovy specific syntax sugar?

junchuanwang:
Looks good to me, I don't see any problem

PR #264, opened 2020-04-21
DarkCluster schema changes: adding multiplierStrategyList and transportClientProperties

davidhoa:
Adding multiplierStrategyList and transportClientProperties for dark clusters.

multiplierStrategyList - this is a list instead of a single value to decoupling adding a new strategy from the rollout. Esp in large services or where a client side dispatching mechanism is chosen, this will allow the new strategy to be picked up as its available, instead of requiring all dark clients to be upgraded first before turning it on. This is a lesson learned from when DegraderLoadBalancerStrategyName was deprecated in favor of DegraderLoadBalancerStrategyList for the same reason.

There are two strategies: RELATIVE_TRAFFIC and CONSTANT_QPS. only RELATIVE_TRAFFIC is implemented so far. The Javadocs in the corresponding classes have more details on how those strategies work.

transportClientProperties - by specifying the transportClientProperties, and having them make their way into the dark service transportClientProperties, the d2 client will automatically use these properties.

Because transportClientProperties has it's own Converter and uses different keys for serialization than what the pegasus object (D2TransportClientProperties) uses, I had to make a change away from using the pegasus object (DarkClusterConfigMap) in both the in-mem and serialized-for-zookeeper representation. Map<String, Object> is the new serialized form, and the DarkClusterConfigMap can still be used in-mem. I tried to minimize the changes required by adding a method accessDarkClusters (which won't be used by Jackson) to do the conversion for users expecting a DarkClusterConfigMap. This also has the benefit of not exposing the underlying Map for threadsafety.

After approval, I will initiate schema review.

### DIFF
--- a/d2-schemas/src/main/pegasus/com/linkedin/d2/DarkClusterConfig.pdl
+++ b/d2-schemas/src/main/pegasus/com/linkedin/d2/DarkClusterConfig.pdl
@@ -21,24 +21,6 @@ record DarkClusterConfig {
   dispatcherOutboundMaxRate: int = 2147483647
 
   /**
-   * There are 2 types of strategy: RELATIVE_TRAFFIC, CONSTANT_QPS. These can be specified in prioritized order and
-   * will be picked in that order depending on availability.
-   */
-  enum multiplierStrategyType {
-
-    /**
-     * This strategy will try to keep the level of traffic on a dark cluster instance relative to the dispatching instance,
-     * as indicated by the multiplier.
-     */
-    RELATIVE_TRAFFIC
-
-    /**
-     * This strategy will try to maintain a certain queries per second, as specified by dispatcherOutboundTargetRate.
-     */
-    CONSTANT_QPS
-  }
-
-  /**
    * Prioritized order of dark cluster multiplier strategies.
    */
   multiplierStrategyList: array[multiplierStrategyType] = ["RELATIVE_TRAFFIC"]
@@ -46,5 +28,5 @@ record DarkClusterConfig {
   /**
    * The transport client properties to use for this dark cluster
    */
-  transportClientProperties: D2TransportClientProperties
+  transportClientProperties: D2TransportClientProperties = {}
 }
--- a/d2-schemas/src/main/pegasus/com/linkedin/d2/DarkClusterConfig.pdl
+++ b/d2-schemas/src/main/pegasus/com/linkedin/d2/DarkClusterConfig.pdl
@@ -21,9 +21,11 @@ record DarkClusterConfig {
   dispatcherOutboundMaxRate: int = 2147483647
 
   /**
-   * Prioritized order of dark cluster multiplier strategies.
+   * Prioritized order of dark cluster multiplier strategies. This is a list to support adding new strategies and having the strategy users
+   * pick it up as they upgrade code versions, versus waiting for all strategy users to upgrade first. This is the same reason
+   * DegraderLoadBalancerStrategyName was replaced by DegraderLoadBalancerStrategyList.
    */
-  multiplierStrategyList: array[multiplierStrategyType] = ["RELATIVE_TRAFFIC"]
+  multiplierStrategyList: array[DarkClusterStrategyName] = ["RELATIVE_TRAFFIC"]
 
   /**
    * The transport client properties to use for this dark cluster
--- a/d2-schemas/src/main/pegasus/com/linkedin/d2/DarkClusterConfig.pdl
+++ b/d2-schemas/src/main/pegasus/com/linkedin/d2/DarkClusterConfig.pdl
@@ -19,4 +19,32 @@ record DarkClusterConfig {
    * Max rate dispatcher can send to dark canary. Measured in qps. Will act as upper bound to protect canaries in case of traffic spikes
    */
   dispatcherOutboundMaxRate: int = 2147483647
+
+  /**
+   * There are 2 types of strategy: RELATIVE_TRAFFIC, CONSTANT_QPS. These can be specified in prioritized order and
+   * will be picked in that order depending on availability.
+   */
+  enum multiplierStrategyType {
+
+    /**
+     * This strategy will try to keep the level of traffic on a dark cluster instance relative to the dispatching instance,
+     * as indicated by the multiplier.
+     */
+    RELATIVE_TRAFFIC
+
+    /**
+     * This strategy will try to maintain a certain queries per second, as specified by dispatcherOutboundTargetRate.
+     */
+    CONSTANT_QPS
+  }
+
+  /**
+   * Prioritized order of dark cluster multiplier strategies.
+   */
+  multiplierStrategyList: array[multiplierStrategyType] = ["RELATIVE_TRAFFIC"]
+
+  /**
+   * The transport client properties to use for this dark cluster
+   */
+  transportClientProperties: D2TransportClientProperties
 }
--- a/d2-schemas/src/main/pegasus/com/linkedin/d2/DarkClusterConfig.pdl
+++ b/d2-schemas/src/main/pegasus/com/linkedin/d2/DarkClusterConfig.pdl
@@ -25,7 +25,7 @@ record DarkClusterConfig {
    * pick it up as they upgrade code versions, versus waiting for all strategy users to upgrade first. This is the same reason
    * DegraderLoadBalancerStrategyName was replaced by DegraderLoadBalancerStrategyList.
    */
-  multiplierStrategyList: array[DarkClusterStrategyName] = ["RELATIVE_TRAFFIC"]
+  DarkClusterStrategyPrioritizedList: array[DarkClusterStrategyName] = ["RELATIVE_TRAFFIC"]
 
   /**
    * The transport client properties to use for this dark cluster
--- a/d2-schemas/src/main/pegasus/com/linkedin/d2/multiplierStrategyType.pdl
+++ b/d2-schemas/src/main/pegasus/com/linkedin/d2/multiplierStrategyType.pdl
@@ -1,0 +1,19 @@
+namespace com.linkedin.d2
+
+/**
+ * There are 2 types of strategy: RELATIVE_TRAFFIC, CONSTANT_QPS. These can be specified in prioritized order and
+ * will be picked in that order depending on availability.
+ */
+enum multiplierStrategyType {
+
+  /**
+   * This strategy will try to keep the level of traffic on a dark cluster instance relative to the dispatching instance,
+   * as indicated by the multiplier.
+   */
+  RELATIVE_TRAFFIC
+
+  /**
+   * This strategy will try to maintain a certain queries per second, as specified by dispatcherOutboundTargetRate.
+   */
+  CONSTANT_QPS
+}
--- a/d2-schemas/src/main/pegasus/com/linkedin/d2/multiplierStrategyType.pdl
+++ b/d2-schemas/src/main/pegasus/com/linkedin/d2/multiplierStrategyType.pdl
@@ -7,8 +7,9 @@ namespace com.linkedin.d2
 enum multiplierStrategyType {
 
   /**
-   * This strategy will try to keep the level of traffic on a dark cluster instance relative to the dispatching instance,
-   * as indicated by the multiplier.
+   * This strategy aims to maintain a proportional amount of incoming QPS at the host level between the source and dark clusters.
+   * Configured with "multiplier". For example a multiplier of 1 would mean the average incoming QPS for a source host equals that
+   * of a dark host. A multiplier of 2 means on average a dark host will receive 2x more traffic than a source host.
    */
   RELATIVE_TRAFFIC
 
--- a/d2/src/main/java/com/linkedin/d2/balancer/config/DarkClustersConverter.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/config/DarkClustersConverter.java
@@ -16,17 +16,24 @@
 
 package com.linkedin.d2.balancer.config;
 
+import com.linkedin.d2.D2TransportClientProperties;
+import com.linkedin.d2.DarkClusterConfig;
 import com.linkedin.d2.DarkClusterConfigMap;
-import com.linkedin.d2.balancer.util.JacksonUtil;
+import com.linkedin.d2.MultiplierStrategyTypeArray;
+import com.linkedin.d2.balancer.properties.PropertyKeys;
+import com.linkedin.d2.balancer.properties.util.PropertyUtil;
+import com.linkedin.d2.multiplierStrategyType;
+import com.linkedin.data.DataList;
 import com.linkedin.data.codec.JacksonDataCodec;
 import com.linkedin.data.schema.validation.CoercionMode;
 import com.linkedin.data.schema.validation.RequiredMode;
-import com.linkedin.data.schema.validation.ValidateDataAgainstSchema;
 import com.linkedin.data.schema.validation.ValidationOptions;
-import java.io.IOException;
-import java.util.Collections;
-import java.util.Map;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
 /**
  * This class converts {@link DarkClusterConfigMap} into a Map
@@ -39,6 +46,9 @@ public class DarkClustersConverter
   private static final JacksonDataCodec CODEC = new JacksonDataCodec();
   private static final ValidationOptions VALIDATION_OPTIONS =
       new ValidationOptions(RequiredMode.FIXUP_ABSENT_WITH_DEFAULT, CoercionMode.STRING_TO_PRIMITIVE);
+  private static final float DARK_CLUSTER_MULTIPLIER_DEFAULT = 0;
+  private static final int DARK_CLUSTER_DISPATCHER_OUTBOUND_MAX_RATE = Integer.MAX_VALUE;
+  private static final int DARK_CLUSTER_DISPATCHER_TARGET_RATE = 0;
 
   @SuppressWarnings("unchecked")
   public static Map<String, Object> toProperties(DarkClusterConfigMap config)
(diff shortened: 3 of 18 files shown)


d2/src/main/java/com/linkedin/d2/balancer/config/DarkClustersConverter.java
```diff
@@ -49,36 +59,115 @@
     }
     else
     {
-      try
-      {
-        String json = CODEC.mapToString(config.data());
-        return JacksonUtil.getObjectMapper().readValue(json, Map.class);
```
zhang-chris, line 55: Can we keep this old behavior of using Jackson to serialize everything and selectively replace each transport properties with output of TransportClientPropertiesConverter?

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
```diff
@@ -109,18 +112,38 @@ public DarkClusterStrategy getOrCreate(@Nonnull String darkClusterName, @Nonnull
...
+    if (darkClusterConfig.hasMultiplierStrategyList())
     {
-      // Falling into this clause means that the user does not want traffic to be sent to the dark cluster, temporary or otherwise.
-      return new NoOpDarkClusterStrategy();
+      MultiplierStrategyTypeArray strategyList = darkClusterConfig.getMultiplierStrategyList();
+      for (com.linkedin.d2.multiplierStrategyType multiplierStrategyType : strategyList)
+      {
+        if (RELATIVE_TRAFFIC.equals(multiplierStrategyType) && darkClusterConfig.getMultiplier() > 0)
+        {
+          if (darkClusterConfig.getMultiplier() <= 0)
```
zhang-chris, line 122: this will never be reached because of previous "darkClusterConfig.getMultiplier() > 0"

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
```diff
@@ -109,18 +112,38 @@ public DarkClusterStrategy getOrCreate(@Nonnull String darkClusterName, @Nonnull
...
+            new BaseDarkClusterDispatcherImpl(darkClusterName, _darkClusterDispatcher, _notifier, _verifierManager);
+          return new RelativeTrafficMultiplierDarkClusterStrategy(_sourceClusterName, darkClusterName, darkClusterConfig.getMultiplier(),
+                                                                  baseDarkClusterDispatcher, _notifier, _facilities.getClusterInfoProvider(),
+                                                                  _random);
+        }
+        else if (CONSTANT_QPS.equals(multiplierStrategyType))
+        {
+          // the constant qps strategy is not yet implemented, continue to the next strategy if it exists
+        }
+        else
```
zhang-chris, line 138: This else statement return breaks the use of strategy list since it will never iterate past the first element. but also made a previous comment on not using a list.

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
zhang-chris: I think it would be a lot cleaner if we had the implementations themselves handle these edge cases rather than manually returning NoOpDarkClusterStrategy. In this case always return new RelativeTrafficMultiplierDarkClusterStrategy but since the multiplier is <= 0 then no requests will be sent out.

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
```diff
@@ -109,18 +112,38 @@ public DarkClusterStrategy getOrCreate(@Nonnull String darkClusterName, @Nonnull
...
+        {
+          // Falling into this clause means that the user does not want traffic to be sent to the dark cluster, temporary or otherwise.
+          return new NoOpDarkClusterStrategy();
+        }
+      }
     }
+    // Reaching here means that the user has not specified a strategy. return the NoOp Strategy.
+    return new NoOpDarkClusterStrategy();
   }
```
zhang-chris, line 147: Prefer using switch/case for clarity.

d2-schemas/src/main/pegasus/com/linkedin/d2/multiplierStrategyType.pdl
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * There are 2 types of strategy: RELATIVE_TRAFFIC, CONSTANT_QPS. These can be specified in prioritized order and
+ * will be picked in that order depending on availability.
+ */
+enum multiplierStrategyType {
+
+  /**
+   * This strategy will try to keep the level of traffic on a dark cluster instance relative to the dispatching instance,
+   * as indicated by the multiplier.
```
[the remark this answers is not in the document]
davidhoa: will change. thanks.

d2-schemas/src/main/pegasus/com/linkedin/d2/multiplierStrategyType.pdl
```diff
@@ -0,0 +1,19 @@
...
+   */
+  RELATIVE_TRAFFIC
+
+  /**
+   * This strategy will try to maintain a certain queries per second, as specified by dispatcherOutboundTargetRate.
```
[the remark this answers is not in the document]
davidhoa: I'd rather not go through another schema review just to add this. I expect it will be added soon.

d2/src/main/java/com/linkedin/d2/balancer/config/DarkClustersConverter.java
davidhoa: that seems pretty hacky. Doing what I did (use the Map<String, Object>) is more consistent with what the code was doing for other stuff, anyway. The only reason I was trying to use the pegasus object for both in-mem and for zookeeper storage was because I didn't see the need to have two formats; with the inclusion of transportClientProperties I do need the separate format. Regarding your proposal, that would require retraversing the DarkClusterConfigMap and accessing the dataMap directly to hardcode some mappings, and wouldn't take into account further changes in the TransportClientPropertiesConverter. So I don't think that's a better option. Doing the Map<String, Object> way may be more code, but it's pretty understandable and in line with the rest of the Converter classes.

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
davidhoa: yeah, sorry about that. I'll fix that code. The reasons I did not want to have the implementations handle it was because then every handleRequest invocation has to do an if check, rather than only when a strategy has to be created. The tradeoff of putting it in createStrategy seems to be worth not doing that if statement in the hot path.

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
davidhoa: doing this will still use the strategylist, because if constant qps isn't available, it will go into the else if and not the else:
if (RELATIVE_TRAFFIC) {
}
else if (CONSTANT_QPS) {
}
else {
}

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
davidhoa: sure.

d2-schemas/src/main/pegasus/com/linkedin/d2/DarkClusterConfig.pdl
```diff
@@ -19,4 +19,14 @@ record DarkClusterConfig {
    * Max rate dispatcher can send to dark canary. Measured in qps. Will act as upper bound to protect canaries in case of traffic spikes
    */
   dispatcherOutboundMaxRate: int = 2147483647
-}
```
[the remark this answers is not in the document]
davidhoa: removing would be backwards incompatible, and I think we can implement it sooner rather than later. What's the cleaner approach? I don't think those two are bound to the current constant qps implementation.

d2-schemas/src/main/pegasus/com/linkedin/d2/DarkClusterConfig.pdl
```diff
@@ -19,4 +19,14 @@ record DarkClusterConfig {
...
\ No newline at end of file
+
+  /**
+   * Prioritized order of dark cluster multiplier strategies.
+   */
+  multiplierStrategyList: array[multiplierStrategyType] = ["RELATIVE_TRAFFIC"]
```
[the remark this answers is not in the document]
zhang-chris: I see, can we include this explanation to the doc?

d2-schemas/src/main/pegasus/com/linkedin/d2/multiplierStrategyType.pdl
```diff
@@ -0,0 +1,19 @@
+namespace com.linkedin.d2
+
+/**
+ * There are 2 types of strategy: RELATIVE_TRAFFIC, CONSTANT_QPS. These can be specified in prioritized order and
+ * will be picked in that order depending on availability.
+ */
+enum multiplierStrategyType {
```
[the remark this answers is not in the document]
zhang-chris: Why do you prefer DarkClusterStrategyType over DarkClusterStrategy? I don't think the Type is necessary here.

d2-schemas/src/main/pegasus/com/linkedin/d2/multiplierStrategyType.pdl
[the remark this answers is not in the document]
zhang-chris: Sure, and for the doc what about

This strategy will try to maintain a certain queries per second to the entire dark cluster.
Configured with "dispatcherOutboundTargetRate".

d2/src/main/java/com/linkedin/d2/balancer/config/DarkClustersConverter.java
zhang-chris: ok

d2-schemas/src/main/pegasus/com/linkedin/d2/DarkClusterConfig.pdl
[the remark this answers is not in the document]
davidhoa: we discussed offline. We will leave it as is, and consider deprecating it later if not needed. To remove it would require a major version bump of pegasus, which is undesirable if it can be avoided.

d2-schemas/src/main/pegasus/com/linkedin/d2/DarkClusterConfig.pdl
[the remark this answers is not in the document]
davidhoa: done.

d2-schemas/src/main/pegasus/com/linkedin/d2/multiplierStrategyType.pdl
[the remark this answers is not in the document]
davidhoa: as discussed, DarkClusterStrategy can't be used because we already have a java class named that. We decided on DarkClusterStrategyName instead.

d2-schemas/src/main/pegasus/com/linkedin/d2/multiplierStrategyType.pdl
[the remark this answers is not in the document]
davidhoa: done.

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
```diff
@@ -109,18 +109,35 @@ public DarkClusterStrategy getOrCreate(@Nonnull String darkClusterName, @Nonnull
...
+      {
+        switch(multiplierStrategyType)
+        {
+          case RELATIVE_TRAFFIC:
+            if (darkClusterConfig.getMultiplier() <= 0)
+            {
+              // it is a valid case to set RELATIVE_TRAFFIC but set the multiplier to zero to
+              // temporarily set the traffic to zero; don't log anything.
+              return new NoOpDarkClusterStrategy();
+            }
```
zhang-chris, line 125: I'm still a fan of exposing an API in DarkClusterStrategy#isValidConfig that the factory can just call into for each strategy type :) 
That way everything is self-contained in each impl and the factory doesn't need to know these details.

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
```diff
@@ -109,18 +109,35 @@ public DarkClusterStrategy getOrCreate(@Nonnull String darkClusterName, @Nonnull
...
-                                                                                              _notifier, _verifierManager);
-      return new RelativeTrafficMultiplierDarkClusterStrategy(_sourceClusterName, darkClusterName, darkClusterConfig.getMultiplier(),
-                                                              baseDarkClusterDispatcher, _notifier, _facilities.getClusterInfoProvider(), _random);
-    }
-    else
-    {
-      // Falling into this clause means that the user does not want traffic to be sent to the dark cluster, temporary or otherwise.
-      return new NoOpDarkClusterStrategy();
+      DarkClusterStrategyNameArray strategyList = darkClusterConfig.getMultiplierStrategyList();
+      for (com.linkedin.d2.DarkClusterStrategyName multiplierStrategyType : strategyList)
```
zhang-chris, line 115: update name

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
```diff
@@ -109,18 +109,35 @@ public DarkClusterStrategy getOrCreate(@Nonnull String darkClusterName, @Nonnull
...
+            BaseDarkClusterDispatcher baseDarkClusterDispatcher =
+              new BaseDarkClusterDispatcherImpl(darkClusterName, _darkClusterDispatcher, _notifier, _verifierManager);
+            return new RelativeTrafficMultiplierDarkClusterStrategy(_sourceClusterName, darkClusterName, darkClusterConfig.getMultiplier(),
+                                                                    baseDarkClusterDispatcher, _notifier, _facilities.getClusterInfoProvider(),
+                                                                    _random);
+          case CONSTANT_QPS:
+            // the constant qps strategy is not yet implemented, continue to the next strategy if it exists
+            break;
+          default:
+            return new NoOpDarkClusterStrategy();
```
zhang-chris, line 135: not necessary since we return no op strategy at end anyway.

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
```diff
@@ -109,18 +109,35 @@ public DarkClusterStrategy getOrCreate(@Nonnull String darkClusterName, @Nonnull
...
+        }
+      }
     }
+    // Reaching here means that the user has not specified a strategy. return the NoOp Strategy.
```
zhang-chris, line 139: can remove comment

d2/src/main/java/com/linkedin/d2/balancer/properties/PropertyKeys.java
```diff
@@ -179,6 +179,8 @@
   public static final String DARK_CLUSTER_MULTIPLIER = "multiplier";
   public static final String DARK_CLUSTER_OUTBOUND_TARGET_RATE = "dispatcherOutboundTargetRate";
   public static final String DARK_CLUSTER_OUTBOUND_MAX_RATE = "dispatcherOutboundMaxRate";
+  public static final String DARK_CLUSTER_MULTIPLIER_STRATEGY_LIST = "multiplierStrategyList";
```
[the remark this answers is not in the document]
davidhoa: updated to DARK_CLUSTER_STRATEGY_LIST = "darkClusterStrategyList"

d2/src/test/java/com/linkedin/d2/balancer/config/DarkClustersConverterTest.java
```diff
@@ -111,11 +129,69 @@ public void testEntriesInClusterConfig()
...
+    Assert.assertEquals(Objects.requireNonNull(returnedTransportClientProperties.getRequestTimeout()).longValue(),
+                        1000, "expected 1000 request Timeout");
+
+  }
+
+  @Test
+  public void testMultipleStrategies()
+  {
+    DarkClusterConfigMap configMap = new DarkClusterConfigMap();
+    DarkClusterStrategyNameArray multiplierStrategyTypeArray = new DarkClusterStrategyNameArray();
```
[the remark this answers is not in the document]
davidhoa: done.

d2/src/test/java/com/linkedin/d2/balancer/config/DarkClustersConverterTest.java
```diff
@@ -111,11 +129,69 @@ public void testEntriesInClusterConfig()
...
+    multiplierStrategyTypeArray.add(DarkClusterStrategyName.CONSTANT_QPS);
+    DarkClusterConfig config = new DarkClusterConfig()
+      .setMultiplierStrategyList(multiplierStrategyTypeArray);
+
+    configMap.put(DARK_CLUSTER_KEY, config);
+
+    // these are defaults that will be set if the fields are missing.
+    config.setMultiplier(0.0f);
+    config.setDispatcherOutboundTargetRate(0);
+    config.setDispatcherOutboundMaxRate(Integer.MAX_VALUE);
```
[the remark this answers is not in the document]
davidhoa: done.

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
davidhoa: changed to darkClusterStrategyName

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
davidhoa: done.

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
davidhoa: ok.

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
davidhoa: sure, I can do this. that gives us the best of both; the factory doesn't need to know these details, and we can optimize for the common codepath at the expense of just one additional call when we create the strategy (instead of at each getOrCreateStrategy)

d2/src/main/java/com/linkedin/d2/balancer/properties/ClusterProperties.java
```diff
@@ -95,9 +102,17 @@ public ClusterProperties(String clusterName,
...
   }
 
+  /**
+   * @deprecated Use the constructor that uses a Map instead of DarkClusterConfigMap. Using this object is not flexible enough to hold
+   * transportClientProperties, because {@link com.linkedin.d2.balancer.config.TransportClientPropertiesConverter} uses different
+   * keys in it's serialization than how Jackson would serialize D2TransportClientProperties. That is problematic since ClusterProperties
+   * already should have had all necessary conversions done, but in this case the pegasus objects don't have a mechanism to allow the conversions.
+   * The solution is to use a Map<String, Object> to pass in the darkClusters.
+   */
+  @Deprecated
```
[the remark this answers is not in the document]
davidhoa: I don't think this is a good idea. It'd be confusing since the getter that jackson would use to serialize would have a Map<String, Object> but the constructor would have something different.

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
davidhoa: ended up declaring a static method in the RelativeTrafficMultiplierDarkClusterStrategy, because you can't have a static method in the interface that you can override in the impl.

PR #265, opened 2020-04-21
PDL Migration.

karthikbalasub:
This PR migrates most of the PDSC schemas to PDL. Only PDSC files that
are needed for tests that require them in PDSC format are not migrated.


restli-tools/src/main/java/com/linkedin/restli/tools/data/SchemaFormatTranslator.java
```diff
@@ -365,8 +365,7 @@ private static NamedDataSchema checkForErrorsAndGetTopLevelSchema(
       return  null;
     }
     DataSchema sourceSchema = topLevelSchemas.get(0);
-    if (!(sourceSchema instanceof NamedDataSchema) ||
-        !((NamedDataSchema) sourceSchema).getFullName().equals(schemaFullname))
+    if (!(sourceSchema instanceof NamedDataSchema))
```
evanw555, line 368: Why did you loosen this check to ignore incorrect filenames?

karthikbalasub: That was not intentional. reverted.
Since some of the test pdscs are in resources/external, the package was different when being translated. But I used custom tasks with paths configured correctly. Forgot to remove this change.

PR #266, opened 2020-04-21
Refactor stream encoder and Add data encoder for protobuf

aman1309:


data/src/main/java/com/linkedin/data/codec/AbstractJacksonDataCodec.java
```diff
@@ -248,7 +248,7 @@ public void objectToJsonGenerator(Object object, JsonGenerator generator, boolea
     protected final JsonGenerator _generator;
     private final boolean _orderMapEntriesByKey;
 
-    protected JacksonTraverseCallback(JsonGenerator generator)
+    public JacksonTraverseCallback(JsonGenerator generator)
```
karthikrg, line 251: Why did we make this public?

data/src/main/java/com/linkedin/data/codec/entitystream/AbstractDataEncoder.java
```diff
@@ -0,0 +1,340 @@
...
+    {
+      _traverseCallback = initTraverseCallback(_out);
+    }
+    catch (IOException e)
+    {
+      _writeHandle.error(e);
+    }
+  }
+
+  abstract protected Data.TraverseCallback initTraverseCallback(OutputStream out) throws IOException;
```
karthikrg, line 99: Suggest renaming this to createTraverseCallback for consistency with the blocking codec

data/src/main/java/com/linkedin/data/codec/entitystream/AbstractDataEncoder.java
```diff
@@ -0,0 +1,340 @@
...
+        {
+          DataMap dataMap = preProcessMap((DataMap) current);
+          if (dataMap != null)
+          {
+            _iteratorStack.push(createIterator(dataMap));
+            _traverseCallback.startMap(dataMap);
+          }
+          else
+          {
+            _stack.pop();
```
karthikrg, line 208: This part seems to be the same as the else condition in the list. Suggest moving this to a common method?

data/src/main/java/com/linkedin/data/codec/entitystream/AbstractDataEncoder.java
```diff
@@ -0,0 +1,340 @@
...
+    }
+    catch (IOException e)
+    {
+      _writeHandle.error(e);
+    }
+  }
+
+  abstract protected Data.TraverseCallback initTraverseCallback(OutputStream out) throws IOException;
+
+  abstract protected void flushOutputStream(Data.TraverseCallback traverseCallback) throws IOException;
```
karthikrg, line 101: This method contract seems weird. What if we made TraverseCallback implement Closeable and called close on it?

data/src/main/java/com/linkedin/data/codec/AbstractJacksonDataCodec.java
aman1309: to access in entity stream encoders

data/src/main/java/com/linkedin/data/codec/entitystream/AbstractDataEncoder.java
aman1309: I initially thought it would be backward compatible but since I override Callback in AbstractJacksonDataEncoder now it can be done. Thanks

data/src/main/java/com/linkedin/data/codec/entitystream/AbstractDataEncoder.java
```diff
@@ -0,0 +1,337 @@
...
+        catch (Exception e)
+        {
+          _writeHandle.error(e);
+          break;
+        }
+      }
+    }
+  }
+
+  @SuppressWarnings("unchecked")
```
ndarshanrestless, line 187: I understand that it is moved from "AbstractJacksonDataEncoder"
What runtime unchecked exceptions are expected ?

data/src/test/java/com/linkedin/data/codec/entitystream/TestProtobufDataEncoder.java
```diff
@@ -0,0 +1,59 @@
...
+import com.linkedin.data.codec.CodecDataProviders;
+import com.linkedin.data.codec.ProtobufDataCodec;
+import com.linkedin.entitystream.CollectingReader;
+import com.linkedin.entitystream.EntityStream;
+import com.linkedin.entitystream.EntityStreams;
+import org.testng.annotations.Test;
+
+import static org.testng.Assert.*;
+
+
```
ndarshanrestless, line 35: What percentage of test coverage does this provide for ProtobufDataEncoder.java (in-fact for AbstractDataEncoder)  ?

data/src/main/java/com/linkedin/data/codec/entitystream/AbstractDataEncoder.java
```diff
@@ -0,0 +1,337 @@
...
+      }
+      else
+      {
+        try
+        {
+          generate();
+        }
+        catch (Exception e)
+        {
+          _writeHandle.error(e);
```
ndarshanrestless, line 180: worth logging ?

data/src/main/java/com/linkedin/data/codec/entitystream/AbstractDataEncoder.java
```diff
@@ -0,0 +1,337 @@
...
+          break;
+        }
+      }
+    }
+  }
+
+  @SuppressWarnings("unchecked")
+  private void generate()
+      throws Exception
```
ndarshanrestless, line 189: Consider "defensive coding". If possible, exactly list the specific exception instead of generic "Exception" which will the   caller prepare and handle it better. Or may be in this specific situation that is overkill as you not conditionally handling it in the caller ?

data/src/main/java/com/linkedin/data/codec/entitystream/AbstractDataEncoder.java
aman1309: line 242

data/src/main/java/com/linkedin/data/codec/entitystream/AbstractDataEncoder.java
aman1309: this should be done in writeHandle.error if required

data/src/main/java/com/linkedin/data/codec/entitystream/AbstractDataEncoder.java
aman1309: Exception is used here to avoid missing any kind of exception in any caller and that writeHandle.error method is invoked with certainty.

data/src/test/java/com/linkedin/data/codec/entitystream/TestProtobufDataEncoder.java
aman1309: 82%. missing null and IOException cases. will try to add them

data/src/main/java/com/linkedin/data/codec/entitystream/AbstractJacksonDataEncoder.java
```diff
@@ -43,61 +37,28 @@
...
-    }
-    catch (IOException e)
-    {
-      _writeHandle.error(e);
-    }
+    _generator = _jsonFactory.createGenerator(out);
+    return new JacksonStreamTraverseCallback(_generator);
   }
 
   protected void writeStartObject() throws IOException
```
karthikbalasub, line 64: I find it odd that both the traverse callback and the encoder can write using _generator. Ideally all writing to _generator should be done in the traverse callback (ie no _generator member in the Encoder class). 

These protected methods wouldn't be needed as the overriding codecs can override using the traverse callback now. 
I see these are used only by the LICOR codec. Should we clean this up? This would be breaking API change, so would require a major version bump, but the code would be cleaner.
@karthikrg what do you think?

aman1309: I overrode callback not to wait for major version. should we do major version bump just for this? I can add a todo when we plan to do major version bump this can be added with it.

li-protobuf/src/main/java/com/linkedin/data/protobuf/ProtoWriter.java
```diff
@@ -357,4 +358,11 @@ public void writeString(String value, Function<Integer, Byte> leadingOrdinalGene
       throw new EOFException(String.format("Pos: %d, limit: %d, len: %d", _position, _limit, 1));
     }
   }
+
+  @Override
+  public void close() throws IOException
+  {
+    flush();
+    _out.close();
```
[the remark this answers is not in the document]
aman1309: close method is to be called last, there must be no flush pending after its invoked. so no there shouldn't be any failure.
will update DataCodec using traverse callback.close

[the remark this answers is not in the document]
karthikrg: I agree with Aman. As long as we make DataCodec use TraverseCallback.close(), we should be good here.

data/src/main/java/com/linkedin/data/codec/entitystream/AbstractJacksonDataEncoder.java
karthikrg: I am ok with doing this breaking change and bumping major version. Agree it is cleaner. I didn't raise this before, because we've typically avoided breaking changes to the framework.

data/src/main/java/com/linkedin/data/codec/AbstractJacksonDataCodec.java
```diff
@@ -115,20 +115,11 @@ protected JsonGenerator createJsonGenerator(Writer out) throws IOException
 
   protected void writeObject(Object object, JsonGenerator generator) throws IOException
   {
-    try
+    try (Data.TraverseCallback callback = createTraverseCallback(generator))
     {
-      Data.TraverseCallback callback = createTraverseCallback(generator);
       Data.traverse(object, callback);
       generator.flush();
```
karthikbalasub, line 121: Why do you need to flush here?
The behavior should be consistent: traverseCallback handles flush/close on the generator. And this should be documented on createTraverseCallback method

data/src/main/java/com/linkedin/data/codec/entitystream/AbstractJacksonDataEncoder.java
karthikbalasub: On second thought we don't need to introduce a major version for this. Just deprecate these methods and point to the travers callback as the way to make customizations

data/src/main/java/com/linkedin/data/codec/AbstractJacksonDataCodec.java
aman1309: generate close implementation has a condition on flush before stream close that why kept it here

karthikbalasub: My point is about consistency between different codecs. Protobuf codec assumes the callback handles the flush, while this code has to call it explicitly. 

We should make it consistent and set the expectation that the traverse callback should handle flushing/closing the generator.
Then the codec's would just close the traverse callback.

aman1309: moved flush to callback close

data/src/main/java/com/linkedin/data/codec/entitystream/AbstractDataEncoder.java
```diff
@@ -0,0 +1,337 @@
...
+    {
+      _traverseCallback = createTraverseCallback(_out);
+    }
+    catch (IOException e)
+    {
+      _writeHandle.error(e);
+    }
+  }
+
+  abstract protected Data.TraverseCallback createTraverseCallback(OutputStream out) throws IOException;
```
karthikbalasub, line 99: Please add javadoc on this method.
Document that the callback returned should take ownership of the output stream and close it before the callback is closed.

data/src/main/java/com/linkedin/data/codec/entitystream/AbstractJacksonDataEncoder.java
```diff
@@ -43,330 +37,115 @@
...
   }
 
+  // method is moved to @Data.TraverseCallback. Extend JacksonStreamTraverseCallback to override method behaviour.
+  @Deprecated
   protected void writeStartObject() throws IOException
   {
     _generator.writeStartObject();
   }
 
+  // method is moved to @Data.TraverseCallback. Extend JacksonStreamTraverseCallback to override method behaviour.
```
karthikbalasub, line 71: Make these javadoc comments so they appear in api docs..

PR #267, opened 2020-04-22
Optimize Zookeeper Read during Announcing/DeAnnouncing

nizarm:
1. Introduce a URI hash-based ephemeral prefix instead of the existing ephemoral
2. All the Announcers use the URI hashcode to prefix the ephemeral node getting created during markUp
3. ZookeeperEphemeralStore uses the Prefix Generator to filter out or remove unwanted child reads during markUp/markDown
4. A D2 cluster of Size 1000 nodes can create 2.7 million Zookeeper read requests just from the deploying nodes during a service deploy
5. This enhancement helps to reduce the 2.7 million reads down to just under 3K reads: A reduction by 99%


d2/src/main/java/com/linkedin/d2/balancer/servers/AnnouncerUriHashPrefixGenerator.java
```diff
@@ -0,0 +1,54 @@
...
+  public String getAnnouncedUriHashCode()
+  {
+    if (_announcer == null || _announcer.getUri() == null)
+    {
+      return ZooKeeperEphemeralStore.DEFAULT_PREFIX;
+    }
+
+    int hashCode = Math.abs(_announcer.getUri().hashCode());
+    return Integer.toString(hashCode);
+  }
```
ChaoLinkedIn, line 53: Still prefer to use the full hostName + others as the prefix. I do not see big benefit using hashing -- you need to generated hash value every time so no CPU savings. More importantly, using the host name as the node name can immediately help me to find the right node when I browse the zk directory. That's very helpful for manual checking!

bohhyang: agree on the value of keeping the hostname. Could we append the hostName after the hash code? like "HashCode_HostName_"?

I think the hashCode could save us a few reads on some nodes in the case that many URIs (from multiple applications) are hosted on one host.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZookeeperPrefixChildFilter.java
```diff
@@ -0,0 +1,69 @@
...
+  public ZookeeperPrefixChildFilter(ZookeeperEphemeralPrefixGenerator prefixGenerator)
+  {
+    _prefixGenerator = prefixGenerator;
+  }
+
+  @Override
+  public List<String> filter(List<String> children)
+  {
+    List<String> filteredChildren = new ArrayList<>();
+    for (String child : children)
```
bohhyang, line 46: use java 8 stream?

d2/src/main/java/com/linkedin/d2/balancer/servers/ZookeeperPrefixChildFilter.java
```diff
@@ -0,0 +1,69 @@
...
+    {
+      int separatorIndex = child.lastIndexOf('-');
```
bohhyang, line 48: if you take the previous comment about having "HashCode-HostName-" as the prefix, then this could be searching for the first index instead of the last?

d2/src/main/java/com/linkedin/d2/balancer/servers/AnnouncerUriHashPrefixGenerator.java
bohhyang: I mean "HashCode-HostName-" (with hyphens instead of underscores). The filter class could only match the hashCode part, and leave the hostname for manual debugging.

d2/src/main/java/com/linkedin/d2/discovery/stores/zk/ZooKeeperEphemeralStore.java
```diff
@@ -74,6 +75,7 @@
   private static final Logger                                      _log =
                                                                          LoggerFactory.getLogger(ZooKeeperEphemeralStore.class);
   private static final Pattern PATH_PATTERN    = Pattern.compile("(.*)/(.*)$");
+  public static final String DEFAULT_PREFIX = "ephemoral";
```
bohhyang, line 78: ephemeral?

d2/src/main/java/com/linkedin/d2/discovery/stores/zk/ZooKeeperEphemeralStore.java
```diff
@@ -165,6 +196,8 @@ public ZooKeeperEphemeralStore(ZKConnection client,
       useNewWatcher = true;
     }
 
+    _zookeeperChildFilter = zookeeperChildFilter == null ? (children -> children) : zookeeperChildFilter;
```
bohhyang, line 199: "children -> children" could be Function::identity?

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperUriStoreFactory.java
```diff
@@ -0,0 +1,48 @@
...
+import com.linkedin.d2.discovery.stores.zk.ZookeeperChildFilter;
+import com.linkedin.d2.discovery.stores.zk.ZookeeperEphemeralPrefixGenerator;
+import com.linkedin.d2.discovery.stores.zk.builder.ZooKeeperEphemeralStoreBuilder;
+
+/**
+ * A factory class to create {@link ZooKeeperEphemeralStore < UriProperties >}
+ *
+ * @author Nizar Mankulangara (nmankulangara@linkedin.com)
+ */
+public class ZooKeeperUriStoreFactory implements ZooKeeperConnectionManager.ZKStoreFactory<UriProperties, ZooKeeperEphemeralStore<UriProperties>>
```
rachelhanhan, line 17: Will this class replace the existing class ZKUriStoreFactory.java? Having 2 very similar classes looks confusing to me, if the plan is to use this class going forward, we can probably mark ZKUriStoreFactory as deprecated and add the doc to explain.
(I don't really see any non-testing usage of ZKUriStoreFactory today, looks like container creates its own impl of the factory interface)

d2/src/main/java/com/linkedin/d2/balancer/servers/AnnouncerUriHashPrefixGenerator.java
nizarm: Thanks - If we prefer hostname - I am okay with it too. I want to keep it simple and avoid combining both hostname and hashcode as the value of combining is not that significant. Also we may have other repercussions if the size of the znode is too big than the current one.

d2/src/main/java/com/linkedin/d2/balancer/servers/AnnouncerHostPrefixGenerator.java
```diff
@@ -0,0 +1,52 @@
...
+  {
+    if (hostName == null)
+    {
+      _hostName = null;
+    }
+    else
+    {
+      // Since just want to use the machine name for pre-fix and not the entire FQDN to reduce the size of name
+      int machineNameEndIndex = hostName.indexOf('.');
+      _hostName = machineNameEndIndex > 0 ? hostName.substring(0, machineNameEndIndex) : hostName;
```
ChaoLinkedIn, line 43: cross colo announcement might have duplicate host name. should be fine since we still have unique ID. still we can consider adding fabric name.

nizarm: Thank you. Since the fabric name is part of the machine name - we will be having the fabric name.

PR #275, opened 2020-04-29
Enable tests in Travis, use retries for some tests

evanw555:
Enabling tests in the Travis CI build to give us higher confidence in proposed changes. Addresses tests with CI flakiness:

- Tests that are flaky in Travis are added to the group `ci-flaky`, which is excluded only in the CI environment (i.e. they'll still run locally).
- Some flaky tests were given one retry using the `SingleRetry` class, though they should still be investigated (note that some tests might need a retry even when run locally).
- I've created two internal tracking tickets to track addressing the flaky tests: one for R2/D2-related tests, and one for everything else.

**Other changes:**
- Added a new script `scripts/travis/build.sh` which is run for each tag, master, and each PR. However, the tests are skipped when building on a tag to avoid flaky releases.
- Added extra logging for test failures to show the actual assertion message.
- Added an override in `RestLiIntegrationTest` for the client timeout to be increased from `10000` to `20000` when being run in the CI environment.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -290,6 +290,16 @@ subprojects {
     }
   }
 
+  // Exclude tests which are known to be flaky in the Travis CI environment
+  if (System.getenv('TRAVIS') == 'true' && System.getenv('USER') == 'travis') {
+    afterEvaluate {
+      project.tasks.withType(Test).forEach {
+        it.options.excludeGroups 'ci-flaky'
+        it.systemProperties['test.httpRequestTimeout'] = '20000'
+      }
+    }
+  }
+
   task asyncTests(type: Test) {
     useTestNG() {
       includeGroups 'async'
@@ -389,6 +399,14 @@ project.gradle.addListener(new TestListener() {
 
   @Override
   void afterTest(TestDescriptor testDescriptor, TestResult result) {
+    // Display test failure messages since these aren't shown by default
+    if (result.failures) {
+      logger.error "\nFailure message for ${testDescriptor.getClassName()} > ${testDescriptor.getDisplayName()}:"
+      result.failures.forEach {
+        logger.error "    ${it.getMessage()}"
+      }
+    }
+    // Accumulate info about skipped tests, display at the end of the build as a warning
     if (result.skippedTestCount > 0)
     {
       skippedTests[testDescriptor.className] << testDescriptor.name
@@ -399,10 +417,10 @@ project.gradle.addListener(new TestListener() {
 project.gradle.buildFinished { BuildResult result ->
   if (skippedTests.size() > 0)
   {
-    final StringBuilder msgBuilder = new StringBuilder("\nThe following tests should be executed but actually skipped. Please fix and try again.");
+    final StringBuilder msgBuilder = new StringBuilder("\nThe following tests were skipped or failed and had to retry.");
     skippedTests.each { className, methods ->
-      msgBuilder.append("\nClass: ${className} , Methods: ${methods}")
+      msgBuilder.append("\n    ${className} > ${methods}")
     }
-    throw new GradleException(msgBuilder.toString())
+    logger.warn msgBuilder.toString()
   }
 }
--- a/d2/src/test/java/com/linkedin/d2/balancer/servers/ZookeeperConnectionManagerTest.java
+++ b/d2/src/test/java/com/linkedin/d2/balancer/servers/ZookeeperConnectionManagerTest.java
@@ -1,5 +1,6 @@
 package com.linkedin.d2.balancer.servers;
 
+import com.linkedin.test.util.retry.ThreeRetries;
 import java.io.IOException;
 import java.net.URI;
 import java.util.HashMap;
@@ -352,7 +353,7 @@ public class ZookeeperConnectionManagerTest
     shutdownManager(manager);
   }
 
-  @Test(invocationCount = 10, timeOut = 10000)
+  @Test(invocationCount = 10, timeOut = 10000, retryAnalyzer = ThreeRetries.class)
   public void testMarkUpDuringSessionExpirationManyCallbacks()
     throws Exception
   {
@@ -437,7 +438,7 @@ public class ZookeeperConnectionManagerTest
     executorService.shutdown();
   }
 
-  @Test(invocationCount = 10, timeOut = 10000)
+  @Test(invocationCount = 10, timeOut = 10000, retryAnalyzer = ThreeRetries.class)
   public void testMarkUpAndDownMultipleTimesFinalUp()
     throws Exception
   {
--- a/restli-int-test/src/test/java/com/linkedin/restli/examples/instrumentation/TestLatencyInstrumentation.java
+++ b/restli-int-test/src/test/java/com/linkedin/restli/examples/instrumentation/TestLatencyInstrumentation.java
@@ -45,6 +45,7 @@ import com.linkedin.restli.examples.instrumentation.api.InstrumentationControl;
 import com.linkedin.restli.examples.instrumentation.client.LatencyInstrumentationBuilders;
 import com.linkedin.restli.examples.instrumentation.server.LatencyInstrumentationResource;
 import com.linkedin.restli.server.RestLiServiceException;
+import com.linkedin.test.util.retry.SingleRetry;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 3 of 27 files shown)


d2/src/test/java/com/linkedin/d2/balancer/servers/ZookeeperConnectionManagerTest.java
```diff
@@ -352,7 +353,7 @@ public void testMarkUpDuringSessionExpiration()
     shutdownManager(manager);
   }
 
-  @Test(invocationCount = 10, timeOut = 10000)
+  @Test(invocationCount = 10, timeOut = 10000, retryAnalyzer = SingleRetry.class) // TODO: temporary retry to alleviate CI failures, should be investigated
```
junchuanwang, line 356: It is not necessary but I think 3 times will be better, sort of industry standard?

d2/src/test/java/com/linkedin/d2/balancer/servers/ZookeeperConnectionManagerTest.java
```diff
@@ -437,7 +438,7 @@ public void testMarkUpAndDownMultipleTimesFinalDown()
     executorService.shutdown();
   }
 
-  @Test(invocationCount = 10, timeOut = 10000)
+  @Test(invocationCount = 10, timeOut = 10000, groups = { "ci-flaky" })
```
junchuanwang, line 441: what does this group do? Is there a way to disable all ci-flaky tests after you group them?

test-util/src/main/java/com/linkedin/test/util/VariableRetries.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Allows N retries for a given test method. Subclass implementations must specify the value of N.
+ *
+ * Note that the same instance is used for all iterations of a test method, meaning that even if there are multiple
+ * iterations (e.g. data provider provides multiple sets of input) only N retries will be allowed.
+ *
+ * @author Evan Williams
+ */
+public abstract class VariableRetries implements IRetryAnalyzer
```
junchuanwang, line 31: I am afraid I am not a fan of this naming. How about just `Retires` or `RetrySetup` ?

d2/src/test/java/com/linkedin/d2/balancer/servers/ZookeeperConnectionManagerTest.java
evanw555: All tests in the group `ci-flaky` are conditionally excluded if being run in Travis. If being run locally, they will still be run. I added logic for this in the root `build.gradle`, search for the reference to `ci-flaky`

test-util/src/main/java/com/linkedin/test/util/VariableRetries.java
evanw555: Haha I agree, I originally named it `NRetries` but I thought it looked weird. I can just do `Retries` or something like that.

d2/src/test/java/com/linkedin/d2/balancer/servers/ZookeeperConnectionManagerTest.java
evanw555: I can create another implementation for 3 retries. I think I will add this as part of the dev guide:
- If test is flaky in CI, add single retry.
- If still flaky, use three retries.
- If still flaky, add to `ci-flaky` group.

Or do you think this is excessive?

li-bsoetarman:
I wonder, should we just spend a bit more time investigating and finding the root cause of these flaky tests rather than deferring them?  I recently tried to troubleshoot this directly on the tester build machine; while nothing is really conclusive, it maybe caused by some tests run in parallel and interfering with each other.  We can try removing the gradle parallel settings, and/or change the port allocations from static to finding unused ports during tests (for example: zookeeper port, http port, etc.).

evanw555:
@li-bsoetarman I agree that we should solve the issues rather than ignore them, however we need to have the testing enabled by 5/31, since all releases will be via GitHub after that date and we need testing here to ensure that the releases are safe. The current strategy is to identify and retry/disable the flaky ones, then come back afterward (I've created tickets for this) and address the problems. I personally don't have any bandwidth between now and 5/31 to spend more time on this.

I'll give you some observations based on what I've seen:
- Most failures are due to timeouts.
- I tried disabling the parallel build earlier and it still failed, though perhaps it might help. We'd have to investigate.
- The port issue may explain the problems in different environments, we can look into this.

build.gradle
```diff
@@ -290,6 +290,16 @@ subprojects {
     }
   }
 
+  // Exclude tests which are known to be flaky in the Travis CI environment
+  if (System.getenv('TRAVIS') == 'true' && System.getenv('USER') == 'travis') {
```
nickibi, line 294: Should we check System.getenv('TRAVIS') != null first?

d2/src/test/java/com/linkedin/d2/balancer/servers/ZookeeperConnectionManagerTest.java
nickibi: Second Junchuan, 3 times retry is kind of industry standard. Any reason to set it for 10?

build.gradle
evanw555: Doesn't this check accomplish that? `null != 'true'`, and there's no risk of NPE

restli-int-test/src/test/java/com/linkedin/restli/examples/instrumentation/TestLatencyInstrumentation.java
```diff
@@ -167,37 +168,19 @@ public void afterMethod() throws Exception
...
    * @param forceException whether the upstream and downstream resources should trigger the error response path
    * @param timingImportanceThreshold impacts which keys are included in the request context
    */
-  @Test(dataProvider = "latencyInstrumentation")
+  @Test(dataProvider = "latencyInstrumentation", retryAnalyzer = SingleRetry.class)
   public void testLatencyInstrumentation(boolean useStreaming, boolean forceException, boolean useScatterGather,
       TimingImportance timingImportanceThreshold) throws RemoteInvocationException, InterruptedException
   {
-    AssertionError testFailure = null;
-    for (int iteration = 0; iteration <= MAX_RETRIES; iteration++)
```
karthikbalasub, line 182: remove MAX_RETRIES constant

scripts/travis/build.sh
```diff
@@ -0,0 +1,29 @@
...
+if [ "$TRAVIS" != "true" ] || [ "$USER" != "travis" ]; then
+  echo "This script should only be run by Travis CI."
+  exit 2
+fi
+
+# Output something every 9 minutes, otherwise Travis will abort after 10 minutes of no output
+while sleep 9m; do echo "[Ping] Keeping Travis job alive ($((SECONDS / 60)) minutes)"; done &
+WAITER_PID=$!
+
+# Skip tests if building a tag to prevent flaky releases
```
karthikbalasub, line 13: Why do we need this? With the change to skip flaky tests this shouldn't be a problem, right?
Also, even if release job fails, we can retry again?

d2/src/test/java/com/linkedin/d2/balancer/servers/ZookeeperConnectionManagerTest.java
karthikbalasub: Did you check the tests in ci-flaky group with 3 retires?

scripts/travis/build.sh
evanw555: It seems to me that it would be redundant. All PRs to master are tested, and all commits on master are tested. Given that release tags are always on master, we would essentially be testing the same code 3 times.

d2/src/test/java/com/linkedin/d2/balancer/servers/ZookeeperConnectionManagerTest.java
evanw555: @nickibi The retry here (`SingleRetry`) allows for one retry. The `invocationCount = 10` expression means that the test will run 10 times and must pass all 10 times.

@karthikbalasub Nope, I excluded tests that failed even with a retry. I'll try another iteration that replaces `ci-flaky` with 3 retries. Then we'll see which ones really need to be excluded.

evanw555:
@ChaoLinkedIn I'll do another iteration that disables the parallel build and enables all the tests to see.

scripts/travis/build.sh
karthikbalasub: Okay, makes sense.

PR #278, opened 2020-05-01
D2 load balancer strategy test framework

rachelhanhan:
Creating a new test framework to test d2 load balancer strategies.

Background: We are creating a new load balancer strategy with a different set of d2 properties. There are mainly 2 changes in the new strategy:
1. The new strategy is using the average latency to make the degrading decision. It was previously using a hard-coded manual value. This would reduce the client effort of figuring out the value.
2. We will also refactor the strategy to make the code cleaner.

With the change we are proposing, it is important to have a test framework to help us understand the performance before we release this change. The test framework allows users to input number of hosts, different d2 properties, qps, latency etc. The purpose of the test framework is to help with different test scenarios with least user input and good flexibility. Once the framework is done, we will perform integration tests and simulation tests using the framework to verify the correctness and performance difference of different load balancer strategies.


d2-int-test/build.gradle
```diff
@@ -13,9 +13,11 @@ dependencies {
   compile externalDependency.zookeeper
```
zhang-chris, line 13: Pegasus follows older code style such as bracket on new line. See here: https://iwww.corp.linkedin.com/wiki/cf/display/ENGS/Service+Infrastructure+Java+Coding+Conventions

all files should have the copyright statement at the top ... which reminds me I need to include in my PR too

d2/src/test/java/com/linkedin/d2/balancer/strategies/framework/LoadBalancerStrategyTestRunnerBuilder.java
```diff
@@ -0,0 +1,261 @@
...
+  private Map<URI, Integer> _currentRequestCountMap = new HashMap<>();
+  private Map<URI, Integer> _callCountMap = new HashMap<>();
+  private Map<URI, Long> _latencySumMap = new HashMap<>();
+
+  public LoadBalancerStrategyTestRunnerBuilder(final loadBalancerStrategyType type,
+      Map<String, Object> strategyProperties, final String serviceName, final int numHosts) {
+    if (strategyProperties != null) {
+      _strategyProperties.putAll(strategyProperties);
+    }
+    this._strategyProperties.put(PropertyKeys.CLOCK, _clockedExecutor);
```
zhang-chris, line 60: no need for this prefix

d2-int-test/src/test/java/com/linkedin/d2/loadbalancer/strategies/TestLoadBalancerStrategy.java
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * Integration tests for all the load balancer strategies
+ */
+public class TestLoadBalancerStrategy {
+  private static final String DUMMY_SERVICE_NAME = "dummyService";
+
+  @Test
+  public void testStrategyV3HealthPoints() {
+    long healthyLatency = 10L;
+    long unhealthyLatency = 3100L; // For strategyV3, the default highLatency is 3000
```
zhang-chris, line 26: don't use magic number, reference a constant

d2-int-test/src/test/java/com/linkedin/d2/loadbalancer/strategies/TestLoadBalancerStrategy.java
```diff
@@ -0,0 +1,116 @@
...
+
+    LoadBalancerStrategyTestRunner testRunner = new LoadBalancerStrategyTestRunnerBuilder(loadBalancerStrategyType.DEGRADER, null, DUMMY_SERVICE_NAME, 5)
+        .addStrategyProperty(HTTP_LB_CONSISTENT_HASH_ALGORITHM, "pointBased")
```
zhang-chris, line 29: reference a constant instead of string

d2-int-test/src/test/java/com/linkedin/d2/loadbalancer/strategies/TestLoadBalancerStrategy.java
```diff
@@ -0,0 +1,116 @@
...
+    assertEquals(pointsMap.get(testRunner.getUri(2)).intValue(), 100);
+    assertEquals(pointsMap.get(testRunner.getUri(3)).intValue(), 100);
+    assertEquals(pointsMap.get(testRunner.getUri(4)).intValue(), 100);
+  }
+
+  @Test
+  public void testStrategyV3HealthPointsRecover() {
+    LatencyQPSCorrelation unhealthyLatencyCorrelation = new LatencyQPSCorrelation() {
+      @Override
+      public long getLatency(int requestsPerInterval) {
```
zhang-chris, line 57: convert to lambda here and everywhere else

d2-int-test/src/test/java/com/linkedin/d2/loadbalancer/strategies/TestLoadBalancerStrategy.java
```diff
@@ -0,0 +1,116 @@
...
+import org.testng.annotations.Test;
+
+import static com.linkedin.d2.balancer.properties.PropertyKeys.*;
+import static org.testng.Assert.*;
+
+
+/**
+ * Integration tests for all the load balancer strategies
+ */
+public class TestLoadBalancerStrategy {
```
zhang-chris, line 20: This framework is a great start. In addition to these strategy specific tests, can we define test scenarios generically in such a way that any strategy can be plugged into it? I'm imagining something comparable to a DataProvider in unit tests.

We can define several scenarios which we will consider our "standards" that any strategy should pass. For example:
1. A cluster of 5 hosts has 1 extremely unhealthy. We can assert that in 5(?) intervals, that host should be fully degraded by any strategy using its default configs. 
2. Continuing the previous scenario, a fully degraded host is now perfectly healthy. We can assert that in 5 intervals the host is now fully recovered by any strategy using its default configs.
3. etc, etc

evanw555:
I'd like to discuss a few things with you about this PR:
1. Is the goal of this to provide test utilities in `d2:testArtifacts` or `d2-int-test`? If so, we need to make sure this aligns with changes I'm going to make soon to isolate all test artifacts in pegasus to their own modules. In other words, all publicly consumable test artifacts in `d2` need to be moved to some module like `d2-test-api`. Let's talk offline about this.
2. I noticed you added Guava as a dependency to these modules. You should really avoid doing so unless absolutely necessary. There are some internal consumers which have compatibility issues with Guava, so we had to remove it from almost all modules and add Gradle logic to fail the build if it gets added (which makes me curious why this didn't fail...)

d2-int-test/build.gradle
rachelhanhan: Good to know! Thank you!

d2-int-test/src/test/java/com/linkedin/d2/loadbalancer/strategies/TestLoadBalancerStrategy.java
rachelhanhan: Added a few default test scenarios

d2-int-test/src/test/java/com/linkedin/d2/loadbalancer/strategies/TestLoadBalancerStrategy.java
rachelhanhan: Fixed

d2-int-test/src/test/java/com/linkedin/d2/loadbalancer/strategies/TestLoadBalancerStrategy.java
rachelhanhan: I changed the tests using the default test scenarios

rachelhanhan:
Checked with Evan offline. Looks like the plan is to move the publicly consumable tests only, this test framework is very specific for d2 load balancer strategy, it will not be used by other products.

Removed guava dependency.

Also added a new plugin - plugin errors in the response, so that the integration tests can also test against high/lowErrorRate.

d2/src/main/java/com/linkedin/d2/balancer/strategies/degrader/DegraderLoadBalancerStrategyV3.java
```diff
@@ -1387,6 +1387,15 @@ void setStrategy(int partitionId, PartitionDegraderLoadBalancerState.Strategy st
     partition.setState(newState);
   }
 
+  /**
+   * Method used to verify the points of URIs in tests
+   * @param partition partition ID
+   * @return A URI to points map
+   */
+  public Map<URI, Integer> getPointsMap(int partition) {
```
ChaoLinkedIn, line 1395: Is it possible to make it package only? We do not want to expose this internal data to everyone yet.

d2/src/test/java/com/linkedin/d2/balancer/strategies/framework/LoadBalancerStrategyTestRunner.java
```diff
@@ -0,0 +1,235 @@
...
+    }, 10, LoadBalancerStrategyTestRunnerBuilder.INTERVAL_IN_MILLIS, TimeUnit.MILLISECONDS);
+    return _clockedExecutor.runFor(LoadBalancerStrategyTestRunnerBuilder.INTERVAL_IN_MILLIS * _numIntervals);
+  }
+
+  /**
+   * Execute one interval with the given request count
+   */
+  private void runInterval()
+  {
+    int currentIntervalIndex = (int) (_clockedExecutor.currentTimeMillis() / LoadBalancerStrategyTestRunnerBuilder.INTERVAL_IN_MILLIS);
```
ChaoLinkedIn, line 190: Do we know the clockedExecutor timer starts with 0? If not it should be currentTimeMillis - previousTimeStamp.

d2/src/main/java/com/linkedin/d2/balancer/strategies/degrader/DegraderLoadBalancerStrategyV3.java
rachelhanhan: Tthe testing framework is not using the same package name, because this particular V3Strategy is under /degrader, but the testing framework is one level above, so I can't make it non public.

I also thought about using @VisibleForTesting annotation, but looks like it is in Guava, and no one has ever used it before in pegasus, so I guess it's not proper to use that annotation either.

I will just remove this method. I will directly invoke return _strategy.getState().getPartitionState(DEFAULT_PARTITION_ID).getPointsMap() to get the points map, they actually already provide a public access. Having the method here just makes it a little bit easier.

d2/src/test/java/com/linkedin/d2/balancer/strategies/framework/LoadBalancerStrategyTestRunner.java
rachelhanhan: Yes, each test will always create a new instance of the testRunner with a new clockedExecutor, we won't reuse the same instance again in a different test. Each test will run the test start from time 0.

d2/src/main/java/com/linkedin/d2/balancer/strategies/degrader/DegraderLoadBalancerStrategyV3.java
rachelhanhan: Resolved the comment, let me know if you have more questions about this one

d2/src/test/java/com/linkedin/d2/balancer/strategies/framework/LoadBalancerStrategyTestRunnerBuilder.java
```diff
@@ -0,0 +1,494 @@
...
+            RestException restException = new RestException(restResponseBuilder.build(), new Throwable("internal error"));
+            callback.onResponse(TransportResponseImpl.error(restException));
+            return;
+          }
+          callback.onResponse(TransportResponseImpl.success(restResponseBuilder.build()));
+        }
+      }, latency, TimeUnit.MILLISECONDS);
+
+      // Collect basic stats
+      synchronized (_lock)
```
ChaoLinkedIn, line 468: Will mockTransportClient handle concurrent requests? I'm trying to understand why we need to have a lock.

rachelhanhan:
Replying to Chao's comment:
Because the granularity of the measurement is on each interval level using the average, creating latency or error to only a proportion of the requests may not bring extra value, since all we care about is just the avg in that interval, no matter how much the deviation is.

In terms of having method level latency difference, for now it does not provide too much value because we only care about interval avg for the entire service. It can be useful if one day we have method level load balancing and we can compare the performance difference

d2/src/test/java/com/linkedin/d2/balancer/strategies/framework/LoadBalancerStrategyTestRunnerBuilder.java
rachelhanhan: You're right, there is only 1 thread calling the tranportClient.restRequest sequentially, I don't need this lock, just removed it.

PR #279, opened 2020-05-01
Make the translate schema task configurable

nickibi:
Make the translate schema task configurable.

Add skipTranslateSchema flag, so that translateSchema task for pdl->pdsc can be configured for local build, to reduce the local build time.

### DIFF
--- a/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
+++ b/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
@@ -563,6 +563,8 @@ public class PegasusPlugin implements Plugin<Project>
   private static final String TRANSLATED_SCHEMAS_DIR = "legacyPegasusSchemas";
   // Enable the use of argFiles for the tasks that support them
   private static final String ENABLE_ARG_FILE = "pegasusPlugin.enableArgFile";
+  // Flag used to skip running translate legacy schema (pdl->pdsc) to save local build time
+  private static final String DISABLE_LEGACY_SCHEMA_TRANSLATION = "pegasusPlugin.disableLegacySchemaTranslation";
 
   @SuppressWarnings("unchecked")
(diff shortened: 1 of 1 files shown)


evanw555:
> The "skipTranslateSchema" provided seems to be an opt-out configuration here. But for voyager-api, don't they want an opt-in configuration? By default they should not have task locally. Only command inside "product-spec" need to find a way to run it.

We can't have it configured as opt-in because then in order for web MPs to build, _all_ MPs that web MPs depend on for schemas would need to opt-in for this task.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -563,6 +563,8 @@
   private static final String TRANSLATED_SCHEMAS_DIR = "legacyPegasusSchemas";
   // Enable the use of argFiles for the tasks that support them
   private static final String ENABLE_ARG_FILE = "pegasusPlugin.enableArgFile";
+  // Flag used to skip running translate schema (pdl->pdsc) to save local build time
+  private static final String SKIP_TRANSLATE_SCHEMA = "skipTranslateSchema";
```
evanw555, line 567: 1. The name of this property is misleading, one might think it applies when converting between PDL and PDSC. I suggest you rename it to something like `pegasusPlugin.disableLegacySchemaTranslation` or `pegasusPlugin.disableLegacySchemas`. I wanted to originally rename this task to this, but I couldn't for compatibility reasons.
2. I suggest you prefix this with `pegasusPlugin.` because I believe that's the best practice. If we run into an issue where one of our properties conflicts with a property from another plugin, that may be difficult to resolve.

nickibi: That's a good suggestion, I will update the name of this property.

nickibi:
No need to add this flag to disable TranslateSchemasTask for local build. Close the PR.

PR #285, opened 2020-05-04
Initial pdl schemas for relative load balancing.

zhang-chris:
Created initial schemas for relative load balancing. This is a new set of configs with some copied over from LoadBalancer/DegraderStrategyProperties. 

D2QuarantineProperties - contains quarantine related properties.
D2RingProperties - for ring properties.
RelativeStrategyProperties - for relative strategy specific properties.


d2-schemas/src/main/pegasus/com/linkedin/d2/D2RelativeStrategyProperties.pdl
```diff
@@ -0,0 +1,125 @@
...
+   * If the error rate during the interval exceeds this value, then we will decrease the health score by downStep.
+   * Values are from 0.0 - 1.0.
+   */
+  highErrorRate: optional double
+
+  /**
+   * If the error rate during the interval is under this value, then we will increase the health score by upStep.
+   * Values are from 0.0 - 1.0.
+   */
+  lowErrorRate: optional double
```
rachelhanhan, line 49: What's the default value here?

rachelhanhan: Same applied to all other optional fields

d2-schemas/src/main/pegasus/com/linkedin/d2/D2RelativeStrategyProperties.pdl
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * Relative load balancer strategy properties for a particular D2 service.
+ *
+ * Balances traffic to hosts within a service by dynamically adjusting a server's health score
+ * based on call statitics compared relatively to the performance of the entire cluster.
+ *
+ * Health score is rated on a scale from 0.0 - 1.0, with 0.0 meaning most unhealthy (all traffic
+ * routed away) and 1.0 meaning most healthy (no traffic routed away). Note that this behavior is
+ * inverse of dropRate in the degrader strategy.
```
rachelhanhan, line 11: The term "degrader strategy" can be changed to "load balancer strategy"?

d2-schemas/src/main/pegasus/com/linkedin/d2/D2QuarantineProperties.pdl
```diff
@@ -0,0 +1,27 @@
...
+ * Properties for D2 quarantine.
+ *
+ * Uhealthy hosts can be placed in quarantine, in which a predefined method is used
+ * to check endpoint health without routing real request traffic to them.
+ */
+record D2QuarantineProperties {
+
+  /**
+   * The max percentage of hosts in the cluster that can be quarantined at the same time.
+   * Values are from 0.0 - 1.0. A value of 0.0 means quarantine is disabled.
```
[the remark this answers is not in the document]
zhang-chris: ok

d2-schemas/src/main/pegasus/com/linkedin/d2/D2RingProperties.pdl
```diff
@@ -0,0 +1,92 @@
...
+    failOnNoMatch: boolean = false
+
+    /**
+     *  Optional config value. if false, don't warn on falling back to random if the uri doesn't match the regex
+     */
+    warnOnNoMatch: boolean = true
+  }
+
+  /**
+   * The highest ratio of unused entries over the total entries of the Ring points that d2 maintains.
```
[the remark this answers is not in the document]
zhang-chris: Discussed offline about defaults, I want to not include defaults in pdl to avoid having an extra layer of "defaults". Currently the priority is:
1. user-configured cfg2
2. global cfg2
3. code
4. pdl

I'd rather just get rid of all pdl defaults since they are commonly overridden anyway and will be more misleading than helpful.

d2-schemas/src/main/pegasus/com/linkedin/d2/D2RelativeStrategyProperties.pdl
zhang-chris: "Degrader" strategy is more specific imo since that is the actual name of the strategy using drop rate. Load balancer strategy is generic and could mean both degrader and relative.

d2-schemas/src/main/pegasus/com/linkedin/d2/D2Service.pdl
```diff
@@ -63,15 +69,24 @@ record D2Service includes D2ChangeTimeStamps {
...
    */
   loadBalancerStrategyProperties: D2LoadBalancerStrategyProperties
 
   /**
-   * The degrader properties for this service
+   * The degrader properties for this service.
+   *
+   * Required along with D2DegraderProperties for the degrader strategy.
    */
   degraderProperties: D2DegraderProperties
```
rachelhanhan, line 83: Do we need to mark these 2 old fields as optional since when client migrates, they can only set relativeStrategyProperties and get rid of these 2 old property fields

rachelhanhan: Wondering why this not optional even today, if user specify RANDOM, then these properties are not used either. Is it not optional for a purpose?

zhang-chris:
Since we are adding "RELATIVE" enum this is technically backwards incompatible change. Do we need to bump major version for this?

d2-schemas/src/main/pegasus/com/linkedin/d2/D2Service.pdl
zhang-chris: I assume it's because any normal use case will have these fields. Random is really for testing I think. I will mark these 2 as optional.

d2-schemas/src/main/pegasus/com/linkedin/d2/D2QuarantineProperties.pdl
```diff
@@ -0,0 +1,27 @@
...
+  /**
+   * Specifies the method used for health checking. Format: <METHOD>:<FULL_PATH>. Defaults to OPTIONS method.
+   */
+  quarantineMethod: optional string
+
+  /**
+   * The latency threshold (in milliseconds) for health check requests. Response times longer
+   * than this threshold are considered as unhealthy.
+   */
+  quarantineLatency: optional long
```
ChaoLinkedIn, line 26: This field is not used anymore -- instead we use lowLatency as the threshold. For relative LB will this become required? If yes it should not be optional.

d2-schemas/src/main/pegasus/com/linkedin/d2/D2RingProperties.pdl
```diff
@@ -0,0 +1,92 @@
...
+
+  /**
+   * How fast should load balancer ramp up node that has been 100% degraded and is being recovered. Defaults to 1.0.
+   */
+  ringRampFactor: optional double
+
+  /**
+   * The number of points per weight given for a particular node in a hash ring. Defaults to 100.
+   */
+  pointsPerWeight: optional int
```
ChaoLinkedIn, line 16: Does this field still make sense for relative LB?

d2-schemas/src/main/pegasus/com/linkedin/d2/D2RingProperties.pdl
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * Optional config value. if true, fail if no regex matches, otherwise fall back to random.
+     */
+    failOnNoMatch: boolean = false
+
+    /**
+     *  Optional config value. if false, don't warn on falling back to random if the uri doesn't match the regex
+     */
+    warnOnNoMatch: boolean = true
+  }
```
ChaoLinkedIn, line 49: I thought relative LB does not support affinity routing? If that's the case you do not need these fields.

d2-schemas/src/main/pegasus/com/linkedin/d2/D2RingProperties.pdl
```diff
@@ -0,0 +1,92 @@
...
+
+  /**
+   * The highest ratio of unused entries over the total entries of the Ring points that d2 maintains.
+   */
+  hashRingPointCleanupRate: optional double
+
+  /**
+   * Consistent hash algorithm the d2 load balancer should use. Defaults to POINT_BASED.
+   */
+  consistentHashAlgorithm: optional enum ConsistentHashAlgorithm {
```
ChaoLinkedIn, line 59: Ditto

d2-schemas/src/main/pegasus/com/linkedin/d2/D2RingProperties.pdl
zhang-chris: It is supported. Underlying ring behavior is unchanged. Relative LB is for adjusting health of servers (points on ring).

d2-schemas/src/main/pegasus/com/linkedin/d2/D2RingProperties.pdl
zhang-chris: See above.

d2-schemas/src/main/pegasus/com/linkedin/d2/D2RingProperties.pdl
zhang-chris: Yes, see comment below.

d2-schemas/src/main/pegasus/com/linkedin/d2/D2QuarantineProperties.pdl
zhang-chris: Removed. 

"For relative LB will this become required?" What are you referring to?

d2-schemas/src/main/pegasus/com/linkedin/d2/D2RingProperties.pdl
ChaoLinkedIn: OK I misunderstood that part.

d2-schemas/src/main/pegasus/com/linkedin/d2/D2RelativeStrategyProperties.pdl
```diff
@@ -0,0 +1,125 @@
...
+   * The health score threshold that determines whether D2 should perform slow start
+   * (by doubling the health score) or the normal stepUp/Down operation.
+   * Values are from 0.0 - 1.0. A value of 0.0 means slow start is disabled.
+   */
+  slowStartThreshold: optional double
+
+  /**
+   * Regular expression used to match a request status code and indicate a server-side error.
+   */
+  errorStatusRegex: optional string
```
asakusuma, line 109: Have you thought about using an array of tuples to describe a list of ranges, rather than using a regex?

d2-schemas/src/main/pegasus/com/linkedin/d2/D2QuarantineProperties.pdl
```diff
@@ -0,0 +1,21 @@
...
+  /**
+   * The max percentage of hosts in the cluster that can be quarantined at the same time.
+   * Values are from 0.0 - 1.0. A value of 0.0 means quarantine is disabled.
+   */
+  quarantineMaxPercent: optional double
+
+  /**
+   * Specifies the method used for health checking. Format: <METHOD>:<FULL_PATH>. Defaults to OPTIONS method.
+   */
+  quarantineMethod: optional string
```
asakusuma, line 20: Could this possibly be an enum for the method, combined with another field for the path? Rather than just a string.

d2-schemas/src/main/pegasus/com/linkedin/d2/D2RelativeStrategyProperties.pdl
zhang-chris: an array of tuples is not as flexible so I'd rather use regex.

d2-schemas/src/main/pegasus/com/linkedin/d2/D2QuarantineProperties.pdl
zhang-chris: Actually, thanks for pointing this out. I will drop this field altogether as I realized now that it is not necessary.

zhang-chris: Yes, I can separate it out into 2 fields:
1. healthCheckMethod: optional enum (Do you know of a PDL for http methods?)
2. healthCheckPath: optional string

d2-schemas/src/main/pegasus/com/linkedin/d2/D2RelativeStrategyProperties.pdl
asakusuma: Given that the domain is pretty narrow (3 digit integers), can you give an example of how a regex is more flexible than a list of tuples?

d2-schemas/src/main/pegasus/com/linkedin/d2/D2QuarantineProperties.pdl
asakusuma: Looks good, thanks

d2-schemas/src/main/pegasus/com/linkedin/d2/D2RelativeStrategyProperties.pdl
zhang-chris: For example, if I wanted to consider any 5xx and 404 as errors I would write (5..|404). I'm not sure I understand your suggestion. What would it look like using an array of tuples?

asakusuma: You could encode that requirement as:
```
[[500,599], [404]]
```
As in `array[array[int]]`

You could also introduce a more sophisticated record type to model a set of error codes. Like:
```
record ErrorCodeRange {
  lowerBound: int
  upperBound: int
}

errorStatusFilter: array[ErrorCodeRange]
```

The point here is that it would be nice if we could model a set of ints in a stricter fashion than a string.

zhang-chris: I see, I like the ErrorCodeRange suggestion. I will do that, thanks.

d2-schemas/src/main/pegasus/com/linkedin/d2/D2RelativeStrategyProperties.pdl
```diff
@@ -0,0 +1,136 @@
...
+
+  /**
+   * The health score threshold that determines whether D2 should perform slow start
+   * (by doubling the health score) or the normal stepUp/Down operation.
+   * Values are from 0.0 - 1.0. A value of 0.0 means slow start is disabled.
+   */
+  slowStartThreshold: optional double
+
+  /**
+   * A request status code that falls within the provided ranges will be considered as a server-side error.
```
asakusuma, line 108: Just to be extra explicit and avoid confusion, can you change "falls within the provided ranges" to "falls within any of the provided ranges"

Basically saying that this is OR, not AND logic. Seems obvious but this sort of thing can trip up external readers.

zhang-chris: good point, will do.

koushikkothagal:
DMRC shippit 2 out of 2. DMRC approved.

PR #286, opened 2020-05-04
Set up automated releases, improve helper scripts

evanw555:
Sets up automatic Bintray publications in Travis whenever a tag is pushed. Improves/moves the `local-release` and `release` scripts. Also bumps the Bintray Gradle plugin, and bumps the project version to `28.3.4`.

Also adds an explicit dependency for `assemble` on `buildTestArtifacts` for projects where test artifacts are published (i.e. `d2`, `darkcluster`). The only reason this wasn't a problem earlier is because `d2` happened to add some custom tasks which resulted in `testJar` ultimately getting run. Once `darkcluster` was added, this wasn't the case and the test artifacts weren't being built.

**Note:**
- This makes the release workflow and helper scripts exactly the same as in ParSeq.
- Once this is merged, I'll test the process by releasing `28.3.4`


gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=28.3.3
+version=28.3.4
```
nickibi, line 1: 28.3.4 is already in master, please rebase master

evanw555: Thanks for letting me know

.travis.yml
```diff
@@ -1,20 +1,40 @@
...
-  - master
-  - /^v\d+\.\d+\.\d+$/
+    - master
+    - /^v\d+\.\d+\.\d+$/
+
+# Set environment variables (includes Bintray credentials)
+env:
+  global:
+    - secure: 'DJbT0+O3W57jxaXmpVvuY4AH5VDqAgvmAwtw9GLK7fHhntmKMh8o43xWU7Tmv5LUPDU/UChXzte5fRPSe75KvkS72Rie02M/MSizO9B9fWGQ5trzWqCqGO4mI38v9e2PfInOWa7R/FA3z1NYas11CulkTXQI+a10po7nzIMrEnM='
+    - secure: 'b4Q5mdqAQJcyuDyvxZK6MiImxho18HteKMDT0UANpIyrFXEfzWKIlB110unofXSXKiI0bAg+ob/PLRULkB/wuj69WasXRiH83+Ao2WpNbWyMoqPwiY/DL8MN0M1RhHbuSMso/O6C34eRImhj5NDcN+nkI+py8pHAq9FNrmQ92O4='
```
aman1309, line 24: can you add super_secret, plain text credentials to our internal wiki? for hotfix releases and bookkeeping
also, are these required here? can we not add these in repository settings?

local-release
```diff
@@ -1,25 +1,5 @@
...
-./gradlew -Dmaven.repo.local=$LOCAL_REPO publishReleasePublicationToMavenLocal
-
-if [ $? -eq 0 ]; then
-  echo "Published pegasus $VERSION to $LOCAL_REPO"
-else
-  exit 1
-fi
 
+# TODO: delete this eventually
+echo 'This script has been moved to ./scripts/local-release'
```
nickibi, line 4: What's the plan to delete this one?

evanw555: Wait until we're comfortable with using the new scripts, then delete these. Probably in a month or two.

nickibi: Good to know that.

.travis.yml
evanw555: I won't add these as plaintext, because these are my personal credentials. I'll add instructions for how to set them yourself or how to replace them.
As for configuring them via repo settings, I'd have to do it in Travis settings rather than GitHub settings (I don't think they can be accessed here). However, I don't have Travis admin permissions for LinkedIn. I see other repos setting it like this, so it seems to be a common practice.

.travis.yml
```diff
@@ -1,20 +1,40 @@
...
 
 # Below skips the installation step completely (https://docs.travis-ci.com/user/customizing-the-build/#Skipping-the-Installation-Step)
-# If we don't skip it, Travis runs unnecessary Gradle tasks like './gradlew assemble'
-install:
-  - true
+# If we don't skip it Travis runs unnecessary Gradle tasks like './gradlew assemble'
+install: true
 
-# Only build on master and release tags
+# Build and test commits to master, PRs to master, and release tags
```
karthikbalasub, line 12: The script is still doing only build (-x test), so this comment should be updated?

evanw555: Oops you're right. I blindly copied this from parseq. I'll update it

.travis.yml
aman1309: is it possible possible to create the account using our team email id? or can we have 1 common account?

scripts/help-text/local-release.txt
```diff
@@ -0,0 +1,11 @@
+Usage: ./scripts/local-release [OPTION]...
+Publishes Rest.li's Maven artifacts to ~/local-repo.
+This script must be run from the root project directory.
+
+Options:
+  -h, --help                    print this help text and exit
+  -s, --snapshot                release a snapshot version
+
+Examples:
+  ./scripts/local-release       publish x.y.z artifacts to ~/local-repo
```
junchuanwang, line 10: here what is the difference between x.y.z and x.y.z-snapshot version? I though both are local snapshots?

.travis.yml
evanw555: No, I asked Tools and the only way to authenticate as of now is to use personal credentials.

PR #293, opened 2020-05-13
Creating a light-weight representation of InJarFileDataSchemaLocation

logancarmody:
When parsing schemas, the AbstractDataSchemaResolver stores bad schema locations in an instance variable. If the DataSchemaLocation is a InJarFileDataSchemaLocation, that means we are storing the entire jar in the _badLocations set, and not allowing it to be garbage collected.
We have seen this cause undue memory pressure / OOMs when resolving Schemas.

This introduces a light-weight representation of InJarFileDataSchemaLocation, by just storing the path to the jar and the path once inside the jar. This fulfills the equals and hashCode requirements of DataSchemaLocation without the overhead of the entire jar.

Discovered by analyzing a heap dump using jxray:
Screen Shot 2020-05-13 at 12 50 12 PM

### DIFF
--- a/data/build.gradle
+++ b/data/build.gradle
@@ -14,6 +14,7 @@ dependencies {
   testCompile externalDependency.commonsIo
   testCompile externalDependency.testng
   testCompile externalDependency.junit
+  testCompile externalDependency.mockito
 
   antlr externalDependency.antlr
 }
--- a/data/src/main/java/com/linkedin/data/schema/DataSchemaLocation.java
+++ b/data/src/main/java/com/linkedin/data/schema/DataSchemaLocation.java
@@ -36,6 +36,19 @@ public interface DataSchemaLocation
    */
   File getSourceFile();
 
+  /**
+   * Return {@link DataSchemaLocation} that is a lightweight representation of this location for storing in-memory.
+   *
+   * For example, {@link com.linkedin.data.schema.resolver.InJarFileDataSchemaLocation}
+   * contains an entire Jar-file in it's impl; this isn't necessary for storing as a {@link DataSchemaLocation},
+   * so it could return a lighter-weight implementation.
+   *
+   * @return a light-weight representation of this DataSchemaLocation. Default is {@code this}
+   */
+  default DataSchemaLocation getLightweightRepresentation() {
+    return this;
+  }
+
   final DataSchemaLocation NO_LOCATION = new DataSchemaLocation()
   {
     @Override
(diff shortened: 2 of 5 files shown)


data/src/main/java/com/linkedin/data/schema/resolver/JarFileDataSchemaLocation.java
```diff
@@ -0,0 +1,69 @@
...
+    _jarFileName = jarFileName;
+    _pathInJar = pathInJar;
+  }
+
+  @Override
+  public boolean equals(Object o)
+  {
+    if (this == o)
+      return true;
+    if (o instanceof JarFileDataSchemaLocation == false)
```
karthikbalasub, line 45: This would break the functionality when a location is checked agains known bad locations.

In that case InJarFileDataSchemaLocation will be checked and it would not match JarFileDataSchemaLocation

logancarmody: Updated to account for this and added unit tests

evanw555:
This is good, however IMO the logic to convert from `InJar...` to `Jar...` (i.e. hardcoding this specific conversion) isn't totally intuitive when reading the resolver code, and the similarity in class names doesn't really help with this. I'm wondering if there's a more elegant way to accomplish this. I have a couple ideas, but they may seem a little excessive so bear with me (it may also be tough given that it's blocking you).

1. We shouldn't do a specific conversion from `InJar...` to `Jar...` in the abstract resolver. One way to resolve this would be to add a method to the `DataSchemaLocation` interface (e.g. `#getLightweightRepresentation()`) so that when adding to the bad locations map, we can just do `_badLocations.add(location.getLightweightRepresentation())`. The interface can provide a default implementation that just returns `this`.
2. An even more excessive approach to the previous idea would be to have the `DataSchemaLocation` interface implement another interface called `Lightweight<DataSchemaLocation>` which adds this `#getLightweightRepresentation()` method. This makes it non-specific to the location type.
3. The naming of the lightweight class is not intuitive in my opinion. It could be renamed to something like `LightweightJarDataSchemaLocation`, but I also feel as though making it an inner static class of `InJarFileDataSchemaLocation` would also be appropriate, being called something like `InJarFileDataSchemaLocation.LightweightRepresentation`.

Please let me know what you think.

logancarmody:
Addressed @evanw555 's comments and added a `getLightweightRepresentation` method

PR #294, opened 2020-05-14
change errors in ChannelPoolLifecycle to warn

davidhoa:
Errors in dark canary dispatchers can throw off validation frameworks. In ChannelPoolLifecycle, if the dark canary is unreachable, you will get errors from netty that are logged in this class. By changing this to warn, it still logs the problem, but at a warn level. If it's truly a problem, such as failing a call, that can and will be detected higher up and can be recorded as an error there. For instance, even outside of dark canary, if it's a RetriableException, a new channel can be created to service the request, so this shouldn't be logged as an error. I will send further details separately.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-version=28.3.8
+version=28.3.9
 group=com.linkedin.pegasus
 org.gradle.configureondemand=true
 org.gradle.parallel=true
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=28.3.7
+version=28.3.8
```
nizarm, line 1: There is a 28.3.8 already released. You might want to consider changing this?

davidhoa: thanks, bumped again.

PR #295, opened 2020-05-14
Stop publishing test artifacts & internal modules

evanw555:
Isolates the consumable test artifacts for `d2` and `darkcluster` into the new modules `d2-test-api` and `darkcluster-test-api`, respectively. Also discontinues publishing "internal" modules to Bintray, with these modules determined as modules with no main source set and no known consumers. Both these changes are backward-incompatible.

**Notable changes:**
- Bumps pegasus to `29.0.0`
- Stops publishing the following modules to Bintray (i.e. stops publishing them at all):
  - `'d2-benchmark'`
  - `'d2-int-test'`
  - `'generator-test`
  - `'log-test-config'`
  - `'r2-int-test'`
  - `'r2-perf-test'`
  - `'restli-internal-testutils'`
- Stops publishing test artifacts (e.g. `testArtifacts` configuration, `-tests` classifier) for all modules (only applies to `d2` and `darkcluster`):
  - Moves consumable classes (e.g. `ZKTestUtils`) to modules `d2-test-api` and `darkcluster-test-api`
  - To make some inner classes (`TrackerClientTest.TestClient`, `WarmupLoadBalancerTest.TestLoadBalancer`) accessible, I moved them to being top-level classes (`TestClient`, `TestLoadBalancer`).

**Open Questions:**
- Since the inner classes are now top-level (`TestClient`, `TestLoadBalancer`), should they be renamed to be less ambiguous?


d2-test-api/src/main/java/com/linkedin/d2/balancer/clients/TestClient.java
```diff
@@ -0,0 +1,150 @@
...
+  public static final int DEFAULT_REQUEST_TIMEOUT = 500;
+  public StreamRequest streamRequest;
+  public RestRequest restRequest;
+  public RequestContext restRequestContext;
+  public Map<String, String> restWireAttrs;
+  public TransportCallback<StreamResponse> streamCallback;
+  public TransportCallback<RestResponse>   restCallback;
+  public ScheduledExecutorService _scheduler;
+
+  public boolean                         shutdownCalled;
```
junchuanwang, line 53: I know this file is not written by you, but this variables naming is not consistent with other class variables, can you fix this by this chance?

d2-test-api/src/main/java/com/linkedin/d2/balancer/clients/TestClient.java
```diff
@@ -0,0 +1,150 @@
...
+  private final boolean _emptyResponse;
+  private boolean _dontCallCallback;
```
junchuanwang, line 55: Also I noticed this, I actually seldom see variable contains "don't", I will rename to "notToCallCallback"... just saying.

junchuanwang:
Overall I have no problem with the remove modules.

d2-test-api/src/main/java/com/linkedin/d2/balancer/clients/TestClient.java
evanw555: Sure, may as well while I'm at it

d2-test-api/src/main/java/com/linkedin/d2/balancer/clients/TestClient.java
evanw555: Yeah, it's sort of weird. I'll consider renaming.

d2-test-api/src/main/java/com/linkedin/d2/balancer/clients/TestClient.java
evanw555: Actually I decided to keep it as-is because these variables are public.

d2-test-api/src/main/java/com/linkedin/d2/balancer/clients/TestClient.java
```diff
@@ -0,0 +1,153 @@
+/*
+   Copyright (c) 2020 LinkedIn Corp.
```
karthikbalasub, line 2: From what I remember, the policy is to keep the original copyright date if this is a simple refactor of the code.
This is the year of first publication.

PR #296, opened 2020-05-14
Fix maxContentLength Issue & Log HttpProtocol used per Service

nizarm:
1. Fix MaxContentLength bug in HttpChannelInitializer
2. Add info log to log information on http protocol used

### DIFF
--- a/r2-netty/src/main/java/com/linkedin/r2/netty/client/http/HttpChannelInitializer.java
+++ b/r2-netty/src/main/java/com/linkedin/r2/netty/client/http/HttpChannelInitializer.java
@@ -72,8 +72,8 @@ class HttpChannelInitializer extends ChannelInitializer<NioSocketChannel>
   private final int _maxInitialLineLength;
   private final int _maxHeaderSize;
   private final int _maxChunkSize;
-  private final int _maxContentLength;
   private final int _sslHandShakeTimeout;
+  private final long _maxContentLength;
   private final boolean _ssl;
   private final boolean _enableSSLSessionResumption;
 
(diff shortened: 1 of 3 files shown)


r2-netty/src/main/java/com/linkedin/r2/netty/client/http/HttpChannelInitializer.java
```diff
@@ -86,7 +86,7 @@
     _maxInitialLineLength = maxInitialLineLength;
     _maxHeaderSize = maxHeaderSize;
     _maxChunkSize = maxChunkSize;
-    _maxContentLength = maxContentLength > Integer.MAX_VALUE ? Integer.MAX_VALUE : (int) maxContentLength;
+    _maxContentLength = maxContentLength;
```
rachelhanhan, line 89: This is in Http2ChannelInitilizer too, should we fix that file as well?

nizarm: We can't fix it in Http2ChannelInitializer as Netty expects an integer value there and the one in Http2ChannelInitializer is not used in StreamHandler

nizarm: The one used in Http2ChannelInitializer is used for parent channel and it is not used to serve the same purpose in HttpChannelInitializer. The http2 equivalent of the same can be found @ Http2StreamChannelInitializer

rachelhanhan: Got it, thanks

PR #299, opened 2020-05-18
Implement Host Override List feature in D2.

ssheng:
**Requirements**
To support features like Fullstack Hovr, D2 needs to support the following routing override scenarios.
  1. Route requests going to a given cluster to a specific host.
  2. Route requests going to a given service to a specific host.
  3. Provide route overrides for multiple clusters and services.

D2 provides an existing override mechanism called TargetHints. However, TargetHints does not meet the above requirements due to the following restrictions.
TargetHints does not allow multiple overrides for different clusters and services.
TargetHints requires users to provide a full URL as opposed to a URL prefix like the D2 announcer. Therefore, we lose the flexibility of appending the service paths for different D2 services.

**Implementation**
Similar to TargetHints, HostOverrideList is also specified and passed through the RequestContext. HostOverrideList allows the user to add an arbitrary number of overrides for different clusters and services. When the request gets to the SimpleLoadBalancer, the implementation will look for the first matching override for the current cluster and service names. The request URI will be rewritten to be the override URI provided.
  1. hostOverrideList.addClusterOverride(“Cluster1”, “https://host:8080/context/”);
  2. hostOverrideList.addServiceOverride(“Service1”, “https://host:8080/context/”);

The provided override URI is a prefix, similar to the announced URI, as opposed to a full URI in the TargetHints mechanism. The service path will be appended to the end of the URI by the SimpleLoadBalancer when the request URI is rewritten. In the above examples, the written URI will be “https://host:8080/context/service1path”.

The TargetHints mechanism will be deprecated but backward compatibility will be kept. To achieve the original TargetHints behavior, one can add generic override without providing a cluster or service name.
  1. hostOverrideList.addOverride(“https://host:8080/context/”);

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancer.java
@@ -39,6 +39,7 @@ import com.linkedin.d2.balancer.properties.UriProperties;
 import com.linkedin.d2.balancer.strategies.LoadBalancerStrategy;
 import com.linkedin.d2.balancer.util.ClientFactoryProvider;
 import com.linkedin.d2.balancer.util.ClusterInfoProvider;
+import com.linkedin.d2.balancer.util.HostOverrideList;
 import com.linkedin.d2.balancer.util.HostToKeyMapper;
 import com.linkedin.d2.balancer.util.KeysAndHosts;
 import com.linkedin.d2.balancer.util.LoadBalancerUtil;
@@ -84,6 +85,7 @@ import static com.linkedin.d2.discovery.util.LogUtil.warn;
 public class SimpleLoadBalancer implements LoadBalancer, HashRingProvider, ClientFactoryProvider, PartitionInfoProvider, WarmUpService,
                                            ClusterInfoProvider
 {
+  private static final String HOST_OVERRIDE_LIST = "HOST_OVERRIDE_LIST";
   private static final Logger     _log =
                                            LoggerFactory.getLogger(SimpleLoadBalancer.class);
   private static final String     D2_SCHEME_NAME = "d2";
@@ -177,10 +179,17 @@ public class SimpleLoadBalancer implements LoadBalancer, HashRingProvider, Clien
 
         // Check if we want to override the service URL and bypass choosing among the existing
         // tracker clients. This is useful when the service we want is not announcing itself to
-        // the cluster, ie a private service for a set of clients.
+        // the cluster, ie a private service for a set of clients. This mechanism is deprecated;
+        // use host override list instead.
+        @SuppressWarnings("deprecation")
         URI targetService = LoadBalancerUtil.TargetHints.getRequestContextTargetService(requestContext);
 
-        if (targetService == null)
+        // Checks if we have a host override list provided in the request context. If present,
+        // get the override URI available override for the current cluster and service names.
+        HostOverrideList overrides = (HostOverrideList) requestContext.getLocalAttr(HOST_OVERRIDE_LIST);
+        URI override = overrides == null ? null : overrides.getOverride(clusterName, serviceName);
+
+        if (targetService == null && override == null)
         {
           LoadBalancerStateItem<UriProperties> uriItem = getUriItem(serviceName, clusterName, cluster);
           UriProperties uris = uriItem.getProperty();
@@ -201,16 +210,26 @@ public class SimpleLoadBalancer implements LoadBalancer, HashRingProvider, Clien
         }
         else
         {
-          _log.debug("service hint found, using generic client for target: {}", targetService);
+          URI target = override == null ? targetService : URI.create(override + service.getPath());
+          if (targetService != null && override != null)
+          {
+            _log.warn("Both TargetHints and HostOverrideList are found. HostOverList will take precedence %s.", target);
+          }
 
-          TransportClient transportClient = _state.getClient(serviceName, targetService.getScheme());
+          if (_log.isDebugEnabled())
+          {
+            _log.debug("Rewrite URI as specified in the TargetHints/HostOverrideList {} for cluster {} and service {}.",
+                target, clusterName, serviceName);
+          }
+          TransportClient transportClient = _state.getClient(serviceName, target.getScheme());
           if (transportClient == null)
           {
-            throw new ServiceUnavailableException(serviceName,
-                "PEGA_1001. Cannot find transportClient for service " + serviceName + " and scheme: " + targetService.getScheme()
-                    + " with service hint" + targetService);
+            throw new ServiceUnavailableException(serviceName, String.format(
+                    "PEGA_1001. Cannot find transportClient for service %s and scheme %s with URI specified in"
+                        + "TargetHints/HostOverrideList %s", serviceName, target.getScheme(), target));
           }
-          clientCallback.onSuccess(new RewriteLoadBalancerClient(serviceName, targetService, transportClient));
+
+          clientCallback.onSuccess(new RewriteLoadBalancerClient(serviceName, target, transportClient));
         }
       }
       catch (ServiceUnavailableException e)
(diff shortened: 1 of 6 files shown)


d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancer.java
```diff
@@ -201,16 +210,17 @@ public void getClient(Request request, RequestContext requestContext, Callback<T
         }
         else
         {
-          _log.debug("service hint found, using generic client for target: {}", targetService);
-
-          TransportClient transportClient = _state.getClient(serviceName, targetService.getScheme());
+          URI target = override == null ? targetService : URI.create(override + service.getPath());
+          _log.debug("Host override found, using generic client for target: {}", target);
```
nizarm, line 214: Eventhough TargetService hint is deprecated - should we include that int he debug (Host override/TargetService Hint found) until we completely remove the TargetService?

d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancer.java
```diff
@@ -201,16 +210,17 @@ public void getClient(Request request, RequestContext requestContext, Callback<T
...
+          TransportClient transportClient = _state.getClient(serviceName, target.getScheme());
           if (transportClient == null)
           {
             throw new ServiceUnavailableException(serviceName,
-                "PEGA_1001. Cannot find transportClient for service " + serviceName + " and scheme: " + targetService.getScheme()
-                    + " with service hint" + targetService);
+                "PEGA_1001. Cannot find transportClient for service " + serviceName + " and scheme: " + target.getScheme()
+                    + " with host override " + target);
```
nizarm, line 220: Ditto - should we include TargetService in the error until we completely remove the TargetHost?

d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancer.java
```diff
@@ -201,16 +210,17 @@ public void getClient(Request request, RequestContext requestContext, Callback<T
         }
         else
         {
-          _log.debug("service hint found, using generic client for target: {}", targetService);
-
-          TransportClient transportClient = _state.getClient(serviceName, targetService.getScheme());
+          URI target = override == null ? targetService : URI.create(override + service.getPath());
```
nizarm, line 213: Do we need to make sure override variable ends with '/' here, to make sure that servicepath is added as a proper suffix?

d2/src/main/java/com/linkedin/d2/balancer/util/HostOverrideList.java
```diff
@@ -0,0 +1,103 @@
...
+    /**
+     * Checks if the provided cluster and service names match this key.
+     * @param cluster Cluster name to check against.
+     * @param service Service name to check against.
+     * @return {@code True} if provided cluster and service name match the key; {@code false} otherwise.
+     */
+    public boolean match(String cluster, String service) {
+      if (_cluster == null && _service == null) {
+        return true;
+      }
```
nizarm, line 91:
```java
if (this.equals(WILDCARD_KEY)) {
 return true;
}
```
?

d2/src/test/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerTest.java
```diff
@@ -623,6 +626,113 @@ public void testGetClient() throws Exception
...
+        "/service1Path",
+        Arrays.asList("degrader"),
+        Collections.<String,Object>emptyMap(),
+        null,
+        null,
+        prioritizedSchemes,
+        null));
+    uriRegistry.put(cluster1, new UriProperties(cluster1, uriData));
+
+    URI override = URI.create("http://override/path");
```
nizarm, line 691: Can we also have a test with override value with an ending slash URI.create("http://override/path/")

d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancer.java
ChaoLinkedIn: Given that override takes precedence, and there are many applications use targetService for now, we probably want to log a warning message here if both present?

d2/src/main/java/com/linkedin/d2/balancer/util/HostOverrideList.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Objects;
+
+
+/**
+ * Stores the list of host overrides for cluster and services. The order of the overrides are stored in the same
+ * order as the additions. Checks for the first match and return the overridden {@link URI}.
+ */
+public class HostOverrideList
```
ChaoLinkedIn, line 30: HostOverrides or HostOverriders?

d2/src/main/java/com/linkedin/d2/balancer/util/HostOverrideList.java
```diff
@@ -0,0 +1,103 @@
...
+{
+  private Map<Key, URI> _overrides = new LinkedHashMap<>();
+
+  public void addClusterOverride(String cluster, URI uri) {
+    ArgumentUtil.notNull(cluster, "cluster");
+    ArgumentUtil.notNull(uri, "uri");
+    _overrides.put(new Key(cluster, null), uri);
```
ChaoLinkedIn, line 37: Since the overrides could be supplied by different app, what if there are conflicts?

d2/src/main/java/com/linkedin/d2/balancer/util/HostOverrideList.java
ssheng: I wanted the word list to imply order of override. HostOverrides is fine, too, if you feel strong about the naming.

d2/src/main/java/com/linkedin/d2/balancer/util/HostOverrideList.java
ssheng: It won't be supplied by different apps since only the current app can add the override. However, it may be supplied by different places in the code. A solution I can think of is having a state of the list such that the first caller can add to the list and seal the list. The second caller will receive an exception adding to a sealed list. Another option is to expose the list contents to the caller so they can decide where to add the override. I don't particularly like either one as both add significant complexity to the logic.

d2/src/main/java/com/linkedin/d2/balancer/util/HostOverrideList.java
ssheng: Better, I think we can just use ==.

d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancer.java
ssheng: @nizarm's question, unit test shows that we don't have to. I will add another test to test with a '/' at the end.
@ChaoLinkedIn, I'll address this comment in the comment below.

d2/src/test/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerTest.java
ssheng: Turned out the implementation is quite fragile that adding a suffix '\' will break the behavior. Given that the normal D2 path leverages the same concatenation. I recommend that we fix that all together in a different PR.

d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancer.java
ssheng: Added as suggested.

d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancer.java
nizarm: thanks

d2/src/test/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerTest.java
nizarm: sure

PR #303, opened 2020-05-19
Handle valueClass NPE issue and give warnings

BrianPin:
When a finder resource class returns a non-record (Data type is POJO or not defined in PDL files) the restspec generation will be halted by seeing a NPE. The NPE is caused by `valueClass` being `null`.

### DIFF
--- a/restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
+++ b/restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
@@ -2628,8 +2628,8 @@ public final class RestLiAnnotationReader
 
     if (valueClass == null)
     {
-      throw new ResourceConfigException("Class '" + resourceModel.getResourceClass().getName()
-          + "' is not supported @Finder method");
+      throw new ResourceConfigException("Class '" + resourceModel.getResourceClass().getSimpleName()
+          + "' does not support @Finder method");
     }
 
     Class<?> returnType, elementType;
--- a/restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
+++ b/restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
@@ -2599,7 +2599,7 @@ public final class RestLiAnnotationReader
     {
       throw new ResourceConfigException("@Finder method '" + method.getName()
           + "' on class '" + resourceModel.getResourceClass().getName()
-          + "' has an invalid return type, The return type must be a RecordTemplate'");
+          + "' has an invalid return type. The return type must be a RecordTemplate");
     }
 
     Class<?> returnType, elementType;
--- a/restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
+++ b/restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
@@ -565,8 +565,8 @@ public final class RestLiAnnotationReader
           {
             continue;
           }
-          throw new ResourceConfigException("Class '" + collectionResourceClass.getName() + "' of a"
-              + "KeyUnstructuredDataResource class does not support for @Finder methods");
+          throw new ResourceConfigException("Class '" + collectionResourceClass.getSimpleName() + "' of a"
+              + " KeyUnstructuredDataResource class does not support @Finder methods");
         }
       }
 
@@ -692,8 +692,8 @@ public final class RestLiAnnotationReader
         {
           continue;
         }
-        throw new ResourceConfigException("Class '" + resourceClass.getName() + "' of a SingleUnstructuredDataResource "
-            + "class does not support for @Finder methods");
+        throw new ResourceConfigException("Class '" + resourceClass.getSimpleName() + "' of a SingleUnstructuredDataResource "
+            + "class does not support @Finder methods");
       }
     }
 
@@ -2629,7 +2629,7 @@ public final class RestLiAnnotationReader
     if (valueClass == null)
     {
       throw new ResourceConfigException("Class '" + resourceModel.getResourceClass().getName()
-          + "' is not supported for @Finder method");
+          + "' is not supported @Finder method");
     }
 
     Class<?> returnType, elementType;
--- a/restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
+++ b/restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
@@ -2597,9 +2597,8 @@ public final class RestLiAnnotationReader
 
     if (valueClass == null)
     {
-      throw new ResourceConfigException("@Finder method '" + method.getName()
-          + "' on class '" + resourceModel.getResourceClass().getName()
-          + "' has an invalid return type. The return type must be a RecordTemplate");
+      throw new ResourceConfigException("Class '" + resourceModel.getResourceClass().getName()
+          + "' where the class is a subtype of KeyUnstructuredDataResource is not supported for finders");
     }
 
     Class<?> returnType, elementType;
--- a/restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
+++ b/restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
@@ -2595,6 +2595,13 @@ public final class RestLiAnnotationReader
     Method method = finderMethodDescriptor.getMethod();
     Class<?> valueClass = resourceModel.getValueClass();
 
+    if (valueClass == null)
+    {
+      throw new ResourceConfigException("@Finder method '" + method.getName()
+          + "' on class '" + resourceModel.getResourceClass().getName()
+          + "' has an invalid return type, The return type must be a RecordTemplate'");
+    }
+
     Class<?> returnType, elementType;
     try
     {
--- a/restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
+++ b/restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
@@ -693,7 +693,7 @@ public final class RestLiAnnotationReader
           continue;
         }
         throw new ResourceConfigException("Class '" + resourceClass.getName() + "' of a SingleUnstructuredDataResource "
-            + "does not support for @Finder methods");
+            + "class does not support for @Finder methods");
       }
     }
 
--- a/restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
+++ b/restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
@@ -554,20 +554,6 @@ public final class RestLiAnnotationReader
         @SuppressWarnings("unchecked")
         Class<? extends KeyUnstructuredDataResource<?>> clazz = (Class<? extends KeyUnstructuredDataResource<?>>) collectionResourceClass;
         actualTypeArguments = ReflectionUtils.getTypeArgumentsParametrized(KeyUnstructuredDataResource.class, clazz);
-        for (Method method : collectionResourceClass.getDeclaredMethods())
-        {
-          if (method.isSynthetic())
-          {
-            continue;
-          }
-          Finder finderAnno = method.getAnnotation(Finder.class);
-          if (finderAnno == null)
-          {
-            continue;
-          }
-          throw new ResourceConfigException("Class '" + collectionResourceClass.getSimpleName() + "' of a"
-              + " KeyUnstructuredDataResource class does not support @Finder methods");
-        }
       }
 
       keyClass = ReflectionUtils.getClass(actualTypeArguments.get(0));
@@ -678,23 +664,6 @@ public final class RestLiAnnotationReader
       Class<? extends SingleObjectResource<?>> clazz = (Class<? extends SingleObjectResource<?>>) resourceClass;
       List<Class<?>> kvParams = ReflectionUtils.getTypeArguments(SingleObjectResource.class, clazz);
       valueClass = kvParams.get(0).asSubclass(RecordTemplate.class);
-    }
-    else
-    {
-      for (Method method : resourceClass.getDeclaredMethods())
-      {
-        if (method.isSynthetic())
-        {
-          continue;
-        }
-        Finder finderAnno = method.getAnnotation(Finder.class);
-        if (finderAnno == null)
-        {
-          continue;
-        }
-        throw new ResourceConfigException("Class '" + resourceClass.getSimpleName() + "' of a SingleUnstructuredDataResource "
-            + "class does not support @Finder methods");
-      }
     }
 
     ResourceType resourceType = getResourceType(resourceClass);
@@ -2626,10 +2595,11 @@ public final class RestLiAnnotationReader
     Method method = finderMethodDescriptor.getMethod();
     Class<?> valueClass = resourceModel.getValueClass();
 
-    if (valueClass == null)
+    if (SingleUnstructuredDataResource.class.isAssignableFrom(resourceModel.getResourceClass())
+        || KeyUnstructuredDataResource.class.isAssignableFrom(resourceModel.getResourceClass()))
     {
       throw new ResourceConfigException("Class '" + resourceModel.getResourceClass().getSimpleName()
-          + "' does not support @Finder method");
+          + "' extends the '" + resourceModel.getResourceClass().getSuperclass().getSimpleName() + "' class does not support @Finder methods");
     }
 
     Class<?> returnType, elementType;
--- a/restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
+++ b/restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
@@ -554,6 +554,20 @@ public final class RestLiAnnotationReader
         @SuppressWarnings("unchecked")
         Class<? extends KeyUnstructuredDataResource<?>> clazz = (Class<? extends KeyUnstructuredDataResource<?>>) collectionResourceClass;
         actualTypeArguments = ReflectionUtils.getTypeArgumentsParametrized(KeyUnstructuredDataResource.class, clazz);
+        for (Method method : collectionResourceClass.getDeclaredMethods())
+        {
+          if (method.isSynthetic())
+          {
+            continue;
+          }
+          Finder finderAnno = method.getAnnotation(Finder.class);
+          if (finderAnno == null)
+          {
+            continue;
+          }
+          throw new ResourceConfigException("Class '" + collectionResourceClass.getName() + "' of a"
+              + "KeyUnstructuredDataResource class does not support for @Finder methods");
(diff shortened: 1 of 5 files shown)


restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
```diff
@@ -2595,6 +2595,13 @@ private static void validateFinderMethod(final ResourceMethodDescriptor finderMe
     Method method = finderMethodDescriptor.getMethod();
     Class<?> valueClass = resourceModel.getValueClass();
 
+    if (valueClass == null)
+    {
+      throw new ResourceConfigException("@Finder method '" + method.getName()
+          + "' on class '" + resourceModel.getResourceClass().getName()
+          + "' has an invalid return type, we expect return type to be a record '");
```
junchuanwang, line 2602: nit: A bit suggestion on the grammar:  

`we expect the return type to be 'Record' `

Otherwise looks good to me.

nickibi:
I am wondering whether it is possible to add test for this change?

restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
evanw555: I agree with Junchuan, however this seems to imply that the return type must be exactly `Record`. I'd suggest just copying the phrasing from other exceptions in this file: `The return type must be a RecordTemplate`.

evanw555:
> I am wondering whether it is possible to add test for this change?

Yes, this is definitely possible and should be tested. See `TestRestLiApiBuilder#testMisconfiguredServiceErrors` for an example of how this is done.

nickibi:
> > I am wondering whether it is possible to add test for this change?
> 
> Yes, this is definitely possible and should be tested. See `TestRestLiApiBuilder#testMisconfiguredServiceErrors` for an example of how this is done.

@BrianPin Can you please add test for this change?

BrianPin:
@nickibi @evanw555 The unit test is added. And message is fine tuned

nickibi:
I think you pulled other commits from the master to this pull request. You should do git pull --rebase to keep your commit on top of the branch. Please only keep your changes in this PR, there are 36 files change in this PR.

BrianPin:
> I think you pulled other commits from the master to this pull request. You should do git pull --rebase to keep your commit on top of the branch. Please only keep your changes in this PR, there are 36 files change in this PR.

I see it, that is not good, I will redo a new branch then, I will probably abandon this PR

evanw555:
> I see it, that is not good, I will redo a new branch then, I will probably abandon this PR

You don't need to abandon this PR. You can just fix it locally then force-push to the same branch and the PR will be corrected.

restli-server/src/test/java/com/linkedin/restli/internal/server/model/TestRestLiApiBuilder.java
```diff
@@ -198,4 +198,33 @@ public void testActionReturnType(Class<?> resourceClass, Class<?> expectedAction
       Assert.assertEquals(resourceMethodDescriptor.getActionReturnType(), expectedActionReturnType);
     }
   }
+
+  @DataProvider(name = "unstructuredFinderReturnTypeData")
+  private Object[][] unstructuredFinderReturnTypeData()
+  {
+    return new Object[][]
+        {
+            {fooBarFinderResource.class, "The return type must be a RecordTemplate"}
```
nickibi, line 207: nit: extra space after "{" and before "}"

restli-server/src/test/java/com/linkedin/restli/internal/server/model/SampleResources.java
```diff
@@ -759,4 +761,16 @@ public void AttachmentParamsIncorrectDataTypeAction(@RestLiAttachmentsParam Stri
...
+
+  @RestLiCollection(name="lucky",keyName="dayOfWeek")
+  public class fooBarFinderResource extends UnstructuredDataCollectionResourceTemplate<Integer>
+  {
+    @Finder("findFooBar")
+    public List<String> findLucky(@PagingContextParam final PagingContext context, @QueryParam("dayOfWeek") Integer dayOfWeek) throws Exception
+    {
+      List<String> res = new LinkedList<>();
+      res.add("gotLuck!");
+      return res;
```
nickibi, line 773: Do you think "return Collections.singletonList("gotLuck!")" is cleaner?

restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
```diff
@@ -2595,6 +2595,13 @@ private static void validateFinderMethod(final ResourceMethodDescriptor finderMe
     Method method = finderMethodDescriptor.getMethod();
     Class<?> valueClass = resourceModel.getValueClass();
 
+    if (valueClass == null)
+    {
+      throw new ResourceConfigException("@Finder method '" + method.getName()
+          + "' on class '" + resourceModel.getResourceClass().getName()
+          + "' has an invalid return type, The return type must be a RecordTemplate'");
```
evanw555, line 2602: nits:
1. either make the `T` lowercase or replace the preceding comma with a period.
2. remove the dangling `'` at the end of the message.

restli-server/src/test/java/com/linkedin/restli/internal/server/model/SampleResources.java
```diff
@@ -759,4 +761,16 @@ public void AttachmentParamsIncorrectDataTypeAction(@RestLiAttachmentsParam Stri
 
   @RestLiCollection(name = "collectionComplexKeyTask")
   class CollectionComplexKeyTask extends ComplexKeyResourceTaskTemplate<EmptyRecord, EmptyRecord, EmptyRecord> {}
+
+  @RestLiCollection(name="lucky",keyName="dayOfWeek")
+  public class fooBarFinderResource extends UnstructuredDataCollectionResourceTemplate<Integer>
```
evanw555, line 766: Given that this file is bloated with resource classes, I'd suggest naming this something meaningful to clarify what it's testing. I'd suggest `_____Resource` where the prefix describes what scenario it covers (e.g. `NonRecordFinderResource` or `UnsupportedReturnTypeFinderResource`)

restli-server/src/test/java/com/linkedin/restli/internal/server/model/TestRestLiApiBuilder.java
```diff
@@ -198,4 +198,33 @@ public void testActionReturnType(Class<?> resourceClass, Class<?> expectedAction
       Assert.assertEquals(resourceMethodDescriptor.getActionReturnType(), expectedActionReturnType);
     }
   }
+
+  @DataProvider(name = "unstructuredFinderReturnTypeData")
+  private Object[][] unstructuredFinderReturnTypeData()
```
evanw555, line 203: I'd recommend avoiding the term "unstructured" here, since it may be confused with another Rest.li feature called "unstructured resources"

restli-server/src/test/java/com/linkedin/restli/internal/server/model/TestRestLiApiBuilder.java
```diff
@@ -198,4 +198,33 @@ public void testActionReturnType(Class<?> resourceClass, Class<?> expectedAction
...
+  }
+
+  /**
+   * Ensures that when finder methods are processed, if the return type is not of a Record, then it will be warned.
+   * For instance, it should recognize that the "logical" return type for a method
+   * {@code Task<ActionResult<String>> doFoo();} is {@code String.class}.
+   *
+   * @param resourceClass resource used as an input
+   */
+  @Test(dataProvider = "unstructuredFinderReturnTypeData")
```
evanw555, line 218: Given that there's only one case being tested, it's acceptable to hard code the inputs rather than using a data provider. Caveat would be if this case will likely get extended. Your call.

BrianPin: I will add more resources class to test since it addresses this concern and also provide more coverage

restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
```diff
@@ -2595,6 +2626,12 @@ private static void validateFinderMethod(final ResourceMethodDescriptor finderMe
     Method method = finderMethodDescriptor.getMethod();
     Class<?> valueClass = resourceModel.getValueClass();
 
+    if (valueClass == null)
+    {
+      throw new ResourceConfigException("Class '" + resourceModel.getResourceClass().getName()
+          + "' is not supported for @Finder method");
```
karthikbalasub, line 2632: Fix the error message

restli-server/src/test/java/com/linkedin/restli/internal/server/model/TestRestLiApiBuilder.java
```diff
@@ -198,4 +198,62 @@ public void testActionReturnType(Class<?> resourceClass, Class<?> expectedAction
...
+    return new Object[][]
+        {
+            { FinderSupportedAssociationDataResource.class },
+            { FinderSupportedComplexKeyDataResource.class },
+            { FinderWithActionResource.class }
+        };
+  }
+
+  @Test(dataProvider = "finderSupportedResourceTypeData")
+  public void testFinderSupportedResourceType(Class<?> resourceClass)
```
karthikbalasub, line 247: You can use the expectedExceptions and expectedExceptionsMessageRegExp attributes of the @Test annotation

restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
```diff
@@ -554,6 +554,20 @@ else if (KeyUnstructuredDataResource.class.isAssignableFrom(collectionResourceCl
...
+          {
+            continue;
+          }
+          Finder finderAnno = method.getAnnotation(Finder.class);
+          if (finderAnno == null)
+          {
+            continue;
+          }
+          throw new ResourceConfigException("Class '" + collectionResourceClass.getName() + "' of a"
+              + "KeyUnstructuredDataResource class does not support for @Finder methods");
```
karthikbalasub, line 569: nit: remove "for"

.. does not support @Finder methods"

restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
```diff
@@ -665,6 +679,23 @@ private static ResourceModel processSimpleResource(final Class<?> resourceClass,
...
+        {
+          continue;
+        }
+        Finder finderAnno = method.getAnnotation(Finder.class);
+        if (finderAnno == null)
+        {
+          continue;
+        }
+        throw new ResourceConfigException("Class '" + resourceClass.getName() + "' of a SingleUnstructuredDataResource "
+            + "class does not support for @Finder methods");
```
karthikbalasub, line 696: nit: remove "for"

.. does not support @Finder methods"

restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
BrianPin: I suppose it is meant to remove "for"

restli-server/src/test/java/com/linkedin/restli/internal/server/model/TestRestLiApiBuilder.java
BrianPin: This test is to ensure other resources classes other then the: [KeyUnstructuredDataResource, SingleUnstructuredDataResource] are not getting an exception, so the code in the test is not expecting exceptions. But the annotation of the two you suggested I will put them in my first test which there it expects exceptions.

restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
```diff
@@ -2629,7 +2629,7 @@ private static void validateFinderMethod(final ResourceMethodDescriptor finderMe
     if (valueClass == null)
     {
       throw new ResourceConfigException("Class '" + resourceModel.getResourceClass().getName()
-          + "' is not supported for @Finder method");
+          + "' is not supported @Finder method");
```
karthikbalasub: nit: this should be "does not support @Finder"

restli-server/src/test/java/com/linkedin/restli/internal/server/model/TestRestLiApiBuilder.java
```diff
@@ -216,20 +216,13 @@ public void testActionReturnType(Class<?> resourceClass, Class<?> expectedAction
...
-      Assert.fail("For the finder resource class with a non RecordTemplate sub class, we shall throw an exception");
-    }
-    catch (ResourceConfigException resourceConfigException)
-    {
-      Assert.assertTrue(resourceConfigException.getMessage().contains(expectedPartialMessage),
-          String.format("Expected %s with message containing \"%s\" but instead found message \"%s\"",
-              ResourceConfigException.class.getSimpleName(), expectedPartialMessage, resourceConfigException.getMessage()));
-    }
+    RestLiApiBuilder.buildResourceModels(Collections.singleton(resourceClass));
+    Assert.fail("For the finder resource class with a non RecordTemplate sub class, we shall throw an exception");
```
karthikbalasub, line 225: Not needed as the test framework would fail if expected exception is not thrown

restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
```diff
@@ -554,6 +554,20 @@ else if (KeyUnstructuredDataResource.class.isAssignableFrom(collectionResourceCl
         @SuppressWarnings("unchecked")
         Class<? extends KeyUnstructuredDataResource<?>> clazz = (Class<? extends KeyUnstructuredDataResource<?>>) collectionResourceClass;
         actualTypeArguments = ReflectionUtils.getTypeArgumentsParametrized(KeyUnstructuredDataResource.class, clazz);
+        for (Method method : collectionResourceClass.getDeclaredMethods())
```
evanw555, line 557: This logic seems overly complicated to me, and it's also duplicated. Could you not simply have a check in `validateFinderMethod` which accesses the class type via the `ResourceModel`? Perhaps I'm missing something...

restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
```diff
@@ -554,6 +554,20 @@ else if (KeyUnstructuredDataResource.class.isAssignableFrom(collectionResourceCl
...
+          if (method.isSynthetic())
+          {
+            continue;
+          }
+          Finder finderAnno = method.getAnnotation(Finder.class);
+          if (finderAnno == null)
+          {
+            continue;
+          }
+          throw new ResourceConfigException("Class '" + collectionResourceClass.getSimpleName() + "' of a"
```
evanw555, line 568: nit: `Class X of a Y class` doesn't really make sense. It would be more clear to say `Class X extending Y`

Also, I'd suggest not hard-coding the class names into the String message. You should programmatically get the class name. This lets the navigate/refactor tool find this reference.

restli-server/src/test/java/com/linkedin/restli/internal/server/model/TestRestLiApiBuilder.java
```diff
@@ -198,4 +198,54 @@ public void testActionReturnType(Class<?> resourceClass, Class<?> expectedAction
...
+  {
+    return new Object[][]
+        {
+            { FinderUnsupportedKeyUnstructuredDataResource.class, "KeyUnstructuredDataResource class does not support @Finder methods" },
+            { FinderUnsupportedSingleUnstructuredDataResource.class, "SingleUnstructuredDataResource class does not support @Finder methods" }
+        };
+  }
+
+  /**
+   * Ensures that when finder methods are processed, if the return type is not of a Record, then it will be warned.
```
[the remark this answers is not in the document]
BrianPin: I re-write it to "Ensures that when finder methods are processed, when the resource value class is a SingleUnstructuredDataResource or a KeyUnstructuredDataResource, we will end up with a ResourceConfigException because we don't support that righ tnow." Hope that seems like a good doc string.

restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
BrianPin: Done

restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
BrianPin: Consolidated to the method 'validateFinderMethod'

restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
```diff
@@ -2595,6 +2595,13 @@ private static void validateFinderMethod(final ResourceMethodDescriptor finderMe
     Method method = finderMethodDescriptor.getMethod();
     Class<?> valueClass = resourceModel.getValueClass();
 
+    if (SingleUnstructuredDataResource.class.isAssignableFrom(resourceModel.getResourceClass())
+        || KeyUnstructuredDataResource.class.isAssignableFrom(resourceModel.getResourceClass()))
+    {
+      throw new ResourceConfigException("Class '" + resourceModel.getResourceClass().getSimpleName()
+          + "' extends the '" + resourceModel.getResourceClass().getSuperclass().getSimpleName() + "' class does not support @Finder methods");
```
evanw555, line 2602: I just realized that this will not have the intended effect. If class `A` extends some template class, and that extends `KeyUnstructuredDataResource`, then this will print the name of the template class (I think). I would suggest not printing the name of the superclass and instead just append `...because it's an unstructured data resource` at the end of the message.

restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
```diff
@@ -2595,6 +2595,13 @@ private static void validateFinderMethod(final ResourceMethodDescriptor finderMe
     Method method = finderMethodDescriptor.getMethod();
     Class<?> valueClass = resourceModel.getValueClass();
 
+    if (SingleUnstructuredDataResource.class.isAssignableFrom(resourceModel.getResourceClass())
```
evanw555, line 2598: Although this is alright, an alternative would be to leverage the logic in `ResourceModel#getResourceEntityType` to do this check without it being class-specific. You can avoid logic duplication this way.

restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
BrianPin: yep. make sense

restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
BrianPin: cool, makes me know one more thing

PR #305, opened 2020-05-21
D2 Zookeeper client Retry do not work for uri getchildren

nizarm:
D2 Zookeeper client retry does not work with the version of the getChildren method which uses AsyncCallback.Children2Callback. This is what leveraged in ZookeeperEphemeral store watcher to read and set watches on uri store.


cx-super (commented):
Looks good. Can we have a test as well?

bohhyang (approved):

ChrisCarini:

PR #308, opened 2020-05-22
add protobuf stream data codec and use in PROTOBUF2

aman1309:
1. Add streamProtobuf data codec and use it in content type
2. Expose codec interfaces required to extend for application specific value ordinals


restli-common/src/main/java/com/linkedin/restli/common/ContentType.java
```diff
@@ -107,10 +111,12 @@
         new SymbolTableBasedContentTypeProvider(PROTOBUF,
```
karthikbalasub, line 111: Update this codec also.

aman1309: updated, left it initially since it is deprecated.

PR #312, opened 2020-05-29
When server builds URI and projection fields are included, convert ra…

BrianPin:
…w string to MaskTree

### DIFF
--- a/restli-common/src/main/java/com/linkedin/restli/internal/common/URIParamUtils.java
+++ b/restli-common/src/main/java/com/linkedin/restli/internal/common/URIParamUtils.java
@@ -70,13 +70,18 @@ public class URIParamUtils
     {
       if (dataMap.containsKey(parameterName))
       {
-        if (dataMap.get(parameterName) instanceof String)
+        Object projectionParameters = dataMap.get(parameterName);
+        if (projectionParameters instanceof String)
         {
-          result.put(parameterName, dataMap.getString(parameterName));
+          result.put(parameterName, (String) projectionParameters);
         }
-        else if (dataMap.get(parameterName) instanceof DataMap)
+        else if (projectionParameters instanceof DataMap)
         {
-          result.put(parameterName, URIMaskUtil.encodeMaskForURI(dataMap.getDataMap(parameterName)));
+          result.put(parameterName, URIMaskUtil.encodeMaskForURI((DataMap) projectionParameters));
+        }
+        else
+        {
+          throw new IllegalArgumentException("Invalid projection field data type");
         }
       }
     }
--- a/restli-common/src/main/java/com/linkedin/restli/internal/common/URIParamUtils.java
+++ b/restli-common/src/main/java/com/linkedin/restli/internal/common/URIParamUtils.java
@@ -53,6 +53,14 @@ public class URIParamUtils
   private static final Pattern NORMALIZED_URI_PATTERN = Pattern.compile("(^/|/$)");
   private static final Pattern URI_SEPARATOR_PATTERN = Pattern.compile("/+");
 
+  /**
+   * Return the string encoded version of query parameters.
+   * For projection parameters stored in dataMap, this function handles both cases when the value is a original string
+   * or a structured {@link DataMap}
+   *
+   * @param dataMap the {@link DataMap} which contains projection parameters as keys
+   * @return a {@link Map} from query param key to value in encoded string
+   */
   private static Map<String, String> dataMapToQueryParams(DataMap dataMap)
   {
     final Map<String, String> result = encodeDataMapParameters(dataMap);
@@ -62,11 +70,11 @@ public class URIParamUtils
     {
       if (dataMap.containsKey(parameterName))
       {
-        try
+        if (dataMap.get(parameterName) instanceof String)
         {
           result.put(parameterName, dataMap.getString(parameterName));
         }
-        catch (ClassCastException exception)
+        else if (dataMap.get(parameterName) instanceof DataMap)
         {
           result.put(parameterName, URIMaskUtil.encodeMaskForURI(dataMap.getDataMap(parameterName)));
         }
--- a/restli-common/src/main/java/com/linkedin/restli/internal/common/URIParamUtils.java
+++ b/restli-common/src/main/java/com/linkedin/restli/internal/common/URIParamUtils.java
@@ -62,7 +62,14 @@ public class URIParamUtils
     {
       if (dataMap.containsKey(parameterName))
       {
-        result.put(parameterName, URIMaskUtil.encodeMaskForURI(dataMap.getDataMap(parameterName)));
+        try
+        {
+          result.put(parameterName, dataMap.getString(parameterName));
+        }
+        catch (ClassCastException exception)
+        {
+          result.put(parameterName, URIMaskUtil.encodeMaskForURI(dataMap.getDataMap(parameterName)));
+        }
       }
     }
 
--- a/restli-server/src/main/java/com/linkedin/restli/internal/server/response/BatchFinderResponseBuilder.java
+++ b/restli-server/src/main/java/com/linkedin/restli/internal/server/response/BatchFinderResponseBuilder.java
@@ -151,11 +151,7 @@ public class BatchFinderResponseBuilder
     }
     catch (CloneNotSupportedException exception)
     {
-      BatchFinderEntry entry = new BatchFinderEntry(
-          new RestLiServiceException(S_500_INTERNAL_SERVER_ERROR, "Batch finder response builder failed when rebuild projection URI"));
-      return new RestLiResponseDataImpl<>(new BatchFinderResponseEnvelope(S_500_INTERNAL_SERVER_ERROR, collectionResponse),
-          headers,
-          cookies);
+      throw new RestLiServiceException(S_500_INTERNAL_SERVER_ERROR, "Batch finder response builder failed when rebuild projection URI");
     }
   }
 
--- a/restli-server/src/main/java/com/linkedin/restli/internal/server/response/BatchFinderResponseBuilder.java
+++ b/restli-server/src/main/java/com/linkedin/restli/internal/server/response/BatchFinderResponseBuilder.java
@@ -107,19 +107,19 @@ public class BatchFinderResponseBuilder
     TimingContextUtil.beginTiming(routingResult.getContext().getRawRequestContext(),
         FrameworkTimingKeys.SERVER_RESPONSE_RESTLI_PROJECTION_APPLY.key());
 
-    for (Object criteriaParam : criteriaParams.values())
+    try
     {
-      RecordTemplate criteria = new AnyRecord((DataMap) criteriaParam);
-      BatchFinderEntry entry;
-      if (result.getResults().containsKey(criteria))
+      for (Object criteriaParam : criteriaParams.values())
       {
-        CollectionResult<RecordTemplate, RecordTemplate> cr = result.getResult(criteria);
-
-        //Process elements
-        List<AnyRecord> elements = buildElements(cr, resourceContext);
-
-        try
+        RecordTemplate criteria = new AnyRecord((DataMap) criteriaParam);
+        BatchFinderEntry entry;
+        if (result.getResults().containsKey(criteria))
         {
+          CollectionResult<RecordTemplate, RecordTemplate> cr = result.getResult(criteria);
+
+          //Process elements
+          List<AnyRecord> elements = buildElements(cr, resourceContext);
+
           //Process paging
           final CollectionMetadata projectedPaging =
               buildPaginationMetaData(routingResult, criteria, resourceContext, request, cr);
@@ -129,31 +129,34 @@ public class BatchFinderResponseBuilder
 
           entry = new BatchFinderEntry(elements, projectedPaging, projectedCustomMetadata);
         }
-        catch (CloneNotSupportedException exception)
+        else if (result.getErrors().containsKey(criteria))
+        {
+          entry = new BatchFinderEntry(result.getErrors().get(criteria));
+        }
+        else
         {
           entry = new BatchFinderEntry(
-              new RestLiServiceException(S_500_INTERNAL_SERVER_ERROR, "Batch finder response builder failed when rebuild projection URI"));
+              new RestLiServiceException(S_404_NOT_FOUND, "The server didn't find a representation for this criteria"));
         }
-      }
-      else if (result.getErrors().containsKey(criteria))
-      {
-        entry = new BatchFinderEntry(result.getErrors().get(criteria));
-      }
-      else
-      {
-        entry = new BatchFinderEntry(
-            new RestLiServiceException(S_404_NOT_FOUND, "The server didn't find a representation for this criteria"));
+
+        collectionResponse.add(entry);
       }
 
-      collectionResponse.add(entry);
+      TimingContextUtil.endTiming(routingResult.getContext().getRawRequestContext(),
+          FrameworkTimingKeys.SERVER_RESPONSE_RESTLI_PROJECTION_APPLY.key());
+
+      return new RestLiResponseDataImpl<>(new BatchFinderResponseEnvelope(HttpStatus.S_200_OK, collectionResponse),
+          headers,
+          cookies);
     }
-
-    TimingContextUtil.endTiming(routingResult.getContext().getRawRequestContext(),
-        FrameworkTimingKeys.SERVER_RESPONSE_RESTLI_PROJECTION_APPLY.key());
-
-    return new RestLiResponseDataImpl<>(new BatchFinderResponseEnvelope(HttpStatus.S_200_OK, collectionResponse),
-                                        headers,
-                                        cookies);
+    catch (CloneNotSupportedException exception)
+    {
+      BatchFinderEntry entry = new BatchFinderEntry(
+          new RestLiServiceException(S_500_INTERNAL_SERVER_ERROR, "Batch finder response builder failed when rebuild projection URI"));
+      return new RestLiResponseDataImpl<>(new BatchFinderResponseEnvelope(S_500_INTERNAL_SERVER_ERROR, collectionResponse),
+          headers,
+          cookies);
+    }
   }
 
   private List<AnyRecord> buildElements(CollectionResult<RecordTemplate, RecordTemplate> cr,
--- a/restli-server/src/main/java/com/linkedin/restli/internal/server/response/BatchFinderResponseBuilder.java
+++ b/restli-server/src/main/java/com/linkedin/restli/internal/server/response/BatchFinderResponseBuilder.java
@@ -27,6 +27,8 @@ import com.linkedin.restli.common.BatchCollectionResponse;
 import com.linkedin.restli.common.CollectionMetadata;
 import com.linkedin.restli.common.CollectionResponse;
 import com.linkedin.restli.common.HttpStatus;
+import com.linkedin.restli.common.RestConstants;
+import com.linkedin.restli.internal.common.AllProtocolVersions;
 import com.linkedin.restli.internal.common.URIParamUtils;
 import com.linkedin.restli.internal.server.ResourceContextImpl;
 import com.linkedin.restli.internal.server.RoutingResult;
@@ -116,14 +118,22 @@ public class BatchFinderResponseBuilder
         //Process elements
         List<AnyRecord> elements = buildElements(cr, resourceContext);
 
-        //Process paging
-        final CollectionMetadata projectedPaging =
-            buildPaginationMetaData(routingResult, criteria, resourceContext, request, cr);
+        try
+        {
+          //Process paging
+          final CollectionMetadata projectedPaging =
+              buildPaginationMetaData(routingResult, criteria, resourceContext, request, cr);
 
-        //Process metadata
-        final AnyRecord projectedCustomMetadata = buildMetaData(cr, resourceContext);
+          //Process metadata
+          final AnyRecord projectedCustomMetadata = buildMetaData(cr, resourceContext);
 
-        entry = new BatchFinderEntry(elements, projectedPaging, projectedCustomMetadata);
+          entry = new BatchFinderEntry(elements, projectedPaging, projectedCustomMetadata);
+        }
+        catch (CloneNotSupportedException exception)
+        {
+          entry = new BatchFinderEntry(
+              new RestLiServiceException(S_500_INTERNAL_SERVER_ERROR, "Batch finder response builder failed when rebuild projection URI"));
+        }
       }
       else if (result.getErrors().containsKey(criteria))
       {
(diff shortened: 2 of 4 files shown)


restli-common/src/main/java/com/linkedin/restli/internal/common/URIParamUtils.java
```diff
@@ -469,6 +470,7 @@ public static void addSortedParams(UriBuilder uriBuilder, DataMap params, Protoc
 
   public static URI replaceQueryParam(URI uri, String queryParam, DataComplex values, DataMap parameters, ProtocolVersion version)
   {
+    // This is only used by BatchFinderResponseBuilder right now
```
evanw555, line 473: This is a public method, so you can't necessarily make this assumption.

restli-common/src/main/java/com/linkedin/restli/internal/common/URIParamUtils.java
```diff
@@ -478,6 +480,30 @@ public static URI replaceQueryParam(URI uri, String queryParam, DataComplex valu
     return builder.build();
   }
 
+  private static void convertRawProjectionFieldsToMaskTree(DataMap queryParams)
```
evanw555, line 483: Is this extra logic necessary? Isn't the parsed mask tree available in the response builder?

restli-common/src/main/java/com/linkedin/restli/internal/common/URIParamUtils.java
```diff
@@ -486,6 +512,7 @@ public static URI replaceQueryParam(URI uri, String queryParam, DataComplex valu
    */
   public static void addSortedParams(UriBuilder uriBuilder, DataMap params)
```
evanw555, line 513: Once again, this method is public so you should be careful about changing behavior.

restli-common/src/main/java/com/linkedin/restli/internal/common/URIParamUtils.java
```diff
@@ -486,6 +512,7 @@ public static URI replaceQueryParam(URI uri, String queryParam, DataComplex valu
...
   {
+    convertRawProjectionFieldsToMaskTree(params);
```
karthikbalasub, line 515: Agree with Evan, also the projection parameters are already available as MaskTree in resourContext. So a better fix would be to update the `buildCriteriaURI()` method in BatchFinderResponseBuilder to replace the projection params with masktree objects.

restli-common/src/main/java/com/linkedin/restli/internal/common/URIParamUtils.java
BrianPin: The change is moved to the BatchFinderResponseBuilder class now, and it is only limited in that class use case

restli-common/src/main/java/com/linkedin/restli/internal/common/URIParamUtils.java
BrianPin: yes, you are right,  the mastrees are generated during the class ResourceContextImpl constructor, so to avoid this extra step, I just need a reference from ResourceContextImpl instance

restli-common/src/main/java/com/linkedin/restli/internal/common/URIParamUtils.java
BrianPin: yes, it is resolved

restli-common/src/main/java/com/linkedin/restli/internal/common/URIParamUtils.java
BrianPin: Yes, changed the way I extract projection MaskTree by directly referencing those projection fields which already in ResourceContextImpl

restli-server/src/main/java/com/linkedin/restli/internal/server/response/BatchFinderResponseBuilder.java
```diff
@@ -207,14 +208,33 @@ private DataList getCriteriaParameters(RoutingResult routingResult)
...
+                                           batchParamName,
                                            criteriaList,
-                                           resourceContext.getParameters(),
+                                           queryParams,
                                            resourceContext.getRestliProtocolVersion());
   }
+
+  private DataMap extractQueryParamsFromResourceContext(ResourceContextImpl resourceContext)
+  {
+    DataMap queryParams = resourceContext.getParameters();
```
karthikbalasub, line 225: You should create a copy instead of modifying the underlying parameters. There might be logic elsewhere (and custom filter) that depend on these parameters parsed from URI.

karthikbalasub:
Also, add unit tests for this case

restli-server/src/main/java/com/linkedin/restli/internal/server/response/BatchFinderResponseBuilder.java
BrianPin: Got it, DataMap has two copy related method, clone and copy, the clone seems to be a shallow copy and the copy seems to be suit for this use case

restli-server/src/main/java/com/linkedin/restli/internal/server/response/BatchFinderResponseBuilder.java
```diff
@@ -195,26 +196,52 @@ private AnyRecord buildMetaData(CollectionResult<RecordTemplate, RecordTemplate>
...
-                                           resourceContext.getParameters(),
+                                           queryParams,
                                            resourceContext.getRestliProtocolVersion());
   }
+
+  private DataMap extractQueryParamsFromResourceContext(ResourceContextImpl resourceContext)
+  {
+    try
+    {
+      DataMap queryParams = resourceContext.getParameters().copy();
```
karthikbalasub, line 227: Deep copy might be expensive and for this usecase, a shallow copy is sufficient.

restli-server/src/main/java/com/linkedin/restli/internal/server/response/BatchFinderResponseBuilder.java
```diff
@@ -195,26 +196,52 @@ private AnyRecord buildMetaData(CollectionResult<RecordTemplate, RecordTemplate>
...
+      }
+      if (queryParams.containsKey(RestConstants.METADATA_FIELDS_PARAM))
+      {
+        queryParams.put(RestConstants.METADATA_FIELDS_PARAM, resourceContext.getMetadataProjectionMask().getDataMap());
+      }
+      return queryParams;
+    }
+    catch (CloneNotSupportedException cloneNotSupportedException)
+    {
+      return resourceContext.getParameters();
```
karthikbalasub, line 244: This is not a good way to recover this error. I'd prefer failing the request as this is unexpected.

restli-server/src/test/java/com/linkedin/restli/internal/server/response/TestBatchFinderResponseBuilder.java
```diff
@@ -341,4 +344,62 @@ private static Foo generateMetaData(Boolean onError)
...
   }
+
+  @DataProvider(name = "build_uri_test_cases")
+  public Object[][] testBuildCriteriaURIDataProvider() throws RestLiSyntaxException
+  {
+    String[] requestUrls =
+    {
+      "/greetings?bq=searchGreetings&criteria=List((id:1,tone:SINCERE))&message=hello",
+      "/talent/api/talentHiringProjectCandidates?bq=candidates&candidates=List((candidate:urn%3Ali%3Ats_hire_identity%3A88156577))",
+      "/talent/api/talentHiringProjectCandidates?bq=candidates&candidates=List((candidate:urn%3Ali%3Ats_hire_identity%3A88156577))&fields=candidate",
```
BrianPin, line 355: These tested URL has not:
1. Be real batch criterias at all (just single criteria) I will add to later test
2. These URLs are not exhaustive, I will add more URL test cases

restli-server/src/main/java/com/linkedin/restli/internal/server/response/BatchFinderResponseBuilder.java
BrianPin: Instead of `copy` I used `clone`, according to the comment and short code trace, clone should be the shallow copy

restli-server/src/main/java/com/linkedin/restli/internal/server/response/BatchFinderResponseBuilder.java
```diff
@@ -195,26 +196,52 @@ private AnyRecord buildMetaData(CollectionResult<RecordTemplate, RecordTemplate>
...
+      }
+      if (queryParams.containsKey(RestConstants.METADATA_FIELDS_PARAM))
+      {
+        queryParams.put(RestConstants.METADATA_FIELDS_PARAM, resourceContext.getMetadataProjectionMask().getDataMap());
+      }
+      return queryParams;
+    }
+    catch (CloneNotSupportedException exception)
+    {
+      throw new RuntimeException(exception);
```
BrianPin, line 244: In this part of error handling, because I don't expect to change all the callers of buildCriteriaURI with `throws`
so I instead throws a RuntimeException

restli-server/src/test/java/com/linkedin/restli/internal/server/response/TestBatchFinderResponseBuilder.java
```diff
@@ -341,4 +344,62 @@ private static Foo generateMetaData(Boolean onError)
...
+    };
+
+    String[] batchCriteriaParameterNames =
+    {
+      "criteria",
+      "candidates",
+      "candidates"
+    };
+
+    ProtocolVersion[] protocolVersions =
```
karthikrg, line 365: Why is this an array instead of a constant? Will we have more values here?

BrianPin: Yeah, it is an array because I found we have two protocol versions in the code. So I assume it is needed to test in both protocol versions.

BrianPin: The detail is, if there exists some http client and it send a request with a http header saying it is protocol version 100, then our ResourceContextImpl class internally will parse query parameters differently.

karthikrg: I understand that we need to support both 1.0 and 2.0, however all your array values are 2.0

BrianPin: yes, I met some issues where it only happens in 1.0 so I will include that fix in this PR and then that is complete

restli-server/src/main/java/com/linkedin/restli/internal/server/response/BatchFinderResponseBuilder.java
karthikbalasub: I think throwing a RestLiServiceException with 500 status code and cleaner error message would be good here. For reference you can check the other response builders.

restli-server/src/test/java/com/linkedin/restli/internal/server/response/TestBatchFinderResponseBuilder.java
karthikbalasub: As we discussed, let us get the fix in for the projection issue with protocol 2.0.0 support and raise a separate PR for 1.0.0 issue.

restli-server/src/main/java/com/linkedin/restli/internal/server/response/BatchFinderResponseBuilder.java
BrianPin: I just fixed it, one point I am curious is that, for each entry in the batch, is it better to put in 500 error code for each entry, or just fail the whole batch in one 500 code?

restli-server/src/test/java/com/linkedin/restli/internal/server/response/TestBatchFinderResponseBuilder.java
BrianPin: yes, the 1.0.0 part of the issue is long existing one

restli-server/src/main/java/com/linkedin/restli/internal/server/response/BatchFinderResponseBuilder.java
karthikbalasub: For this, just one 500 error is fine. The paging parameters are common for all criteria.

BrianPin: got it I will make another change then, right now i put failure in one entry, the other will continue to process or they will generate another 500.

restli-server/src/test/java/com/linkedin/restli/internal/server/response/TestBatchFinderResponseBuilder.java
```diff
@@ -341,4 +344,81 @@ private static Foo generateMetaData(Boolean onError)
...
+
+    String[] expectedURIs =
+    {
+        "/greetings?bq=searchGreetings&criteria=List((id:1,tone:SINCERE))&message=hello",
+        "/talent/api/talentHiringProjectCandidates?bq=candidates&candidates=List((candidate:urn%3Ali%3Ats_hire_identity%3A88156577))",
+        "/talent/api/talentHiringProjectCandidates?bq=candidates&candidates=List((candidate:urn%3Ali%3Ats_hire_identity%3A88156577))&fields=candidate",
+        "/talent/api/talentHiringProjectCandidates?bq=candidates&candidates=List((candidate:urn%3Ali%3Ats_hire_identity%3A88156577))",
+        "/talent/api/talentHiringProjectCandidates?bq=candidates&candidates=List((candidate:urn;aba%3Ats_hire_identity%3A88156588))",
+        "/talent/api/talentHiringProjectCandidates?bq=candidates&candidates=List((candidate:urn%3Ali%3Ats_hire_identity%3A88156577))&fields=candidate",
+        "/talent/api/talentHiringProjectCandidates?bq=candidates&candidates=List((candidate:urn;aba%3Ats_hire_identity%3A88156588))&fields=candidate",
```
BrianPin, line 377: Note the original case's %3B became ;
that is minor but I just want to highlight it in case something is wrong

restli-server/src/main/java/com/linkedin/restli/internal/server/response/BatchFinderResponseBuilder.java
```diff
@@ -105,45 +107,56 @@ public RestLiResponse buildResponse(RoutingResult routingResult,
...
+          headers,
+          cookies);
+    }
+    catch (CloneNotSupportedException exception)
+    {
+      BatchFinderEntry entry = new BatchFinderEntry(
+          new RestLiServiceException(S_500_INTERNAL_SERVER_ERROR, "Batch finder response builder failed when rebuild projection URI"));
+      return new RestLiResponseDataImpl<>(new BatchFinderResponseEnvelope(S_500_INTERNAL_SERVER_ERROR, collectionResponse),
+          headers,
+          cookies);
```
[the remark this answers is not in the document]
BrianPin: Sounds good

evanw555:
Also another comment: this bug was originally not caught because the int test `TestBatchFinderResource` didn't have any cases with projections. You should improve this test class to have cases with projections

restli-common/src/main/java/com/linkedin/restli/internal/common/URIParamUtils.java
```diff
@@ -62,7 +62,14 @@
     {
       if (dataMap.containsKey(parameterName))
       {
-        result.put(parameterName, URIMaskUtil.encodeMaskForURI(dataMap.getDataMap(parameterName)));
+        try
```
evanw555, line 65: Using try-catches to dictate normal logic flow (rather than just for exceptions) is bad practice. I would suggest retrieving the generic object from the map and checking its type.

Overall though this is better IMO.

restli-server/src/main/java/com/linkedin/restli/internal/server/response/BatchFinderResponseBuilder.java
BrianPin: @karthikbalasub , all the changes regarding clone and exception handling are reverted since I use a simpler way to address the original issue

restli-server/src/main/java/com/linkedin/restli/internal/server/response/BatchFinderResponseBuilder.java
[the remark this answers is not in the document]
BrianPin: I addressed the original issue with another simpler approach, so returning 500 as server side error is not needed now

restli-server/src/main/java/com/linkedin/restli/internal/server/response/BatchFinderResponseBuilder.java
```diff
@@ -105,45 +107,52 @@ public RestLiResponse buildResponse(RoutingResult routingResult,
     TimingContextUtil.beginTiming(routingResult.getContext().getRawRequestContext(),
         FrameworkTimingKeys.SERVER_RESPONSE_RESTLI_PROJECTION_APPLY.key());
 
-    for (Object criteriaParam : criteriaParams.values())
+    try
```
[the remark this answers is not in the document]
BrianPin: As we discussed earlier, this logic is not needed since we addressed the problem in the other way

restli-common/src/main/java/com/linkedin/restli/internal/common/URIParamUtils.java
BrianPin: Thanks for the advice, fixed it.

restli-server/src/main/java/com/linkedin/restli/internal/server/response/BatchFinderResponseBuilder.java
```diff
@@ -195,26 +204,47 @@ private AnyRecord buildMetaData(CollectionResult<RecordTemplate, RecordTemplate>
...
   private DataList getCriteriaParameters(RoutingResult routingResult)
   {
-    String batchParamName = getBatchParamterName(routingResult);
-    return(DataList)routingResult.getContext().getStructuredParameter(batchParamName);
+    String batchParameterName = getBatchParameterName(routingResult);
+    return(DataList)routingResult.getContext().getStructuredParameter(batchParameterName);
   }
 
-  private URI buildCriteriaURI(ResourceContextImpl resourceContext, RecordTemplate criteria, String queryParam, URI uri)
+  protected static URI buildCriteriaURI(ResourceContextImpl resourceContext, RecordTemplate criteria, String batchParameterName, URI uri)
```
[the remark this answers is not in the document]
BrianPin: Done

BrianPin:
> Also another comment: this bug was originally not caught because the int test `TestBatchFinderResource` didn't have any cases with projections. You should improve this test class to have cases with projections

@evanw555 , The request builder (RootBuilderWrapper)  in that test class `TestBatchFinderResource` does not seem to be able to set projection fields in its APIs. 
So I think I am not able to extend that test case, instead I have my buildCriteriaURI test cases in the PR, will that be sufficient?

restli-common/src/main/java/com/linkedin/restli/internal/common/URIParamUtils.java
```diff
@@ -62,7 +70,14 @@
     {
       if (dataMap.containsKey(parameterName))
       {
-        result.put(parameterName, URIMaskUtil.encodeMaskForURI(dataMap.getDataMap(parameterName)));
+        if (dataMap.get(parameterName) instanceof String)
```
evanw555, line 73: Minor comments:
1. Add a final `else` case which throws an exception. If one of these projection params is present but is not a string or data map, then it should be treated as an illegal argument.
2. A minor optimization could be made here. You have 4 different `get` accessor calls here, you could do one initial get then perform the checks/casting on that object.

BrianPin: Fixed

restli-common/src/main/java/com/linkedin/restli/internal/common/URIParamUtils.java
```diff
@@ -53,6 +53,14 @@
   private static final Pattern NORMALIZED_URI_PATTERN = Pattern.compile("(^/|/$)");
   private static final Pattern URI_SEPARATOR_PATTERN = Pattern.compile("/+");
 
+  /**
+   * Return the string encoded version of query parameters.
+   * For projection parameters stored in dataMap, this function handles both cases when the value is a original string
```
karthikbalasub, line 58: This is a cleaner solution, Nice!

restli-server/src/test/java/com/linkedin/restli/internal/server/response/TestBatchFinderResponseBuilder.java
```diff
@@ -341,4 +344,94 @@ private static Foo generateMetaData(Boolean onError)
...
+
+    return testData;
+  }
+
+  @Test(dataProvider = "build_uri_test_cases")
+  public void testBuildCriteriaURI(ResourceContextImpl resourceContext,
+                                   RecordTemplate criteria,
+                                   String batchParameterName,
+                                   URI uri,
+                                   URI expectedURI) throws CloneNotSupportedException
```
aman1309, line 432: remove CloneNotSupportedException

BrianPin: Done

restli-server/src/main/java/com/linkedin/restli/internal/server/response/BatchFinderResponseBuilder.java
```diff
@@ -195,24 +195,24 @@ private AnyRecord buildMetaData(CollectionResult<RecordTemplate, RecordTemplate>
...
   private DataList getCriteriaParameters(RoutingResult routingResult)
   {
-    String batchParamName = getBatchParamterName(routingResult);
-    return(DataList)routingResult.getContext().getStructuredParameter(batchParamName);
+    String batchParameterName = getBatchParameterName(routingResult);
+    return (DataList) routingResult.getContext().getStructuredParameter(batchParameterName);
   }
 
-  private URI buildCriteriaURI(ResourceContextImpl resourceContext, RecordTemplate criteria, String queryParam, URI uri)
+  static URI buildCriteriaURI(ResourceContextImpl resourceContext, RecordTemplate criteria, String batchParameterName, URI uri)
```
nickibi, line 210: Is this for testing purpose only? If yes, we should add a comment "package-private for testing purpose".

BrianPin: Sorry just saw this, will do in my next PR.

BrianPin: This is for testing only right now

PR #313, opened 2020-05-29
Enabling warmup in LastSeen Loadbalancer

nizarm:
Enabling warmup in LastSeen Loadbalancer

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/zkfs/LastSeenLoadBalancerWithFacilities.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/zkfs/LastSeenLoadBalancerWithFacilities.java
@@ -22,6 +22,7 @@ import com.linkedin.common.util.None;
 import com.linkedin.d2.balancer.Directory;
 import com.linkedin.d2.balancer.KeyMapper;
 import com.linkedin.d2.balancer.LoadBalancerWithFacilities;
+import com.linkedin.d2.balancer.WarmUpService;
 import com.linkedin.d2.balancer.properties.ClusterProperties;
 import com.linkedin.d2.balancer.properties.ServiceProperties;
 import com.linkedin.d2.balancer.properties.UriProperties;
@@ -50,7 +51,8 @@ import org.slf4j.LoggerFactory;
  *
  * @author Francesco Capponi (fcapponi@linkedin.com)
  */
-public class LastSeenLoadBalancerWithFacilities implements LoadBalancerWithFacilities {
+public class LastSeenLoadBalancerWithFacilities implements LoadBalancerWithFacilities, WarmUpService
+{
   private static final Logger LOG = LoggerFactory.getLogger(LastSeenLoadBalancerWithFacilities.class);
 
   private final ZKFSDirectory _directory;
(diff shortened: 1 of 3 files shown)


nizarm:
> In theory LastSeen should already have a warmup integrated. Adding it twice shouldn't hurt but it might be symptom of some malfunctioning.
> 
> However, if you can try this out before shipping it and this work can be a short term patch.
> However we should have a plan to fix the larger problem of a not using the right executor IMHO.
> 
> If we don't have time for that we can move forward with this

I verified that this fix works. This is to bring feature parity with existing load balancer and avoid the issue every time there is a service restart. 

If LastSeen already had warmup integrated this fix wouldn't have been required. Can you please point towards where warmup happening in LastSeen?

I agree that this is not a permanent fix but a workaround to address the current situation while we figure out the actual fix. Having this fix in place will help us to continue to rollout this important feature.

FreCap:
> > In theory LastSeen should already have a warmup integrated. Adding it twice shouldn't hurt but it might be symptom of some malfunctioning.
> > However, if you can try this out before shipping it and this work can be a short term patch.
> > However we should have a plan to fix the larger problem of a not using the right executor IMHO.
> > If we don't have time for that we can move forward with this
> 
> I verified that this fix works. This is to bring feature parity with existing load balancer and avoid the issue every time there is a service restart.
> 
> If LastSeen already had warmup integrated this fix wouldn't have been required. Can you please point towards where warmup happening in LastSeen?
> 
> I agree that this is not a permanent fix but a workaround to address the current situation while we figure out the actual fix. Having this fix in place will help us to continue to rollout this important feature.

At start of the balancer, if they are already present on disk, it would reload the services
https://github.com/linkedin/rest.li/blob/66e5abc68f28bbff2010a147f987c42e30579668/d2/src/main/java/com/linkedin/d2/discovery/stores/zk/LastSeenZKStore.java#L168

nizarm:
> At start of the balancer, if they are already present on disk, it would reload the services
> https://github.com/linkedin/rest.li/blob/66e5abc68f28bbff2010a147f987c42e30579668/d2/src/main/java/com/linkedin/d2/discovery/stores/zk/LastSeenZKStore.java#L168

The issue here is that the warmup happens at the store level and the load balancer states and its subscribers are not set up unless it is asked to from the higher layers. So the warmup we do at store level does not actually populate the in-memory LoadBalancerState.

d2/src/main/java/com/linkedin/d2/balancer/zkfs/LastSeenLoadBalancerWithFacilities.java
```diff
@@ -50,7 +51,8 @@
  *
  * @author Francesco Capponi (fcapponi@linkedin.com)
  */
-public class LastSeenLoadBalancerWithFacilities implements LoadBalancerWithFacilities {
+public class LastSeenLoadBalancerWithFacilities implements LoadBalancerWithFacilities, WarmUpService
```
ChaoLinkedIn, line 54: Given that WarmUpService interface is supported everywhere, can we move this interface into LoadBalancer (or merge them) and have an impl in LoadBalancerWithFacilitiesDelegator to save a lot of code?

nizarm: Great suggestion. Can we do this refactoring as part of a different pull request so that we can unblock the LastSeen rollout quickly?

PR #314, opened 2020-06-01
Added pathspec user guide

junchuanwang:
Add a page introduce how to use PathSpec in Rest.Li

Added to "How to" section. Named as "Use PathSpec in Rest.Li"

### DIFF
--- a/src/_data/user_guide_navigation.yml
+++ b/src/_data/user_guide_navigation.yml
@@ -67,6 +67,8 @@ toc:
         url: /rest.li/Send-Rest_li-Request-Query-In-Request-Body
       - page: Use Projections
         url: /rest.li/How-to-use-projections-in-Java
+      - page: Use PathSpec in Rest.Li
+        url: /rest.li/pathspec_user_guide
       - page: Avro Conversion
         url: /rest.li/Rest_li_Avro_conversions
       - page: Compression
--- a/src/_data/user_guide_navigation.yml
+++ b/src/_data/user_guide_navigation.yml
@@ -43,6 +43,8 @@ toc:
         url: /rest.li/Rest_li-Filters
       - page: Projections
         url: /rest.li/Projections
+      - page: PathSpecs
+        url: /rest.li/pathspec
       - page: Validation
         url: /rest.li/Validation-in-Rest_li
       - page: Attachment Streaming
@@ -67,8 +69,6 @@ toc:
         url: /rest.li/Send-Rest_li-Request-Query-In-Request-Body
       - page: Use Projections
         url: /rest.li/How-to-use-projections-in-Java
-      - page: Use PathSpec in Rest.Li
-        url: /rest.li/use_pathspec
       - page: Avro Conversion
         url: /rest.li/Rest_li_Avro_conversions
       - page: Compression
--- a/src/_data/user_guide_navigation.yml
+++ b/src/_data/user_guide_navigation.yml
@@ -68,7 +68,7 @@ toc:
       - page: Use Projections
         url: /rest.li/How-to-use-projections-in-Java
       - page: Use PathSpec in Rest.Li
-        url: /rest.li/pathspec_user_guide
+        url: /rest.li/use_pathspec
       - page: Avro Conversion
         url: /rest.li/Rest_li_Avro_conversions
       - page: Compression
--- a/src/pathspec_user_guide.md
+++ b/src/pathspec_user_guide.md
@@ -1,0 +1,444 @@
+---
+layout: guide
+title: PathSpec User Guide
+permalink: /pathspec_user_guide
+excerpt: Guide for use PathSpec within Rest.Li
+---
+
+# PathSpec User Guide
+
+## What is PathSpec
+PathSpec represents a path to a component within a complex data object within Rest.Li framework. It generates uniform references in hierarchical data maps (used in the data-transform) component. It is an abstract path concept which has two forms of concrete existences.
+
+1. The PathSpec path can be represented as a string, current example usages of its string form can be found in [Validation annotation](https://linkedin.github.io/rest.li/Validation-in-Rest_li#specifying-restli-validation-annotations), where the PathSpec string is used in annotation and the [Annotation Reader](https://github.com/linkedin/rest.li/blob/master/restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java#L185) would interpret it.
+2. The PathSpec path also have a binded java class: [PathSpec.java](https://github.com/linkedin/rest.li/blob/master/data/src/main/java/com/linkedin/data/schema/PathSpec.java). This java class can be passed as arguments in Restli Framework. Its Java class form is now mainly used for [Projection in Rest.Li framework](https://linkedin.github.io/rest.li/How-to-use-projections-in-Java). The PathSpec java class also has a method to return the correct path string by calling `toString()`.
+
+## Current usage
+### PathSpec in Projection
+PathSpec's java class binding can be used for projection. Users can get PathSpec object that represents the fields in data object. For example, it could be obtained from generated RecordTemplate subclasses using the .fields() method. 
+For example:
+```java
+PathSpec pathSpec = Foo.fields().bar(); 
+```
+
+More concrete examples can be found [here in the wiki](https://linkedin.github.io/rest.li/How-to-use-projections-in-Java#getting-the-pathspec-of-a-field).
+
+This capability is provided by Rest.Li auto-generated code, as data object representation for data in Pegasus schema should extend [RecordTemplate.java](https://github.com/linkedin/rest.li/blob/master/data/src/main/java/com/linkedin/data/template/RecordTemplate.java), and will define a inner class called "fields", which extended PathSpec.java. By passing PathSpec object to the reqeust builder, the PathSpec is then used by [MaskCreator](https://github.com/linkedin/rest.li/blob/master/data-transform/src/main/java/com/linkedin/data/transform/filter/request/MaskCreator.java) to create a [MaskTree](https://github.com/linkedin/rest.li/blob/master/data-transform/src/main/java/com/linkedin/data/transform/filter/request/MaskTree.java). Thus, PathSpec can be used to control the projection behavior.
+
+### PathSpec in Validation
+PathSpec's string form is used for [Request Validation](https://linkedin.github.io/rest.li/Validation-in-Rest_li) in Rest.Li Resource. A string path can be added in annotation such as "CreateOnly" and "ReadOnly". For example:
+```java
+@CreateOnly({"/id", "/EXIF"})
+public class PhotoResource extends CollectionResourceTemplate<Long, Photo>
+{
+    ...
+}
+```
+
+This string is then parsed by [RestLiAnnotationReader](https://github.com/linkedin/rest.li/blob/master/restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java), and [DataSchemaUtil](https://github.com/linkedin/rest.li/blob/master/data/src/main/java/com/linkedin/data/schema/DataSchemaUtil.java) will do corresponding validation against this path within the DataMap.
+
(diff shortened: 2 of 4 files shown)


src/pathspec_user_guide.md
```diff
@@ -0,0 +1,444 @@
+---
```
evanw555, line 1: General comments for this whole PR:

1. I'd suggest rethinking how you frame what a PathSpec is. You say that a PathSpec is a concept with "two forms". This isn't exactly true. A PathSpec is a specification with an intended meaning, and is language agnostic. There could be PathSpec support in Python, Go, etc. You should frame PathSpecs as just a specification, and it happens to have implemented bindings in Java (not sure about other languages). I'd recommend highlighting the concept and the specification, and treating the Java binding as just an implementation.

2. All your links are full links to the live website. I'd advise against this. You should either be using absolute path links (e.g. `/rest.li/something`) or better yet relative links using the built-in Jekyll `relative_url` plugin (e.g. `{{ 'something' | relative_url }}`. This second approach is best-practice because it takes into account the "base URL" of the website (e.g. `"/"`, `"/rest.li/"`). See [here](https://github.com/linkedin/rest.li/blob/gh-pages-source/src/_includes/head.html#L28) for an example.

3. I see inconsistent formatting of Rest.li's name (e.g. Rest.Li, Restli, RestLi). I'd suggest referencing Rest.li using the standard format: "Rest.li". Can be solved using simple search-and-replace.

src/_data/user_guide_navigation.yml
```diff
@@ -67,6 +67,8 @@ toc:
         url: /rest.li/Send-Rest_li-Request-Query-In-Request-Body
       - page: Use Projections
         url: /rest.li/How-to-use-projections-in-Java
+      - page: Use PathSpec in Rest.Li
```
evanw555, line 70: I don't think this is appropriate for the "How to" subheading. This is less of a guide and more of a specification for the PathSpec concept. Tbh I think this would be more appropriate in most of the other sections (eg. Data Schemas, Data, Main Concepts, etc.)

src/pathspec_user_guide.md
```diff
@@ -0,0 +1,444 @@
...
+```
+[Check example code here](https://github.com/linkedin/rest.li/blob/master/generator-test/src/test/java/com/linkedin/pegasus/generator/test/pdl/fixtures/CustomRecord.java#L62). 
+Therefore it is very easy to get the PathSpec java binded class. Let's say you have a `Foo` schema which has `bar` fields. You can get the PathSpec by following
+```java
+PathSpec pathSpec = Foo.fields().bar();
+```
+This has also been documented in [How to Use Projections in Java](https://linkedin.github.io/rest.li/How-to-use-projections-in-Java#getting-the-pathspec-of-a-field)
+
+
+## More Resources
```
evanw555, line 443: nit: "More Examples"

src/_data/user_guide_navigation.yml
junchuanwang: I thought about putting this file there. But I had two reasons
(1) most of the contents in this files are just examples, looks more fitting to "how to" section.
(2) This concept is not very big and has been mentioned in `projection` or `validation`. I don't think there needs to be a separate page for that. In fact if we have had too many files in those two secionts `Data Schema` and `Main Concept`, it adds burden to the users to understand.

I mainly want to keep this is a reference guide.

src/pathspec_user_guide.md
junchuanwang: (1) reworded, thanks!
(2) Changed, thanks!
(3) updated, thanks.

src/pathspec_user_guide.md
```diff
@@ -0,0 +1,444 @@
+---
+layout: guide
+title: PathSpec User Guide
+permalink: /pathspec_user_guide
+excerpt: Guide for use PathSpec within Rest.Li
```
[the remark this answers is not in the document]
junchuanwang: reworded, thanks!

src/pathspec_user_guide.md
```diff
@@ -0,0 +1,444 @@
...
+---
+
+# PathSpec User Guide
+
+## What is PathSpec
```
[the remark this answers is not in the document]
junchuanwang: Added an example

src/pathspec_user_guide.md
```diff
@@ -0,0 +1,444 @@
...
+PathSpec represents a path to a component within a complex data object within Rest.Li framework. It generates uniform references in hierarchical data maps (used in the data-transform) component. It is an abstract path concept which has two forms of concrete existences.
```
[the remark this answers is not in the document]
junchuanwang: changed. Thanks

src/pathspec_user_guide.md
```diff
@@ -0,0 +1,444 @@
...
+
+1. The PathSpec path can be represented as a string, current example usages of its string form can be found in [Validation annotation](https://linkedin.github.io/rest.li/Validation-in-Rest_li#specifying-restli-validation-annotations), where the PathSpec string is used in annotation and the [Annotation Reader](https://github.com/linkedin/rest.li/blob/master/restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java#L185) would interpret it.
+2. The PathSpec path also have a binded java class: [PathSpec.java](https://github.com/linkedin/rest.li/blob/master/data/src/main/java/com/linkedin/data/schema/PathSpec.java). This java class can be passed as arguments in Restli Framework. Its Java class form is now mainly used for [Projection in Rest.Li framework](https://linkedin.github.io/rest.li/How-to-use-projections-in-Java). The PathSpec java class also has a method to return the correct path string by calling `toString()`.
```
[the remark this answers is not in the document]
junchuanwang: I understand your concern and also mentioned this is not language specific.

I just feel that given the fact that "java" is now the only available language in the wiki docs now, it became harder to explain.

But I also think it is not very ambiguous. Rest.li wiki readers who are seeking answers shouldn't get confused.

src/pathspec_user_guide.md
```diff
@@ -0,0 +1,444 @@
...
+
+## Current usage
```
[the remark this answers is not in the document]
junchuanwang: changed. Thanks

src/pathspec_user_guide.md
```diff
@@ -0,0 +1,444 @@
...
+        "intB", "UnionFieldWithInlineRecord/com.linkedin.restli.examples.greetings.api.myRecord/foo2", "MapWithTyperefs/*/id"));
+    annotations.put("readOnly", Arrays.asList("stringA", 
+        "intA", "UnionFieldWithInlineRecord/com.linkedin.restli.examples.greetings.api.myRecord/foo1", "ArrayWithInlineRecord/*/bar1", "validationDemoNext/stringB", "validationDemoNext/UnionFieldWithInlineRecord"));
+    ...
+
+    validator = new RestLiDataValidator(annotations, ValidationDemo.class, ResourceMethod.CREATE, validatorClassMap);
+
+```
+
+## PathSec Semantics
```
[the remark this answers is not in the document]
junchuanwang: now there are only two sections now, `PathSpec Syntax in its string form` and `PathSpec Syntax in its java binded class form`

src/pathspec_user_guide.md
```diff
@@ -0,0 +1,444 @@
...
+
+PathSpec has been defined as a path to a component within a complex data object path. This provides a way to traverse the Pegasus data object. The abstract data object, in most case, is a form of [DataMap](https://github.com/linkedin/rest.li/blob/master/data/src/main/java/com/linkedin/data/DataMap.java) internally in RestLi framework, but PathSpec should be meaningful for the same data object in other forms, for example, a json representing the same data object.
+
+The PathSpec format is represented by separators(`'/'`) and segments in between. 
+```
+/demoRecord/innerRecordField/nestedInnerRecordField
+```
+
+The path segment for some types, such as array, could have some attributes, which following the `&` sign,
```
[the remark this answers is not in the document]
junchuanwang: you are right and I am aware of that. But I don't think it meaningful to introduce non-reserved attribute since they are not interpreted anyway. It will add to confusion.

But I have re-worded this part.

src/use_pathspec.md
```diff
@@ -0,0 +1,479 @@
+---
+layout: guide
+title: Use PathSpec in Rest.Li
```
karthikbalasub, line 3: This should be "How to use.."

src/use_pathspec.md
```diff
@@ -0,0 +1,479 @@
...
+## PathSpec Syntax in its string form
+
+PathSpec has been defined as a path to a component within a complex data object path. This provides a way to traverse the Pegasus data object. The abstract data object, in most case, is a form of [DataMap](https://github.com/linkedin/rest.li/blob/master/data/src/main/java/com/linkedin/data/DataMap.java) internally in Rest.li framework, but PathSpec should be meaningful for the same data object in other forms, for example, a json representing the same data object.
+
+The PathSpec string format is represented by separators(`'/'`) and segments in between.
+```
+/demoRecord/innerRecordField/nestedInnerRecordField
+```
+
+The path segment could use attribute syntax to carry some meaningful attributes. They attributes could be added following the `&` sign. Users can add any attributes but for some types, there are reserved attributes. For example, for array type, one can specify `start` and `count` attributes and these two attributes are used in specifying projections.
```
karthikbalasub, line 100: These attributes added to the string form using '?' and '&' separators.

src/use_pathspec.md
```diff
@@ -0,0 +1,479 @@
...
+```
+/arrayOfIntFieldE?start=0&count=10
+```
+
+For some collection types, such as maps and arrays, the path segment could also be replaced by the wildcard. A wildcard means that this segment path string can be replaced by any applicable segment string, for example.
```
karthikbalasub, line 105: For collection types..

src/use_pathspec.md
```diff
@@ -0,0 +1,479 @@
...
+```
+/mapOfRecordField/*/innerRecordField
+```
+Above examples points to the `innerRecordField` field of the `map` value in a `map` schema. `map` is a collection schema type, here `*` wildcarded its keys.
+
+Each PathSpec has a corresponding string form.
```
karthikbalasub, line 111: remove this line?

src/use_pathspec.md
```diff
@@ -0,0 +1,479 @@
...
+
+Another common use case is that you want to use wildcard (represented by symbol "*" ) to "select all keys", if the interested path is to a field for all entries in this map, then it would be
+
+```
+/mapField/*/<recordField>
+```
+
+It is worth noting that in Pegasus, the map keys are always of string types.
+
+Here is an concrete example of a record, containing a map field, and that map field's map has value of record type. The example also defines a "recordInlineMap" field for similar demenstration purpose
```
karthikbalasub, line 212: demonstration

src/use_pathspec.md
```diff
@@ -0,0 +1,479 @@
...
+  }]
+}
+```
+Then similarly if you are interested in some fields only, you can use wildCard to select all elements so you can refer to a field
+```
+/recordArray/*/location
+/recordInlineArray/*/f
+```
+
+One big difference between Array and Map is that in Array, we support reference to the range. It is achieved by specifying attributes aforementioned. 1
```
karthikbalasub, line 257: It is achieved by using the 'start' and 'count' attributes.

(Hard to remember which attributes you are referring to, or if someone reads this section directly)

src/use_pathspec.md
```diff
@@ -0,0 +1,479 @@
...
+```
+In this example, the union also contains array and map, and the reference to them would be 
+```
+/unionArray/*/map
+/unionArray/*/array
+```
+If the schema defines more than one array or more than one map in the union, they need to have defined alias for each (see Alias usage in next section).
+
+##### Alias
+Alias are used for refer to types (such as record type) that cannot be directly defined in the field due to same name conflict. Here is an example PDSC with name "AliasTest" in "AliasTest.pdsc"
```
karthikbalasub, line 344: s/pdsc/pdl

src/use_pathspec.md
```diff
@@ -0,0 +1,479 @@
...
+/**
+ * Aliase examples
+ */
+@aliases = ["com.linkedin.pegasus.examples.AliasTest2"]
+record AliasTest {
+  a1: AliasTest
+  a2: AliasTest
+}
+```
+Here the AliasTest2 is an alias for another record with same name "AliasTest" in "AliasTest2.pdsc"
```
karthikbalasub, line 357: pdsc -> pdl

src/_data/user_guide_navigation.yml
karthikbalasub: I agree with Evan on this. 
Normally "How to" pages should answer how to solve a specific problem. This can go under the "Main Concepts" section?

src/use_pathspec.md
junchuanwang: Renamed and moved to main concepts

PR #319, opened 2020-06-03
Add option to force fully qualified names in PDSC encoder.

karthikbalasub:


data/src/main/java/com/linkedin/data/schema/SchemaToJsonEncoder.java
```diff
@@ -95,6 +95,7 @@ public static String schemasToJson(Collection<DataSchema> schemas, JsonBuilder.P
   protected final JsonBuilder _builder;
   protected String _currentNamespace = "";
   protected String _currentPackage = "";
+  protected boolean _alwaysUseFullyQualifedName = false;
```
aman1309, line 98: private?

data/src/main/java/com/linkedin/data/schema/SchemaToJsonEncoder.java
```diff
@@ -127,6 +128,14 @@ public String getCurrentNamespace()
     return _currentNamespace;
   }
 
+  /**
+   * Configure the encoder to always use fully qualified names when encoding named type references.
```
aman1309, line 132: can you add when this should be set

restli-tools/src/main/java/com/linkedin/restli/tools/data/SchemaFormatTranslator.java
```diff
@@ -90,6 +89,10 @@
     OPTIONS.addOption(OptionBuilder.withLongOpt("skip-verification")
         .withDescription("Skip verification of the translated schemas. Be cautious, as incorrect translations will not be caught.")
         .create('k'));
+
+    OPTIONS.addOption(OptionBuilder.withLongOpt("force-pdsc-fully-qualified-names")
+        .withDescription("Forces generated PDSC schemas to always use fully qualified names.")
```
aman1309, line 94: same here

data/src/main/java/com/linkedin/data/schema/SchemaToJsonEncoder.java
nickibi: This is for LinkedIn specific use case, we should avoid adding those information in the open source code base.

karthikbalasub: Updated the javadoc to provide more details.

PR #321, opened 2020-06-04
fix dark cluster strategy refresh

davidhoa:
All the properties relevant to the dark cluster strategy is contained within the source d2 cluster properties, which means that the dark cluster properties themselves will not get changed when a strategy property of an existing darkcluster get's changed. For example, if the multiplier gets changed from 0.1 to 1.0, only the source d2 cluster will get changed. This PR fixes the code to only listen to the source d2 cluster, and fixes up the relevant testcases.


darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
```diff
@@ -150,30 +150,34 @@ public void onClusterAdded(String updatedClusterName)
...
           {
-            // just update the dark cluster that changed
-            _darkStrategyMap.put(darkClusterName, createStrategy(darkClusterName,
-                                                                 darkConfigMap.get(darkClusterName)));
+            // we need to refresh every dark cluster in this "source" cluster, because we don't have an easy way to tell
+            // what changed in the source cluster. It could have been a dark cluster change, or it could have been an unrelated
+            // change in the source cluster. Because there are so few changes in cluster znodes, this is not a
+            // big concern at the moment. If it does become a concern, storing the original darkClusterConfig and comparing with the
+            // new one is one way to reduce the strategy refreshes. However, since the strategy itself is so lightweight, it probably won't be a
+            // problem or worth the effort.
```
zhang-chris, line 169: can be more concise: "For simplicity, we refresh all strategies since we expect cluster updates to be rare and refresh to be cheap."

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
```diff
@@ -150,30 +150,34 @@ public void onClusterAdded(String updatedClusterName)
...
+            String darkClusterName = entry.getKey();
+            DarkClusterConfig darkClusterConfig = entry.getValue();
+            _darkStrategyMap.put(darkClusterName, createStrategy(darkClusterName, darkClusterConfig));
```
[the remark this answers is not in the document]
davidhoa: made the changes we talked about. Also, I realized that the StrategyFactoryImpl's getOrCreate() shouldn't modify the strategyMap, because if there's constant traffic, it would make it impossible to remove a dark cluster strategy if the getOrCreate() added to the strategyMap.

darkcluster/src/test/java/com/linkedin/darkcluster/TestDarkClusterStrategyFactory.java
```diff
@@ -259,6 +311,18 @@ public void onClusterAdded(String clusterName)
       // deleting whatever the first cluster listener added. It would have been more straightforward to have
       // a handle directly to the other clusterListener, but there's no good reason for the StrategyFactory to
       // expose that or allow it to be passed in, as the clusterListener needs to manipulate internal state.
+      try
```
davidhoa, line 314: this is not a new change, nothing has changed in this function. I don't know why github shows a diff here, the file looks fine.

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
```diff
@@ -150,43 +150,52 @@ public void onClusterAdded(String updatedClusterName)
...
-        // darkClusterConfigMap on the parent d2 cluster, because that has the properties needed
-        // to recreate the dark cluster strategies, such as the multiplier.
-        String darkClusterName = updatedClusterName;
         try
         {
-          DarkClusterConfigMap darkConfigMap = _facilities.getClusterInfoProvider().getDarkClusterConfigMap(_sourceClusterName);
-          if (darkConfigMap.containsKey(darkClusterName))
+          DarkClusterConfigMap updatedDarkConfigMap = _facilities.getClusterInfoProvider().getDarkClusterConfigMap(_sourceClusterName);
+
+          // get the existing entries in the strategy map
```
zhang-chris, line 161: comments at 161 and 163 can be removed.

PR #322, opened 2020-06-05
Introduce pegasus-all meta-project

evanw555:
Adds a new module, `pegasus-all`, which is simply a meta-project that
programmatically depends on all other consumer-facing modules. This
makes it easier for certain use cases to pull in the entire dependency
tree.


pegasus-all/build.gradle
```diff
@@ -0,0 +1,12 @@
+/*
+    This is a meta-project that programmatically depends on all other consumer-facing modules
```
junchuanwang, line 2: Should we add any documentations to other places as well?

evanw555: I'm not sure. We could add documentation to the website, but there isn't currently documentation on which artifacts do what. We could potentially add documentation to the description of each module's POM artifact, but that's a change for another day.

pegasus-all/build.gradle
```diff
@@ -0,0 +1,12 @@
...
+    so that the entire dependency tree of pegasus may be pulled in easily. Be warned that this
+    module is not intended to be directly consumed, since it'll bloat the consumer's dependencies.
```
junchuanwang, line 4: Based on the wording here, it's usage is somewhat non-standard? I will recommend adding to somewhere to let consumer's know.

evanw555: I'm not too sure how we can surface this to people. I think I'll merge this for now and ponder that afterward.

PR #323, opened 2020-06-08
Fix non-deterministic issue for annotations in the generated data template code by using right schema location for typeref class template

BrianPin:
When rest.li process schema files to generate data template java code, the order of the input source file will cause the change in the annotation of the generated java class. For example:
When a PDL or PDSC file A contains a data schema definition in B, and B is a typeref, when the input is in order [A, B], the generated B's class annotation comment will have a different class name compare to the two file in order [B, A]
In depth, this is caused by parsing result's _schemaAndLocations entry was populated in different order when parsing is processed.


DPUkyle:
Amazing! Thanks Brian. Was your new test `TestPegasusDataTemplateGenerator` able to reproduce the problem before you implemented the fix?

cc: @Astro03

generator/src/test/java/com/linkedin/pegasus/generator/TestPegasusDataTemplateGenerator.java
```diff
@@ -205,4 +210,78 @@ private void testRunGenerator(String pegasusFilename, Map<String, String> expect
             file -> file.getName().replace(".java", ""),
             Function.identity()));
   }
+
+  @DataProvider(name = "withMultiDataTemplateCases")
+  private Object[][] createDataTemplateCases()
+  {
+    return new Object[][]
+        {
+            {"IsoDuration.pdsc", "PremiumService.pdsc"},
```
junchuanwang, line 219: can you test using PDL? not to say pdsc is not good for this kind of testing, But since we are migrating to PDL. Need to make sure PDL is actually working. Also if we use PDL now, we don't need to worry about later they are migrated in this MP

generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
```diff
@@ -712,6 +712,7 @@ private ClassTemplateSpec generateFixed(FixedDataSchema schema)
 
   private TyperefTemplateSpec generateTyperef(TyperefDataSchema schema, TyperefDataSchema originalTyperefSchema)
   {
+    pushCurrentLocation(_schemaResolver.nameToDataSchemaLocations().get(schema.getFullName()));
```
junchuanwang, line 715: as we discussed offline, Please confirm if the inline typeref's dataSchemaLocation is also using enclosingSchema's location.

BrianPin: Junchuan, I confirmed this in my new added test case

generator/src/test/java/com/linkedin/pegasus/generator/TestPegasusDataTemplateGenerator.java
BrianPin: Yes, PDL cases are added

BrianPin:
> Amazing! Thanks Brian. Was your new test `TestPegasusDataTemplateGenerator` able to reproduce the problem before you implemented the fix?
> 
> cc: @Astro03

I was able to reproduce manually.
The way I test to repro is:
(inspired from talent solution api rmb diff comparison result)
1. Have a A.pdsc contain a field b which is defined in another pdsc called B
2. In B.pdsc, it has no field but itself is a typeref pointing to some primitive like string or long.
3. In the unit test of TestPegasusDataTemplateGenerator permute the input differently:
[A, B]
and 
[B, A]

the annotation comment of B.java will be different.
*It will be different only when I restart the test all over again.
In a subsequent test case, it will be the same.

BrianPin:
> Amazing! Thanks Brian. Was your new test `TestPegasusDataTemplateGenerator` able to reproduce the problem before you implemented the fix?
> 
> cc: @Astro03

generator/src/test/java/com/linkedin/pegasus/generator/TestPegasusDataTemplateGenerator.java
```diff
@@ -205,4 +208,123 @@ private void testRunGenerator(String pegasusFilename, Map<String, String> expect
...
+    String[] javaFilenames = Arrays.copyOfRange(testArgs, 2, testArgs.length);
+    testDataTemplateGenerationDeterministic(pegasusFilenames, javaFilenames);
+  }
+
+  @DataProvider(name = "with_inline_field_in_record_pdl")
+  private Object[][] createCasesWithInlineFieldInRecord()
+  {
+    return new Object[][]
+        {
+            {"BRecord.pdl", "InlineRecord.java", "BRecord.java"},
```
Astro03, line 255: I think the only 2 cases you need are 

  {"BRecord.pdl", "InlineRecord.java"},
  {"InlineRecord.java", "BRecord.pdl"},"

Unless you want to test duplicates  like you have here.

generator/src/test/resources/generator/BRecord.pdl
```diff
@@ -0,0 +1,3 @@
+record BRecord {
+  bField: record InlineRecord {}
+}
```
Astro03, line 3: nit: newline to end the file

generator/src/test/resources/referenceJava/ARecord.java
```diff
@@ -0,0 +1,135 @@
...
+import com.linkedin.data.schema.RecordDataSchema;
+import com.linkedin.data.schema.SchemaFormatType;
+import com.linkedin.data.template.DataTemplateUtil;
+import com.linkedin.data.template.GetMode;
+import com.linkedin.data.template.RecordTemplate;
+import com.linkedin.data.template.SetMode;
+
+
+/**
+ *
```
Astro03, line 15: nit: actual comments in the javadoc comment section

generator/src/test/resources/referenceJava/BRecord.java
```diff
@@ -0,0 +1,135 @@
...
+import com.linkedin.data.schema.RecordDataSchema;
+import com.linkedin.data.schema.SchemaFormatType;
+import com.linkedin.data.template.DataTemplateUtil;
+import com.linkedin.data.template.GetMode;
+import com.linkedin.data.template.RecordTemplate;
+import com.linkedin.data.template.SetMode;
+
+
+/**
+ *
```
Astro03, line 15: nit: fill javadoc comment out?

generator/src/test/java/com/linkedin/pegasus/generator/TestPegasusDataTemplateGenerator.java
```diff
@@ -205,4 +208,123 @@ private void testRunGenerator(String pegasusFilename, Map<String, String> expect
...
+  @Test(dataProvider = "with_inline_field_in_record_pdl")
+  public void testCasesWithInlineFieldInRecord(String[] testArgs)
+      throws Exception
+  {
+    String[] pegasusFilenames = Arrays.copyOfRange(testArgs, 0, 1);
+    String[] javaFilenames = Arrays.copyOfRange(testArgs, 1, testArgs.length);
+    testDataTemplateGenerationDeterministic(pegasusFilenames, javaFilenames);
+  }
+
+  private void testDataTemplateGenerationDeterministic(String[] pegasusFilenames, String[] javaFilenames)
```
Astro03, line 268: nit: testDataTemplateGenerationDeterminism  (the Deterministic in the sentence doesn't make sense)

generator/src/test/java/com/linkedin/pegasus/generator/TestPegasusDataTemplateGenerator.java
```diff
@@ -205,4 +208,123 @@ private void testRunGenerator(String pegasusFilename, Map<String, String> expect
...
+  {
+    return new Object[][]
+        {
+            {"IsoDuration.pdsc", "PremiumService.pdsc", "IsoDuration.java", "PremiumService.java", "PremiumServiceState.java", "AccountBalanceType.java"},
+            {"PremiumService.pdsc", "IsoDuration.pdsc", "IsoDuration.java", "PremiumService.java", "PremiumServiceState.java", "AccountBalanceType.java"},
+        };
+  }
+
+  @Test(dataProvider = "with_typeref_field_in_record_pdsc")
+  public void testDataTemplateGenerationOrderWithResolver(String[] testArgs)
```
Astro03, line 223: Oh, sorry, I just realized what you are doing with all these tests. 

Since you only really have one test with a nuance on the inputs, the better practice would be to have 1 DataProvider with your different tests as it's inputs. 

i.e. 

data resolver
{
  // typered fields with pdsc files
  { ...} 
  {...}
  // typeref field with pdl files
 { ...}
 { ... } 
 // inline field with pdsc files
 { ...} 
 {...}
 //inline field with pdl files
 { ...}
 { ...}
}

then your test would be just the one test you had.

generator/src/test/java/com/linkedin/pegasus/generator/TestPegasusDataTemplateGenerator.java
karthikbalasub, line 268: A better way to test this (and avoid having to update java files everytime code generation logic changes) is to take two arrays of Pegasus schemas as input (in different order) and then generate the source files and compare they are same.

BrianPin: That was my first attempt, sometime today in my chat with Astro, I changed to current way.
I think ...... I can do the original way

karthikbalasub: I think you can just generate MD5 for all the generated files together by xor-ing individual MD5s and compare. It would be simpler than current logic.
See https://stackoverflow.com/questions/47253864/order-independent-hash-in-java

generator/src/test/resources/generator/BRecord.pdl
BrianPin: done

generator/src/test/resources/referenceJava/BRecord.java
BrianPin: this is removed (by addressing Karthik's comment)

generator/src/test/java/com/linkedin/pegasus/generator/TestPegasusDataTemplateGenerator.java
BrianPin: thanks, addressed

generator/src/test/java/com/linkedin/pegasus/generator/TestPegasusDataTemplateGenerator.java
BrianPin: Yes, thanks, addressed into a single data provider "test_schema_permutation_deterministic"

BrianPin:
> Update the PR description to match the current logic.

sure

generator/src/test/java/com/linkedin/pegasus/generator/TestPegasusDataTemplateGenerator.java
```diff
@@ -205,4 +206,62 @@ private void testRunGenerator(String pegasusFilename, Map<String, String> expect
...
   }
+
+  @DataProvider(name = "test_schema_permutation_deterministic")
+  private Object[][] createPermutedDataTemplateCases()
+  {
+    return new Object[][]
+        {
+            {"IsoDuration.pdsc", "PremiumService.pdsc", "PremiumService.pdsc", "IsoDuration.pdsc"},
+            {"AField.pdl", "ARecord.pdl", "ARecord.pdl", "AField.pdl"},
+            {"BRecord.pdl", "BField.pdl", "BField.pdl", "BRecord.pdl"},
```
karthikbalasub, line 217: I suggest you return two arrays of strings as input. 
Eg
```
{{"IsoDuration.pdsc", "PremiumService.pdsc"}, {"PremiumService.pdsc", "IsoDuration.pdsc"}}
```
And add a comment explaining why order is important for this test.

generator/src/test/java/com/linkedin/pegasus/generator/TestPegasusDataTemplateGenerator.java
```diff
@@ -205,4 +206,62 @@ private void testRunGenerator(String pegasusFilename, Map<String, String> expect
...
+    System.setProperty(AbstractGenerator.GENERATOR_RESOLVER_PATH, String.format("@%s", argFile.toPath()));
+    String tempDirectoryPath1 = _tempDir.getAbsolutePath();
+    String[] mainArgs = new String[pegasusFilenames.length + 1];
+    mainArgs[0] = tempDirectoryPath1;
+    for (int i = 0; i < pegasusFilenames.length; i++)
+    {
+      mainArgs[i+1] = new File(pegasusDir + FS + pegasusFilenames[i]).getAbsolutePath();
+    }
+    PegasusDataTemplateGenerator.main(mainArgs);
+    File[] generatedFiles = _tempDir.listFiles((File dir, String name) -> name.endsWith(".java"));
```
karthikbalasub, line 247: Is this guaranteed to return the files sorted alphabetically?

generator/src/test/java/com/linkedin/pegasus/generator/TestPegasusDataTemplateGenerator.java
Astro03: I see that you changed to a single data provider but you still have only 1 input arrray into your test, then within your test, you split it up by dividing by 2. 

You should have two separate arrays, one for each variable. It makes it a lot easier to understand the logic within your code/test.

generator/src/test/java/com/linkedin/pegasus/generator/TestPegasusDataTemplateGenerator.java
Astro03: I was talking about this comment. What did i suggest?

generator/src/test/java/com/linkedin/pegasus/generator/TestPegasusDataTemplateGenerator.java
BrianPin: Both of the two points are addressed, I added a comment on the DataProvider, Please let me know if the comment is clear

generator/src/test/java/com/linkedin/pegasus/generator/TestPegasusDataTemplateGenerator.java
BrianPin: the returning files are not needed to be sorted, (if it sorts that is fine), the requirement is that for two consecutive calls of `_tempDir.listFiles` , the  returning files order shall be the same.

generator/src/test/java/com/linkedin/pegasus/generator/TestTemplateSpecGenerator.java
```diff
@@ -72,14 +74,18 @@
...
   public void initMocks() {
     MockitoAnnotations.initMocks(this);
     _uniqueNumberGenerator = new AtomicInteger();
-    Mockito.when(_resolver.nameToDataSchemaLocations()).thenReturn(Collections.singletonMap(INPUT_SCHEMA_NAME, _location));
+    _location = new StringDataSchemaLocation("location");
+    HashMap<String, DataSchemaLocation> schemaToLocation = new HashMap<>();
+    schemaToLocation.put(INPUT_SCHEMA_NAME, _location);
+    schemaToLocation.put(CUSTOM_TYPE_1.getFullName(), _location);
+    schemaToLocation.put(CUSTOM_TYPE_2.getFullName(), _location);
+    Mockito.when(_resolver.nameToDataSchemaLocations()).thenReturn(schemaToLocation);
```
BrianPin, line 88: This change here is happening is only because the test wasn't need to mock the schema location before, now it has to, so I have to patch this test, it is not because of any additional logic change

generator/src/test/java/com/linkedin/pegasus/generator/TestPegasusDataTemplateGenerator.java
karthikbalasub: I don't see that consecutive calls would return the same order: https://docs.oracle.com/javase/7/docs/api/java/io/File.html#listFiles()

I suggest sorting the generateFiles array to be sure this cannot cause flakiness.

generator/src/test/java/com/linkedin/pegasus/generator/TestPegasusDataTemplateGenerator.java
```diff
@@ -205,4 +206,64 @@ private void testRunGenerator(String pegasusFilename, Map<String, String> expect
...
+    return new Object[][]
+        {
+            {new String[]{"IsoDuration.pdsc", "PremiumService.pdsc"}, new String[]{"PremiumService.pdsc", "IsoDuration.pdsc"}},
+            {new String[]{"AField.pdl", "ARecord.pdl"}, new String[]{"ARecord.pdl", "AField.pdl"}},
+            {new String[]{"BRecord.pdl", "BField.pdl"}, new String[]{"BField.pdl", "BRecord.pdl"}},
+        };
+  }
+
+  @Test(dataProvider = "test_schema_permutation_determinisim")
+  public void testDataTemplateGenerationDeterminism(String[][] testArgs)
```
karthikbalasub, line 227: I guess I made a mistake in my suggestion, what I wanted is for this method to take 2 args:
(String[] inputSchemasOrder1, String[] inputSchemasOrder2)

Astro03: I agree, is this not how this test framework organizes the inputs for DataProviders?

BrianPin: This is addressed

BrianPin:
> Maybe extract the Comparator?
> 
> ```
> Comparator<String> byAbsolutePath = Comparator.comparing(File::getAbsolutePath);
> Arrays.sort(generatedFiles1, byAbsolutePath);
> Arrays.sort(generatedFiles2, byAbsolutePath);
> ```

@DPUkyle , When I do the suggestion of yours, the JDK says 
```
Non-static method cannot be referenced from a static context
```
I think it only works in unnamed comparator object where it can use the `File` instance for the method `getAbsolutePath`

BrianPin:
> Amazing! Thanks Brian. Was your new test `TestPegasusDataTemplateGenerator` able to reproduce the problem before you implemented the fix?
> 
> cc: @Astro03

@DPUkyle  @Astro03 
I am able to reproduce the problem if I remove my fix,
The key fact is I used the LinkedinHashMap in the parser result in order keep iterating over the HashMap consistently. 
So I think I have addressed all comments now

Astro03:
Is it still only true that you can reproduce the issue without your fix if the inputs to your test is exactly right? 

Did you resolve the issue where the tests need to start a new JVM for it to run the other way?

BrianPin:
> Is it still only true that you can reproduce the issue without your fix if the inputs to your test is exactly right?
> 
> Did you resolve the issue where the tests need to start a new JVM for it to run the other way?

@Astro03 , I resolved the issue where I don't need the JVM or test to restart to reproduce the issue.
It is because I changed the use of HashMap to LinkedHashMap
This way the order we insert things to the hash map is predictable.

For JVM internal reason, Java HashMap's iteration order is unpredictable and we can't rely on that.

Astro03:
Looks great!

PR #324, opened 2020-06-08
Added annotation processor introduction

junchuanwang:
This needs to be published after https://github.com/linkedin/rest.li/pull/314


src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
+---
+layout: guide
+title: PDL Schema
+permalink: /annotation_processor
+excerpt: Documentation of using annotation processor in Rest.Li
+---
+
```
nickibi, line 7: Can we add a content table here?

src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
...
+# Annotation Processor
+This sections introduces a generic schema annotation processing tool in Rest.Li. Before reading this, it is recommended to be familiar with the concept of PathSpec in Rest.Li. [Insert link here]
+
+## Annotating Rest.Li Schemas
```
karthikbalasub, line 11: Should we use "pegasus schemas" instead of Rest.Li?
Pegasus schemas can be used without rest.li framework and the annotation can still be applied

src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
...
+In Rest.Li, schema writers are able to add annotations to schema fields or the schema itself.
+
+**Example**: annotate on the schema fields
```
karthikbalasub, line 14: annotate -> Annotation

src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
...
+  @data_classification = "MEDIUM"
+  lastName: string,
+  @data_classification = "LOW"
+  userId: long,
+  @data_classification = "HIGH"
+  socialSecurityNumber: string,
+}
+```
+
+**Example**: annotate directly on the schema
```
karthikbalasub, line 28: annotation

src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
...
+```pdl
+@data_classification = "HIGH"
+record UserPersonallyIdentifiableInformation {
+  firstName: string,
+  lastName: string,
+  userId: long,
+  socialSecurityNumber: string,
+}
+```
+Note that these annotations are on the schema level.
```
karthikbalasub, line 38: Remove this line( redundant) or better say: "The data_classifcation annotation is specified at record level"

src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
...
+title: PDL Schema
+permalink: /annotation_processor
+excerpt: Documentation of using annotation processor in Rest.Li
+---
+
+# Annotation Processor
+This sections introduces a generic schema annotation processing tool in Rest.Li. Before reading this, it is recommended to be familiar with the concept of PathSpec in Rest.Li. [Insert link here]
+
+## Annotating Rest.Li Schemas
+In Rest.Li, schema writers are able to add annotations to schema fields or the schema itself.
```
karthikbalasub, line 12: "are able to add annotations to schema" -> are able to add annotations using @<annotation name> syntax to schema

src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
...
+record UserPersonallyIdentifiableInformation {
+  firstName: string,
+  lastName: string,
+  userId: long,
+  socialSecurityNumber: string,
+}
+```
+Note that these annotations are on the schema level.
+
+The support that Rest.Li provided for processing annotation in Rest.Li schema has been quite limited. These annotations in above examples are later stored as a `DataSchema`'s class attribute called "property". Just as the example shows, both field and the DataSchema can have this "property". We let the user chooses where they want to have the schema annotated and how they want to interpret. There were no preferrabled way regarding how annotation should be written, read and interpreted.
```
karthikbalasub, line 40: Keep all text in third person, so avoid "we" "you" etc.

We let the user chooses where they want to have the schema annotated and how they want to interpret. There were no preferrabled way regarding how annotation should be written, read and interpreted.
->
The Rest.Li framework did not provide specification on how these annotations should be interpreted and if was left to the user to add logic for interpreting them.

(Written syntax and read as properties are specified by framework)

src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Inherit and override schema annotations
+But Rest.Li users found it useful to process annotations during schema processing. One use case is to introduce "inheritance" and "overrides" to annotations so those annotations can be dynamically processed in the way user defines, when for example the schemas were reused. This gives annotation extensibility and adds to schema reusability.
```
karthikbalasub, line 43: Do not start a new section with "But", try to re-word this similar to my suggestion above.

src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
...
+
+Here are examples:
+Example case 1: Users might want the annotation of a field be inherited. The fields can reuse the annotations from upper level.
```
karthikbalasub, line 46: "inherited from the  parent record."
Don't say re-use.. inherited is the right work. 
And we don't allow inhertiance other than from records. righ?

src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
...
+```pdl
+@persistencePolicyInDays = 30
+record  UserVisitRecord {
+  visitedUrls: array[URL]
+  visitedUserProfiles: array[UserRecord]
+}
+```
+Reading from the schema, we might find both `visitedUrls` and `visitedUserProfiles` have `persistencePolicyInDays` annotated as 30.
+//<inheritance example>
```
karthikbalasub, line 55: The // is visible in doc. check.

src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
...
+
+Example case 2: Users might want the annotation of a field be overriden. Override is where the annotation on a field or a schema might get updated when other annotations assign it another value.
+//<override example>
```
karthikbalasub, line 58: same for //

src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
...
+```pdl
+@persistencePolicyInDays = 30
+record  UserVisitRecord {
+  @persistencePolicyInDays = 10
+  visitedUrls: array[URL]
+  visitedUserProfiles: array[UserRecord]
+}
+```
+Reading from the schema, a user might find the field `visitedUserProfiles`s `persistencePolicyInDays` annotation is 30 but `visitedUrls`'s `persistencePolicyInDays` annotation has value 10, which overrides the original value inherited(which was 30).
```
karthikbalasub, line 67: some 's are not formatted correctly.

src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
...
+  //...
+  @persistencePolicyInDays = {"/recycledChatHistories" : 3650}
+  visitRecord: UserVisitRecord
+  //...
+}
+```
+In this example, schema `EnterpriseUserRecord` reused `UserVisitRecord` and its annotation in the field `visitRecord`, but it overrides the annotation value of `recycledChatHistories`.
+
+
+All above examples shows that inheritance and overrides gives more extensibility to the schema annotations. Users should be free to define their own rules regarding how those annotations can be read. There should be an imambigous annotation value for the fields after user's rules are applied. We call this processing step to figure out the eventual value of a field or schema's annotation "resolution" and we call such value "resolved" value.
```
karthikbalasub, line 93: imambigous -> unambiguous 

Also use third person:
We call this processing step to figure out the eventual value of a field or schema's annotation "resolution" and we call such value "resolved" value.

This processing step .. is called "resolution" and the final value is "resolved" value.

src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
...
+
+What is more, users could have their own customized logic to process the annotations and define the custom behavior when annotations is overriden or inheritted, and even have the flexibility to set a customized resolved value by calling another local function or remote procedure.  The need for such extensibility gives the motivation for Annotation Processor. We aim to build a tool that can process annotations during schema resolution, process all interested annotations by plugin-ing user's own logic and save the "resolved" value back to corresponding DataSchema.
```
karthikbalasub, line 95: "annotations is" -> "annotations are"

src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
...
+
+### A common application of overriding annotation using PathSpec
+In the above `EnterpriseUserRecord` example, we use `@persistencePolicyInDays = {"/recycledChatHistories" : 3650}` do denote the field we are overriding is `recycledChatHistories`. This is basically using PathSpec to reference the field or child schema that needs to be overriden. PathSpec <insert-linke-here> can be used as a path to associate relations between fields among nested schemas. Users can use PathSpec to inambigously specifcy all the paths to the child fields that they want to override.
```
karthikbalasub, line 98: inambigously -> unambiguously

src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
...
+
+
+In Linkedin, such usage is so common that we want to use 
+we provided `PathSpecBaseVisitor`.
```
karthikbalasub, line 102: Needs rewording. 
Suggestion: " The above usage is very common at LinkedIn so an implementation of annotation processing based on this behavior is provided as part of annotation framework"

karthikbalasub, line 102: Introducing Visitor without explaining what it is. I suggest moving the references to PathSpecBaseVisitor to a later seciton.

src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
...
+(2) All *overrides* are applied on fields, not on the record
+
+For more examples regarding the syntax, one can read the java doc from `PathSpecBaseVisitor`.//<insert Link about > and the test cases //<insert Link about >
+
+
+One can seek to extend this case to adapt to their own use cases. Please see next section regarding what class to extend to fit the best use cases.
+
+## Use the schema annotation processor
+We have created a paradigm of processing Rest.Li schema annotations and wrapped them into `com.linkedin.data.schema.annotation` package.
+First thing to understand is that `DataSchema` object internally is stored recursively. We added `resolvedProperties` attribute, in order to store the final reesolved annotation.
```
karthikbalasub, line 115: reesolved -> resolved

karthikbalasub, line 115: You need to clean up this section. 
Not all schemas are recursive.

Also even without recursion in schemas, there are scenarios where you need differnt resolved Properties. Eg, a record used as two fields with different overrides.

src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
...
+```pdl
+record Employee {
+  id: long
+  supervisor: Employee
+}
+```
+
+Its memory presentation is as followed:
+<center>
+<b>DataSchema Example</b><br><img src="{{ 'assets/images/DataSchema_with_resolvedProperties.png' | relative_url }}" />
```
karthikbalasub, line 128: Should this example show different resolved values for Employee::id field?

src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
...
+</center>
+
+The SchemaAnnotationProcessor will process `DataSchema` using `DataSchemaRichContextTraverser` and resolve the annotation for fields in the schemas.
```
karthikbalasub, line 131: Add links for the classes referenced

src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
...
+
+
+<center>
+<b>Sequence Diagram</b><br><img src="{{ 'assets/images/AnnotationProcessor_UML_sequence_diagram.png' | relative_url }}"  />
+</center>
+
+The `DataSchemaRichContextTraverser` travsers the data schema and in turn calls an implementation of `SchemaVisitor`. It is the `SchemaVisitor`' that resolves the annotations for fields in the schemas, based on the context provided by the `DataSchemaRichContextTraverser`. `SchemaVisitor` may also create copy of original data schema if the oroginal data schema needs to be immutable.
```
karthikbalasub, line 138: extra quote for SchemaVisitor

oroginal -> original

Comment about immutable not very clear. Can you add more details?

src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
...
+
+If the overriding of annotations are specified using the `PathSpec` syntax, the `PathSpecBaseVisitor` and `SchemaHandler` class are implemented for such use case. If a user want to implement their own logic, one should look for reimplementing the `SchemaVisitor`, or extending `PathSpecBaseVisitor`, or simply implement `SchemaHandler`
+
```
karthikbalasub, line 141: Maybe adding a table or a flow chart to pick the approach would be nicer.

Annotation based on pathspec based overriding -> create new handler
Annotation based on pathspec with custom traversal -> override visitor and create handler
Annotation not using pathspec based overriding -> create visitor and handler

src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
...
+<center>
+<b>Class Diagram</b><br><img src="{{ 'assets/images/AnnotationProcessor_UML_class_diagram.png' | relative_url }}"  />
+</center>
+
+
+
+
+For more, please ook at the test cases. <insert-me-the-link>
```
karthikbalasub, line 149: ook -> look

src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
+---
+layout: guide
+title: PDL Schema
```
evanw555, line 3: I think you meant to update this to `Schema Annotation Processing`

src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
...
+permalink: /annotation_processor
+excerpt: Documentation of using annotation processor in Rest.Li
+---
+
+# Annotation Processor
+This sections introduces a generic schema annotation processing tool in Rest.Li. Before reading this, it is recommended to be familiar with the concept of PathSpec in Rest.Li. [Insert link here]
```
evanw555, line 9: I notice that you strictly refer to Rest.li as "Rest.Li", I'd suggest doing a search-and-replace to change all instances of "Rest.Li" to "Rest.li"

src/annotation_processor.md
evanw555, line 46: nits:
- `be inherited` -> `to be inherited`
- `from upper level` -> `from the upper level`

src/annotation_processor.md
evanw555, line 12: The way you say "schema writers" sounds impersonal and almost implies that the "writers" are Java Writer classes. I would say "authors"

src/annotation_processor.md
evanw555, line 40: What does the first sentence mean? It sounds like you're saying support was and still is currently limited. Also, I'd recommend not presenting information about features as a story detailing the old support and the new support. This will confuse new users; why does it matter to them that support _used_ to be one way, but _now_ it's another way? It doesn't provide much of a value for users to know that we previously didn't support this. That's my opinion.

nits:
- `let the user chooses` -> `let the user choose`
- `how they want to interpret` -> `how they want it to be interpreted`
- `no preferrabled way` -> `no preferable ways`
- `written, read and interpreted` -> `written, read, and interpreted`

src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
...
+@persistencePolicyInDays = 30
+record  UserVisitRecord {
+  visitedUrls: array[URL]
+  visitedUserProfiles: array[UserRecord]
+}
+```
+Reading from the schema, we might find both `visitedUrls` and `visitedUserProfiles` have `persistencePolicyInDays` annotated as 30.
+//<inheritance example>
+
+Example case 2: Users might want the annotation of a field be overriden. Override is where the annotation on a field or a schema might get updated when other annotations assign it another value.
```
evanw555, line 57: nit: `field be overridden` -> `field to be overridden`

src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
...
+//<override example>
+```pdl
+@persistencePolicyInDays = 30
+record  UserVisitRecord {
```
evanw555, line 61: nit: too much whitespace

src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
...
+  userName: UserName
+  //...
+  @persistencePolicyInDays = {"/recycledChatHistories" : 3650}
+  visitRecord: UserVisitRecord
+  //...
+}
+```
+In this example, schema `EnterpriseUserRecord` reused `UserVisitRecord` and its annotation in the field `visitRecord`, but it overrides the annotation value of `recycledChatHistories`.
+
+
```
evanw555, line 92: nits:
- `examples shows` -> `examples show`
- `overrides gives` -> `overrides give`
- `to the schema annotations` -> `to schema annotations`
- `how those annotations can be read` -> `how those annotations are read`
- `imambiguous` -> `unambiguous`
- `user's rules` -> `the user's rules`
- `call such value "resolved" value` -> `call this value the "resolved" value`

src/annotation_processor.md
evanw555, line 95: nits:
- `inheritted` -> `inherited`
- First sentence is a run-on sentence.
- Extra space at the beginning of the second sentence.
- `interested annotations` -> `relevant annotations` or `applicable annotations`
- `plugin-ing` -> `plugging in`
- `user's own logic` -> `the user's logic`
- `and save the` -> `and saving the`
- `back to corresponding DataSchema` -> `back to the corresponding DataSchema`

src/annotation_processor.md
evanw555, line 98: nits:
- `do denote` -> `to denote`
- `the field we are overriding is` -> `that the field being overridden is`
- `associate relations` -> `define relationships` or `define associations`
- `specifcy` -> `specifcy`
- `to the child fields` -> `to child fields`

src/annotation_processor.md
```diff
@@ -0,0 +1,149 @@
...
+### A common application of overriding annotation using PathSpec
+In the above `EnterpriseUserRecord` example, we use `@persistencePolicyInDays = {"/recycledChatHistories" : 3650}` do denote the field we are overriding is `recycledChatHistories`. This is basically using PathSpec to reference the field or child schema that needs to be overriden. PathSpec <insert-linke-here> can be used as a path to associate relations between fields among nested schemas. Users can use PathSpec to inambigously specifcy all the paths to the child fields that they want to override.
+
+
+In Linkedin, such usage is so common that we want to use
```
evanw555, line 101: nit: `Linkedin` -> `LinkedIn`

src/annotation_processor.md
junchuanwang: I have changed. Meanwhile I really think this is subtle difference and reader might get confused by the term "Pegasus"

src/annotation_processor.md
junchuanwang: Good advice. Thanks

src/annotation_processor.md
junchuanwang: `we don't allow inhertiance other than from records`

I think this is also our own interpretation. It is not necessarily the case for others.

src/annotation_processor.md
junchuanwang: Yes. These were actually marker for myself and forgot to remove, thanks

src/annotation_processor.md
junchuanwang: I think they are. for example `visitedUrls` itself is one word. It is not plural.

src/annotation_processor.md
junchuanwang: Sorry this is incomplete. But I think your reply is good and I will adopt.

src/annotation_processor.md
junchuanwang: Yes you are right and I replaced them all now.

src/annotation_processor.md
junchuanwang: All re-worded

src/annotation_processor.md
junchuanwang: I think this is going to complicate this example(because different resolved value in recursive schema would be an advanced example). I really meant to show how DataSchema is stored as a reference here.

PR #326, opened 2020-06-16
Fix the way schema files get moved to the build directory

Astro03:
This fixes a bug when building the data template jar without doing a gradle clean task, pdsc's that were moved or deleted will still be present in the jar. 

The actual issue is that the pegasus plugin uses the Gradle Copy type task instead of a Sync type.

Things in this PR:
* Created new junitTest and testngTest tasks -> previously the test task did only testng tests
* Added ability to run functional tests in the gradle-plugin module
* Updated the PegasusPlugin to use the sync task to copy the schemas instead of the copy task
* Created functional tests to validate PegasusPlugin copies schemas properly

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -273,9 +273,13 @@ subprojects {
     testOutputDir = sourceSets.test.output.classesDirs.getSingleFile()
   }
 
-  test {
+  task junitTests(type: Test) {
     maxHeapSize = '4g'
+    useJUnit()
+  }
 
+  task testngTests(type: Test) {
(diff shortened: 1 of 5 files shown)


DPUkyle:
I want to use the spock framework for testing pegasus-plugins; it is the _de facto_ standard in the Gradle community. @Astro03 and I will sync-up offline.

Astro03:
Will change this to two separate PR's
1) adding integTest into gradle-plugin with default test
2) the changes for the fix

PR #329, opened 2020-06-17
Only include the underlying exception message in BufferedReaderInputStream

alexjinghn:
only propagate the exception message instead of the entire exception here

### DIFF
--- a/r2-filter-compression/src/main/java/com/linkedin/r2/filter/compression/streaming/BufferedReaderInputStream.java
+++ b/r2-filter-compression/src/main/java/com/linkedin/r2/filter/compression/streaming/BufferedReaderInputStream.java
@@ -54,7 +54,9 @@ class BufferedReaderInputStream extends InputStream implements Reader
 
     if (_throwable != null)
     {
-      throw new IOException(_throwable);
+      // Underlying network layer might throw an exception here and in certain frameworks, the exception class might not be classloaded (e.g. J2EE servlet containers separate server classes from application classes)
+      // however logging framework might try to load the class when logging this exception, creating performance problems.
+      throw new IOException(_throwable.getMessage());
     }
     else if (done())
     {


r2-filter-compression/src/main/java/com/linkedin/r2/filter/compression/streaming/BufferedReaderInputStream.java
```diff
@@ -54,7 +54,9 @@ public int read() throws IOException
 
     if (_throwable != null)
     {
-      throw new IOException(_throwable);
+      // Underlying network layer might throw an exception here and in certain frameworks, the exception class might not be classloaded (e.g. J2EE servlet containers separate server classes from application classes)
+      // however logging framework might try to load the class when logging this exception, creating performance problems.
+      throw new IOException(_throwable.getMessage());
```
ChaoLinkedIn, line 59: This means that you're going to lose the call stacks come with the throwable, right?

Why performance is a concern when there are errors?

alexjinghn: If the throwable is included, log4j might try to classload it (to figure out the jar name of the class) when it is logged. However, this throwable may come from underlying network layer and sometimes classes from that layer (e.g. Jetty) are not visible to the application. In those cases, log4j will never load the class and will repeat trying, each time scanning through the classpath.

ChaoLinkedIn: Understood. If classloader returns after some retries, ie performance issue, it should be ok. However if it stuck here forever, we'll have to figure out how to handle it. Can you confirm which case is this?

alexjinghn: It’s the first case. It’s a nice-to-have feature in log4j. If it can’t find the jar, it will give up and just put a ? there

ChaoLinkedIn: So error does not happen here. It's log4j trying to figure out the unknown stack. I'd vote not to change it here because we might lose some useful info. The app should fix the errors instead of worry about the performance when it happens.

alexjinghn: Errors here are not app level errors. Most of them are network IO errors. eg when client times out, it sends an error to the server which is thrown here. I think we can omit the stacktrace from server here because application really only need to know network io issue happened here. server details shouldn’t be exposed to application anyway. And lastly, the same error would have been logged in the server already.

ChaoLinkedIn: OK then. Approved.

PR #333, opened 2020-06-19
Add new changelog

evanw555:
Replaces the old changelog with a new changelog conforming to the
"[keep-a-changelog](https://keepachangelog.com/en/1.0.0/)" standard, along with convenient links for seeing
which commits are included in each release. Adds a helper script for
updating the changelog and adding links as necessary.

The process to update the changelog is now as follows:
- For **ALL** changes, add entries under the "Unreleased" heading.
- If bumping the project version (i.e. releasing a new version), run the script at `./scripts/update-changelog` to insert a new heading and update links accordingly.

We should try to keep the changelog "semantic" and consumer-facing, meaning that we shouldn't just copy-paste commit messages. We should explicitly mention what behavior has changed and what APIs have changed.


nickibi:
> * If bumping the project version (i.e. releasing a new version), run the script at `./scripts/bump-changelog` to insert a new heading and update links accordingly.

If I understand it correctly, when we release a new version, we should also run `./scripts/bump-changelog`? Do you think this part should be triggered by release cmd?

evanw555:
@karthikbalasub @nickibi That would make sense from a usability perspective, however it doesn't sound possible. The release command is run on a commit that's already been merged, while the changelog command needs to be run before even opening the PR (since the changelog edits need to be committed).

nickibi:
> @karthikbalasub @nickibi That would make sense from a usability perspective, however it doesn't sound possible. The release command is run on a commit that's already been merged, while the changelog command needs to be run before even opening the PR (since the changelog edits need to be committed).

I think it makes sense. However, I think the changelog command should be run just before committing the PR. Since the changelog might be updated by other PRs and this should be the last step before merging the PR.

scripts/help-text/bump-changelog.txt
```diff
@@ -0,0 +1,9 @@
+Usage: ./scripts/bump-changelog [OPTION]...
+Moves everything under the "Unreleased" heading of the changelog to a new version heading based on the version number
```
karthikbalasub, line 2: Does this mean user should bump the version in gradle.properties before running this?  If so, document that.

evanw555: Yes, and I'll make sure this is explicitly stated. I'll also explicitly state this in the dev guide.

As for the naming, I chose `bump-changelog` because it should only be used when bumping to a new version, however `update-changelog` would make more sense to whoever is using it. I'll change it, and I'll also ensure that if someone tries to run it when not bumping the project version, it'll explain to them that the script is only needed when bumping the project version.

PR #328, opened 2020-06-17
Add spock integration testing to the project and gradle-plugin module.

Astro03:
This script Create a script to allow spock integration testing.

As per https://guides.gradle.org/testing-gradle-plugins/#integration-tests:
Integration tests or "functional tests verify the correctness of the [Gradle]
plugin end-to-end. In practice that means applying, configuring and
executing the functionality of the plugin implementation".

* Create a script that adds spock integration testing for a module
* Add integration testing to the gradle-plugin module
* Create an integ test that validates the pegasus plugin can be applied

PR #337, opened 2020-06-22
Update Pegasus Plugin's CopySchema tasks to delete stale schemas 

Astro03:
This PR needs to go into master after https://github.com/linkedin/rest.li/pull/328/

The Issue:
The issue is that the pegasus plugin, when preparing the schemas for
publication, will copy the schemas to the build directory without
removing stale schemas. Without the fix, these stale schemas will be
packaged into the data template jar -> which is a problem.

The current workaround, which is implemented, is that we have a delete
task and a copy task. The delete task is triggered by a gradle project
flag. The copy moves the files as intended. This workaround is not
viable since it breaks the UP-TO-DATE checks between builds. In
particular, it breaks the UP-TO-DATE checks during a HOT RELOAD
(runPlay task of the Play Framework).

The Fix:
Instead of the workaround as above, we "skip" the delete task and make
the copy task become a sync task. The task name stays the same but the
class type will be a sync task which will "Synchronizes the contents of
a destination directory with some source directories and files
[https://docs.gradle.org/current/dsl/org.gradle.api.tasks.Sync.html]".
This task will now delete stale files in the destination and will pass
the 'up-to-date' between builds.

Testing:
Integ test:
1) Validate cacheability/up-to-date of tasks between builds when running the mainDataTemplateJar task. 
2) Validate the mainCopySchemas task will delete stale schemas (this is the issue that was specified above). This was tested using the mainDataTemplateJar task which has a dependency on the mainCopySchemas task.

Unit test:
1) Validate the mainCopySchemas task is a Sync task

In addition, I worked with @DPUkyle to create multiple configurations
to allow the integ test to work. (Kyle's original PR: https://github.com/Astro03/rest.li/pull/2)
 - Pass necessary configurations using sysprops and custom configurations
 - Custom task to keep pegasus-version.properties out of the integTest classpath
 - Override TemporaryFolder to prevent cleanup
 - Use settings.gradle to prevent randomly-generated values in project under test

### DIFF
--- a/gradle-plugins/src/integTest/groovy/com/linkedin/pegasus/gradle/PegasusPluginIntegrationTest.groovy
+++ b/gradle-plugins/src/integTest/groovy/com/linkedin/pegasus/gradle/PegasusPluginIntegrationTest.groovy
@@ -1,10 +1,12 @@
 package com.linkedin.pegasus.gradle
 
+import groovy.json.JsonOutput
 import org.gradle.testkit.runner.GradleRunner
-import org.gradle.testkit.runner.TaskOutcome
 import org.junit.Rule
 import org.junit.rules.TemporaryFolder
 import spock.lang.Specification
+
+import static org.gradle.testkit.runner.TaskOutcome.SUCCESS
 
 class PegasusPluginIntegrationTest extends Specification {
   @Rule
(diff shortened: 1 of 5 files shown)


DPUkyle:
Excellent work, @Astro03. In nearly all cases in Gradle, the `Sync` task type is a superior alternative to `Copy` - and keeps you sane as well. Nice stuff!

gradle-plugins/src/integTest/groovy/com/linkedin/pegasus/gradle/PegasusPluginIntegrationTest.groovy
```diff
@@ -0,0 +1,98 @@
...
+        .withPluginClasspath()
+        .withArguments('mainDataTemplateJar')
+        .forwardOutput()
+        .build()
+
+    then:
+    result.task(':mainDataTemplateJar').outcome == SUCCESS
+  }
+
+  def 'mainCopySchema task will remove stale pdsc'() {
```
DPUkyle, line 32: Nice, this is an excellent test to illustrate the issue behavior.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -57,9 +57,7 @@
 import org.gradle.api.plugins.JavaBasePlugin;
 import org.gradle.api.plugins.JavaPlugin;
 import org.gradle.api.plugins.JavaPluginConvention;
-import org.gradle.api.tasks.Copy;
-import org.gradle.api.tasks.SourceSet;
-import org.gradle.api.tasks.SourceSetContainer;
+import org.gradle.api.tasks.*;
```
DPUkyle, line 60: I would undo this change to the imports (probably IJ organized the imports for you).

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1631,26 +1629,28 @@ protected GenerateDataTemplateTask configureDataTemplateGeneration(Project proje
...
+    });
 
     // Dummy task to maintain backward compatibility, as this task was replaced by CopySchemas
     // TODO: Delete this task once use cases have had time to reference the new task
-    Task copyPdscSchemasTask = project.getTasks().create(sourceSet.getName() + "CopyPdscSchemas");
+    Task copyPdscSchemasTask = project.getTasks().create(sourceSet.getName() + "CopyPdscSchemas", Copy.class);
     copyPdscSchemasTask.dependsOn(destroyStaleFiles);
+    copyPdscSchemasTask.onlyIf(task -> {
+      project.getLogger().lifecycle("{} task is a NO-OP task.", task.getPath());
+      return false;
```
DPUkyle, line 1646: Good use of `Task#onlyIf` to leverage logging

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
Astro03: Updated as stated.  and yes, it was automatic after the 5th one I believe.

Astro03:
@junchuanwang  @evanw555  It looks like the last 2 travis CI checks fail for non-related issues. 

How do you trigger another Travis Run without having to force commit again?

evanw555:
@Astro03 On the build page there's a "Restart Build" button, though you may need to be logged in with access to Travis to press it. I restarted it for you.

junchuanwang:
> This PR needs to go into master after #328
> 
> The Issue:
> The issue is that the pegasus plugin, when preparing the schemas for
> publication, will copy the schemas to the build directory without
> removing stale schemas. Without the fix, these stale schemas will be
> packaged into the data template jar -> which is a problem.
> 
> The current workaround, which is implemented, is that we have a delete
> task and a copy task. The delete task is triggered by a gradle project
> flag. The copy moves the files as intended. This workaround is not
> viable since it breaks the UP-TO-DATE checks between builds. In
> particular, it breaks the UP-TO-DATE checks during a HOT RELOAD
> (runPlay task of the Play Framework).
> 
> The Fix:
> Instead of the workaround as above, we "skip" the delete task and make
> the copy task become a sync task. The task name stays the same but the
> class type will be a sync task which will "Synchronizes the contents of
> a destination directory with some source directories and files
> [https://docs.gradle.org/current/dsl/org.gradle.api.tasks.Sync.html]".
> This task will now delete stale files in the destination and will pass
> the 'up-to-date' between builds.
> 
> Testing:
> Integ test:
> 
> 1. Validate cacheability/up-to-date of tasks between builds
> 2. Validate the mainCopySchemas task will delete stale schemas
> 
> Unit test:
> 
> 1. Validate the mainCopySchemas task is a Sync task
> 
> In addition, I worked with @DPUkyle to create multiple configurations
> to allow the integ test to work. (Kyle's original PR: [Astro03#2](https://github.com/Astro03/rest.li/pull/2))
> 
> * Pass necessary configurations using sysprops and custom configurations
> * Custom task to keep pegasus-version.properties out of the integTest classpath
> * Override TemporaryFolder to prevent cleanup
> * Use settings.gradle to prevent randomly-generated values in project under test

Would you mind expand the detail of your integration test a little bit.Otherwise it seems fine to me

PR #339, opened 2020-06-23
Add PathSpecSet: a set of PathSpecSet with support for full projection

GuiForget:
The PathSpecSet provides functionalities to more easily build or edit a set of PathSpec. In addition, it offers the support for a full projection so that methods like "contains" always return true making it easier for the client code to write logic with the set.

### DIFF
--- a/data-transform/src/main/java/com/linkedin/data/transform/filter/request/MaskCreator.java
+++ b/data-transform/src/main/java/com/linkedin/data/transform/filter/request/MaskCreator.java
@@ -21,7 +21,7 @@
 package com.linkedin.data.transform.filter.request;
 
 import com.linkedin.data.schema.PathSpec;
-
+import com.linkedin.data.schema.PathSpecSet;
 import java.util.Arrays;
 import java.util.Collection;
 
@@ -53,6 +53,26 @@ public class MaskCreator
   public static MaskTree createPositiveMask(Collection<PathSpec> paths)
   {
     return createMaskTree(paths, MaskOperation.POSITIVE_MASK_OP);
+  }
+
+  /**
+   * Create a positive mask for the given set.
+   *
+   * @param pathSpecSet the set that should be in the mask
+   * @return a {@link MaskTree}
+   */
+  public static MaskTree createPositiveMask(PathSpecSet pathSpecSet)
+  {
+    if (pathSpecSet.isAllInclusive())
+    {
+      MaskTree maskTree = new MaskTree();
+      maskTree.addOperation(new PathSpec(PathSpec.WILDCARD), MaskOperation.POSITIVE_MASK_OP);
+      return maskTree;
+    }
+    else
+    {
+      return createMaskTree(pathSpecSet.getPathSpecs(), MaskOperation.POSITIVE_MASK_OP);
+    }
   }
 
   /**
--- a/data-transform/src/test/java/com/linkedin/data/transform/filter/TestMaskCreation.java
+++ b/data-transform/src/test/java/com/linkedin/data/transform/filter/TestMaskCreation.java
@@ -23,15 +23,16 @@ package com.linkedin.data.transform.filter;
 
 import com.linkedin.data.DataMap;
 import com.linkedin.data.schema.PathSpec;
+import com.linkedin.data.schema.PathSpecSet;
 import com.linkedin.data.transform.filter.request.MaskCreator;
 import com.linkedin.data.transform.filter.request.MaskOperation;
 import com.linkedin.data.transform.filter.request.MaskTree;
 
 import java.io.IOException;
 
-import java.nio.file.Path;
 import java.util.Map;
 import org.testng.Assert;
+import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 
 import static com.linkedin.data.TestUtil.dataMapFromString;
@@ -372,4 +373,36 @@ public class TestMaskCreation
     mask.addOperation(new PathSpec("a"), MaskOperation.NEGATIVE_MASK_OP);
     Assert.assertEquals(mask.toString(), "{a=0}");
   }
+
+  @Test(dataProvider = "pathSpecSetProvider")
+  public void testPositiveMaskWithPathSpecSet(PathSpecSet input, MaskTree expected)
+  {
+    Assert.assertEquals(MaskCreator.createPositiveMask(input).getOperations(), expected.getOperations());
+  }
+
+  @DataProvider
+  public static Object[][] pathSpecSetProvider()
+  {
+    MaskTree simpleMaskTree = new MaskTree();
+    simpleMaskTree.addOperation(new PathSpec("myField"), MaskOperation.POSITIVE_MASK_OP);
+    MaskTree fullMaskTree = new MaskTree();
+    fullMaskTree.addOperation(new PathSpec(PathSpec.WILDCARD), MaskOperation.POSITIVE_MASK_OP);
+
+    return new Object[][] {
+        {
+            PathSpecSet.of(new PathSpec("myField")),
+            simpleMaskTree
+        },
+        {
+            PathSpecSet.empty(),
+            new MaskTree()
+        },
+        {
+            PathSpecSet.allInclusive(),
+            fullMaskTree
+        }
+    };
+  }
+
+
 }
--- a/data/src/main/java/com/linkedin/data/schema/PathSpecSet.java
+++ b/data/src/main/java/com/linkedin/data/schema/PathSpecSet.java
@@ -1,0 +1,426 @@
+/*
+   Copyright (c) 2021 LinkedIn Corp.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package com.linkedin.data.schema;
+
+import com.linkedin.data.template.RecordTemplate;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+
+
+/**
+ * Represents an immutable set of {@link PathSpec}s.
+ *
+ * <p>Path spec set is a convenient wrapper for a collection of {@link PathSpec}. A few advantages of using this class
+ * over manually passing around {@code Set<PathSpec>} is:
+ *
+ * <ul>
+ *   <li>guaranteed immutable representation</li>
+ *   <li>
+ *     explicitly supports "all inclusive" which translates into fetching everything (default for Rest.li resources),
+ *     as well as "empty" which translates into fetching no fields
+ *   </li>
+ *   <li>mutable builder for incrementally assembling an immutable {@link PathSpecSet}</li>
+ *   <li>built-in utility to mask a {@link RecordTemplate} with the stored path specs</li>
+ * </ul>
+ *
+ * @author Joseph Florencio
+ */
+final public class PathSpecSet
+{
+  private final static PathSpecSet EMPTY = new PathSpecSet(Collections.emptySet(), false);
+  private final static PathSpecSet ALL_INCLUSIVE = new PathSpecSet(Collections.emptySet(), true);
+
+  private final Set<PathSpec> _pathSpecs;
+  private final boolean _allInclusive;
+
+  private PathSpecSet(Builder builder)
+  {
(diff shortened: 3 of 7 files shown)


junchuanwang:
Hey @GuiForget Is this still a draft or published?

GuiForget:
@junchuanwang it's still a draft, I need to finish the unit tests. I made some progress a couple of weeks ago but I have still several code paths missing

GuiForget:
@junchuanwang this is ready to be reviewed. One functionality I wanted to add but I'm not sure where to add it is the ability to construct a PathSpecSet from a MaskTree. This functionality would be handy given that the annotation @ProjectionParam is usable only with MaskTree at the moment. Originally I was hoping to add a method `PathSpeSet.of(MaskTree)` but MaskTree class is not available to the `data` module. Do you have a recommended place I could add it?

data-transform/src/main/java/com/linkedin/data/transform/filter/request/MaskCreator.java
```diff
@@ -57,6 +57,23 @@ public static MaskTree createPositiveMask(Collection<PathSpec> paths)
...
 
+  /**
+   * Create a positive mask for the given set.
+   *
+   * @param pathSpecSet the set that should be in the mask
+   * @return a {@link MaskTree}
+   */
+  public static MaskTree createPositiveMask(PathSpecSet pathSpecSet)
+  {
+    if (pathSpecSet.isFullProjection()) {
```
BrianPin, line 68: braces on new line?

data-transform/src/test/java/com/linkedin/data/transform/filter/TestMaskCreation.java
```diff
@@ -370,4 +372,34 @@ public void testComposingNegativeMaskWithPositiveSubmasks() throws IOException
     mask.addOperation(new PathSpec("a"), MaskOperation.NEGATIVE_MASK_OP);
     Assert.assertEquals(mask.toString(), "{a=0}");
   }
+
+  @Test(dataProvider = "pathSpecSetProvider")
+  public void testPositiveMaskWithPathSpecSet(PathSpecSet input, MaskTree expected) {
```
BrianPin, line 377: braces on new line?

data-transform/src/main/java/com/linkedin/data/transform/filter/request/MaskCreator.java
GuiForget: Is there a code style I can refer to?

BrianPin: Good q. I think it has been left braces on next line ever since restli first version.
Although I don't think there is strict rules regarding coding style, (for example, the line width)
I just don't want to see in the future there will be many places where mixed type is happening inside the repo

data/src/main/java/com/linkedin/data/schema/PathSpecSet.java
```diff
@@ -0,0 +1,377 @@
+package com.linkedin.data.schema;
```
BrianPin, line 1: Might as well add copy-right file headers?

BrianPin (commented):
Overall looks good to me, but please change the style to match restli style.

data/src/main/java/com/linkedin/data/schema/PathSpecSet.java
```diff
@@ -0,0 +1,377 @@
...
+ *   </li>
+ *   <li>mutable builder for incrementally assembling an immutable {@link PathSpecSet}</li>
+ *   <li>built-in utility to mask a {@link RecordTemplate} with the stored path specs</li>
+ * </ul>
+ *
+ * @author Joseph Florencio
+ */
+final public class PathSpecSet {
+  private final static PathSpecSet _EMPTY = new PathSpecSet(Collections.emptySet(), false);
+  private final static PathSpecSet _FULL_PROJECTION = new PathSpecSet(Collections.emptySet(), true);
```
karthikbalasub, line 38: remove the _ prefix (not the convention followed in this project)

data/src/main/java/com/linkedin/data/schema/PathSpecSet.java
```diff
@@ -0,0 +1,377 @@
...
+
+    if (this.isEmpty()) {
+      return PathSpecSet.empty();
+    }
+
+    Builder builder = newBuilder();
+
+    this.getPathSpecs().stream()
+        .map(childPathSpec -> {
+          ArrayList<String> list = new ArrayList<>();
```
karthikbalasub, line 227: initialize size to parent path size + child path size

data/src/main/java/com/linkedin/data/schema/PathSpecSet.java
```diff
@@ -0,0 +1,377 @@
...
+   * @return a {@link PathSpecSet} with elements starting with the input {@link PathSpec} prefix
+   */
+  public PathSpecSet copyAndRemovePrefix(PathSpec prefix) {
+    if (isFullProjection() || isEmpty()) {
+      // full or empty projections stay the same
+      return this;
+    }
+
+    // if we contain the exact prefix PathSpec, it should be a full projection
+    if (getPathSpecs().contains(prefix)) {
```
karthikbalasub, line 267: This should check for the prefix or any parent of the perfix, right?

data/src/main/java/com/linkedin/data/schema/PathSpecSet.java
evanw555: +1. Add copyright comment at the top of all new source files. See existing files for reference.

data-transform/src/test/java/com/linkedin/data/transform/filter/TestMaskCreation.java
evanw555: +1. This is the convention in Rest.li

data-transform/src/main/java/com/linkedin/data/transform/filter/request/MaskCreator.java
evanw555: Yeah, AFAIK the Rest.li code style isn't documented formally. It's just based on convention

data/src/main/java/com/linkedin/data/schema/PathSpecSet.java
GuiForget: yes you are correct. I'll modify accordingly

GuiForget:
I'll look at the build failure... But all the tests pass locally for me after the rebase

evanw555:
@GuiForget Looks like you've been running into some unfortunate flakiness from the d2/darkcluster tests. You can trigger a re-run in the CI by force-pushing.

GuiForget:
@evanw555 today's rebase passed!

data/src/main/java/com/linkedin/data/schema/PathSpecSet.java
```diff
@@ -0,0 +1,440 @@
...
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+
+
+/**
```
evanw555, line 34: For all javadocs in this class, I'd suggest providing examples using the serialized pathspec syntax (e.g. `/foo/bar`, `/foo`) for clarity, especially in more complicated examples (e.g. copy with scope, remove prefix)

data/src/main/java/com/linkedin/data/schema/PathSpecSet.java
```diff
@@ -0,0 +1,440 @@
+/*
+   Copyright (c) 2012 LinkedIn Corp.
```
evanw555, line 2: Should say 2020

data/src/main/java/com/linkedin/data/schema/PathSpecSet.java
```diff
@@ -0,0 +1,440 @@
...
+/**
+ * Represents an immutable set of {@link PathSpec}s.
+ *
+ * <p>Path spec set is a convenient wrapper for a collection of {@link PathSpec}. A few advantages of using this class
+ * over manually passing around {@code Set<PathSpec>} is:
+ *
+ * <ul>
+ *   <li>guaranteed immutable representation</li>
+ *   <li>
+ *     explicitly supports "full projection" which translates into fetch everything (default for rest.li resources),
```
evanw555, line 43: typo: explicitly

evanw555, line 43: nit: capitalize as "Rest.li"

data/src/main/java/com/linkedin/data/schema/PathSpecSet.java
```diff
@@ -0,0 +1,440 @@
...
+ *
+ * @author Joseph Florencio
+ */
+final public class PathSpecSet
+{
+  private final static PathSpecSet EMPTY = new PathSpecSet(Collections.emptySet(), false);
+  private final static PathSpecSet FULL_PROJECTION = new PathSpecSet(Collections.emptySet(), true);
+
+  private final Set<PathSpec> _pathSpecs;
+  private final boolean _fullProjection;
```
evanw555, line 58: I'm skeptical about using the terminology "full projection", as PathSpecs are not inherently tied to Rest.li projections (that's just one application). However, one could make the argument that the term "projection" is a generic term and isn't implying that it's used for _Rest.li_ projections. What do you think?

I'm also curious to know how a "full projection" PathSpecSet is used. Do you have an example of one use case?

junchuanwang: maybe we can change to name "universal". Otherwise the "full" return a empty set and NULL in toArray seems unnatural to me as well

junchuanwang:
I have no other concerns other than the "fullProjection" boolean“ field.

data/src/main/java/com/linkedin/data/schema/PathSpecSet.java
GuiForget: In Rest.li world, when you have a PathSpecSet you may have logic that is skipped if the projection doesn't include a specific field:
```
if (pathSpecSet.contains(MyRecord.fields().subrecord()) {
    // do a bunch of stuff that is costly but unnecessary if the field subrecord isn't requested
}
```

And then given the ResourceContext, you can easily build a PathSpecSet with `PathSpecSet.of(resourceContext.getProjectionMask())`. 

When no projection was specified, the ProjectionMask is null and you want to make sure that `contains(PathSpec)` returns true for any PathSpec and that's the meaning of a fullProjection.

GuiForget: For the `toArray()` returning `null`, this is consistent with the behavior of the method `fields(PathSpec[])` in the Rest.li builders.

GuiForget: I'm used to full projection so it is difficult for me to come up with an alternative. I'm not sure universal as an adjective works but I could see `universe` or `universeSet` as suitable. Here are some alternatives that I found in a thesaurus dictionary:
- globalSet
- allInclusiveSet
- blanketSet

data/src/main/java/com/linkedin/data/schema/PathSpecSet.java
GuiForget: I edited the javadoc, let me know if this is what you had in mind

data/src/main/java/com/linkedin/data/schema/PathSpecSet.java
GuiForget: I can't find the typo

data/src/main/java/com/linkedin/data/schema/PathSpecSet.java
evanw555: > When no projection is specified, the ProjectionMask is null and you want to make sure that `contains(PathSpec)` returns true for any PathSpec and that's the meaning of a fullProjection.

In this specific use case it makes sense, but this object is simply a set of path specs. An empty path spec set articulates to a full projection when it's used in the specific application of projection, but path specs aren't only for projections. What does an empty path spec set mean when it's being used for indexing values from a complex data object, or matching ReadOnly/CreateOnly predicates? I don't necessarily think that an empty path spec set contains everything for all use cases.

junchuanwang: thanks for the explanation, I will vote for "allInclusiveSet".

evanw555: Synced up with @GuiForget about this offline. He clarified that an empty set is not necessarily the same as an "all inclusive" or "full projection" set, which must be created explicitly as such.

As for naming, I say we avoid using the terminology "full projection", and instead use something that implies inclusivity. `allInclusive`/`isAllInclusive` works for me. `complete` may also be good, but I'm not sure if that has a different technical definition when it comes to sets.

data/src/main/java/com/linkedin/data/schema/PathSpecSet.java
evanw555: Adding `/foo` directly into code segments may be confusing, as it's not proper Java syntax. However, I still think it's more clear to users. I originally meant adding this in addition to the code examples, like `set of {/a} contains /a` or something like that.

data/src/main/java/com/linkedin/data/schema/PathSpecSet.java
GuiForget: I renamed fullProjection to allInclusive and propagated the changes throughout the classes and unit tests

evanw555 (approved):
Thanks for making the suggested changes and thanks for contributing, LGTM. Should probably wait for approval from the other reviewers before merging.

(also nit: but please add an entry to the change for this if you have time)

BrianPin (approved):

PR #341, opened 2020-06-24
[DO NOT CHECK IN] Revert "Add spock integration testing to the project and gradle-plugi…

Astro03:
…ns module. (#328)"

This reverts commit 779fd16dcbf65a630b929ff50d35de73055abeb9.

Using this to see if there is something wrong with these changes that
are causing the Travis CI builds after this commit to all fail.


Astro03:
closing since this was not necessary.

PR #346, opened 2020-07-02
Disable string interning in Jackson JSON since it causes GC issues

karthikrg:
1. Make JsonFactory parameterized so that multiple codec instances can share the same factory if necessary
2. Make the default factory a static singleton for factory reuse to minimize garbage

Interning seems to be messing with batch_get where the keys are not schema but are values, resulting in a large number of strings being interned.

### DIFF
--- a/data/src/main/java/com/linkedin/data/codec/AbstractJacksonDataCodec.java
+++ b/data/src/main/java/com/linkedin/data/codec/AbstractJacksonDataCodec.java
@@ -51,6 +51,18 @@ import java.util.TreeMap;
  */
 public abstract class AbstractJacksonDataCodec implements DataCodec
 {
+  /**
+   * Default factory to be shared between all jackson JSON codec instances. This is done to maximize factory reuse for
+   * performance reasons as recommended by Jackson authors.
+   *
+   * <a href="https://github.com/FasterXML/jackson-docs/wiki/Presentation:-Jackson-Performance">Jackson Performance</a>
+   */
+  public static final JsonFactory JSON_FACTORY = new JsonFactory();
+  static {
+    // Disable string interning by default since it causes GC issues.
+    JSON_FACTORY.disable(JsonFactory.Feature.INTERN_FIELD_NAMES);
+  }
+
   protected static final int DEFAULT_BUFFER_SIZE = 4096;
 
   protected final JsonFactory _factory;
--- a/data/src/main/java/com/linkedin/data/codec/JacksonDataCodec.java
+++ b/data/src/main/java/com/linkedin/data/codec/JacksonDataCodec.java
@@ -48,7 +48,7 @@ public class JacksonDataCodec extends AbstractJacksonDataCodec implements TextDa
 
   public JacksonDataCodec()
   {
-    this(new JsonFactory());
+    this(JSON_FACTORY);
   }
 
   public JacksonDataCodec(JsonFactory jsonFactory)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.4.4] - 2020-07-02
+- Disable string interning in Jackson JSON since it causes GC issues (#346)
+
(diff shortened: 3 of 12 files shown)


data/src/main/java/com/linkedin/data/codec/entitystream/AbstractJacksonDataEncoder.java
```diff
@@ -46,12 +46,16 @@ protected AbstractJacksonDataEncoder(JsonFactory jsonFactory, DataMap dataMap, i
...
+    // Disable field name interning since it causes GC issues.
+    _jsonFactory.disable(JsonFactory.Feature.INTERN_FIELD_NAMES);
   }
 
   protected AbstractJacksonDataEncoder(JsonFactory jsonFactory, DataList dataList, int bufferSize)
   {
     super(dataList, bufferSize);
     _jsonFactory = jsonFactory;
+    // Disable field name interning since it causes GC issues.
+    _jsonFactory.disable(JsonFactory.Feature.INTERN_FIELD_NAMES);
```
aman1309, line 58: this is shouldn't be there in abstract class right? what if someone wants to override it?
same for other places

data/src/main/java/com/linkedin/data/codec/JacksonDataCodec.java
```diff
@@ -42,13 +42,24 @@
...
+  /**
+   * Default factory to be shared by all codec instances. This is done to maximize factory reuse for
+   * performance reasons as recommended by Jackson authors.
+   *
+   * <a href="https://github.com/FasterXML/jackson-docs/wiki/Presentation:-Jackson-Performance">Jackson Performance</a>
+   */
+  private static final JsonFactory JSON_FACTORY = new JsonFactory();
+  static {
+    JSON_FACTORY.disable(JsonFactory.Feature.INTERN_FIELD_NAMES);
+  }
```
aman1309, line 54: not required if you're doing it in abstract classes

data/src/main/java/com/linkedin/data/codec/entitystream/AbstractJacksonDataEncoder.java
karthikrg: Safe to do it here since its the topmost. Subclasses can change this if they want to.

karthikbalasub: This wouldn't allow someone to override the interning feature. Subclasses cannot override this behavior. 
Same for the decoder.

data/src/main/java/com/linkedin/data/codec/AbstractJacksonDataCodec.java
```diff
@@ -51,13 +51,26 @@
...
+
   protected static final int DEFAULT_BUFFER_SIZE = 4096;
 
   protected final JsonFactory _factory;
 
   protected AbstractJacksonDataCodec(JsonFactory factory)
   {
     _factory = factory;
+    // Disable string interning since it causes GC issues.
+    _factory.disable(JsonFactory.Feature.INTERN_FIELD_NAMES);
```
karthikbalasub, line 73: This means it wouldn't be possible to override this behavior to enable interning

karthikbalasub:
Also, please update the commit description to include the batch_get issue wrt interning

data/src/main/java/com/linkedin/data/codec/AbstractJacksonDataCodec.java
karthikrg: You can. You can always inject a custom factory and in your constructor enable interning by explicitly invoking it after calling super.

karthikbalasub: It should be possible to customize the jsonfactory with the existing codec classes, without having to create a subclass.

For example, it would help some service to override the default behavior. 
Since we are disabling interning when the default constructor is used, it should cover the default behavior. When a custom json factory is provided, we should treat it as override and not force behavior on it.

PR #347, opened 2020-07-14
D2 relative strategy integration test

rachelhanhan:
Changes include:
1. A test framework to provide different call count, latency, error rate in multiple intervals to test load balancer strategies. Users can plugin different behaviors to perform all the scenario tests.
2. The actual integration test for RelativeLoadBalancer. Some of the tests run against both relative strategy and the old degrader strategy, and verify the expected behaviors are the same.

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/RelativeLoadBalancerTestHelper.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/RelativeLoadBalancerTestHelper.java
@@ -17,6 +17,7 @@
 package com.linkedin.d2.balancer.strategies.relative;
 
 import java.net.URI;
+import java.util.Collections;
 import java.util.Map;
 
 
@@ -34,6 +35,6 @@ public class RelativeLoadBalancerTestHelper {
    */
   public static Map<URI, Integer> getPointsMap(RelativeLoadBalancerStrategy strategy, int partitionId)
   {
-    return strategy.getPointsMap(partitionId);
+    return Collections.unmodifiableMap(strategy.getPointsMap(partitionId));
   }
 }
--- a/d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/RelativeLoadBalancerTestHelper.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/RelativeLoadBalancerTestHelper.java
@@ -1,5 +1,5 @@
 /*
-   Copyright (c) 2012 LinkedIn Corp.
+   Copyright (c) 2020 LinkedIn Corp.
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
@@ -20,7 +20,18 @@ import java.net.URI;
 import java.util.Map;
 
 
+/**
+ * The helper class for {@link RelativeLoadBalancerStrategy} related tests
+ */
 public class RelativeLoadBalancerTestHelper {
+
+  /**
+   * Get points map for a given partition
+   *
+   * @param strategy The object of the strategy
+   * @param partitionId The id of the partition
+   * @return The points map
+   */
   public static Map<URI, Integer> getPointsMap(RelativeLoadBalancerStrategy strategy, int partitionId)
   {
     return strategy.getPointsMap(partitionId);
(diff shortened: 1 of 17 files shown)


d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/RelativeLoadBalancerTestHelper.java
```diff
@@ -0,0 +1,39 @@
...
+  /**
+   * Get points map for a given partition
+   *
+   * @param strategy The object of the strategy
+   * @param partitionId The id of the partition
+   * @return The points map
+   */
+  public static Map<URI, Integer> getPointsMap(RelativeLoadBalancerStrategy strategy, int partitionId)
+  {
+    return strategy.getPointsMap(partitionId);
```
ChaoLinkedIn, line 37: Wrap with Collections.unmodifiableMap

d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/RelativeLoadBalancerTestHelper.java
```diff
@@ -0,0 +1,39 @@
...
+public class RelativeLoadBalancerTestHelper {
+
+  /**
+   * Get points map for a given partition
+   *
+   * @param strategy The object of the strategy
+   * @param partitionId The id of the partition
+   * @return The points map
+   */
+  public static Map<URI, Integer> getPointsMap(RelativeLoadBalancerStrategy strategy, int partitionId)
```
ChaoLinkedIn, line 35: We should have an interface to expose pointsMap (not only for testing) because we'll have MBean to expose the values to inGraphs.

rachelhanhan: I'm actually thinking about if the points are really meaningful to be collected in the inGraph, for now I prefer not adding points into the inGraph any more, simply because I did not see much value to give the total points of a ring. I started a doc to add some new monitors: https://docs.google.com/document/d/19DpiTNuzXpdSaK-kYPRYg9RJenxJfuIy5zYX_l_1gtA/edit#heading=h.r3ucjtwt0vl4, I will spend some time discuss with you and other folks on the inGraphs to be added.

If this is only for testing, I don't want to expose it as an interface.

d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/RelativeLoadBalancerTestHelper.java
rachelhanhan: Sure.

Also replying to the facilities question here: the new relative load balancer is not an actual "load balancer" in the implementation, it is just a new strategy, it still uses the existing SimpleLoadBalancer, and I don't think simpleLoadBalancer extends Facilities, but LastSeenLoadBalancer does, and it uses the SimpleLoadBalancer internally.

ChaoLinkedIn: ok

d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/RelativeLoadBalancerTestHelper.java
ChaoLinkedIn: If you want to remove this one, it'll still be good to have something to replace.

The total points can reflect the internal state changes that we do not necessarily see just from the latency ingraphs. For example, we can see if the client degrades some hosts, or if the servers never got out of slowStart state. Just looking into error rate or latency will not get your idea. Also note that ingraphs has the granularity of 1 min, while d2 update state update every 5s. Some times the difference can reflect some minor changes.

rachelhanhan: I'm thinking about adding the number of unhealthy hosts (any host that has health score < 1 would count) in the inGraph.
Previously it uses total points in hash ring, which does not tell how many hosts have total points less than 100, the total points just does not make sense to me. We can also discuss more about this offline.

ChaoLinkedIn: Number of unhealthy hosts are good, but it cannot reflect how fast the servers are changing (degrade/recover). You may consider to add those metrics as well.

PR #350, opened 2020-07-17
Improve performance of uri mask encoding and decoding

karthikbalasub:
Fixes the following issues:
Encoding:
 - Projection parameters were first encoded using default logic and then overwritten using the encoded value from URIMask encoding. Avoiding the unnecessary double encoding for projection parameters.
 - During mask creation, we iterate through the entire datamap to cleanup empty masks. Since {'a': 1 } and { 'a' : { '$*': 1} } are functionally equivalent, we can avoid the extra iteration by setting $*:1.
   This might increase the URI size slightly, but the chance of someone using array start/count with default values for both is going to be very low, so doing this optimization makes sense.
 - During encoding a string is created for each sub datamap and then appended to the parent string builder. This is avoided by passing in the string builder to encode into.
Decoding:
 - The decoding logic was poorly designed. It was deleting characters from the front of string builder. This is avoided by passing in the string param directly and checking chars using charAt.

### DIFF
--- a/data-transform/src/main/java/com/linkedin/data/transform/filter/MaskComposition.java
+++ b/data-transform/src/main/java/com/linkedin/data/transform/filter/MaskComposition.java
@@ -90,6 +90,11 @@ public class MaskComposition implements Interpreter
                          instrCtx);
           }
         }
+        // This can happen if the mask is for an array field and the merged start/count resulted in default values.
+        // Setting the wildcard mask to represent all items are included.
+        if (data.isEmpty()) {
+          data.put(FilterConstants.WILDCARD, FilterConstants.POSITIVE);
+        }
       }
     }
   }
--- a/data-transform/src/main/java/com/linkedin/data/transform/filter/MaskComposition.java
+++ b/data-transform/src/main/java/com/linkedin/data/transform/filter/MaskComposition.java
@@ -92,7 +92,8 @@ public class MaskComposition implements Interpreter
         }
         // This can happen if the mask is for an array field and the merged start/count resulted in default values.
         // Setting the wildcard mask to represent all items are included.
-        if (data.isEmpty()) {
+        if (data.isEmpty())
+        {
           data.put(FilterConstants.WILDCARD, FilterConstants.POSITIVE);
         }
       }
--- a/restli-common/src/main/java/com/linkedin/restli/internal/common/URIMaskUtil.java
+++ b/restli-common/src/main/java/com/linkedin/restli/internal/common/URIMaskUtil.java
@@ -14,19 +14,15 @@
    limitations under the License.
 */
 
-/**
- * $Id: $
- */
 package com.linkedin.restli.internal.common;
 
+import com.linkedin.data.DataMap;
 import com.linkedin.data.transform.filter.FilterConstants;
+import com.linkedin.data.transform.filter.request.MaskOperation;
+import com.linkedin.data.transform.filter.request.MaskTree;
 import java.util.ArrayDeque;
 import java.util.Deque;
 import java.util.Map;
-
-import com.linkedin.data.DataMap;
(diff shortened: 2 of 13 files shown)


data-transform/src/main/java/com/linkedin/data/transform/filter/MaskComposition.java
```diff
@@ -90,6 +90,11 @@ public void interpret(InterpreterContext instrCtx)
                          instrCtx);
           }
         }
+        // This can happen if the mask is for an array field and the merged start/count resulted in default values.
+        // Setting the wildcard mask to represent all items are included.
+        if (data.isEmpty()) {
```
BrianPin, line 95: although i like this coding style, but it seems it is not consistent with rest of the code.

BrianPin: I mean the left bracket should be in a new line

restli-common/src/main/java/com/linkedin/restli/internal/common/URIMaskUtil.java
```diff
@@ -115,137 +110,162 @@ else if (entry.getValue()
...
               }
 
-              ii++;
+              fieldIndex++;
 
               // Aggressively consume the numerical value for the range parameter as this is a special case.
-              StringBuilder rangeValue = new StringBuilder();
-              while (ii < toparse.length())
+              int rangeValue = 0;
+              while (fieldIndex < toparse.length() && nextToken == -1)
```
BrianPin, line 201: This is more like a question, `nextToken` is set when we see a comma ',',  and this while loop is condition to enter when the `nextToken` is set to `-1`. Do we design to process only the token before comma ','?

If the token list looks like "token1, token2"  then it seems this token2 is not going to be processed or I missed the fact that the use case will not happen.

restli-common/src/main/java/com/linkedin/restli/internal/common/URIMaskUtil.java
```diff
@@ -115,137 +110,162 @@ else if (entry.getValue()
...
-              + toparse + "' while traversing");
+              + toparse.substring(index) + "' while traversing");
         }
-        toparse.delete(0, 1);
+        index++;
         state = ParseState.PARSE_FIELDS;
         break;
       case DESCEND:
-        if (toparse.indexOf(":(") != 0)
+        if (toparse.charAt(index) != ':' || toparse.charAt(index + 1) != '(')
```
logancarmody, line 152: check for index out of bounds

data-transform/src/main/java/com/linkedin/data/transform/filter/MaskComposition.java
karthikbalasub: Fixed.

restli-common/src/main/java/com/linkedin/restli/internal/common/URIMaskUtil.java
karthikbalasub: This code block handles the start and count params in the projection. 
For example: $*:(c),$start:2,$count:4

We are parsing for digits after we see the field name as "$start" or "$count" (see lines 189-190). 
So we move to the next token if we see "," or ")" (valid), parse the number if we see digit (valid) and anything else is invalid.

karthikbalasub: This code block handles the start and count params in the projection. 
For example: $*:(c),$start:2,$count:4

We are parsing for digits after we see the field name as "$start" or "$count" (see lines 189-190). 
So we move to the next token if we see "," or ")" (valid), parse the number if we see digit (valid) and anything else is invalid.

PR #352, opened 2020-07-22
D2 CallTracker Long Array Tracking Optimization

rickzx:
Changes:
Separates percentile tracking from basic tracking (min/max, average, call count, std), because percentile tracking has memory efficiency issues with resizable long array.

The new LongTrackingWithPercentile class is the same as the old LongTracking class. The new LongTracking class removes percentile tracking information and is memory efficient. Users should always use LongTracking when percentile information is not essential.


pegasus-common/src/main/java/com/linkedin/common/stats/LongTracking.java
```diff
@@ -1,53 +1,13 @@
...
- * Maintain a collection of values and provide the count, average, standard deviation,
- * minimum, maximum, percentile values for the collection.
+ * Tracks count, average, standard deviation, minimum and maximum
+ * in a memory-efficient way.
  *
- * This class implementation is not synchronized. If concurrent access is required, it
- * must be synchronized externally.
+ * If percentile information is required, use {@link LongTrackingWithPercentile}.
  */
 public class LongTracking implements LongTracker
```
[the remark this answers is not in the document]
rickzx: I see. Changed LongTrackingWithPercentile back to LongTracking, changed LongTracking to SimpleLongTracking.

pegasus-common/src/main/java/com/linkedin/common/stats/LongTracking.java
```diff
@@ -1,53 +1,13 @@
-/*
-   Copyright (c) 2012 LinkedIn Corp.
```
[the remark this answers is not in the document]
rickzx: Fixed

pegasus-common/src/main/java/com/linkedin/common/stats/LongTracking.java
```diff
@@ -120,202 +33,53 @@ else if (value > _max)
...
-  }
-
-  public int getMaxCapacity()
-  {
-    return _maxCapacity;
+  @Override
+  public LongStats getStats() {
+    return new LongStats(getCount(), getAverage(), getStandardDeviation(),
+        getMinimum(), getMaximum(),
+        -1L, -1L, -1L, 1L);
```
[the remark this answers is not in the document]
rickzx: Good catch. Fixed

degrader/src/main/java/com/linkedin/util/degrader/CallTrackerImpl.java
```diff
@@ -439,13 +440,13 @@ public void trackCallWithError(long duration)
...
     private int _concurrentMax;
-    private final LongTracking _callTimeTracking;
+    private final LongTracker _callTimeTracking;
     //this map is used to store the number of specific errors that happened in one interval only
     private final Map<ErrorType, Integer> _errorTypeCounts;
 
     private Tracker()
     {
-      _callTimeTracking = new LongTracking();
+      _callTimeTracking = new LongTrackingWithPercentile();
```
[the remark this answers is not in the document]
rickzx: Logic added.

pegasus-common/src/main/java/com/linkedin/common/stats/LongTracking.java
```diff
@@ -80,22 +62,21 @@ public LongTracking(final int maxCapacity, int initialCapacity, double growthFac
     {
       growthFactor = DEFAULT_GROWTH_FACTOR;
     }
+
     _buffer = new long[initialCapacity];
     _bufferSize = initialCapacity;
     _initialCapacity = initialCapacity;
     _growthFactor = growthFactor;
     _maxCapacity = maxCapacity;
+    _simpleLongTracking = new SimpleLongTracking();
```
rachelhanhan, line 71: Wondering what's the difference between doing this and let this class extend the SimpleLongTracking.java, and use super.xxx() instead. Which one is a better option?

pegasus-common/src/main/java/com/linkedin/common/stats/SimpleLongTracking.java
```diff
@@ -0,0 +1,106 @@
+/*
+   Copyright (c) 2012 LinkedIn Corp.
```
rachelhanhan, line 2: Minor: 2020 instead

pegasus-common/src/main/java/com/linkedin/common/stats/LongTracking.java
```diff
@@ -271,51 +212,4 @@ private void grow()
     _buffer = newBuffer;
     _bufferSize = newBufferSize;
   }
-
-  private static double safeDivide(final double numerator, final double denominator)
-  {
-    return denominator != 0 ? numerator / denominator : 0;
-  }
-
-  public static class Config
```
rachelhanhan, line 280: Just want to double check with you that there is no other mps that is using config to create new instance. If not, feel free to directly drop this comment.

pegasus-common/src/main/java/com/linkedin/common/stats/LongTracking.java
rickzx: I personally prefer delegation over inheritance, because I would like to separate out the basic tracking logic from percentile tracking logic. The SimpleLongTracking instance acts just like a blackbox. I think the difference is pretty subtle here though. Feel free to share your opinion on this.

pegasus-common/src/main/java/com/linkedin/common/stats/SimpleLongTracking.java
rickzx: Fixed

pegasus-common/src/main/java/com/linkedin/common/stats/LongTracking.java
rickzx: It only occurs in a test file for LongTracking in the internal `util` mp. Please me know if this class should be added back.

rachelhanhan: That should be fine then, util test should use the util config instead of the pegasus config class.

pegasus-common/src/main/java/com/linkedin/common/stats/LongTracking.java
rachelhanhan: OK, I don't have a strong preference here

PR #353, opened 2020-07-23
Make addFields on AbstractRequestBuilder public.

ianvkoeppe:
We have a use case where we want a method to be able to take any type of
request builder (e.g. BatchGetRequestBuilder or FindRequestBuilder) and
we want to be able to add fields to the request.

Currently, we can accomplish this by using
setParam(RestConstants.FIELDS_PARAM, newHashSet(fields)), but it would
be simpler to just expose the addFields method instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
+- Provide public method in the AbstractRequestBuilder for adding field projections (#353)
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
+- Provide public method in the AbstractRequestBuilder for adding field projections.
```
nickibi, line 16: Can you please include the PR# in the changelog, like (#353).

PR #356, opened 2020-07-29
Remove stacktrace from RestException and StreamException 

xuli-in:
Allow RestRestliServer and StreamRestliServer throw RestException and StreamException with no stacktrace
RestException and StreamException are nested exceptions wrapped around low level throwable, usually a restliServerException.
Remove stack trace for outter nested exception can save time and reduce logging effort on server side.


restli-server/src/main/java/com/linkedin/restli/internal/server/response/ResponseUtils.java
```diff
@@ -148,7 +148,8 @@ public static RestException buildRestException(RestLiResponseException restLiRes
     }
 
     RestResponse restResponse = responseBuilder.build();
-    return new RestException(restResponse, restLiResponseException.getCause());
+    Throwable cause = restLiResponseException.getCause();
+    return new RestException(restResponse, cause==null ? null : cause.toString(), cause, writableStackTrace);
```
nizarm, line 152: Shouldn't the cause.toString() end put printing the entire stack trace? You probably just want to include the exception message for 'message' parameter (cause.getMessage())

restli-server/src/main/java/com/linkedin/restli/internal/server/response/ResponseUtils.java
```diff
@@ -130,7 +130,7 @@ private static RestResponseBuilder encodeResult(String mimeType,
     return builder;
   }
 
-  public static RestException buildRestException(RestLiResponseException restLiResponseException)
+  public static RestException buildRestException(RestLiResponseException restLiResponseException, boolean writableStackTrace)
```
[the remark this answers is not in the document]
xuli-in: Thanks. Updated

restli-server/src/main/java/com/linkedin/restli/internal/server/response/ResponseUtils.java
xuli-in: public Throwable(Throwable cause) {
        fillInStackTrace();
        detailMessage = (cause==null ? null : cause.toString());
        this.cause = cause;
    }

I am thinking of keep the message the same as before and per throwable that is use cause.toString()

xuli-in: For RestLiServiceException, the toString is overritten and not include stacktrace.

nizarm:
Thanks for the change, LGTM

PR #357, opened 2020-07-29
Provide a default symbol table provider implementation

karthikrg:
Provide a default symbol table provider implementation that doesn't use symbol tables for requests/responses of its own, but is able to retrieve remote symbol tables to decode responses from other services


data/src/test/java/com/linkedin/data/codec/symbol/TestDefaultSymbolTableProvider.java
```diff
@@ -0,0 +1,105 @@
...
+    provider.getSymbolTable(_symbolTableName);
+    verify(connection).setRequestProperty(eq("Accept"), eq(ProtobufDataCodec.DEFAULT_HEADER));
+  }
+
+  @Test
+  public void testLocalSymbolTableSuccess()
+  {
+    DefaultSymbolTableProvider provider = new DefaultSymbolTableProvider();
+    // Inject the table into the local cache.
+    provider._cache.put("tableName", _symbolTable);
```
karthikbalasub, line 95: minor: Do you think there will be usecase to inject local symbol table? If so consider adding an API to do that.

data/src/main/java/com/linkedin/data/codec/symbol/DefaultSymbolTableProvider.java
```diff
@@ -0,0 +1,137 @@
...
+
+      // If this is not a remote table, and we didn't find it in the cache, cry foul.
+      if (!isRemote)
+      {
+        throw new IllegalStateException("Unable to fetch symbol table with name: " + symbolTableName);
+      }
+
+      // Ok, we didn't find it in the cache, let's go query the service the table was served from.
+      String url = serverNodeUri + "/" + SYMBOL_TABLE_URI_PATH + "/" + tableName;
+      HttpURLConnection connection = openConnection(url);
```
aman1309, line 102: minor: disconnect finally since this is only one use.

data/src/test/java/com/linkedin/data/codec/symbol/TestDefaultSymbolTableProvider.java
karthikrg: Good idea. Added.

data/src/main/java/com/linkedin/data/codec/symbol/DefaultSymbolTableProvider.java
karthikrg: Good catch. Without this we would have leaked sockets on error responses. Fixed.

PR #358, opened 2020-07-30
Implement support for always projecting fields by name.

karthikbalasub:


data-transform/src/main/java/com/linkedin/data/transform/filter/AbstractFilter.java
```diff
@@ -248,21 +260,33 @@ private Object filterDataMap(DataMap opNode,
     for (Map.Entry<String, Object> entry : valueDataMap.entrySet())
     {
       final String name = entry.getKey();
+      final String nameEscaped = Escaper.replaceAll(name, "$", "$$");
```
BrianPin, line 263: If the name is already escaped(replaced "$" to "$$"), could it happen?

restli-server/src/main/java/com/linkedin/restli/internal/server/util/RestUtils.java
```diff
@@ -251,6 +250,36 @@ public static String pickBestEncoding(String acceptHeader, Set<String> customMim
    */
   public static DataMap projectFields(final DataMap dataMap, final ProjectionMode projectionMode,
       final MaskTree projectionMask)
+  {
```
BrianPin, line 253: Do you think since this adding field is in response, maybe this `projectFields` should be in Response Util class instead of Rest Utils?

data-transform/src/main/java/com/linkedin/data/transform/filter/AbstractFilter.java
karthikbalasub: Field names in the data are never escaped. We only escape them in the mask tree as it has special meaning for $ (eg $*, $start etc)

restli-server/src/main/java/com/linkedin/restli/internal/server/util/RestUtils.java
karthikbalasub: We can move, unfortunately this is a public API and so we cannot delete it. 

If this is not a big concern, I prefer to do this refactoring in a different PR. The current PR is already large.

PR #359, opened 2020-07-30
[darkcluster] Adding identical traffic multiplier strategy

srikrish-19:
This changeset contains a new dark cluster traffic routing strategy called identical traffic multiplier which is meant to serve a usecase where we want to identical requests to be dispatched to all dark clusters configured with this strategy.


darkcluster/src/main/java/com/linkedin/darkcluster/impl/IdenticalTrafficMultiplierDarkClusterStrategy.java
```diff
@@ -0,0 +1,132 @@
...
+import com.linkedin.d2.DarkClusterConfig;
+import com.linkedin.d2.balancer.ServiceUnavailableException;
+import com.linkedin.d2.balancer.util.ClusterInfoProvider;
+import com.linkedin.darkcluster.api.BaseDarkClusterDispatcher;
+import com.linkedin.darkcluster.api.DarkClusterStrategy;
+import com.linkedin.r2.message.RequestContext;
+import com.linkedin.r2.message.rest.RestRequest;
+import java.util.Random;
+
+
```
davidhoa, line 13: javadoc. this should mention that it is similar to RelativeTraffic, etc.

darkcluster/src/main/java/com/linkedin/darkcluster/impl/IdenticalTrafficMultiplierDarkClusterStrategy.java
```diff
@@ -0,0 +1,132 @@
...
+   *
+   * Example 1:
+   * There are 3 dark clusters: A, B and C all of which are configured with same multiplier of 0.1.
+   * There is 1 source instance and 1 dark instance in each cluster.
+   * Assume that the strategy is called for A, B and C in the same order.
+   * For A, there will be no random number persisted in requestContext since we're seeing this request for the first time
+   * So we compute random number, say 0.05 and persist the same in requestContext
+   * Avg#DarkRequests = 1 * 0.1 / 1 = 0.1
+   * since 0.05 < 0.1, request will be sent to A
+   * When it comes to B, the random number is already present and since it is < 0.2, request will be sent to B
```
davidhoa, line 80: why is one "< 0.2" and the other "< 0.3" if they have the same multiplier?

PR #365, opened 2020-08-05
Relax validation for upsert usecase: UPDATE used for create or update.

karthikbalasub:
Fields marked as ReadOnly will be optional for UPDATE methods.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
+
+## [29.4.11] - 2020-08-06
 - Relax validation of read-only fields for upsert usecase: UPDATE used for create or update. Fields marked as ReadOnly will be treated as optional for UPDATE methods.
 
 ## [29.4.10] - 2020-08-05
@@ -4573,7 +4575,8 @@ patch operations can re-use these classes for generating patch messages.
 
 ## [0.14.1]
 
-[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.4.10...master
+[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.4.11...master
+[29.4.11]: https://github.com/linkedin/rest.li/compare/v29.4.10...v29.4.11
 [29.4.10]: https://github.com/linkedin/rest.li/compare/v29.4.9...v29.4.10
 [29.4.9]: https://github.com/linkedin/rest.li/compare/v29.4.8...v29.4.9
 [29.4.8]: https://github.com/linkedin/rest.li/compare/v29.4.7...v29.4.8
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
+- Relax validation of read-only fields for upsert usecase: UPDATE used for create or update. Fields marked as ReadOnly will be treated as optional for UPDATE methods.
 
 ## [29.4.7] - 2020-07-30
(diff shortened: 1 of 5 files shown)


junchuanwang:
I understand this is a relaxation. But does it conflict with 
https://linkedin.github.io/rest.li/Validation-in-Rest_li#validation-for-update-requests
which says
"@ReadOnly and @CreateOnly does not affect input data validation for Rest.li update requests. This is because update is a PUT method and should contain the whole entity, unlike partial update where non-modified fields can be omitted."

restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataValidator.java
```diff
@@ -106,7 +106,7 @@
 
   // ReadOnly fields are treated as optional for these types of requests
   private static final Set<ResourceMethod> readOnlyOptional = new HashSet<>(Arrays.asList(
```
evanw555, line 108: Is this mapping even needed anymore? If you look at the logic path, `#validateInput(RecordTemplate)` already has a switch to check that the resource method is one of `[create, batch_create, update, batch_update]`, so it seems like the "read-only optional" logic is applied for all "input" method types. partial_update and delete have their own separate code paths.

restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataValidator.java
```diff
@@ -247,8 +250,7 @@ public RestLiDataValidator(Map<String, List<String>> annotations,
       for (Map.Entry<String, List<String>> entry : annotations.entrySet())
       {
         String annotationName = entry.getKey();
-        if (annotationName.equals(ReadOnly.class.getAnnotation(RestSpecAnnotation.class).name())
-            && readOnlyRestrictedMethods.contains(resourceMethod))
+        if (annotationName.equals(ReadOnly.class.getAnnotation(RestSpecAnnotation.class).name()))
```
evanw555, line 253: Should you also make the same change for the following if-statement (for createOnly) so that the way these predicates are constructed is consistent?

nickibi: > Should you also make the same change for the following if-statement (for createOnly) so that the way these predicates are constructed is consistent?

+1

restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataValidator.java
karthikbalasub: Ack, fixed it.

restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataValidator.java
karthikbalasub: Done

karthikbalasub:
> I understand this is a relaxation. But does it conflict with
> https://linkedin.github.io/rest.li/Validation-in-Rest_li#validation-for-update-requests
> which says
> "@readonly and @CreateOnly does not affect input data validation for Rest.li update requests. This is because update is a PUT method and should contain the whole entity, unlike partial update where non-modified fields can be omitted."

We would have to update this doc. This doc and the original implementation was not considering the upsert usecase for UPDATE methods.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
+- Relax validation of read-only fields for upsert usecase: UPDATE used for create or update. Fields marked as ReadOnly will be treated as optional for UPDATE methods.
```
nickibi, line 16: I would recommend to add PR# in the changelog, it will be easier for us to look back the history.

PR #366, opened 2020-08-06
Start adding support for multiple input dirs to pegasus (multiple da…

jplaisted:
…ta schema dirs).

This is useful for generating schemas. Our use case is for metadata (DataHub), we want to annotate metadata models with an annotation, and then auto generate event schemas based on those annotations (and then from there avro + java, etc). We want all models (hand written metadata models + auto generated events) in the same module.

Note that there seems to be a lack of tests in this area, I was able to add one but let me know if there are some other kind of tests or procedures I'm missing...

Change details:
- Add a file collection to the `PegasusOptions` object  for the data schema dirs. This defaults to the existing data schema dir the plugin would look for.
- Adjust tasks that used the one data schema dir to accept file collections instead.
  - Tasks that executed some jars may be called one per data schema dir. This could be made more optimal by changing those applications to accept multiple inputs as well, but for now this works.
- Added a test that shows off what I sort of want: have a task that generates a pegasus file, and then wire that up to the rest of the pegasus plugin.

### DIFF
--- a/gradle-plugins/build.gradle
+++ b/gradle-plugins/build.gradle
@@ -46,5 +46,8 @@ integTest {
   dependsOn configurations.dataTemplateForTesting, configurations.pegasusPluginForTesting
   systemProperty 'integTest.dataTemplateCompileDependencies', "'${configurations.dataTemplateForTesting.join("', '")}'"
   systemProperty 'integTest.pegasusPluginDependencies', "'${configurations.pegasusPluginForTesting.join("', '")}'"
+  doFirst {
+    systemProperty 'debug-jvm', debug
+  }
 }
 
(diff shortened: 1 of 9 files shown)


jplaisted:
Also worth noting: I was able to do some hacks before to get some of this generate pdl stuff to work without this change. However, this is MUCH easier now, as you can say what a `ConfigurableFileCollection` is generated by (a set of tasks). Giving that to pegasus in one line is much easier than trying to figure out the several different tasks pegasus creates to wire up dependencies.

e.g. in the test you can see this:

```
    | task generateSomePegasus() {
    |   doFirst {
    |     def out = file('build/generated/pegasus/Foo.pdsc')
    |     out.parentFile.mkdirs()
    |     out.text = '{"name":"Foo","type": "record", "fields": []}'
    |   }
    | }
    |
    |pegasus.main.dataSchemaDirs.from('build/generated/pegasus').builtBy(generateSomePegasus)
```

Without this change that last line becomes much more complicated, since you now have to know each task that is generated by the plugin and how to change their inputs. This change introduces a much easier to use API with a clearer contract.

junchuanwang:
@jplaisted Is this ready for review?

PR #372, opened 2020-08-11
add protobuf content type in request options

aman1309:


src/guide_restli_client.md
```diff
@@ -1012,6 +1012,10 @@ This will set "Content-Type" header value as "application/json".
 
 This will set "Content-Type" header value as "application/x-pson"
 
+#### PROTOBUF2
```
evanw555, line 1015: Are you capitalizing because this is part of an enum? If not, this doesn't need to be in all-caps

aman1309: yup, let's keep it capitalized.

aman1309:
> Do you think it makes sense to include info about general protobuf performance insights or not?

I think that should be separate from guide, may be as blog post or something.

PR #376, opened 2020-08-12
Fix the non-deterministic issue

BrianPin:
Previous fix didn't take care the schema type of Map and Array etc. The non-consistency problem happens in the following scenario:
Suppose there are 3 PDL files, 
Foo1.pdl which has a field of array with type Bar
Foo2.pdl which also has a field of array with type Bar
This 3 schema file will generate Foo1.java Foo2.java and Bar.java and also BarArray.java

It is the annotation in the BarArray.java appeared to have non-consistency in terms appearing to have 2 different generated-by files. Sometimes it is Foo1.pdl, sometimes it is Foo2.pdl.

The solution can be multiple:
1. Just add "pushSchemaLocation" calls to put Bar.pdl location in the BarArray generated class. But this is wrong since the BarArray class should be generated by Foo1 or Foo2

2nd solution is to have the input always sorted. This is way we take little hit on build-time performance, but we gain consistency where once we build, the following many times we can rely on build cache.

### DIFF
--- a/generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
+++ b/generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
@@ -602,7 +602,6 @@ public class TemplateSpecGenerator
 
     final MapTemplateSpec mapClass = (MapTemplateSpec) classInfo.definedClass;
     registerClassTemplateSpec(schema, mapClass);
-
     mapClass.setValueClass(processSchema(valueSchema, enclosingClass, memberName));
     mapClass.setValueDataClass(determineDataClass(valueSchema, enclosingClass, memberName));
 
@@ -706,7 +705,6 @@ public class TemplateSpecGenerator
     fixedClass.setClassName(schema.getName());
     fixedClass.setModifiers(ModifierSpec.PUBLIC);
     registerClassTemplateSpec(schema, fixedClass);
-
     return fixedClass;
   }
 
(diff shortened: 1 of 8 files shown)


generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
```diff
@@ -602,7 +602,6 @@ private MapTemplateSpec generateMap(MapDataSchema schema, ClassTemplateSpec encl
 
     final MapTemplateSpec mapClass = (MapTemplateSpec) classInfo.definedClass;
     registerClassTemplateSpec(schema, mapClass);
-
```
Astro03, line 605: nit: Typically if there are no changes in a file, i would not do any whitespace fixes.

BrianPin:
> LGTM.
> 
> In the ticket, you stated that we missed "array, map, union, enum, etc".
> 
> Are all of these covered now?
> Do we have unit tests to cover all these?

Astro, yes, this fix is the swiss nife to solve all types of schemas. Because it sorts the input.
So all the above mentioned schemas are covered,
I have unit test for map and array, no union and enum, maybe I can add that!

generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
BrianPin: Thanks, I will remove this

PR #378, opened 2020-08-12
Add JMX monitors for RelativeLoadBalancerStrategy

rickzx:
Adding a couple of JMX monitors for relative load balancer strategy, including
1. Standard Deviation, average absolute deviation, median absolute deviation and max absolute deviation of cluster latency.
2. Number of unhealthy hosts
3. Number of hosts in quarantine
4. Total points in the hash ring

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/jmx/RelativeLoadBalancerStrategyJmx.java
+++ b/d2/src/main/java/com/linkedin/d2/jmx/RelativeLoadBalancerStrategyJmx.java
@@ -1,0 +1,168 @@
+package com.linkedin.d2.jmx;
+
+import com.linkedin.d2.balancer.clients.TrackerClient;
+import com.linkedin.d2.balancer.strategies.LoadBalancerQuarantine;
+import com.linkedin.d2.balancer.strategies.relative.RelativeLoadBalancerStrategy;
+import com.linkedin.d2.balancer.strategies.relative.TrackerClientState;
+import com.linkedin.d2.balancer.util.partitions.DefaultPartitionAccessor;
+import com.linkedin.util.degrader.CallTracker;
+import java.net.URI;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+
+public class RelativeLoadBalancerStrategyJmx implements RelativeLoadBalancerStrategyJmxMBean
+{
+  private final RelativeLoadBalancerStrategy _strategy;
+
+  public RelativeLoadBalancerStrategyJmx(RelativeLoadBalancerStrategy strategy)
+  {
+    _strategy = strategy;
+  }
+
+  @Override
+  public double getLatencyStandardDeviation()
+  {
+    Map<TrackerClient, TrackerClientState> stateMap =
+        _strategy.getPartitionState(DefaultPartitionAccessor.DEFAULT_PARTITION_ID).getTrackerClientStateMap();
+
+    List<Double> weightedLatencies = stateMap.keySet().stream()
+        .map(trackerClient -> getWeightedLatency(trackerClient.getLatestCallStats()))
+        .collect(Collectors.toList());
+
+    double avgLatency = weightedLatencies.stream()
+        .mapToDouble(Double::doubleValue)
+        .average()
+        .orElse(0);
+
+    double variance = weightedLatencies.stream()
+        .map(latency -> Math.pow(latency - avgLatency, 2))
+        .mapToDouble(Double::doubleValue)
+        .average()
+        .orElse(0);
+
+    return Math.sqrt(variance);
+  }
+
+  @Override
+  public double getLatencyAverageAbsoluteDeviation()
+  {
+    Map<TrackerClient, TrackerClientState> stateMap =
+        _strategy.getPartitionState(DefaultPartitionAccessor.DEFAULT_PARTITION_ID).getTrackerClientStateMap();
+
+    List<Double> weightedLatencies = stateMap.keySet().stream()
+        .map(trackerClient -> getWeightedLatency(trackerClient.getLatestCallStats()))
+        .collect(Collectors.toList());
+
+    double avgLatency = weightedLatencies.stream()
+        .mapToDouble(Double::doubleValue)
+        .average()
+        .orElse(0);
+
+    return weightedLatencies.stream()
+        .map(latency -> Math.abs(latency - avgLatency))
+        .mapToDouble(Double::doubleValue)
+        .average()
+        .orElse(0);
+  }
+
+  @Override
+  public double getLatencyMedianAbsoluteDeviation()
+  {
+    Map<TrackerClient, TrackerClientState> stateMap =
+        _strategy.getPartitionState(DefaultPartitionAccessor.DEFAULT_PARTITION_ID).getTrackerClientStateMap();
+
+    List<Double> weightedLatencies = stateMap.keySet().stream()
+        .map(trackerClient -> getWeightedLatency(trackerClient.getLatestCallStats()))
+        .collect(Collectors.toList());
+
+    double medianLatency = getMedian(weightedLatencies);
+
+    List<Double> medianAbsolutes = weightedLatencies.stream()
+        .map(latency -> Math.abs(latency - medianLatency))
+        .collect(Collectors.toList());
+
+    return getMedian(medianAbsolutes);
+  }
+
+  @Override
+  public double getLatencyMaxAbsoluteDeviation()
+  {
+    Map<TrackerClient, TrackerClientState> stateMap =
+        _strategy.getPartitionState(DefaultPartitionAccessor.DEFAULT_PARTITION_ID).getTrackerClientStateMap();
+
+    List<Double> weightedLatencies = stateMap.keySet().stream()
+        .map(trackerClient -> getWeightedLatency(trackerClient.getLatestCallStats()))
+        .collect(Collectors.toList());
+
+    double avgLatency = weightedLatencies.stream()
+        .mapToDouble(Double::doubleValue)
+        .average()
+        .orElse(0);
+
+    return weightedLatencies.stream()
+        .map(latency -> Math.abs(latency - avgLatency))
+        .mapToDouble(Double::doubleValue)
+        .max()
+        .orElse(0);
+  }
+
+  @Override
+  public int getUnhealthyHostsCount()
+  {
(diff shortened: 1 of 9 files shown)


d2/src/main/java/com/linkedin/d2/jmx/RelativeLoadBalancerStrategyJmx.java
```diff
@@ -0,0 +1,168 @@
...
+    Map<TrackerClient, TrackerClientState> stateMap =
+        _strategy.getPartitionState(DefaultPartitionAccessor.DEFAULT_PARTITION_ID).getTrackerClientStateMap();
+
+    List<Double> weightedLatencies = stateMap.keySet().stream()
+        .map(trackerClient -> getWeightedLatency(trackerClient.getLatestCallStats()))
+        .collect(Collectors.toList());
+
+    double avgLatency = weightedLatencies.stream()
+        .mapToDouble(Double::doubleValue)
+        .average()
```
rachelhanhan, line 37: Is there a way to get the average of all requests in stead of the average of all hosts? If host A gets 100 requests and host B gets 10 requests, we probably want (100 * latencyA + 10 * latencyB) / 110 instead of (latencyA + latencyB) / 2. That will be consistent to how to calculate the cluster average

rachelhanhan: BTW you can get avgClusterLatency from PartitionStats in PartitionData, we can probably see if we can reuse PartitionStats, because it's also used for metrics emitting purpose

d2/src/main/java/com/linkedin/d2/jmx/RelativeLoadBalancerStrategyJmx.java
```diff
@@ -0,0 +1,168 @@
...
+
+    List<Double> medianAbsolutes = weightedLatencies.stream()
+        .map(latency -> Math.abs(latency - medianLatency))
+        .collect(Collectors.toList());
+
+    return getMedian(medianAbsolutes);
+  }
+
+  @Override
+  public double getLatencyMaxAbsoluteDeviation()
```
rachelhanhan, line 91: On a second thought, I started thinking about for the max deviation, we probably care more about the ones that is greater than the average, if a host has 0 latency, it may impact this max absolute deviation. Should we not use absolute?

d2/src/main/java/com/linkedin/d2/jmx/RelativeLoadBalancerStrategyJmx.java
rachelhanhan, line 37: Is it possible that a host does not get any traffic so the latency is 0, will it be counted here?

rachelhanhan: Can we add some unit test on this to cover some different cases?

rickzx: > BTW you can get avgClusterLatency from PartitionStats in PartitionData, we can probably see if we can reuse PartitionStats, because it's also used for metrics emitting purpose

I've thought of using PartitionStats. But it only keeps the cluster level latency/call count/error rate. If I want to find out the call stats for each individual host, I still have to go through the `TrackerClient` map. So I'm thinking using the call stats from the map directly would produce a more accurate result.

rickzx: > Is it possible that a host does not get any traffic so the latency is 0, will it be counted here?

Since I had the `orElse(0)` clause, it shouldn't be a problem it a host doesn't get any traffic.

rickzx: > Can we add some unit test on this to cover some different cases?

Sure, I've written some local unit tests. Would you like me to also submit those as official code changes?

d2/src/main/java/com/linkedin/d2/jmx/RelativeLoadBalancerStrategyJmx.java
rickzx: Agreed. 

Maybe we could have
1. Standard Deviation and MeanAbsoluteDeviation to capture the entire variance
2. MeanSignedDeviation and MaxSignedDeviation to capture the positive outliers
3. We could also have a metric for (maxLatency - avgLatency) / avgLatency, that directly calculates the relative health factor.

Please feel free to share your thoughts on this.

d2/src/main/java/com/linkedin/d2/jmx/RelativeLoadBalancerStrategyJmx.java
rachelhanhan: OK I will leave this implementation decision to you, as long as we are using the correct way to calculate avg cluster latency :)

rachelhanhan: I mean for the host that does not have any traffic, we probably don't want to count it in the deviation. For example if we have 5 hosts in total, for some reason 4 of them have avg host latency 30, 32, 32, 34 and the other one does not have any traffic, hence the host latency is 0. When we calculate the deviation, I'm thinking about if we should only use 30, 32, 32, 34 as data points.

rachelhanhan: Yes it would be great to always have some unit test coverage

rickzx:
Iterate with Ruxin's review.

A few things:
1. Add unit tests for the new Jmx monitor.
2. Remove some absolute deviation monitors, because we mostly care about the latencies that are above average.
3. Add in aboveAverageLatencyStandardDeviation that calculates the std of hosts in the higher latency range.
4. Add in maxLatencyRelativeFactor and nthPercentileLatencyRelativeFactor, which intuitively gives the max and nth-percentile relative factor. The latter removes host-level outliers.

Please feel free to share your thoughts!

d2/src/main/java/com/linkedin/d2/jmx/RelativeLoadBalancerStrategyJmx.java
rickzx: Fixed in the new commit

d2/src/main/java/com/linkedin/d2/jmx/RelativeLoadBalancerStrategyJmx.java
rickzx: Addressed in the new commit. Please see the comment below.

d2/src/main/java/com/linkedin/d2/jmx/RelativeLoadBalancerStrategyJmx.java
rickzx: Unit tests added in the new commit.

d2/src/main/java/com/linkedin/d2/jmx/RelativeLoadBalancerStrategyJmx.java
```diff
@@ -0,0 +1,189 @@
...
+        .map(trackerClient -> StateUpdater.getAvgHostLatency(trackerClient.getCallTracker().getCallStats()))
+        .mapToLong(Long::longValue)
+        .max()
+        .orElse(0L);
+
+    return avgLatency == 0 ? 0 : maxLatency / avgLatency;
+  }
+
+  @Override
+  public double getNthPercentileLatencyRelativeFactor(double pct)
```
rachelhanhan, line 99: Wondering how do you plan to implement this in container? Do you plan to go with multiple percentile relative factors?

d2/src/main/java/com/linkedin/d2/jmx/RelativeLoadBalancerStrategyJmx.java
```diff
@@ -0,0 +1,189 @@
...
+  public double getLatencyStandardDeviation()
+  {
+    Map<TrackerClient, TrackerClientState> stateMap =
+        _strategy.getPartitionState(DefaultPartitionAccessor.DEFAULT_PARTITION_ID).getTrackerClientStateMap();
+
+    return calculateStandardDeviation(stateMap.keySet());
+  }
+
+  @Override
+  public double getLatencyMeanAbsoluteDeviation()
```
rachelhanhan, line 52: As we discussed offline, I think it's fair to remove mean, since the calculation is a bit more complicated, also it did not provide too much additional value

d2/src/test/java/com/linkedin/d2/jmx/RelativeLoadBalancerStrategyJmxTest.java
```diff
@@ -0,0 +1,133 @@
...
+        Arrays.asList(20, 20), Arrays.asList(10, 10), Arrays.asList(200L, 200L), Arrays.asList(100L, 100L), Arrays.asList(0, 0));
+
+    RelativeLoadBalancerStrategyJmx jmx = mockRelativeLoadBalancerStrategyJmx(trackerClientsEqual);
+    assertEquals(jmx.getLatencyStandardDeviation(), 0.0);
+    assertEquals(jmx.getLatencyMeanAbsoluteDeviation(), 0.0);
+    assertEquals(jmx.getAboveAverageLatencyStandardDeviation(), 0.0);
+
+    List<TrackerClient> trackerClientsDiverse1 = TrackerClientMockHelper.mockTrackerClients(3,
+        Arrays.asList(20, 20, 20), Arrays.asList(10, 10, 10),
+        Arrays.asList(100L, 150L, 200L), Arrays.asList(50L, 75L, 100L), Arrays.asList(0, 0, 0));
```
rachelhanhan, line 67: Let's also have a unit test where the call count and latency for some of the hosts are 0.

rickzx: Added in the latest commit.

d2/src/main/java/com/linkedin/d2/jmx/RelativeLoadBalancerStrategyJmx.java
rickzx: I'm not sure about which percentile to use. I can start by trying out p99, p95. So having a more general interface here might be easier for later experimentations.

PR #379, opened 2020-08-13
add callback based ClusterInfoProvider.getDarkClusterConfigMap

davidhoa:
Add a callback method for ClusterInfoProvider.getDarkClusterConfigMap.

bumping the minor version because I technically made backwards incompatible changes to the Facilities interface (by not provide a default implementation of ClusterInfoProvider) and to ClusterInfoProvider (by adding a new interface method).

Also note that I did not deprecate the old interface method; there's a bug in the jdk that is preventing me from doing that. I suspect that the deprecation suppress works across modules but not within the same module, because I didn't get the warning for the MockClusterInfoProvider, but I did get it for ZKFSLoadBalancer and TogglingLoadBalancer.

Using this callback in DarkClusterStrategyFactoryImpl solves the timeout issue in startup, verified through pulling these changes into an app. I tried various ways to reproduce this in unit tests, but was unable to, even though we know the startup thread was blocked until timeout.

### DIFF
--- a/darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
+++ b/darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
@@ -16,6 +16,7 @@
 
 package com.linkedin.darkcluster.impl;
 
+import com.linkedin.common.callback.Callback;
 import java.util.Map;
 import java.util.Random;
 import java.util.Set;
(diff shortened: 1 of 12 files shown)


darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
```diff
@@ -85,14 +85,11 @@ public DarkClusterStrategyFactoryImpl(@Nonnull Facilities facilities,
...
-    // associated dark clusters. The order matters here: we directly invoke onClusterAdded to
-    // directly query zookeeper for the DarkClusterConfigMap, and then we add the listener.
-    // Doing it in reverse causes a timeout. Ideally, registerClusterListener would have some
-    // functionality similar to SimpleLoadBalancer.listenToCluster that both adds a listener and
-    // requests the data through the eventbus, which will trigger ClusterSubscriber::handlePut and
-    // and iterate through the ClusterListeners.
-    _clusterListener.onClusterAdded(_sourceClusterName);
+    // associated dark clusters. While registering the cluster listener is enough,
+    // we also "warm up" the strategies directly by triggering the clusterListener so that
+    // we retrieve the dark clusters before any inbound request. This is a
```
li-bsoetarman, line 90: Is this comment got cut-off at the end of the line?

davidhoa: oops, will remove that, thanks for catching.

PR #380, opened 2020-08-13
Provide an option in SmoothRateLimiter to not drop tasks if going above

FreCap:
the max buffered. Dropping tasks might be more diruptive to workflows
compared to just not ratelimit.


r2-core/src/main/java/com/linkedin/r2/transport/http/client/SmoothRateLimiter.java
```diff
@@ -107,7 +133,17 @@ public void submit(Callback<None> callback) throws RejectedExecutionException
...
-      throw new RejectedExecutionException("Cannot submit callback because the buffer is full at " + _maxBuffered);
+      if (_bufferOverflowMode == BufferOverflowMode.DROP)
+      {
+        throw new RejectedExecutionException(
+          String.format("PEGA_2000: Cannot submit callback because the buffer is full at %d tasks for ratelimiter: %s", _maxBuffered, _rateLimiterName));
+      }
+      else
+      {
+        _rateLimitedLoggerOverBuffer.error(String.format(
+          "PEGA_2001: the buffer is full at %d tasks for ratelimiter: %s. Executing a request immediately to avoid overflowing and dropping the task.", _maxBuffered,
```
rachelhanhan, line 144: Should this be a warning instead? And "PEGA_xxx" is normally only returned with exception right? Without the exception I don't see why we want to give this code in a rate limited log

r2-core/src/main/java/com/linkedin/r2/transport/http/client/SmoothRateLimiter.java
```diff
@@ -78,19 +97,26 @@ public SmoothRateLimiter(ScheduledExecutorService scheduler, Executor executor,
     _executor = executor;
     _pendingCallbacks = pendingCallbacks;
     _maxBuffered = maxBuffered;
+    _bufferOverflowMode = bufferOverflowMode;
+    _rateLimiterName = rateLimiterName;
```
rachelhanhan, line 101: What is the rateLimiterName used for? Do we expect CapacityAwareClient to pass the name?

r2-core/src/main/java/com/linkedin/r2/transport/http/client/SmoothRateLimiter.java
```diff
@@ -207,6 +250,13 @@ public void loop()
         return;
       }
 
+      // the size of the pending cannot be ever greater than the maxBuffered. We prefer
+      // running above the limit then risking a leak
```
rachelhanhan, line 254: Just wondering if the queue is full, does it mean the enqueue rate and dequeue rate would be exactly the same at that point? Basically if the queue is full, we would not perform any rate limiting? Just want to get some clarification on the expected behavior.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/SmoothRateLimiter.java
FreCap: In theory, this is a "Serious problem" that they should solve, so I made it an error. But I can make this a warning if you prefer

r2-core/src/main/java/com/linkedin/r2/transport/http/client/SmoothRateLimiter.java
FreCap: Yes, it is used for logging purposes. I realized that it is very useful to know it when you throw the exception or log the warning, since otherwise you don't know the queue it was happening.
e.g. you have several services that are using the queue, but only one is the offender. How do you identify it?

It is the same concept as the threadName, without it you're completely lost at debug time

r2-core/src/main/java/com/linkedin/r2/transport/http/client/SmoothRateLimiter.java
FreCap: Exactly!

PR #381, opened 2020-08-14
Put extension schemas into the dataTemplate jar under /extensions path

nickibi:
Before we put extension schemas into the extensionSchema jar.
Now we would like to put the extension schemas into the same jar as base schema but under /extensions path.

### DIFF
--- a/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
+++ b/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
@@ -569,6 +569,8 @@ public class PegasusPlugin implements Plugin<Project>
   // Enable the use of argFiles for the tasks that support them
   private static final String ENABLE_ARG_FILE = "pegasusPlugin.enableArgFile";
 
+  private static final String PEGASUS_PLUGIN_CONFIGURATION = "pegasusPlugin";
+
   @SuppressWarnings("unchecked")
   private Class<? extends Plugin<Project>> _thisPluginType = (Class<? extends Plugin<Project>>)
       getClass().asSubclass(Plugin.class);
@@ -669,7 +671,7 @@ public class PegasusPlugin implements Plugin<Project>
     ConfigurationContainer configurations = project.getConfigurations();
 
     // configuration for getting the required classes to make pegasus call main methods
-    configurations.maybeCreate("pegasusPlugin");
+    configurations.maybeCreate(PEGASUS_PLUGIN_CONFIGURATION);
 
     // configuration for compiling generated data templates
     Configuration dataTemplateCompile = configurations.maybeCreate("dataTemplateCompile");
@@ -729,11 +731,6 @@ public class PegasusPlugin implements Plugin<Project>
     // and validate.
     Configuration schemaAnnotationHandler = configurations.maybeCreate("schemaAnnotationHandler");
 
-    // configuration for parsing and validating extension schemas during build time.
-    //
-    // publish extension schemas into extension schema Jar once the validation passes.
-    Configuration extensionSchema = configurations.maybeCreate("extensionSchema");
-
     // configuration for publishing jars containing rest idl and generated client builders
     // to the project artifacts for including in the ivy.xml
     //
@@ -761,10 +758,10 @@ public class PegasusPlugin implements Plugin<Project>
 
       String version = properties.getProperty("pegasus.version");
 
-      project.getDependencies().add("pegasusPlugin", "com.linkedin.pegasus:data:" + version);
-      project.getDependencies().add("pegasusPlugin", "com.linkedin.pegasus:data-avro-generator:" + version);
-      project.getDependencies().add("pegasusPlugin", "com.linkedin.pegasus:generator:" + version);
-      project.getDependencies().add("pegasusPlugin", "com.linkedin.pegasus:restli-tools:" + version);
+      project.getDependencies().add(PEGASUS_PLUGIN_CONFIGURATION, "com.linkedin.pegasus:data:" + version);
+      project.getDependencies().add(PEGASUS_PLUGIN_CONFIGURATION, "com.linkedin.pegasus:data-avro-generator:" + version);
+      project.getDependencies().add(PEGASUS_PLUGIN_CONFIGURATION, "com.linkedin.pegasus:generator:" + version);
+      project.getDependencies().add(PEGASUS_PLUGIN_CONFIGURATION, "com.linkedin.pegasus:restli-tools:" + version);
     }
     else
     {
@@ -773,8 +770,8 @@ public class PegasusPlugin implements Plugin<Project>
           + " are available on the configuration pegasusPlugin",
           project.getPath());
     }
-    project.getDependencies().add("pegasusPlugin", "org.slf4j:slf4j-simple:1.7.2");
-    project.getDependencies().add("pegasusPlugin", project.files(System.getProperty("java.home") + "/../lib/tools.jar"));
+    project.getDependencies().add(PEGASUS_PLUGIN_CONFIGURATION, "org.slf4j:slf4j-simple:1.7.2");
+    project.getDependencies().add(PEGASUS_PLUGIN_CONFIGURATION, project.files(System.getProperty("java.home") + "/../lib/tools.jar"));
 
     // this call has to be here because:
     // 1) artifact cannot be published once projects has been evaluated, so we need to first
@@ -826,8 +823,6 @@ public class PegasusPlugin implements Plugin<Project>
         configureSchemaAnnotationValidation(project, sourceSet, generateDataTemplateTask);
       }
 
-      configureExtensionSchemaValidationAndPublishTasks(project, sourceSet);
-
       Task cleanGeneratedDirTask = project.task(sourceSet.getTaskName("clean", "GeneratedDir"));
       cleanGeneratedDirTask.doLast(new CacheableAction<>(task ->
       {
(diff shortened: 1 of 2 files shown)


gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1693,6 +1646,34 @@ protected GenerateDataTemplateTask configureDataTemplateGeneration(Project proje
     prepareLegacySchemasForPublishTask.dependsOn(destroyStaleFiles);
     dataTemplateJarDepends.add(prepareLegacySchemasForPublishTask);
 
+    // extension schema directory
+    File extensionSchemaDir = project.file(getExtensionSchemaPath(project, sourceSet));
+
+    if (!SharedFileUtils.getSuffixedFiles(project, extensionSchemaDir, PDL_FILE_SUFFIX).isEmpty()) {
+      // Validate extension schemas if extension schemas are provided.
+      ValidateExtensionSchemaTask validateExtensionSchemaTask = project.getTasks()
+          .create(sourceSet.getTaskName("validate", "ExtensionSchema"), ValidateExtensionSchemaTask.class, task -> {
```
evanw555, line 1655: It would make more sense if it was called `validateExtensionSchemas` IMO

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1693,6 +1646,34 @@ protected GenerateDataTemplateTask configureDataTemplateGeneration(Project proje
...
+            task.setClassPath(project.getConfigurations().getByName("pegasusPlugin"));
+            if (isPropertyTrue(project, ENABLE_ARG_FILE)) {
+              task.setEnableArgFile(true);
+            }
+          });
+
+      Task prepareExtensionSchemasForPublishTask = project.getTasks()
+          .create(sourceSet.getName() + "CopyExtensionSchemas", Sync.class, task ->
+          {
+            task.from(extensionSchemaDir, syncSpec -> syncSpec.include("**/*" + PDL_FILE_SUFFIX));
```
evanw555, line 1668: So we're only supporting extensions as PDLs from the get-go? Good 😃

nickibi: Yes, this feature is only supported for PDL.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
nickibi: fixed

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1693,6 +1646,34 @@ protected GenerateDataTemplateTask configureDataTemplateGeneration(Project proje
...
+    File extensionSchemaDir = project.file(getExtensionSchemaPath(project, sourceSet));
+
+    if (!SharedFileUtils.getSuffixedFiles(project, extensionSchemaDir, PDL_FILE_SUFFIX).isEmpty()) {
+      // Validate extension schemas if extension schemas are provided.
+      ValidateExtensionSchemaTask validateExtensionSchemaTask = project.getTasks()
+          .create(sourceSet.getTaskName("validate", "ExtensionSchemas"), ValidateExtensionSchemaTask.class, task -> {
+            task.setInputDir(extensionSchemaDir);
+            task.setResolverPath(
+                getDataModelConfig(project, sourceSet).plus(project.files(getDataSchemaPath(project, sourceSet))));
+            task.setClassPath(project.getConfigurations().getByName("pegasusPlugin"));
```
aman1309, line 1659: nit: move "pegasusPlugin" to constants

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1693,6 +1646,34 @@ protected GenerateDataTemplateTask configureDataTemplateGeneration(Project proje
     prepareLegacySchemasForPublishTask.dependsOn(destroyStaleFiles);
     dataTemplateJarDepends.add(prepareLegacySchemasForPublishTask);
 
+    // extension schema directory
+    File extensionSchemaDir = project.file(getExtensionSchemaPath(project, sourceSet));
+
+    if (!SharedFileUtils.getSuffixedFiles(project, extensionSchemaDir, PDL_FILE_SUFFIX).isEmpty()) {
```
BrianPin, line 1652: I think in rest.li the left curly brace is in next line

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1693,6 +1646,34 @@ protected GenerateDataTemplateTask configureDataTemplateGeneration(Project proje
...
+            task.setInputDir(extensionSchemaDir);
+            task.setResolverPath(
+                getDataModelConfig(project, sourceSet).plus(project.files(getDataSchemaPath(project, sourceSet))));
+            task.setClassPath(project.getConfigurations().getByName("pegasusPlugin"));
+            if (isPropertyTrue(project, ENABLE_ARG_FILE)) {
+              task.setEnableArgFile(true);
+            }
+          });
+
+      Task prepareExtensionSchemasForPublishTask = project.getTasks()
```
BrianPin, line 1665: Question, what is the previous task of task: `prepareExtensionSchemasForPublishTask` and what is the next task?
I feel `prepare` should be part of some task, not itself a task. But let me know if I miss understand something.

BrianPin: I see `validateExtensionSchemaTask` and `copyPdscSchemasTask` is the previous task of this one.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
nickibi: Good catch.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
nickibi: Used constants instead.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
nickibi: Fixed

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
nickibi: Discussed offline, not a question any more.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1693,6 +1648,37 @@ protected GenerateDataTemplateTask configureDataTemplateGeneration(Project proje
     prepareLegacySchemasForPublishTask.dependsOn(destroyStaleFiles);
     dataTemplateJarDepends.add(prepareLegacySchemasForPublishTask);
 
+    // extension schema directory
+    File extensionSchemaDir = project.file(getExtensionSchemaPath(project, sourceSet));
+
+    if (!SharedFileUtils.getSuffixedFiles(project, extensionSchemaDir, PDL_FILE_SUFFIX).isEmpty())
```
BrianPin, line 1654: There is a hardcoded "extensions"  literal in the function `getExtensionSchemaPath`, I feel it will be left alone if we somehow decide to refactor all the extension schema path to something else. And there is probably somewhere else use the same purpose identifier "extensions" , if we can create or find existing identifier that represent "extensions" path then it will be ideal.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1710,6 +1696,11 @@ protected GenerateDataTemplateTask configureDataTemplateGeneration(Project proje
               copySpec.eachFile(fileCopyDetails ->
                   fileCopyDetails.setPath(TRANSLATED_SCHEMAS_DIR + File.separatorChar + fileCopyDetails.getPath())));
 
+          // Copy all extension schemas as-is into the root extensions directory in the JAR
+          task.from(publishableExtensionSchemasBuildDir, copySpec ->
```
BrianPin, line 1700: in gradle, copy and sync task are alike except sync will remove existing files in the destination dir.
I think maybe in this we should use sync task instead of copy to remove existing data template to keep the destination dir clean?

nickibi: I thought about removing schemas from destination dir. However, I think for debugging purpose, we should keep the schema in the build dir. So that we clearly know what has been packed in the data template jar.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
nickibi: For the first question, are you talking about we should make "extensions" as a constant？I could update it to use constant instead. "extensions" directory is like "pegasus" directory, we defined how the schema directory will be structured. Also we provide the config to customers, if they prefer to use other directory for extension schemas. I think what you said should not be a problem here.

nickibi: I checked PegasusPlugin, we defined all the "getDirectoryPath" methods in this way, e.g pegasus, idl, snapshot etc. For coding style consistent purpose, I would prefer to leave it as-is. We may need to clean the code to use constants all over the places, that should be in a separate PR.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -669,7 +671,7 @@ public void apply(Project project)
     ConfigurationContainer configurations = project.getConfigurations();
 
     // configuration for getting the required classes to make pegasus call main methods
```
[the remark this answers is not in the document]
nickibi: We agree with you the comment is not that intuitive. But it is out of scope of this change. We'd better address it in another PR.

PR #382, opened 2020-08-14
Allow publishing release candidate versions

evanw555:
This change will allow the maintainers of Rest.li to publish release
candidate versions (e.g. `1.2.3-rc.1`) from non-master branches in order
to test experimental features in a safe way as dictated by industry
convention as per the [SemVer](https://semver.org/) spec. Also cleans up the changelog a bit.

Notes:
- Changes to `scripts/release` add support and assertions for the local CLI.
- Changes to `scripts/travis/publish-tag.sh` add support on the Travis CI side.
- The change to `.travis.yml` enables deployment (i.e. bintray publishing) for RC tags.

### DIFF
--- a/scripts/help-text/release.txt
+++ b/scripts/help-text/release.txt
@@ -1,9 +1,13 @@
 Usage: ./scripts/release [OPTION]... [TARGET_COMMIT]
-Releases a new version of Rest.li by creating and pushing a tag at TARGET_COMMIT (defaults to HEAD).
-This script must be run from the root project directory.
+Releases a new version of Rest.li by creating and pushing a tag at TARGET_COMMIT (defaults to HEAD). This script must be
+run from the root project directory. TARGET_COMMIT must be an ancestor of master, unless the version being released is a
+release candidate version.
+
+Please note that the version used to create the tag will be the project version defined at HEAD, not at TARGET_COMMIT,
+though this should be fixed in the future.
 
 Options:
-  -h, --help                    print this help text and exit
+  -h, --help                    print this help text and exit.
 
 Examples:
   ./scripts/release             create and push a release tag at HEAD
--- a/scripts/release
+++ b/scripts/release
@@ -17,7 +17,7 @@ for ARG in "$@"; do
     cat ./scripts/help-text/release.txt
     exit 0
   else
-    TARGET="$1"
+    TARGET="$ARG"
   fi
 done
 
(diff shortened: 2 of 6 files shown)


scripts/help-text/release.txt
```diff
@@ -1,9 +1,13 @@
 Usage: ./scripts/release [OPTION]... [TARGET_COMMIT]
-Releases a new version of Rest.li by creating and pushing a tag at TARGET_COMMIT (defaults to HEAD).
-This script must be run from the root project directory.
+Releases a new version of Rest.li by creating and pushing a tag at TARGET_COMMIT (defaults to HEAD). This script must be
+run from the root project directory. TARGET_COMMIT must be an ancestor of master, unless the version being released is a
+release candidate version.
```
BrianPin, line 4: Does it mean if TARGET is at non-master branch, then it can be released as a rc?

scripts/release
```diff
@@ -45,9 +43,64 @@ if [ -z "$VERSION" ]; then
...
+  fi
+  # Assert that there exists no available lower RC number (e.g. don't allow 1.2.3-rc.2 if 1.2.3-rc.1 is available)
+  RC_NUMBER=1
+  while true; do
+    RC_VERSION="$SUFFIXLESS_VERSION-rc.$RC_NUMBER"
+    # Break once we've reached the current RC number
+    if [ "$VERSION" = "$RC_VERSION" ]; then break; fi
+    # If this rc number is available,
+    if [ -z "$(git tag --list "v$RC_VERSION")" ]; then
+      echo "Cannot create release candidate $VERSION, as $RC_VERSION doesn't exist."
```
BrianPin, line 93: Just curious, when we release a rc, the prefix version should be in the future or in the past?
I think the common should be:
say we are now at v0.1 
we are going to v0.2 but have some rc,
so we have
v0.2-rc1
v0.2-rc2
then when everything works,
v0.2 is the released.

We shall have more rc\d  and probably the prefix version string should never happened?

I verified other rc conventions by looking at linux
https://github.com/torvalds/linux/releases

scripts/help-text/release.txt
evanw555: It means:
- Normal releases must be an ancestor of master.
- Candidate releases can be on any branch.

scripts/release
evanw555: I agree with your assessment, that's how it should be. The release script is designed with this pattern in mind, and it fails in the following cases:
- Trying to release `0.1-rc.1` after `0.1` has already been released (line `81`)
- Trying to release `0.1-rc.2` before `0.1-rc.1` has been released (applies for any combination of numbers; line `93`)

The logic you've highlighted is the latter.

PR #383, opened 2020-08-14
Treat ReadOnly fields as optional in PU patches

evanw555:
This fix treats `ReadOnly` required fields as optional when validating
`(BATCH)PARTIAL_UPDATE` patch inputs. Without this fix, users were unable
to perform patch set operations on record fields where the record
contains a `ReadOnly` field. Were the field included, it would fail
`ReadOnly` validation; were it omitted, it would fail since it's required.
With this fix, users can omit the field from the patch input.

**NOTE:** This PR is incomplete, as I still need to add integration tests.

### DIFF
--- a/restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataValidator.java
+++ b/restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataValidator.java
@@ -464,15 +464,20 @@ public class RestLiDataValidator
    * Checks that if the patch is applied to a valid entity, the modified entity will also be valid.
    * This method
    * (1) Checks that required/ReadOnly/CreateOnly fields are not deleted.
-   * (2) Checks that new values for record templates contain all required fields.
+   * (2) Checks that new values for record templates contain all required fields (treating ReadOnly fields as optional).
    * (3) Applies the patch to an empty entity and validates the entity for custom validation rules
    * and Rest.li annotations (Allows required fields to be absent by using {@link RequiredMode#IGNORE},
    * because a patch does not necessarily contain all fields).
    *
-   * NOTE: Updating a part of an array is not supported. So if the array contains a required field that is
-   * readonly or createonly, the field cannot be present (no partial updates on readonly/createonly)
-   * but cannot be absent either (no missing required fields). This means the array cannot be changed by a
-   * partial update request. This is something that should be fixed.
+   * NOTE: There are two remaining support gaps with this method:
+   * (1) Updating a part of an array is not supported if the object contains a descendant required CreateOnly field.
+   * Because of the current lack of support for patching array elements, setting an array field is always treated as
+   * setting the entire array with all-new elements. Since this is effectively a "Create", we should support setting
+   * CreateOnly fields in array patch operations. TODO: allow setting CreateOnly fields in array elements
+   * (2) Similar to the above problem, if an array element contains a ReadOnly field, then the user currently has no way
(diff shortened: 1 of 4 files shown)


evanw555:
@aman1309 Added a test case in the existing PU/BPU validation tests

restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataValidator.java
```diff
@@ -464,15 +464,15 @@ protected ValidationResult validateOutputAgainstSchema(RecordTemplate dataTempla
...
    * because a patch does not necessarily contain all fields).
    *
-   * NOTE: Updating a part of an array is not supported. So if the array contains a required field that is
-   * readonly or createonly, the field cannot be present (no partial updates on readonly/createonly)
-   * but cannot be absent either (no missing required fields). This means the array cannot be changed by a
-   * partial update request. This is something that should be fixed.
+   * NOTE: Updating a part of an array or other complex object is not supported if the object contains a required field
+   * that is CreateOnly. In this case, the field cannot be present (no partial updates on CreateOnly fields), but cannot
+   * be absent either (no missing required fields).
+   * TODO: This is something that should be fixed.
```
karthikbalasub, line 475: I think what the original comment was trying to say is: "Partial update is not supported on arrays, as we always replace the entire array". 
So in the patch, the entire array record has to be present (full update).

This is not true for other complex types or Maps.  So the original comment is correct and this issue is specific to arrays.
So in partial update, the required createOnly fields must be present (it must be treated as update, so for existing items they should've been present in original and for new items they behave like create) and readonly fields should be treated as optional (can be present for items being updated and can be absent for new array items)

For Maps/other complex types, patch is supported, so if the operation is $set, then they should be treated as CREATE (ie, CreateOnly fields are required, ReadOnly fields should not be present).

restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataValidator.java
```diff
@@ -534,7 +540,9 @@ private ValidationResult checkNewRecordsAreNotMissingFields(RecordTemplate entit
         // Replace $set with the field name to get the full path
         path[path.length - 1] = message.getFormat();
         DataElement element = DataElementUtil.element(new SimpleDataElement(entity.data(), entity.schema()), path);
-        ValidationResult result = ValidateDataAgainstSchema.validate(element, new ValidationOptions());
+        ValidationOptions validationOptions = new ValidationOptions();
+        validationOptions.setTreatOptional(_readOnlyOptionalPredicate);
```
karthikbalasub, line 544: This should be done only for $set on array fields, see my comment above.

evanw555: I understand why this would apply to arrays, but I don't think this applies to just arrays. If a field contains an inner record, then setting that field is basically the same as updating the entire record and all its descendants as new records. This logic applies to both arrays and non-arrays. Otherwise, setting a field as an entire record is impossible if that record contains ReadOnly fields.

Also, the user who originally reported this bug is experiencing it for a union-of-a-record type, not an array.

karthikbalasub: Treating them as optional is fine as they won't be present. 

What I meant is that we should also check that ReadOnly fields are **not present** for non array types. ie, Run the readOnlyRestricted predicate. (we cannot do that for arrays due to the reason in my above comment).

PR #387, opened 2020-08-17
Release 29.6.0-rc.1 as the release candidate

rachelhanhan:
SI team has been working on the relative load balancer feature in d2_relative_strategy branch for a while. Now we plan to release all the commits from d2_relative_strategy as a new version with rc suffix, which indicates the version may not be finalized/stable. We will use this version to test with LinkedIn internal services. Once we verify there is no problem, we will release it as a regular version.


CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.6.0-rc.1] - 2020-08-17
```
evanw555, line 17: This is invalid, since the link `[29.6.0-rc.1]` doesn't have an associated link value at the bottom of the document. Normally, we'd use `./scripts/update-changelog` to update the changelog, but I'm honestly not sure how the RC version should be reflected in the changelog

gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=29.5.2
+version=29.6.0-rc.1
```
evanw555, line 1: Version format looks good

CHANGELOG.md
rachelhanhan: I see! I think we can still reflect it in the changelog for this special version. Once we're ready to generate a regular version for this, we will use master branch, so this won't appear in the master branch.

PR #392, opened 2020-08-20
Adding ability to let users determine if a request should be dispatched to dark clusters

srikrish-19:
@davidhoa PR for adding a DarkDispatcherVerifier


darkcluster/src/main/java/com/linkedin/darkcluster/api/DarkDispatcherVerifier.java
```diff
@@ -0,0 +1,20 @@
...
+
+import com.linkedin.r2.message.RequestContext;
+import com.linkedin.r2.message.rest.RestRequest;
+
+
+/**
+ * Interface that lets users define custom logic to determine if a given request is to be dispatched to dark cluster or not.
+ */
+@FunctionalInterface
+public interface DarkDispatcherVerifier
```
davidhoa, line 11: this name can be confused with the other Verifier ... and you can't tell what's what when configuring config properties. What about DarkCustomGatekeeper? DarkGatekeeper?

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterManagerImpl.java
```diff
@@ -54,19 +54,27 @@
...
+      @Nonnull DarkClusterStrategyFactory strategyFactory, String whiteListRegEx,
+      String blackListRegEx, @Nonnull Notifier notifier, DarkDispatcherVerifier darkDispatcherVerifier) {
     _whiteListRegEx = whiteListRegEx == null ? null : Pattern.compile(whiteListRegEx);
     _blackListRegEx = blackListRegEx == null ? null : Pattern.compile(blackListRegEx);
     _notifier = notifier;
     _facilities = facilities;
     _sourceClusterName = sourceClusterName;
     _darkClusterStrategyFactory = strategyFactory;
     _uriRewriterMap = new HashMap<>();
+    _darkDispatcherVerifier = darkDispatcherVerifier;
```
davidhoa, line 77: lets  create a no op implementation and do the null check here, instead of on every request.

darkcluster/src/main/java/com/linkedin/darkcluster/api/DarkDispatcherVerifier.java
srikrish-19: Will call it DarkGateKeeper

PR #395, opened 2020-08-26
Fix an issue where NullPointerException is raised when excluded hosts hint is not set at retry

rickzx:


d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
```diff
@@ -225,16 +225,23 @@ public void onError(Throwable e)
...
         {
           Set<URI> exclusionSet = ExcludedHostHints.getRequestContextExcludedHosts(_context);
-          int attempts = exclusionSet.size();
-          if (attempts <= _limit)
+          if (exclusionSet == null || exclusionSet.isEmpty())
           {
-            LOG.warn("A retriable exception happens. Going to retry. This is attempt {}. Current exclusion set: ",
-                attempts, ". Current exclusion set: " + exclusionSet);
-            retry = doRetryRequest(_request, _context);
+            LOG.warn("Excluded hosts hint for retry is not set or is empty. This request will fail.");
```
nizarm, line 230: How about "Excluded hosts hint for retry is not set or is empty. This request will not be retried." ?

rickzx: Sure, it's clearer.

PR #396, opened 2020-08-27
Add pdsc support for extension schema resolver

BrianPin:
In the extension schema resolver, since we are multi format resolver, we should be able to read / parse pdsc files as well.

### DIFF
--- a/restli-server/src/main/java/com/linkedin/restli/server/ResourceContext.java
+++ b/restli-server/src/main/java/com/linkedin/restli/server/ResourceContext.java
@@ -236,6 +236,10 @@ public interface ResourceContext extends CustomRequestContext
    * 2. the server is configured to return default value for every request-response pair
    * @return true if the service need to fill in default value when the field of the return value record is absent
    */
-  boolean isFillInDefaultsRequested();
-  void setFillInDefaultValues(boolean fillInDefaultValues);
+  default boolean isFillInDefaultsRequested()
+  {
+    return false;
+  }
+
+  default void setFillInDefaultValues(boolean fillInDefaultValues) {}
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
+
+## [29.5.7] - 2020-08-26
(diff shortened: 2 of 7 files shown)


restli-server/src/main/java/com/linkedin/restli/server/ResourceContext.java
```diff
@@ -236,6 +236,10 @@
...
    * @return true if the service need to fill in default value when the field of the return value record is absent
    */
-  boolean isFillInDefaultsRequested();
-  void setFillInDefaultValues(boolean fillInDefaultValues);
+  default boolean isFillInDefaultsRequested()
+  {
+    return false;
+  }
+
+  default void setFillInDefaultValues(boolean fillInDefaultValues) {}
```
junchuanwang, line 244: why is this no-op function a default method?

BrianPin: This pair of set/get appeared to cause learning-api MP's not override interface method issue.
```
frontend/test/com/linkedin/learning/api/enterprise/mock/MockResourceContext.java:21: error: MockResourceContext is not abstract and does not override abstract method setFillInDefaultValues(boolean) in ResourceContext
public class MockResourceContext implements ResourceContext {
```
So I have this to represent any other similar situation

junchuanwang: But before you made this change, the original method is also not default, how did it work previously for learning-api?

BrianPin: Something probably changed in deco, this is the wc-test I did for default serialization on Monday, https://crt.prod.linkedin.com/#/testing/executions/3150a0af-7b96-44e5-8f9a-3e4e6d352c53/child-executions
and I can't find learning-api in child execution, so I think it might be new

BrianPin: I will speak with offline about it

nickibi: I think this one is unrelated to "Add pdsc support for extension schema resolver", right?

BrianPin: yes

PR #399, opened 2020-08-28
Change R2's HTTP client interface to support other Netty EventLoopGroup in addition to NioEventLoopGroup

wyr123:
Currently, R2's HTTP client only accept Netty's `NioEventLoopGroup` and hardcode it in its API interface. In order to use other types of event loop group like `EpollEventLoopGroup` in the container, we replace the `NioEventLoopGroup` to its parent interface `EventLoopGroup` in the function signature of all public API.

This change is backward compatible because:
* All existing code passing in `NioEventLoopGroup` would still work
* There is no public getter for the `eventLoopGroup` in R2's client-side API so changing its type would not break existing code


wyr123:
Hi @karthikbalasub , could you please help me merge the changes because i don't have permission to do that? Thanks.

PR #400, opened 2020-08-31
Update ExtensionSchemaAnnotation schema

nickibi:
Update ExtensionSchemaAnnotation schema: 
Remove resource field.
Add versionSuffix as an optional field.


nickibi:
> No test updates/additions?

We do not have tests for this ExtensionSchemaAnnotation schema. I could add some tests associated with ExtensionSchemaValidationCmdLineApp class, it will be part of the coming change for ExtensionSchemaValidation.

PR #403, opened 2020-09-02
Update ExtensionSchemaValidationCmdLineApp with more validations

nickibi:
Add following validations for extension schemas
1. The extension schema name has to follow the naming convention - 'baseSchemaName' + "Extensions"
2. The extension schema can only include base schema
3. The extension schema's fields can only be Typeref or array of Typeref
4. The extension schema's field's annotation - versionSuffix value has to match the versionSuffix value in "resourceKey" annotation.

And tests

### DIFF
--- a/restli-tools/src/main/java/com/linkedin/restli/tools/data/ExtensionSchemaValidationCmdLineApp.java
+++ b/restli-tools/src/main/java/com/linkedin/restli/tools/data/ExtensionSchemaValidationCmdLineApp.java
@@ -15,14 +15,19 @@
 */
 package com.linkedin.restli.tools.data;
 
+import com.linkedin.data.DataList;
 import com.linkedin.data.DataMap;
 import com.linkedin.data.schema.ArrayDataSchema;
 import com.linkedin.data.schema.DataSchema;
 import com.linkedin.data.schema.DataSchemaResolver;
 import com.linkedin.data.schema.NamedDataSchema;
 import com.linkedin.data.schema.RecordDataSchema;
+import com.linkedin.data.schema.TyperefDataSchema;
 import com.linkedin.data.schema.grammar.PdlSchemaParser;
+import com.linkedin.data.schema.resolver.ExtensionsDataSchemaResolver;
 import com.linkedin.data.schema.resolver.MultiFormatDataSchemaResolver;
+import com.linkedin.pegasus.generator.DataSchemaParser;
+import com.linkedin.pegasus.generator.FileFormatDataSchemaParser;
 import com.linkedin.restli.internal.tools.RestLiToolsUtils;
 import com.linkedin.data.schema.validation.CoercionMode;
 import com.linkedin.data.schema.validation.RequiredMode;
@@ -51,11 +56,15 @@ import org.slf4j.LoggerFactory;
 
 
 /**
- * This class is used to validate extension schema, the validation covers following parts:
- * 1. The extension schema is a validate schema.
- * 2. The extension schema's fields should only be annotated by "extension" namespace.
- * 3. The extension schema's fields annotations should be {@link ExtensionSchemaAnnotation}.
- * 4. The extension schema's field's schema only be annotated by "resourceKey" namespace.
+ * This class is used to validate extension schemas, the validation covers following parts:
+ * 1. The extension schema is a valid schema.
+ * 2. The extension schema name has to follow the naming convention: <baseSchemaName> + "Extensions"
+ * 3. The extension schema can only include the base schema.
+ * 4. The extension schema's field annotation keys must be in the "extension" namespace
+ * 5. The extension schema's field annotations must conform to {@link ExtensionSchemaAnnotation}.
+ * 6. The extension schema's fields can only be Typeref or array of Typeref
+ * 7. The extension schema's field schema's annotation keys must be in the "resourceKey" namespace.
+ * 8. The extension schema's field annotation versionSuffix value has to match the versionSuffix value in "resourceKey" annotation on the field schema.
  *
  *
  * @author Yingjie Bi
@@ -67,6 +76,8 @@ public class ExtensionSchemaValidationCmdLineApp
   private static final String PDL = "pdl";
   private static final String EXTENSION_ANNOTATION_NAMESPACE= "extension";
   private static final String RESOURCE_KEY_ANNOTATION_NAMESPACE = "resourceKey";
+  private static final String EXTENSIONS_SUFFIX = "Extensions";
+  private static final String VERSION_SUFFIX = "versionSuffix";
 
   static
   {
@@ -77,7 +88,8 @@ public class ExtensionSchemaValidationCmdLineApp
 
   public static void main(String[] args) throws Exception
   {
-    try {
+    try
+    {
       final CommandLineParser parser = new GnuParser();
       CommandLine cl = parser.parse(_options, args);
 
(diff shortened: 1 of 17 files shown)


restli-tools/src/test/extensions/validCase/BarExtensions.pdl
```diff
@@ -0,0 +1,4 @@
+record BarExtensions includes Bar {
+  @extension.using = "finder: test"
+  injectedField: array[DummyKey]
+}
```
BrianPin, line 4: make new lines at each of the new files

restli-tools/src/main/java/com/linkedin/restli/tools/data/ExtensionSchemaValidationCmdLineApp.java
```diff
@@ -112,44 +124,58 @@ public static void main(String[] args) throws Exception
       _logger.error("Invalid arguments: " + e.getMessage());
       System.exit(1);
     }
+    catch (InvalidExtensionSchemaException e)
+    {
+      _logger.error(e.getMessage());
```
BrianPin, line 129: Maybe give some hint in the beginning, just like the previous parseexception, so when see the log, we know clearly this is from extension schema validator

restli-tools/src/main/java/com/linkedin/restli/tools/data/ExtensionSchemaValidationCmdLineApp.java
```diff
@@ -179,48 +205,101 @@ private static void checkExtensionSchemaFields(List<RecordDataSchema.Field> exte
...
-        System.exit(1);
+       throw new InvalidExtensionSchemaException("Field schema: " + fieldSchema.toString() + " is not annotated with 'resourceKey'");
     }
   }
 
-  private static boolean arrayAndItemIsAnnotatedWithResourceKey(DataSchema schema)
+  private static void checkExtensionSchemaFieldSchemaType(DataSchema fieldSchema)
+      throws InvalidExtensionSchemaException
+  {
+    if (fieldSchema.getType() == DataSchema.Type.ARRAY) {
```
BrianPin, line 243: new line

restli-tools/src/test/extensions/invalidExtensionAnnotation/FooExtensions.pdl
```diff
@@ -0,0 +1,4 @@
+record FooExtensions includes Foo {
```
BrianPin, line 1: If we can add a comment say this is valid or not valid, then it is better, 
or we can use the name of the extension schema directly to tell it is valid or not valid.

BrianPin: I see you put all invalid in one directory, so for my prev comment it is fine,
but if you can add some reason in the comment say why it is not valid. Then it is perfect.

restli-tools/src/test/java/com/linkedin/restli/tools/data/TestExtensionSchemaValidationCmdLineApp.java
```diff
@@ -0,0 +1,98 @@
...
+package com.linkedin.restli.tools.data;
+
+import java.io.File;
+
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+
+public class TestExtensionSchemaValidationCmdLineApp {
```
BrianPin, line 25: new line

restli-tools/src/test/pegasus/Bar.pdl
```diff
@@ -0,0 +1,3 @@
+record Bar {
```
BrianPin, line 1: What is the purpose of the following test schema doing?
It is neither in invalid or valid directory.

restli-tools/src/main/java/com/linkedin/restli/tools/data/ExtensionSchemaValidationCmdLineApp.java
```diff
@@ -53,9 +58,13 @@
 /**
  * This class is used to validate extension schema, the validation covers following parts:
```
evanw555, line 59: Rephrasing suggestions for clarity:
1. I don't understand what this means. What is a `validate schema`? Is this supposed to say `valid schema`?
2. Makes sense. (nit: replace hyphen `-` with colon `:`)
3. I would use a definite article like `can only include the base schema` to imply that there is only one base schema for an extension.
4. What does it mean to be annotation by a namespace? Does this mean to say all field annotation keys must be in the `extension` namespace? (nit: be careful of the distinction between "should" and "must")
5. It would be more accurate to say that the annotations "must conform to `{@link ExtensionSchemaAnnotation}`" (nit: be careful of the distinction between "should" and "must")
6. Makes sense.
7. Similar comment as in bullet 4.
8. Clarify which `resourceKey` annotation. Corresponding to what? (nit: remove the hyphen `-`)

Also, in the text at the top, nit: `to validate extension schema` -> `to validate extension schemas`

restli-tools/src/test/pegasus/Bar.pdl
nickibi: This is base schema which is included in the extension schema. I will add comment doc.

restli-tools/src/main/java/com/linkedin/restli/tools/data/ExtensionSchemaValidationCmdLineApp.java
nickibi: Fixed

restli-tools/src/main/java/com/linkedin/restli/tools/data/ExtensionSchemaValidationCmdLineApp.java
nickibi: Fixed

restli-tools/src/main/java/com/linkedin/restli/tools/data/ExtensionSchemaValidationCmdLineApp.java
nickibi: Fixed

restli-tools/src/test/extensions/invalidExtensionAnnotation/FooExtensions.pdl
nickibi: Added comment docs with more details.

restli-tools/src/test/extensions/validCase/BarExtensions.pdl
nickibi: Fixed

restli-tools/src/test/java/com/linkedin/restli/tools/data/TestExtensionSchemaValidationCmdLineApp.java
nickibi: Fixed

BrianPin (approved):

restli-tools/src/main/java/com/linkedin/restli/tools/data/ExtensionSchemaValidationCmdLineApp.java
```diff
@@ -112,44 +124,58 @@ public static void main(String[] args) throws Exception
...
-      }
+ static void parseAndValidateExtensionSchemas(String resolverPath, File inputDir)
+     throws IOException, InvalidExtensionSchemaException
+ {
+   // Parse each extension schema and validate it
+   Iterator<File> iterator = FileUtils.iterateFiles(inputDir, new String[]{PDL}, true);
+   while (iterator.hasNext())
+   {
+     File inputFile = iterator.next();
+     DataSchemaResolver resolver = MultiFormatDataSchemaResolver.withBuiltinFormats(resolverPath);
```
karthikbalasub, line 142: move this outside the loop

restli-tools/src/main/java/com/linkedin/restli/tools/data/ExtensionSchemaValidationCmdLineApp.java
```diff
@@ -179,48 +205,102 @@ private static void checkExtensionSchemaFields(List<RecordDataSchema.Field> exte
...
+  private static void checkExtensionSchemaFieldSchema(DataSchema fieldSchema, Map<String, Object> extensionAnnotations)
+      throws InvalidExtensionSchemaException
   {
-    if (!isAnnotatedWithResourceKey(fieldSchema) && !arrayAndItemIsAnnotatedWithResourceKey(fieldSchema))
+    checkExtensionSchemaFieldSchemaType(fieldSchema);
+    if (!isAnnotatedWithResourceKey(fieldSchema, extensionAnnotations) && !arrayAndItemIsAnnotatedWithResourceKey(fieldSchema, extensionAnnotations))
     {
-        _logger.error("Field schema: [{}] is not annotated with 'resourceKey'", fieldSchema.toString());
-        System.exit(1);
+       throw new InvalidExtensionSchemaException("Field schema: " + fieldSchema.toString() + " is not annotated with 'resourceKey'");
```
karthikbalasub, line 236: This error message might be misleading for Arrays a it would say the Array schema is missing annotation, whereas the correct message is for the array item's type. 

I think you can inline the logic in checkExtensionSchemaFieldSchemaType here, and then just use the modified fieldSchema to call isAnnotatedWithResourceKey

You can delete arrayAndItemIsAnnotatedWithResourceKey

you should also consider in-lining isAnnotatedWithResourceKey logic. It is odd that at some places you are throwing exception and some methods return boolean.

restli-tools/src/main/java/com/linkedin/restli/tools/data/ExtensionSchemaValidationCmdLineApp.java
nickibi: Fixed

restli-tools/src/main/java/com/linkedin/restli/tools/data/ExtensionSchemaValidationCmdLineApp.java
nickibi: Updated the code based on the suggestion.

restli-tools/src/main/java/com/linkedin/restli/tools/data/ExtensionSchemaValidationCmdLineApp.java
```diff
@@ -179,48 +205,91 @@ private static void checkExtensionSchemaFields(List<RecordDataSchema.Field> exte
...
   }
 
-  private static void checkExtensionSchemaFieldSchema(DataSchema fieldSchema)
+  private static void checkExtensionSchemaFieldSchema(DataSchema fieldSchema, Map<String, Object> extensionAnnotations)
+      throws InvalidExtensionSchemaException
   {
-    if (!isAnnotatedWithResourceKey(fieldSchema) && !arrayAndItemIsAnnotatedWithResourceKey(fieldSchema))
+
+    checkExtensionSchemaFieldSchemaType(fieldSchema, extensionAnnotations);
+  }
```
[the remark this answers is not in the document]
nickibi: removed

karthikbalasub (approved):

PR #405, opened 2020-09-03
Fix a bug when check with entry prefix

BrianPin:
When we are parsing schemas from a jar, the entry from both "pegasus" and "extension" should be considered. But in the code when we check if it is "pegasus" , I use
```
entry.getName().startsWith(_schemaResolver.getSchemasDirectoryName().getName()) || isExtensionEntry(entry)
```
where the resolver will return "extension" when getSchemasDirectoryName is called. Which is an error while it should be 
```
entry.getName().startsWith("pegasus") || isExtensionEntry(entry)
```

The semantic is to say: if the entry is either started with "pegasus" or started with "extension", I will parse this entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
-- Fix a bug in FileFormatDataSchemaParser, where jar entry match failed because of case
+- Fix a bug in `FileFormatDataSchemaParser` and remove `isExtensionEntry` method call to simplify the logic.
 
 ## [29.6.3] - 2020-09-03
 - Updated HTTP/2 parent channel idle timeout logging level to info from error 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
+- Fix a bug in FileFormatDataSchemaParser, where jar entry match failed because of case
 
 ## [29.6.3] - 2020-09-03
 - Updated HTTP/2 parent channel idle timeout logging level to info from error 
--- a/generator/src/main/java/com/linkedin/pegasus/generator/FileFormatDataSchemaParser.java
+++ b/generator/src/main/java/com/linkedin/pegasus/generator/FileFormatDataSchemaParser.java
@@ -94,7 +94,7 @@ public class FileFormatDataSchemaParser {
                 final JarEntry entry = entries.nextElement();
                 if (!entry.isDirectory() &&
                     entry.getName().endsWith(_schemaParserFactory.getLanguageExtension()) &&
-                    (entry.getName().toLowerCase().startsWith(_schemaResolver.getSchemasDirectoryName().name().toLowerCase())))
+                    (entry.getName().toLowerCase().startsWith(_schemaResolver.getSchemasDirectoryName().name().toLowerCase() + "/")))
                 {
                   parseJarEntry(jarFile, entry, result);
                   result.getSourceFiles().add(sourceFile);
--- a/generator/src/main/java/com/linkedin/pegasus/generator/FileFormatDataSchemaParser.java
+++ b/generator/src/main/java/com/linkedin/pegasus/generator/FileFormatDataSchemaParser.java
@@ -148,21 +148,6 @@ public class FileFormatDataSchemaParser {
   }
 
   /**
-   * Schema files when archived in a jar can start with pegasus/... or extensions/
-   * For the resolver that is supporting extensions as directory name, we like it to be included in the parsing
-   * @param entry an entry in the jar file, e.g. "pegasus/..." or "extensions/...."
-   * @return true if this entry starts with "extensions"
-   */
-  private boolean isExtensionEntry(JarEntry entry)
-  {
-    if (_schemaResolver.getSchemasDirectoryName() == SchemaDirectoryName.EXTENSIONS)
-    {
-      return entry.getName().startsWith(EXTENSION_PATH_ENTRY);
-    }
-    return false;
-  }
-
-  /**
    * Parse a source that specifies a file (not a fully qualified schema name).
    *
    * @param schemaSourceFile provides the source file.
--- a/generator/src/main/java/com/linkedin/pegasus/generator/FileFormatDataSchemaParser.java
+++ b/generator/src/main/java/com/linkedin/pegasus/generator/FileFormatDataSchemaParser.java
@@ -94,7 +94,7 @@ public class FileFormatDataSchemaParser {
                 final JarEntry entry = entries.nextElement();
                 if (!entry.isDirectory() &&
                     entry.getName().endsWith(_schemaParserFactory.getLanguageExtension()) &&
-                    (entry.getName().toLowerCase().startsWith(_schemaResolver.getSchemasDirectoryName().name().toLowerCase() + "/")))
+                    (entry.getName().startsWith(_schemaResolver.getSchemasDirectoryName().getName() + "/")))
                 {
                   parseJarEntry(jarFile, entry, result);
                   result.getSourceFiles().add(sourceFile);
--- a/generator/src/main/java/com/linkedin/pegasus/generator/FileFormatDataSchemaParser.java
+++ b/generator/src/main/java/com/linkedin/pegasus/generator/FileFormatDataSchemaParser.java
@@ -92,8 +92,9 @@ public class FileFormatDataSchemaParser {
               while (entries.hasMoreElements())
               {
                 final JarEntry entry = entries.nextElement();
-                if (!entry.isDirectory() && entry.getName().endsWith(_schemaParserFactory.getLanguageExtension()) &&
-                    (entry.getName().startsWith(_schemaResolver.getSchemasDirectoryName().getName()) || isExtensionEntry(entry)))
+                if (!entry.isDirectory() &&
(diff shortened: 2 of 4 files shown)


aman1309 (commented):
LGTM. tests?

BrianPin:
> LGTM. tests?

Added for base schema parsing in extensions

generator/src/main/java/com/linkedin/pegasus/generator/FileFormatDataSchemaParser.java
```diff
@@ -92,8 +92,10 @@ public FileFormatDataSchemaParser(String resolverPath, DataSchemaResolver schema
               while (entries.hasMoreElements())
               {
                 final JarEntry entry = entries.nextElement();
-                if (!entry.isDirectory() && entry.getName().endsWith(_schemaParserFactory.getLanguageExtension()) &&
-                    (entry.getName().startsWith(_schemaResolver.getSchemasDirectoryName().getName()) || isExtensionEntry(entry)))
+                if (!entry.isDirectory() &&
+                    entry.getName().endsWith(_schemaParserFactory.getLanguageExtension()) &&
+                    (entry.getName().startsWith(SchemaDirectoryName.PEGASUS.getName()) ||
```
karthikbalasub, line 97: I don't see why "parseSources" should load from both locations.
Based on my understanding: When SchemaDirectoryName == EXTENSION, the ExtensionSchemaResolver will be configured to resolve from both locations.

And parseSource should only parse the extension schemas, right?

junchuanwang:
The commit description better be "what is the status after your fix", for example:
`Fix a bug when check with entry prefix, so that "pegasus" entry is not ignored`

evanw555:
> The commit description better be "what is the status after your fix", for example:
> `Fix a bug when check with entry prefix, so that "pegasus" entry is not ignored`

It should also be more specific. Entry in what? It's not clear at first glance that this change has to do with resolving schemas in _JAR_ entries. Also please update the changelog entry accordingly.

generator/src/main/java/com/linkedin/pegasus/generator/FileFormatDataSchemaParser.java
BrianPin: The reason for doing this change is for users who don't want to create two parsers to call parse sources twice.
The other reason is that when I add the condition in this FileFormatDataSchemaParser, I had the pegasus base entry in mind.
I also feel that extension schema only also make sense.

For the other question that resolve from both location, it is true, and it is in my existing unit test cases.

evanw555, line 97: I'd like to make a comment on the original condition logic itself. It took me a while to understand what this if-statement is checking for because it's unnecessarily convoluted. The original logic basically said:

if `entryName.startsWith(schemasDirectoryName)` or `isExtensionEntry()`
...where `isExtensionEntry()` checks that the schemas directory name is `extensions` and that the entry name starts with `extensions/`

Couldn't this just be reduced to

if `entryName.startsWith(schemasDirectoryName + '/')`? Since if the schema directory name is `extensions`, then it would be checking that it starts with `extensions/` anyway

Ignoring the fact that the name `isExtensionEntry()` is misleading and inaccurate (since it also checks the resolver's schemas directory name value, not just the entry), it doesn't even seem like this method is needed.

Since you're changing this logic, though, my suggestion is that you simplify this if-statement to make it more comprehensible.

BrianPin: After offline discussion, I removed the support of parsing pegasus PDL by ExtensionSchemaResolver

BrianPin: @evanw555 , I modified the logic, please have a look

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
+- Fix a bug in FileFormatDataSchemaParser, where jar entry match failed because of case
```
nickibi, line 16: Update the changelog according to current version.

nickibi (approved):

generator/src/main/java/com/linkedin/pegasus/generator/FileFormatDataSchemaParser.java
```diff
@@ -92,8 +92,9 @@ public FileFormatDataSchemaParser(String resolverPath, DataSchemaResolver schema
               while (entries.hasMoreElements())
               {
                 final JarEntry entry = entries.nextElement();
-                if (!entry.isDirectory() && entry.getName().endsWith(_schemaParserFactory.getLanguageExtension()) &&
-                    (entry.getName().startsWith(_schemaResolver.getSchemasDirectoryName().getName()) || isExtensionEntry(entry)))
+                if (!entry.isDirectory() &&
+                    entry.getName().endsWith(_schemaParserFactory.getLanguageExtension()) &&
+                    (entry.getName().toLowerCase().startsWith(_schemaResolver.getSchemasDirectoryName().name().toLowerCase())))
```
evanw555, line 97: 1. I'm not sure what the original logic (before the `SchemaDirectoryName` change) looked like, but would it make sense to add a `+ "/"` to the check? Otherwise cases like `pegasusFoo/Foo.pdl` would not be filtered out. However, we need to make sure we don't break existing behavior.

2. I'm curious about this bug. Why do we want to allow case-insensitive matches? I'll follow up offline.

BrianPin: For #1 , yeah more or less your idea is great, but the original logic is like this
For #2, it is because we add to jar with lower cases, but the directory name defined Resolver is in upper case. That is why we need case insensitive. Follow up with me offline anytime

junchuanwang: If this is the case, ignoring the case is generally ok.
but why did we add to jar using lower case ?

BrianPin: It is one of the original choice I guess

BrianPin: When a enum is being used, the method `name()` will return its original form which in our case is all upper cases. On the other hand, the method `getName()` will return the actual coded content literal, which in our is all lower cases.
In the beginning of my very first PR of this parser, I was using `name()` so it is not matching the configured entry prefix in jar, but I have the isExtensionEntry to help to rescue.

generator/src/test/java/com/linkedin/pegasus/generator/TestDataSchemaParser.java
```diff
@@ -172,6 +172,54 @@ public void testSchemaFilesInExtensionPathInJar(String[] files, String[] expecte
...
+                    "pegasus/Bar.pdl",
+                    "pegasus/Fuzz.pdsc"
+                },
+                new String[]{}
+            },
+        };
+  }
+
+  @Test(dataProvider = "PegasusSchemasWithExtensionResolver")
+  public void testBaseSchemaFilesInExtensionResolverInJar(String[] files, String[] expected) throws Exception
```
karthikbalasub, line 192: I don't think this test is covering the change made in the code.
Can you add a test case that also has extension schemas in the jar?
And a case where the extension schema includes a schema in /pegasus/ directory.

BrianPin: Karthik, The scenario you described is already in the same test class.  For extension schema project

There were 2 test scenario here
1.  extensions and pegasus in jar
2. extensions and pegasus in path

The one I added here is to guarantee we dont parse pegasus schema in the jar
Which is a conclusion after viewing your comment and discuss with the team

karthikbalasub: In that case, can we merge this testcase with the dataprovider above? No need to keep them separate

BrianPin: Yes, merged the tests, resolving this comment for now.

generator/src/main/java/com/linkedin/pegasus/generator/FileFormatDataSchemaParser.java
```diff
@@ -92,8 +92,9 @@ public FileFormatDataSchemaParser(String resolverPath, DataSchemaResolver schema
               while (entries.hasMoreElements())
               {
                 final JarEntry entry = entries.nextElement();
-                if (!entry.isDirectory() && entry.getName().endsWith(_schemaParserFactory.getLanguageExtension()) &&
-                    (entry.getName().startsWith(_schemaResolver.getSchemasDirectoryName().getName()) || isExtensionEntry(entry)))
+                if (!entry.isDirectory() &&
+                    entry.getName().endsWith(_schemaParserFactory.getLanguageExtension()) &&
+                    (entry.getName().toLowerCase().startsWith(_schemaResolver.getSchemasDirectoryName().name().toLowerCase() + "/")))
```
karthikbalasub, line 97: Just use SchemaDirectorName::getName here. Also I agree that we shouldn't ignore case.
Following check should be enough:

entry.getName().startsWith(_schemaResolver.getSchemasDirectoryName().getName() + "/")

BrianPin: Ok

BrianPin: Addressed

karthikbalasub (approved):

PR #407, opened 2020-09-03
Update RestLiValidatorFilter and RestLiDataValidator to expose creation of restli validators

vshwnth2:
There are three main changes in this PR:

- in RestLiValidatorFilter, exposes the creation of the request and response RestLiDataValidator. This will allow for other use cases that extend the RestLiValidatorFilter to override this creation and have their own implementation

- Within RestLiDataValidator, a new DataSchemaAnnotationValidator each time 'validateOutput' is called. Inside the RestLiDataSchemaDataValidator, now it creates the validator in the constructor and pass that instance along each time.

- Within RestLiDataValidator, exposes the creation of the input entity validator

Existing unit tests pass. No functionality should be changed in the PR.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-version=29.6.2
+version=29.6.3
 group=com.linkedin.pegasus
 org.gradle.configureondemand=true
 org.gradle.parallel=true
--- a/restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataSchemaDataValidator.java
+++ b/restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataSchemaDataValidator.java
@@ -16,8 +16,10 @@
 
 package com.linkedin.restli.common.validation;
 
+import com.linkedin.common.callback.Function;
 import com.linkedin.data.schema.DataSchema;
 import com.linkedin.data.schema.validation.ValidationResult;
+import com.linkedin.data.schema.validator.DataSchemaAnnotationValidator;
 import com.linkedin.data.template.RecordTemplate;
 import com.linkedin.data.transform.filter.request.MaskTree;
 import com.linkedin.restli.common.ResourceMethod;
@@ -32,6 +34,7 @@ import java.util.Collections;
  */
 public class RestLiDataSchemaDataValidator extends RestLiDataValidator {
   private final DataSchema _validatingSchema;
+  private final Function<DataSchema, DataSchemaAnnotationValidator> _validatorInitFunc;
 
   /**
    * Constructor.
@@ -53,6 +56,7 @@ public class RestLiDataSchemaDataValidator extends RestLiDataValidator {
     }
 
     _validatingSchema = validatingSchema;
+    _validatorInitFunc = DataSchemaAnnotationValidator::new;
   }
 
   /**
@@ -63,7 +67,7 @@ public class RestLiDataSchemaDataValidator extends RestLiDataValidator {
   @Override
   public ValidationResult validateOutput(RecordTemplate dataTemplate)
   {
-    return super.validateOutputAgainstSchema(dataTemplate, _validatingSchema);
+    return super.validateOutputAgainstSchema(dataTemplate, _validatingSchema, _validatorInitFunc);
   }
 
   /**
--- a/restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataSchemaDataValidator.java
+++ b/restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataSchemaDataValidator.java
@@ -34,7 +34,6 @@ import java.util.Collections;
  */
 public class RestLiDataSchemaDataValidator extends RestLiDataValidator {
   private final DataSchema _validatingSchema;
-  private final DataValidator _inputDataValidator;
   private final DataSchemaAnnotationValidator _outputSchemaValidator;
 
   /**
@@ -57,7 +56,6 @@ public class RestLiDataSchemaDataValidator extends RestLiDataValidator {
     }
 
     _validatingSchema = validatingSchema;
-    _inputDataValidator = new DataValidator(_validatingSchema);
     _outputSchemaValidator = new DataSchemaAnnotationValidator(_validatingSchema);
   }
 
@@ -87,18 +85,6 @@ public class RestLiDataSchemaDataValidator extends RestLiDataValidator {
     else
     {
       return super.getValidatorForOutput(validatingSchema);
-    }
-  }
-
-  @Override
-  protected Validator getValidatorForInput(DataSchema validatingSchema) {
-    if (_validatingSchema.equals(validatingSchema))
-    {
-      return _inputDataValidator;
-    }
-    else
-    {
-      return super.getValidatorForInput(validatingSchema);
     }
   }
 
--- a/restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataSchemaDataValidator.java
+++ b/restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataSchemaDataValidator.java
@@ -29,6 +29,7 @@ import java.util.Collections;
 
 /**
  * Extension of {@link RestLiDataValidator} to allow for validation against a provided data schema.
+ * This validator is only used for output validation.
  *
  * @author Evan Williams
  */
--- a/restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataSchemaDataValidator.java
+++ b/restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataSchemaDataValidator.java
@@ -16,7 +16,6 @@
 
 package com.linkedin.restli.common.validation;
 
-import com.linkedin.common.callback.Function;
 import com.linkedin.data.schema.DataSchema;
 import com.linkedin.data.schema.validation.ValidationResult;
 import com.linkedin.data.schema.validator.DataSchemaAnnotationValidator;
@@ -34,7 +33,7 @@ import java.util.Collections;
  */
 public class RestLiDataSchemaDataValidator extends RestLiDataValidator {
   private final DataSchema _validatingSchema;
-  private final Function<DataSchema, DataSchemaAnnotationValidator> _validatorInitFunc;
+  private final DataSchemaAnnotationValidator _schemaValidator;
 
   /**
    * Constructor.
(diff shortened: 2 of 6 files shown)


restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataValidator.java
```diff
@@ -436,6 +437,19 @@ public ValidationResult validateOutput(RecordTemplate dataTemplate, MaskTree pro
...
+      throw new IllegalArgumentException("Validating schema is null");
+    }
+
+    return validateOutputAgainstSchema(dataTemplate,
+        () -> ValidateDataAgainstSchema.validate(dataTemplate.data(), validatingSchema, new ValidationOptions(),
+            new DataSchemaAnnotationValidator(validatingSchema)));
+  }
+
+  protected ValidationResult validateOutputAgainstSchema(RecordTemplate dataTemplate,
+      Supplier<ValidationResult> validationResultSupplier)
```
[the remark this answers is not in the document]
vshwnth2: In regards to `protected Validator getValidatorForOutput(DataSchema validatingSchema);`, we would be overriding this in `RestLiDataSchemaDataValidator` right?

In this case, the overridden method would look something like 
```protected Validator getValidatorForOutput(DataSchema validatingSchema) { return _schemaValidator; }```

It felt a little strange to me to do something like this, since we are ignoring the input validatingSchema parameter.

File: restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataSchemaDataValidator.java [hunk 2]
```diff
@@ -32,6 +34,8 @@
  */
 public class RestLiDataSchemaDataValidator extends RestLiDataValidator {
   private final DataSchema _validatingSchema;
+  private final DataValidator _inputDataValidator;
+  private final DataSchemaAnnotationValidator _outputSchemaValidator;
```
karthikbalasub, line 38: nit: this can be DataValidator

restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataSchemaDataValidator.java
```diff
@@ -53,6 +57,8 @@ public RestLiDataSchemaDataValidator(Annotation[] annotations,
     }
 
     _validatingSchema = validatingSchema;
+    _inputDataValidator = new DataValidator(_validatingSchema);
```
karthikbalasub, line 60: This is not going to be useful. The validatingSchema is for output, based on projection.
and for inputs it would be the resource entity schema or a subset based on patch for partial_update.

Instead of initializing in the constructor, you should just use getValidatorFor* methods, but cache them after first invocation? 

Also this validator class is used only for output validation (input uses RestLiDataValidator directly). so you can remove all the input validator logic and remove the override for getValidatorForInput

karthikbalasub: I actually don't see the need for changes in this file. The changes in other two classes should be enough to override the validator construction logic.

vshwnth2: Is the suggestion to sub-class `RestLiDataSchemaDataValidator` and `RestLiDataValidator` ourselves in our own implementation and use them in our sub-class of `RestLiValidator` by overriding `createRequestRestLiDataValidator` and `createResponseRestLiDataValidator`?

For the output use case, is there a reason we need to do it explicitly via a cache? For each `_validatingSchema` instantiation, it should be the same `DataSchemaAnnotationValidator` from my understanding, so it feels that this doesn't have to be done via a cache on an outer layer

karthikbalasub: If you plan to re-use this validator, then we can leave the caching. However, this validator is used only for output validation (maybe the javadoc should be updated to mention this). So you should remove the override for getValidatorForInput.

vshwnth2: Removed the override and updated the javadoc

restli-server/src/main/java/com/linkedin/restli/server/validation/RestLiValidationFilter.java
```diff
@@ -487,4 +467,46 @@ private boolean shouldValidateOnResponse(FilterRequestContext requestContext)
     future.completeExceptionally(t);
     return future;
   }
+
+  /**
+   * Creates a {@link RestLiValidationFilter} to use for validation onRequest.
```
junchuanwang, line 472: I think you meant to say "{@link RestLiDataValidator}", same as line 476, 484, 488

restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataValidator.java
```diff
@@ -608,13 +608,21 @@ private ValidationResult validateInputEntity(RecordTemplate entity)
...
   private ValidationResult validateOutputEntity(RecordTemplate entity, DataSchema validatingSchema)
   {
-    DataSchemaAnnotationValidator validator = new DataSchemaAnnotationValidator(validatingSchema);
-    return ValidateDataAgainstSchema.validate(entity.data(), validatingSchema, new ValidationOptions(), validator);
+    return ValidateDataAgainstSchema.validate(entity.data(), validatingSchema, new ValidationOptions(),
+        getValidatorForOutput(validatingSchema));
+  }
+
+  protected Validator getValidatorForOutput(DataSchema validatingSchema) {
+    return new DataSchemaAnnotationValidator(validatingSchema);
```
junchuanwang, line 621: Sorry can you remind me why this changes from instantiating "DataValidator" to "DataSchemaAnnotationValidator" now?

restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataSchemaDataValidator.java
```diff
@@ -66,6 +71,24 @@ public ValidationResult validateOutput(RecordTemplate dataTemplate)
...
   }
 
+  /**
+   * Validator to use to validate the output.
+   * The validator is instantiated in the constructor, so directly returns that if input is equal to _validatingSchema.
+   * @param validatingSchema schema to validate against
+   * @return validator
+   */
+  @Override
+  protected Validator getValidatorForOutput(DataSchema validatingSchema) {
```
junchuanwang, line 81: nit: I personally think "getValidatorForOutput" an incomplete name, it makes more sense to me to say for example "getValidatorForOutputEntityValidation"

junchuanwang (commented):
I personally think the commit title should remove the "potential performance gain" since the PR made nothing about it. 

Or did I miss anything from reading commit description?

restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataSchemaDataValidator.java
vshwnth2: sounds good, will update

restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataValidator.java
vshwnth2: Both before and after are instantiating a `DataSchemaAnnotationValidator`, it was just abstracted into its own method to allow for override access

restli-server/src/main/java/com/linkedin/restli/server/validation/RestLiValidationFilter.java
vshwnth2: good catch, thanks

vshwnth2: good catch, thanks

restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataValidator.java
vshwnth2: both before and after instantiate `DataSchemaAnnotationValidator`, this just abstracts it out into its own method

restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataSchemaDataValidator.java
vshwnth2: updated

restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataSchemaDataValidator.java
```diff
@@ -66,6 +71,24 @@ public ValidationResult validateOutput(RecordTemplate dataTemplate)
...
 
+  /**
+   * Validator to use to validate the output.
+   * The validator is instantiated in the constructor, so directly returns that if input is equal to _validatingSchema.
+   * @param validatingSchema schema to validate against
+   * @return validator
+   */
+  @Override
+  protected Validator getValidatorForOutputEntityValidation(DataSchema validatingSchema) {
+    if (_validatingSchema == validatingSchema)
```
idarmans, line 82: if the strict pointer equality check is on purpose, please add an inline doc on why it's the case

restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataValidator.java
junchuanwang: you are right. I previously misunderstood

junchuanwang (approved):
Looks good to me.

restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataSchemaDataValidator.java
karthikbalasub: +1 on Ivan's suggestion

karthikbalasub (approved):

BrianPin:
Would you mind make the coding style consistent with existing Restli style?

BrianPin:
Also please update CHANGE.log, if this needs to be used right away, please also advance the version.
Please checkout Pegasus git workflow for details

gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=29.6.2
+version=29.6.3
```
nickibi, line 1: Please rebase master, it is 29.6.4 in the master. And update CHANGELOG accordingly.

PR #409, opened 2020-09-07
Remove guava from data-avro.

tjni:
The only usage of Guava in data-avro is the CaseFormat class, which is
used to convert from lower case camel-case to upper case camel-case. I
don't think it justifies pulling in the entire library (which is 2.1 MB
in version 18 and 2.6 MB in version 28.1 plus around 1.5 MB of source).

The CaseFormat class has some dependencies, including one on CaseMatcher
that isn't easy to remove, so I've copied over that code from the latest
version of Guava for now (manually shading them) and updated a minimum
amount of code to remove other dependencies on Preconditions, Converter,
and annotations.

Note: Please add the backwards-incompatible label.


evanw555:
This is good. There was an effort in the past to remove Guava completely from Rest.li, since supposedly Guava often causes transitive dependency problems. It was left in a couple complex cases, this being one of them, however the intention was to completely remove it. This is a step in the right direction.

As for where to place the shadowed code, that's a little tricky. In the past, we've included shadowed code in separate modules such as `li-jersey-uri` and `li-protobuf`, however adding a whole new module adds complexity. I'm not too sure about this. It would be helpful to get another opinion here.

tjni:
Thanks for reviewing! What you wrote makes sense. I don't mind either way (and don't have a strong opinion either).

data-avro/src/main/java/com/linkedin/data/avro/PegasusUnionToAvroRecordConvertCallback.java
```diff
@@ -1,8 +1,8 @@
 package com.linkedin.data.avro;
 
-import com.google.common.base.CaseFormat;
 import com.linkedin.data.Data;
 import com.linkedin.data.DataMap;
+import com.linkedin.data.avro.util.guava.CaseFormat;
```
karthikbalasub, line 5: The only place we use CaseFormat is to convert from lowerCamelCase to UpperCamelCase for field names which will always be ascii.
So should we avoid copying the utilities and just write a utility fn that replaces first char with uppercase?

tjni:
I updated the code to remove usage of CaseFormat and reused the code from Guava to transform a character from lowercase to uppercase.

karthikbalasub (approved):

evanw555 (approved):
Glad we're able to solve this now without shadowing any Guava source files 😄 

LGTM, we'll bundle this with the next major version release

PR #410, opened 2020-09-08
Loosen validation for array-descendant patch fields

evanw555:
This change allows a `PARTIAL_UPDATE` or `BATCH_PARTIAL_UPDATE` request to
set a `ReadOnly`/`CreateOnly` field if that field is the descendant of an
array. This is needed because there's no real way to "patch" individual
array elements in Rest.li, so array patching actually sets the entire
array. Thus, users need a way to "patch" an array while maintaining the
existing values of `ReadOnly`/`CreateOnly` fields.

**Note:**
- The IDL/snapshot changes are just so I can test the logic for `CreateOnly` by adding the create-only field `bar3` inside the array in the PDL model.
- I renamed `DataValidator` to `RestLiAnnotationDataValidator` because it makes more sense, IMO.

### DIFF
--- a/restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataValidator.java
+++ b/restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataValidator.java
@@ -104,6 +104,10 @@ public class RestLiDataValidator
   private static final Set<ResourceMethod> CREATE_ONLY_RESTRICTED_METHODS = new HashSet<>(Arrays.asList(
       ResourceMethod.PARTIAL_UPDATE, ResourceMethod.BATCH_PARTIAL_UPDATE));
 
+  // ReadOnly and CreateOnly fields descended from an array field can be specified for these types of requests
+  private static final Set<ResourceMethod> ARRAY_DESCENDANT_ACCEPTED_METHODS = new HashSet<>(Arrays.asList(
+      ResourceMethod.PARTIAL_UPDATE, ResourceMethod.BATCH_PARTIAL_UPDATE));
+
   // Resource methods that require validation on response
   public static final Set<ResourceMethod>  METHODS_VALIDATED_ON_RESPONSE = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(
       ResourceMethod.GET, ResourceMethod.CREATE, ResourceMethod.PARTIAL_UPDATE, ResourceMethod.GET_ALL,
@@ -274,6 +278,11 @@ public class RestLiDataValidator
     _validatorClassMap = Collections.unmodifiableMap(validatorClassMap);
   }
 
+  /**
+   * Validates input data and patches using a given resource's {@link ReadOnly} and {@link CreateOnly} annotations.
+   * Since it's an extension of {@link DataSchemaAnnotationValidator}, it also validates the data using whatever custom
+   * validators are defined in the schema.
+   */
   protected class DataValidator extends DataSchemaAnnotationValidator
   {
     protected DataValidator(DataSchema schema)
@@ -286,14 +295,36 @@ public class RestLiDataValidator
     {
       super.validate(context);
       DataElement element = context.dataElement();
-      if (_readOnlyRestrictedPredicate.evaluate(element))
+      if (_readOnlyRestrictedPredicate.evaluate(element) && !grantArrayDescendantException(element))
       {
         context.addResult(new Message(element.path(), "ReadOnly field present in a %s request", _resourceMethod.toString()));
       }
-      if (_createOnlyPredicate.evaluate(element))
+      if (_createOnlyPredicate.evaluate(element) && !grantArrayDescendantException(element))
       {
         context.addResult(new Message(element.path(), "CreateOnly field present in a %s request", _resourceMethod.toString()));
       }
+    }
+
+    /**
+     * Determines if the given data element can be set despite being {@link ReadOnly} or {@link CreateOnly}.
+     * A given field is granted this exception if it's the descendant of an array field and if it's in a patch input
+     * (i.e. partial_update or batch_partial_update).
+     */
+    private boolean grantArrayDescendantException(DataElement element)
+    {
+      if (element == null || !ARRAY_DESCENDANT_ACCEPTED_METHODS.contains(_resourceMethod))
+      {
+        return false;
+      }
+      DataElement currentElement = element.getParent();
+      while (currentElement != null)
+      {
+        if (currentElement.getSchema().getType() == DataSchema.Type.ARRAY) {
+          return true;
+        }
+        currentElement = currentElement.getParent();
+      }
+      return false;
     }
   }
 
@@ -462,22 +493,14 @@ public class RestLiDataValidator
 
   /**
    * Checks that if the patch is applied to a valid entity, the modified entity will also be valid.
-   * This method
+   * This method...
    * (1) Checks that required/ReadOnly/CreateOnly fields are not deleted.
    * (2) Checks that new values for record templates contain all required fields (treating ReadOnly fields as optional).
    * (3) Applies the patch to an empty entity and validates the entity for custom validation rules
-   * and Rest.li annotations (Allows required fields to be absent by using {@link RequiredMode#IGNORE},
-   * because a patch does not necessarily contain all fields).
-   *
-   * NOTE: There are two remaining support gaps with this method:
-   * (1) Updating a part of an array is not supported if the object contains a descendant required CreateOnly field.
-   * Because of the current lack of support for patching array elements, setting an array field is always treated as
-   * setting the entire array with all-new elements. Since this is effectively a "Create", we should support setting
-   * CreateOnly fields in array patch operations. TODO: allow setting CreateOnly fields in array elements
-   * (2) Similar to the above problem, if an array element contains a ReadOnly field, then the user currently has no way
-   * to patch that array element while keeping the server-provided value. The user can omit the field in the patch
-   * request, but since the server has no concept of "patching elements", it can't assume a given element's existing
-   * value. TODO: allow setting ReadOnly fields in array elements
+   *     and Rest.li annotations, allowing the following exceptions:
+   *     - Allows required fields to be absent by using {@link RequiredMode#IGNORE},
+   *       because a patch does not necessarily contain all fields.
+   *     - Allows array-descendant ReadOnly/CreateOnly fields to be set, since there's currently no way to "patch" arrays.
    *
    * @param patchRequest the patch
    * @return the final validation result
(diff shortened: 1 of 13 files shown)


restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataValidator.java
```diff
@@ -274,9 +278,14 @@ else if (annotationName.equals(CreateOnly.class.getAnnotation(RestSpecAnnotation
     _validatorClassMap = Collections.unmodifiableMap(validatorClassMap);
   }
 
-  private class DataValidator extends DataSchemaAnnotationValidator
+  /**
+   * Validates input data and patches using a given resource's {@link ReadOnly} and {@link CreateOnly} annotations.
+   * Since it's an extension of {@link DataSchemaAnnotationValidator}, it also validates the data using whatever custom
+   * validators are defined in the schema.
+   */
+  private class RestLiAnnotationDataValidator extends DataSchemaAnnotationValidator
```
junchuanwang, line 286: I totally agree the naming needs to be enhanced. There are just too many validators IMO.

I was always curious about this naming convention. Before we named validators like "DataSchemaAnnotationValidator" or "DataValidator", now later a few new validators are now with a "RestLi"prefix, does that carry special meaning?

restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataValidator.java
```diff
@@ -286,15 +295,38 @@ public void validate(ValidatorContext context)
     {
       super.validate(context);
       DataElement element = context.dataElement();
-      if (_readOnlyRestrictedPredicate.evaluate(element))
+      System.out.println("Validate: " + element);
```
junchuanwang, line 298: Did you use this for debug? Not recommend to add it here.

restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataValidator.java
```diff
@@ -286,15 +295,38 @@ public void validate(ValidatorContext context)
...
+      if (_readOnlyRestrictedPredicate.evaluate(element) && !grantArrayDescendantException(element))
```
junchuanwang, line 299: why not 
```
if (!grantArrayDescendantException(element)) {
  // Original logic
}
```

BrianPin:
Do you think there will be a similar request for UPDATE method? Since this PR is targeting PARTIAL_UPDATE and BATCH_PARTIAL_UPDATE

restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataValidator.java
evanw555: Since this method traverses the data tree, it can potentially have performance issues. If we did it your way, it would be invoked for each element in the tree. Since my way is only invoked when the predicate is matched, this means the method will only be invoked for `ReadOnly`/`CreateOnly` elements that are mistakenly included (or those which are included as part of an array).

restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataValidator.java
evanw555: Good catch, lol

restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataValidator.java
evanw555: Good question. My reasoning is:
- The `"RestLi"` prefix implies that the context is the Rest.li request/response validation, not just plain Pegasus data.
- `"Annotation"` because the validator triggers the custom schema validators as well as the "Rest.li annotations" (see next bullet point).
- Together, `"RestLiAnnotation"` implies that the validator uses "Rest.li annotations" (i.e. `ReadOnly`/`CreateOnly`). These are annotations that are specific to the Rest.li REST framework, not just the Pegasus data layer.

Naming is often very confusing or ambiguous in Rest.li, so I'm trying to make it more clear.

evanw555:
> Do you think there will be a similar request for UPDATE method? Since this PR is targeting PARTIAL_UPDATE and BATCH_PARTIAL_UPDATE

Nope, since setting these fields when updating the entire entity is already supported. See: https://linkedin.github.io/rest.li/Validation-in-Rest_li#validation-for-update-requests

restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataValidator.java
karthikrg: Is this println by design or by choice? Logging maybe better?

karthikbalasub (approved):

restli-int-test-api/src/main/idl/com.linkedin.restli.examples.greetings.client.autoValidationDemos.restspec.json
```diff
@@ -1,7 +1,7 @@
 {
   "annotations" : {
     "createOnly" : {
-      "value" : [ "stringB", "intB", "UnionFieldWithInlineRecord/com.linkedin.restli.examples.greetings.api.myRecord/foo2", "MapWithTyperefs/*/id" ]
+      "value" : [ "stringB", "intB", "UnionFieldWithInlineRecord/com.linkedin.restli.examples.greetings.api.myRecord/foo2", "MapWithTyperefs/*/id", "ArrayWithInlineRecord/*/bar3" ]
```
BrianPin, line 4: What is the wildcard meaning in here

restli-int-test/src/test/java/com/linkedin/restli/examples/TestRestLiValidation.java
```diff
@@ -557,12 +555,18 @@ public void testPartialUpdateFailure(RestClient restClient, Object builder, Stri
...
             "{\"patch\": {\"$set\": {\"ArrayWithInlineRecord\": [{\"bar2\": \"missing bar1\"}]}}}",
+            "{\"patch\": {\"$set\": {\"UnionFieldWithInlineRecord\": {\"com.linkedin.restli.examples.greetings.api.myRecord\": {}}}}}",
+            "{\"patch\": {\"$set\": {\"validationDemoNext\": {\"stringA\": \"no stringB\"}}}}",
             // Okay to delete a field containing a ReadOnly field
             "{\"patch\": {\"$delete\": [\"ArrayWithInlineRecord\"]}}",
             // Okay to delete a field containing a CreateOnly field
-            "{\"patch\": {\"MapWithTyperefs\": {\"$delete\": [\"key1\"]}}}"
+            "{\"patch\": {\"MapWithTyperefs\": {\"$delete\": [\"key1\"]}}}",
+            // Okay to set a ReadOnly field if it's the descendant of an array
+            "{\"patch\": {\"$set\": {\"ArrayWithInlineRecord\": [{\"bar1\": \"setting ReadOnly field\", \"bar2\": \"foo\"}]}}}",
```
BrianPin, line 567: What does the dollar sign `$` represent in this string-formed request?

restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataValidator.java
junchuanwang: (1) Maybe out of scope but I think "DataSchemaAnnotationValidator" is itself ambiguous, which made RestLiAnnotationDataValidator ambiguous. Do you see room to improve?  I think it is still a "**DataValidator**", and instantiated from "schemaProperty", so IMO it should be something like "SchemaCustom**DataValidator**"
(2) For "RestLiAnnotationDataValidator" according to what you said, I recommend "RestLiResourceAnnotationSchemaCustomDataValidator".

restli-int-test-api/src/main/idl/com.linkedin.restli.examples.greetings.client.autoValidationDemos.restspec.json
evanw555: This is a pathspec indicating which fields are `CreateOnly`. In this schema, the wildcard indicates that it matches any element of the array field `ArrayWithInlineRecord`. Thus, `ArrayWithInlineRecord/*/bar3` means that the `bar3` field of any element in the `ArrayWithInlineRecord` array field is `CreateOnly`.

restli-int-test/src/test/java/com/linkedin/restli/examples/TestRestLiValidation.java
evanw555: This is the request body of the `PARTIAL_UPDATE` request. See here for some details of the exact protocol: https://linkedin.github.io/rest.li/spec/protocol#partial-update

restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataValidator.java
evanw555: @karthikrg It was just an accident, I added this during testing to see the the order in which elements in the tree are validated. Given that this is invoked once for each element in a JSON object, I think it would flood the logs.

restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataValidator.java
evanw555: 1. I agree 100%, I was also confused by this. The `DataSchemaAnnotationValidator` is itself a data validator, not a schema validator (or a schema annotation validator). It doesn't sound smooth, but the most accurate name would be `CustomValidatorDataValidator` or `SchemaValidatorDataValidator` (the usage of "validator" for two different things is annoying). The name could be improved, however since this class is public changing the name has compatibility concerns. We can change it, but it's a little too much for this change IMO.
2. That suggested name is a mouthful, how about `ResourceAnnotationDataValidator` or `RestLiResourceAnnotationDataValidator`? If we get around to changing the name of the other class, we can change this one too to be consistent with the other (it's private so changing it again is no big deal).

BrianPin (approved):

restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataValidator.java
junchuanwang: For (2) RestLiResourceAnnotationDataValidator sounds good enough, can we use that?

For (1), yes, I know. Also don't know if we can really change that one day (since it is Public), lol

junchuanwang (approved):
One minor suggestion left. Otherwise looks good

restli-common/src/main/java/com/linkedin/restli/common/validation/RestLiDataValidator.java
evanw555: Yeah, I'll update the PR to use that.
> yes, I know. Also don't know if we can really change that one day (since it is Public)

Actually, I think we can maintain some compatibility by renaming the class, then preserving the old class as a deprecated subclass. Eventually, we would delete the old class.

PR #413, opened 2020-09-09
Remove final from nested generated classes.

tjni:
This enables certain techniques that use subclassing to work, such as
some using CGLib to intercept method calls.

Creating this as a follow-up to an email thread I started over the weekend.

### DIFF
--- a/generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
+++ b/generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
@@ -792,7 +792,7 @@ public class TemplateSpecGenerator
         // enclosingClass flag indicates whether a class is nested or not.
         classTemplateSpec.setEnclosingClass(enclosingClass);
         classTemplateSpec.setClassName(classInfo.name);
-        classTemplateSpec.setModifiers(ModifierSpec.PUBLIC, ModifierSpec.STATIC, ModifierSpec.FINAL);
+        classTemplateSpec.setModifiers(ModifierSpec.PUBLIC, ModifierSpec.STATIC);
       }
       else
       {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,11 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
+
+## [29.6.6] - 2020-09-17
 - Loosen `ReadOnly`/`CreateOnly` validation when setting array-descendant fields in a patch request.
 - Add generatePegasusSchemeSnapshot task.
+- Remove final from nested generated classes, such as inline unions.
 
 ## [29.6.5] - 2020-09-09
 - Update RestLiValidatorFilter and RestLiDataValidator to expose creation of restli validators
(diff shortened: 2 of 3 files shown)


generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
```diff
@@ -792,7 +792,7 @@ private ClassInfo classInfoForUnnamed(ClassTemplateSpec enclosingClass, String n
         // enclosingClass flag indicates whether a class is nested or not.
         classTemplateSpec.setEnclosingClass(enclosingClass);
         classTemplateSpec.setClassName(classInfo.name);
-        classTemplateSpec.setModifiers(ModifierSpec.PUBLIC, ModifierSpec.STATIC, ModifierSpec.FINAL);
+        classTemplateSpec.setModifiers(ModifierSpec.PUBLIC, ModifierSpec.STATIC);
```
karthikrg, line 795: **Is this safe?** 

I remember that when I removed finality of DataMap a bunch of stuff failed subtly because of == class comparison checks instead of instanceof. I suspect similar issues (that are even harder to track) may exist with these classes.

karthikbalasub: good point, RecordTemplate::putWrapped expects the object's class to be same as the generated class. So you can't put an object created by extending the generated class in a record (or other generated types). 

However, the code-gen already generates non-final classes for Arrays, Maps and Typerefs (including union typerefs). So I think this PR is not changing the behavior that much.

junchuanwang: Agree with Karthik B said.

Do we have enough existing tests to test the safety of this change?

tjni: Thank you for pointing this out.

The reason this is safe and no tests fail is because nothing in rest.li relies on subclassing these types. This only enables subclassing, which is safe in some use cases and not in others.

I have a separate question regarding the class equality checks. I've run into them a few times before. Do you know what the reason is for making them?

BrianPin (commented):
If you want this feature immediately, you probably need to update the version.
https://iwww.corp.linkedin.com/wiki/cf/display/ENGS/Pegasus+Git+Workflow
Please also remember to update CHANGE.log for this work

tjni:
Thank you! I've updated the CHANGELOG and version.

BrianPin (approved):

generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
karthikbalasub: Not sure why the original authors used class equality. One reason could be performance?
Also, for equals implementation atleast, using class equality ensures symmetry. See the discussion here: https://stackoverflow.com/questions/596462/any-reason-to-prefer-getclass-over-instanceof-when-generating-equals

tjni: Performance doesn't feel like a reason because it involves an extra check, which at best is optimized away by the JIT compiler but otherwise requires more work. What I'm referring to is snippets like:
```
protected Object unwrap(E object) throws ClassCastException
{
  ArgumentUtil.notNull(object, "object");
  if (object.getClass() == _elementClass)
  {
    return object.data();
  }
  else
  {
    throw new ClassCastException("Input " + object + " should be a " + _elementClass.getName());
  }
}
```

The point about `equals` is a good one. I need to think about it some more and will raise a separate conversation if enlightenment happens.

karthikrg: It all depends on equals semantics. If we want equals across the class hierarchy, then both should use instanceof and compare using some other mechanism. Symmetry is still possible.

Performance is not the reason. Some of it is just some implicit assumptions made by authors about the impossibility of subclassing (since PDSC/PDL never really supported inheritance)

karthikbalasub (approved):

junchuanwang (approved):

tjni:
Hi all, is this ready for merging?

nickibi:
> Hi all, is this ready for merging?

It looks like this branch has conflicts with master, please fix it. Then we could merge it.

tjni:
I merged the conflicts in the CHANGELOG and prepared it for a new patch version release.

nickibi (approved):

nickibi:
@tjni It has been merged and released in 29.6.6.

PR #414, opened 2020-09-09
Add generatePegasusSchemeSnapshot task

nickibi:
Add generatePegasusSchemaSnapshot task

The task will read pegasus schema files from input directory and generate snapshots of pegasus schemas.

It will be triggered if extension schema exist or if "enablePegasusSchemaCompatibilityCheck" configuration is provided.

Test: Unit test and manual test in other repos.

### DIFF
--- a/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
+++ b/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
@@ -19,10 +19,12 @@ package com.linkedin.pegasus.gradle;
 import com.linkedin.pegasus.gradle.PegasusOptions.IdlOptions;
 import com.linkedin.pegasus.gradle.tasks.ChangedFileReportTask;
 import com.linkedin.pegasus.gradle.tasks.CheckIdlTask;
+import com.linkedin.pegasus.gradle.tasks.CheckPegasusSnapshotTask;
 import com.linkedin.pegasus.gradle.tasks.CheckRestModelTask;
 import com.linkedin.pegasus.gradle.tasks.CheckSnapshotTask;
 import com.linkedin.pegasus.gradle.tasks.GenerateAvroSchemaTask;
 import com.linkedin.pegasus.gradle.tasks.GenerateDataTemplateTask;
+import com.linkedin.pegasus.gradle.tasks.GeneratePegasusSnapshotTask;
 import com.linkedin.pegasus.gradle.tasks.GenerateRestClientTask;
 import com.linkedin.pegasus.gradle.tasks.GenerateRestModelTask;
 import com.linkedin.pegasus.gradle.tasks.PublishRestModelTask;
@@ -32,6 +34,7 @@ import com.linkedin.pegasus.gradle.tasks.ValidateSchemaAnnotationTask;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -65,6 +68,7 @@ import org.gradle.api.tasks.Sync;
 import org.gradle.api.tasks.bundling.Jar;
 import org.gradle.api.tasks.compile.JavaCompile;
 import org.gradle.api.tasks.javadoc.Javadoc;
+import org.gradle.language.base.plugins.LifecycleBasePlugin;
 import org.gradle.plugins.ide.eclipse.EclipsePlugin;
 import org.gradle.plugins.ide.eclipse.model.EclipseModel;
 import org.gradle.plugins.ide.idea.IdeaPlugin;
@@ -571,6 +575,23 @@ public class PegasusPlugin implements Plugin<Project>
 
   private static final String PEGASUS_PLUGIN_CONFIGURATION = "pegasusPlugin";
 
+  // Enable the use of generic pegasus schema compatibility checker
+  private static final String ENABLE_PEGASUS_SCHEMA_COMPATIBILITY_CHECK = "pegasusPlugin.enablePegasusSchemaCompatibilityCheck";
+
+  private static final String PEGASUS_SCHEMA_SNAPSHOT = "PegasusSchemaSnapshot";
+
+  private static final String PEGASUS_EXTENSION_SCHEMA_SNAPSHOT = "PegasusExtensionSchemaSnapshot";
+
+  private static final String PEGASUS_SCHEMA_SNAPSHOT_DIR = "pegasusSchemaSnapshot";
+
+  private static final String PEGASUS_EXTENSION_SCHEMA_SNAPSHOT_DIR = "pegasusExtensionSchemaSnapshot";
+
+  private static final String PEGASUS_SCHEMA_SNAPSHOT_DIR_OVERRIDE = "overridePegasusSchemaSnapshotDir";
+
+  private static final String PEGASUS_EXTENSION_SCHEMA_SNAPSHOT_DIR_OVERRIDE = "overridePegasusExtensionSchemaSnapshotDir";
+
+  private static final String SRC = "src";
+
   @SuppressWarnings("unchecked")
   private Class<? extends Plugin<Project>> _thisPluginType = (Class<? extends Plugin<Project>>)
       getClass().asSubclass(Plugin.class);
@@ -809,6 +830,14 @@ public class PegasusPlugin implements Plugin<Project>
       // rest model generation could fail on incompatibility
       // if it can fail, fail it early
       configureRestModelGeneration(project, sourceSet);
+
+      // Do compatibility check for schemas under "pegasus" directory if the configuration property is provided.
+      if (isPropertyTrue(project, ENABLE_PEGASUS_SCHEMA_COMPATIBILITY_CHECK))
+      {
+        configurePegasusSchemaSnapshotGeneration(project, sourceSet, false);
+      }
+
+      configurePegasusSchemaSnapshotGeneration(project, sourceSet, true);
 
       configureConversionUtilities(project, sourceSet);
 
@@ -1063,6 +1092,32 @@ public class PegasusPlugin implements Plugin<Project>
     if (override == null)
     {
       return "src" + File.separatorChar + sourceSet.getName() + File.separatorChar + "idl";
+    }
+    else
+    {
+      return override;
+    }
+  }
+
+  private static String getPegasusSchemaSnapshotPath(Project project, SourceSet sourceSet)
+  {
+    String override = getOverridePath(project, sourceSet, PEGASUS_SCHEMA_SNAPSHOT_DIR_OVERRIDE);
+    if (override == null)
+    {
+      return SRC + File.separatorChar + sourceSet.getName() + File.separatorChar + PEGASUS_SCHEMA_SNAPSHOT_DIR;
+    }
+    else
+    {
+      return override;
+    }
+  }
+
+  private static String getPegasusExtensionSchemaSnapshotPath(Project project, SourceSet sourceSet)
+  {
+    String override = getOverridePath(project, sourceSet, PEGASUS_EXTENSION_SCHEMA_SNAPSHOT_DIR_OVERRIDE);
+    if (override == null)
+    {
+      return SRC + File.separatorChar + sourceSet.getName() + File.separatorChar + PEGASUS_EXTENSION_SCHEMA_SNAPSHOT_DIR;
     }
     else
     {
@@ -1388,6 +1443,35 @@ public class PegasusPlugin implements Plugin<Project>
         }
       }));
     });
+  }
+
+  protected void configurePegasusSchemaSnapshotGeneration(Project project, SourceSet sourceSet, boolean isExtensionSchema)
+  {
+    File schemaDir = isExtensionSchema? project.file(getExtensionSchemaPath(project, sourceSet))
+        : project.file(getDataSchemaPath(project, sourceSet));
+    Path publishablePegasusSchemaSnapshotDir = project.getBuildDir().toPath().resolve(sourceSet.getName() +
+            (isExtensionSchema ? PEGASUS_EXTENSION_SCHEMA_SNAPSHOT: PEGASUS_SCHEMA_SNAPSHOT));
+
+    Task generatePegasusSchemaSnapshot = generatePegasusSchemaSnapshot(project, sourceSet,
+        isExtensionSchema ? PEGASUS_EXTENSION_SCHEMA_SNAPSHOT: PEGASUS_SCHEMA_SNAPSHOT, schemaDir,
+        publishablePegasusSchemaSnapshotDir.toFile(), isExtensionSchema);
+
+    Task checkSchemaSnapshot = project.getTasks().create(sourceSet.getTaskName("check",
+        isExtensionSchema ? PEGASUS_EXTENSION_SCHEMA_SNAPSHOT: PEGASUS_SCHEMA_SNAPSHOT),
+        CheckPegasusSnapshotTask.class, task ->
+        {
+          task.dependsOn(generatePegasusSchemaSnapshot);
+          // TODO: update CheckPegasusSnapshotTask
+        });
+
+    File pegasusSchemaSnapshotDir = new File(isExtensionSchema ? getPegasusExtensionSchemaSnapshotPath(project, sourceSet)
+        : getPegasusSchemaSnapshotPath(project, sourceSet));
+
+    Task publishPegasusSchemaSnapshot = publishPegasusSchemaSnapshot(project, sourceSet,
+        isExtensionSchema ? PEGASUS_EXTENSION_SCHEMA_SNAPSHOT: PEGASUS_SCHEMA_SNAPSHOT, checkSchemaSnapshot,
+        publishablePegasusSchemaSnapshotDir.toFile(), pegasusSchemaSnapshotDir);
+
+    project.getTasks().getByName(LifecycleBasePlugin.ASSEMBLE_TASK_NAME).dependsOn(publishPegasusSchemaSnapshot);
   }
 
   @SuppressWarnings("deprecation")
@@ -2094,5 +2178,34 @@ public class PegasusPlugin implements Plugin<Project>
         return exclude;
       });
     });
+  }
+
+  private Task generatePegasusSchemaSnapshot(Project project, SourceSet sourceSet, String taskName, File inputDir, File outputDir, boolean isExtensionSchema)
+  {
+    return project.getTasks().create(sourceSet.getTaskName("generate", taskName),
+        GeneratePegasusSnapshotTask.class, task ->
+        {
+          task.setInputDir(inputDir);
+          task.setResolverPath(getDataModelConfig(project, sourceSet).plus(project.files(getDataSchemaPath(project, sourceSet))));
+          task.setClassPath(project.getConfigurations().getByName(PEGASUS_PLUGIN_CONFIGURATION));
+          task.setPegasusSchemaSnapshotDestinationDir(outputDir);
+          task.setEnableArgFile(isExtensionSchema);
+          if (isPropertyTrue(project, ENABLE_ARG_FILE))
+          {
+            task.setEnableArgFile(true);
+          }
+        });
+  }
+
+  private Task publishPegasusSchemaSnapshot(Project project, SourceSet sourceSet, String taskName, Task checkPegasusSnapshotTask,
+      File inputDir, File outputDir)
+  {
+    return project.getTasks().create(sourceSet.getTaskName("publish", taskName),
+        Sync.class, task ->
+        {
+          task.dependsOn(checkPegasusSnapshotTask);
+          task.from(inputDir);
+          task.into(outputDir);
+        });
   }
 }
--- a/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/GeneratePegasusSnapshotTask.java
+++ b/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/GeneratePegasusSnapshotTask.java
@@ -1,0 +1,180 @@
+/*
+   Copyright (c) 2020 LinkedIn Corp.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+package com.linkedin.pegasus.gradle.tasks;
+
+import com.linkedin.pegasus.gradle.PathingJarUtil;
+import com.linkedin.pegasus.gradle.PegasusPlugin;
+import com.linkedin.pegasus.gradle.internal.ArgumentFileGenerator;
+import java.io.File;
+import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+import org.gradle.api.DefaultTask;
+import org.gradle.api.GradleException;
+import org.gradle.api.file.FileCollection;
+import org.gradle.api.file.FileTree;
+import org.gradle.api.tasks.CacheableTask;
+import org.gradle.api.tasks.Classpath;
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.InputDirectory;
+import org.gradle.api.tasks.OutputDirectory;
+import org.gradle.api.tasks.PathSensitive;
+import org.gradle.api.tasks.PathSensitivity;
+import org.gradle.api.tasks.SkipWhenEmpty;
+import org.gradle.api.tasks.TaskAction;
+
+import static com.linkedin.pegasus.gradle.SharedFileUtils.*;
+
+
+/**
+ * Generate pegasus schema's snapshot, which will be used for pegasus schema compatibility check
+ *
+ * @author Yingjie Bi
+ */
+@CacheableTask
+public class GeneratePegasusSnapshotTask extends DefaultTask
+{
+  private File _inputDir;
+  private FileCollection _resolverPath;
+  private File _pegasusSchemaSnapshotDestinationDir;
+  private FileCollection _classPath;
+  private boolean _enableArgFile;
+  private boolean _extensionSchema;
+
+  @TaskAction
+  public void generatePegasusSnapshot()
+  {
+    FileTree inputDataSchemaFiles = _extensionSchema ? getSuffixedFiles(getProject(), _inputDir,
+        PegasusPlugin.PDL_FILE_SUFFIX) : getSuffixedFiles(getProject(), _inputDir,
+        PegasusPlugin.DATA_TEMPLATE_FILE_SUFFIXES);
+
+    List<String> inputDataSchemaFilenames = StreamSupport.stream(inputDataSchemaFiles.spliterator(), false)
+        .map(File::getPath)
+        .collect(Collectors.toList());
+
+    if (inputDataSchemaFilenames.isEmpty())
+    {
+      getLogger().lifecycle("There are no Pegasus schema input files. Skip generating Pegasus schema snapshots.");
+      return;
+    }
+    getLogger().info("Generating Pegasus schema snapshot...");
+
+    String resolverPathStr = _resolverPath.plus(getProject().files(_inputDir)).getAsPath();
+
+    FileCollection _pathedClasspath;
+    try
+    {
+      _pathedClasspath = PathingJarUtil.generatePathingJar(getProject(), getName(),
+          _classPath, false);
+    }
+    catch (IOException e)
+    {
+      throw new GradleException("Error occurred generating pathing JAR.", e);
+    }
+
+    getProject().javaexec(javaExecSpec ->
+    {
+      String resolverPathArg = resolverPathStr;
+      if (isEnableArgFile())
+      {
+        resolverPathArg = ArgumentFileGenerator.getArgFileSyntax(ArgumentFileGenerator.createArgFile(
+            "generatePegasusSchemaSnapshot_resolverPath", Collections.singletonList(resolverPathArg), getTemporaryDir()));
+      }
+      javaExecSpec.setMain("com.linkedin.restli.tools.snapshot.gen.PegasusSchemaSnapshotGenerationCmdLineApp");
+      javaExecSpec.setClasspath(_pathedClasspath);
+      javaExecSpec.args(resolverPathArg);
+      javaExecSpec.args(_inputDir.getAbsolutePath());
+      javaExecSpec.args(_pegasusSchemaSnapshotDestinationDir);
+    });
+  }
+
+  /**
+   * Directory containing pegasus schema files.
+   */
+  @InputDirectory
+  @SkipWhenEmpty
+  @PathSensitive(PathSensitivity.RELATIVE)
+  public File getInputDir()
+  {
+    return _inputDir;
+  }
+
+  public void setInputDir(File inputDir)
+  {
+    _inputDir = inputDir;
+  }
+
+  /**
+   * The resolver path.
+   */
+  @Classpath
+  public FileCollection getResolverPath()
+  {
+    return _resolverPath;
+  }
+
+  public void setResolverPath(FileCollection resolverPath)
+  {
+    _resolverPath = resolverPath;
+  }
+
+  @OutputDirectory
+  public File getPegasusSchemaSnapshotDestinationDir()
+  {
+    return _pegasusSchemaSnapshotDestinationDir;
+  }
+
+  public void setPegasusSchemaSnapshotDestinationDir(File pegasusSchemaSnapshotDestinationDir )
+  {
+    _pegasusSchemaSnapshotDestinationDir = pegasusSchemaSnapshotDestinationDir;
+  }
+
+
+  @Classpath
+  public FileCollection getClassPath()
+  {
+    return _classPath;
+  }
+
+  public void setClassPath(FileCollection classPath)
+  {
+    _classPath = classPath;
+  }
+
+  @Input
+  public boolean isEnableArgFile()
+  {
+    return _enableArgFile;
+  }
+
+  public void setEnableArgFile(boolean enable)
+  {
+    _enableArgFile = enable;
+  }
+
+  @Input
+  public boolean isExtensionSchema()
+  {
+    return _extensionSchema;
+  }
+
+  public void setExtensionSchema(boolean extensionSchema)
+  {
+    _extensionSchema = extensionSchema;
+  }
+}
--- a/restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/gen/PegasusSchemaSnapshotExporter.java
+++ b/restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/gen/PegasusSchemaSnapshotExporter.java
@@ -1,0 +1,127 @@
+/*
+   Copyright (c) 2020 LinkedIn Corp.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+package com.linkedin.restli.tools.snapshot.gen;
+
+import com.linkedin.data.schema.AbstractSchemaEncoder;
+import com.linkedin.data.schema.AbstractSchemaParser;
+import com.linkedin.data.schema.DataSchema;
+import com.linkedin.data.schema.DataSchemaResolver;
+import com.linkedin.data.schema.NamedDataSchema;
+import com.linkedin.data.schema.PegasusSchemaParser;
+import com.linkedin.data.schema.SchemaToPdlEncoder;
+import com.linkedin.data.schema.resolver.MultiFormatDataSchemaResolver;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.StringWriter;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+
+/**
+ * PegasusSchemaSnapshotExporter, generating pegasus schema snapshot(.pdl) files
+ *
+ * @author Yingjie Bi
+ */
+public class PegasusSchemaSnapshotExporter
+{
+  private static final String PDL = ".pdl";
+
+  private static final String PDSC = ".pdsc";
+
+  /**
+   * Generate pegasus schema snapshot(pegasusSchemaSnapshot.pdl) files to the provided output directory
+   * based on the given input pegasus schemas.
+   *
+   * @param resolverPath schema resolver path
+   * @param inputPath input files directory
+   * @param outputDir output files directory
+   * @throws IOException
+   */
+  public void export(String resolverPath, String inputPath, File outputDir) throws IOException
+  {
+    List<DataSchema> dataSchemas = parseDataSchema(resolverPath, inputPath);
+    for (DataSchema dataSchema : dataSchemas)
+    {
+      writeSnapshotFile(outputDir, ((NamedDataSchema) dataSchema).getFullName(), dataSchema);
+    }
+  }
+
+  private static List<DataSchema> parseDataSchema(String resolverPath, String inputPath)
+      throws RuntimeException, IOException
+  {
+    try (Stream<Path> paths = Files.walk(Paths.get(inputPath)))
+    {
+      return paths
+          .filter(path -> path.toString().endsWith(PDL) || path.toString().endsWith(PDSC))
+          .map(path ->
+              {
+                File inputFile = path.toFile();
+                DataSchemaResolver resolver = MultiFormatDataSchemaResolver.withBuiltinFormats(resolverPath);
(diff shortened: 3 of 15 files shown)


gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1390,6 +1435,87 @@ protected void configureRestModelGeneration(Project project, SourceSet sourceSet
     });
   }
 
+  protected void configurePegasusSchemaSnapshotGeneration(Project project, SourceSet sourceSet)
```
BrianPin, line 1438: I feel this task function and the other function below `configurePegasusExtensionSchemaSnapshotGeneration`
share some common logic, would it be better to factor some common logic together to behave the DIY(Dont repeat yourself) pattern?

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/gen/PegasusSchemaSnapshotExporter.java
```diff
@@ -0,0 +1,159 @@
...
+   * Generate pegasus schema snapshot(pegasusSchemaSnapshot.pdl) files to the provided output directory
+   * based on the given input pegasus schemas.
+   *
+   * @param resolverPath schema resolver path
+   * @param inputPath input files directory
+   * @param outputDir output files directory
+   * @return GeneratorResult
+   * @throws IOException
+   */
+  public GeneratorResult export(String resolverPath, String inputPath, File outputDir) throws IOException {
```
BrianPin, line 59: style

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/gen/PegasusSchemaSnapshotExporter.java
```diff
@@ -0,0 +1,159 @@
...
+      {
+        throw new RuntimeException(parser.errorMessage());
+      }
+
+      List<DataSchema> topLevelDataSchemas = parser.topLevelDataSchemas();
+      if (topLevelDataSchemas == null || topLevelDataSchemas.isEmpty() || topLevelDataSchemas.size() > 1)
+      {
+        throw new RuntimeException("Could not parse extension schema : " + inputFile.getAbsolutePath());
+      }
+      DataSchema topLevelDataSchema = topLevelDataSchemas.get(0);
```
BrianPin, line 92: Is `get(0)` meant to get top level schema only?
If say sub-level schema changed, and we just `get(0)`, that means we dont treat sub-level schema change as change? Since the snapshot didn't change.
By sub-level I mean some recursive entity in the top level record's field

BrianPin: thanks for addressing this question offline

BrianPin (approved):
Reviewed, LGTM

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/gen/PegasusSchemaSnapshotExporter.java
nickibi: fixed

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
nickibi: I thought about it. Then I thought that generateExtensionSchemaSnapshot and generateSchemaSnapshot are two parallel code paths. Both of them call GeneratePegasusSnapshotTask to generate the snapshots. But the purposes of two tasks are kind of different. 
Having separate generatePegasusSchemaSnapshot task and generatePegasusExtensionSchemaSnapshot task makes future customized configuration easier.

nickibi: Abstracted common logic to helper methods.

aman1309 (commented):
can we add groovy integ tests also?

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/gen/PegasusSchemaSnapshotExporter.java
```diff
@@ -0,0 +1,162 @@
...
+     result.addModifiedFile(file);
+     result.addTargetFile(file);
+    }
+    return result;
+  }
+
+  private static List<DataSchema> parseDataSchema(String resolverPath, String inputPath)
+      throws RuntimeException, FileNotFoundException
+  {
+    Iterator<File> iterator = FileUtils.iterateFiles(new File(inputPath), new String[]{PDL}, true);
```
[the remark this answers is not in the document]
nickibi: Extensions schema can only be pdl by design.
Since we are going to remove pdsc support in the future, I feel for new feature, we should support pdl only.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -571,6 +573,16 @@
 
   private static final String PEGASUS_PLUGIN_CONFIGURATION = "pegasusPlugin";
 
+  // Enable the use of generic pegasus schema compatibility checker
+  private static final String ENABLE_PEGASUS_SCHEMA_COMPATIBILITY_CHECK = "pegasusPlugin.enablePegasusSchemaCompatibilityCheck";
```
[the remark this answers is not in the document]
nickibi: "pegasusPlugin." prefix is used to make the configuration name unique, just in case the configuration name is duplicate with other configuration name in other plugins. It is the same format as other pegasus configurations. I will keep this part as is.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -571,6 +573,16 @@
...
+
+  private static final String PEGASUS_SCHEMA_SNAPSHOT = "PegasusSchemaSnapshot";
```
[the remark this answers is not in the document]
nickibi: I am kind of confused here. Are you saying PegasusSchemaSnapshot is not an appropriate name? It is used as the gradle task name. For example, generatePegasusSchemaSnapshot.
"Entity" is definitely inappropriate here.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/GeneratePegasusSnapshotTask.java
```diff
@@ -0,0 +1,168 @@
...
+import static com.linkedin.pegasus.gradle.SharedFileUtils.*;
+
+
+/**
+ * Generate pegasus schema's snapshot, which will be used for pegasus schema compatibility check
+ *
+ * To use this plugin, add these three lines to your build.gradle:
+ * <pre>
+ * apply plugin: 'li-pegasus'
+ * </pre>
```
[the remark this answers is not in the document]
nickibi: I would prefer to keep this part as other gradle tasks do. But it should be - `apply  plugin : "pegasus"` for the open source usage.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1390,6 +1435,64 @@ protected void configureRestModelGeneration(Project project, SourceSet sourceSet
...
+          // TODO: update CheckPegasusSnapshotTask
+        });
+
+    File pegasusSchemaSnapshotDir = new File(getPegasusSchemaSnapshotPath(project, sourceSet));
+    pegasusSchemaSnapshotDir.mkdirs();
+
+    Task publishPegasusSchemaSnapshot = publishPegasusSchemaSnapshot(project, sourceSet,
+        PEGASUS_SCHEMA_SNAPSHOT, checkSchemaSnapshot, publishablePegasusSchemaSnapshotDir,pegasusSchemaSnapshotDir);
+
+    project.getTasks().getByName("assemble").dependsOn(publishPegasusSchemaSnapshot);
```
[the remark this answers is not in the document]
nickibi: Used constant instead.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1390,6 +1435,64 @@ protected void configureRestModelGeneration(Project project, SourceSet sourceSet
...
+          });
+
+      Task publishPegasusExtensionSchemaSnapshot = publishPegasusSchemaSnapshot(project, sourceSet, PEGASUS_EXTENSION_SCHEMA_SNAPSHOT,
+          checkPegasusExtensionSchemaSnapshotTask, publishablePegasusExtensionSchemaSnapshotDir, pegasusExtensionSchemaSnapshotDir);
+
+      project.getTasks().getByName("assemble").dependsOn(publishPegasusExtensionSchemaSnapshot);
+    }
+    else
+    {
+      project.getLogger().info("No extension schemas, skip generatePegasusExtensionSchemaSnapshot task");
```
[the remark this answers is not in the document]
nickibi: info is not required, changed it to debug log.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1070,6 +1089,32 @@ private static String getIdlPath(Project project, SourceSet sourceSet)
     }
   }
 
+  private static String getPegasusSchemaSnapshotPath(Project project, SourceSet sourceSet)
+  {
+    String override = getOverridePath(project, sourceSet, "overridePegasusSchemaSnapshotDir");
+    if (override == null)
+    {
+      return "src" + File.separatorChar + sourceSet.getName() + File.separatorChar + "pegasusSchemaSnapshot";
```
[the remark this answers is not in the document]
nickibi: used constant instead

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/gen/PegasusSchemaSnapshotExporter.java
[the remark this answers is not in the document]
karthikbalasub: This task is also used for base schema snapshot and we. still support PDSC in oss. So I suggest adding PDSC support. You can keep the snapshot file format as PDL.

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/gen/PegasusSchemaSnapshotExporter.java
```diff
@@ -0,0 +1,162 @@
...
+    List<DataSchema> schemas = new ArrayList<>();
+    while (iterator.hasNext())
+    {
+      DataSchemaResolver resolver = MultiFormatDataSchemaResolver.withBuiltinFormats(resolverPath);
```
karthikbalasub, line 80: You can move this resolver outside the loop.

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/gen/PegasusSchemaSnapshotExporter.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    try (FileOutputStream out = new FileOutputStream(generatedSnapshotFile))
+    {
+      out.write(stringWriter.toString().getBytes());
+    }
+    return generatedSnapshotFile;
+  }
+
+  private static class pegasusSchemaSnapshotResult implements GeneratorResult
```
karthikbalasub, line 129: What is the purpose of this result? Please add documentation

restli-tools/src/test/java/com/linkedin/restli/tools/ExporterTestUtils.java
```diff
@@ -85,6 +85,14 @@ public static void compareFiles(String actualFileName, String expectedFileName)
     }
   }
 
+  public static void comparePegasusSchemaSnapshotFiles(String actualFileName, String expectedFileName)
+      throws IOException {
+    String actualContent = ExporterTestUtils.readFile(actualFileName);
+    String expectedContent = ExporterTestUtils.readFile(expectedFileName);
+    Assert.assertEquals(actualContent, expectedContent);
+
```
karthikbalasub, line 93: nit: delete empty line

restli-tools/src/test/java/com/linkedin/restli/tools/ExporterTestUtils.java
```diff
@@ -85,6 +85,14 @@ public static void compareFiles(String actualFileName, String expectedFileName)
     }
   }
 
+  public static void comparePegasusSchemaSnapshotFiles(String actualFileName, String expectedFileName)
+      throws IOException {
```
karthikbalasub, line 89: nit: brace to next line

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -571,6 +573,28 @@
...
   private static final String PEGASUS_PLUGIN_CONFIGURATION = "pegasusPlugin";
 
+  // Enable the use of generic pegasus schema compatibility checker
+  private static final String ENABLE_PEGASUS_SCHEMA_COMPATIBILITY_CHECK = "pegasusPlugin.enablePegasusSchemaCompatibilityCheck";
+
+  private static final String PEGASUS_SCHEMA_SNAPSHOT = "PegasusSchemaSnapshot";
+
+  private static final String PEGASUS_EXTENSION_SCHEMA_SNAPSHOT = "PegasusExtensionSchemaSnapshot";
+
+  private static final String PEGASUS_SCHEMA_SNAPSHOT_DIR = "pegasusSchemaSnapshot";
```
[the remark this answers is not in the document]
nickibi: I feel they are two different use cases. Extension schema snapshot is a default feature, regular one is an optional feature. It could be clearer to keep them into separate directories. Also whether there will be a scenario regular schema and extension schema have the same name coincidently?

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/gen/PegasusSchemaSnapshotExporter.java
nickibi: The resolver was outside the loop. 
However, I tested the change in dr-restaurant, in codegen module while parsing the pdl, it threw an error that a schema is already defined. 
I think each input file should have its own resolver to avoid such case.

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/gen/PegasusSchemaSnapshotExporter.java
nickibi: I added it for testing purpose.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -810,6 +834,13 @@ public void apply(Project project)
       // if it can fail, fail it early
       configureRestModelGeneration(project, sourceSet);
 
+      if (isPropertyTrue(project, ENABLE_PEGASUS_SCHEMA_COMPATIBILITY_CHECK))
```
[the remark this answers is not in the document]
nickibi: Added

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -571,6 +573,28 @@
...
+
+  private static final String PEGASUS_EXTENSION_SCHEMA_SNAPSHOT_DIR_OVERRIDE = "overridePegasusExtensionSchemaSnapshotDir";
+
+  private static final String ASSEMBLE = "assemble";
+
+  private static final String SRC = "src";
+
+
+
+
```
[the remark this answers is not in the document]
nickibi: fixed

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1390,6 +1447,64 @@ protected void configureRestModelGeneration(Project project, SourceSet sourceSet
...
+    File pegasusSchemaSnapshotDir = new File(getPegasusSchemaSnapshotPath(project, sourceSet));
+    pegasusSchemaSnapshotDir.mkdirs();
+
+    Task publishPegasusSchemaSnapshot = publishPegasusSchemaSnapshot(project, sourceSet,
+        PEGASUS_SCHEMA_SNAPSHOT, checkSchemaSnapshot, publishablePegasusSchemaSnapshotDir,pegasusSchemaSnapshotDir);
+
+    project.getTasks().getByName(ASSEMBLE).dependsOn(publishPegasusSchemaSnapshot);
+  }
+
+  protected void configurePegasusExtensionSchemaSnapshotGeneration(Project project, SourceSet sourceSet)
```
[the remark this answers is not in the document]
nickibi: Fixed

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -2095,4 +2210,32 @@ private static void configureDataSchemaResourcesRoot(Project project, SourceSet
...
+          task.setClassPath(project.getConfigurations().getByName(PEGASUS_PLUGIN_CONFIGURATION));
+          task.setPegasusSchemaSnapshotDestinationDir(outputDir);
+          if (isPropertyTrue(project, ENABLE_ARG_FILE))
+          {
+            task.setEnableArgFile(true);
+          }
+        });
+  }
+
+  private Task publishPegasusSchemaSnapshot(Project project, SourceSet sourceSet, String taskName, Task DependentTask,
```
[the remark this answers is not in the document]
nickibi: Fixed

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/GeneratePegasusSnapshotTask.java
```diff
@@ -0,0 +1,168 @@
...
+
+
+/**
+ * Generate pegasus schema's snapshot, which will be used for pegasus schema compatibility check
+ *
+ * To use this plugin, add these three lines to your build.gradle:
+ * <pre>
+ * apply plugin: 'pegasus'
+ * </pre>
+ *
```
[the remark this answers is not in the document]
nickibi: Removed

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/GeneratePegasusSnapshotTask.java
```diff
@@ -0,0 +1,168 @@
...
+    getProject().getLogger().info("Generating Pegasus schema snapshot...");
+
+    String resolverPathStr = _resolverPath.plus(getProject().files(_inputDir)).getAsPath();
+
+    FileCollection _pathedClasspath;
+    try {
+      _pathedClasspath = PathingJarUtil.generatePathingJar(getProject(), getName(),
+          _classPath, false);
+    }
+    catch (IOException e) {
```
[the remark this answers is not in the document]
nickibi: Fixed

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/GeneratePegasusSnapshotTask.java
```diff
@@ -0,0 +1,168 @@
...
+      throw new GradleException("Error occurred generating pathing JAR.", e);
+    }
+
+    getProject().javaexec(javaExecSpec -> {
```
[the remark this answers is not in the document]
nickibi: Fixed

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/gen/PegasusSchemaSnapshotExporter.java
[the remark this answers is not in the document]
nickibi: Updated

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/gen/PegasusSchemaSnapshotExporter.java
nickibi: Since it is not used anywhere, removed.

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/gen/PegasusSchemaSnapshotGenerationCmdLineApp.java
```diff
@@ -0,0 +1,116 @@
...
+ * @author Yingjie Bi
+ */
+public class PegasusSchemaSnapshotGenerationCmdLineApp
+{
+  private static final Logger _logger = LoggerFactory.getLogger(
+      PegasusSchemaSnapshotGenerationCmdLineApp.class);
+
+  private static final Options _options = new Options();
+
+
```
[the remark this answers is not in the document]
nickibi: Removed

restli-tools/src/test/java/com/linkedin/restli/tools/ExporterTestUtils.java
nickibi: Fixed

restli-tools/src/test/java/com/linkedin/restli/tools/ExporterTestUtils.java
nickibi: Fixed

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1390,6 +1447,64 @@ protected void configureRestModelGeneration(Project project, SourceSet sourceSet
...
+        {
+          task.dependsOn(generatePegasusSchemaSnapshot);
+          // TODO: update CheckPegasusSnapshotTask
+        });
+
+    File pegasusSchemaSnapshotDir = new File(getPegasusSchemaSnapshotPath(project, sourceSet));
+    pegasusSchemaSnapshotDir.mkdirs();
+
+    Task publishPegasusSchemaSnapshot = publishPegasusSchemaSnapshot(project, sourceSet,
+        PEGASUS_SCHEMA_SNAPSHOT, checkSchemaSnapshot, publishablePegasusSchemaSnapshotDir,pegasusSchemaSnapshotDir);
```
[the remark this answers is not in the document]
nickibi: fixed

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/gen/PegasusSchemaSnapshotExporter.java
```diff
@@ -0,0 +1,130 @@
...
+
+    try (FileOutputStream out = new FileOutputStream(generatedSnapshotFile))
+    {
+      out.write(stringWriter.toString().getBytes());
+    }
+  }
+
+  private static String getFileExtension(String fileName)
+  {
+    return fileName.substring(fileName.lastIndexOf(".") + 1);
```
tjni, line 128: This is honestly just an OCD nitpick, but `lastIndexOf('.')` seems like it should be a little more efficient than `lastIndexOf(".")`.

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/gen/PegasusSchemaSnapshotExporter.java
```diff
@@ -0,0 +1,130 @@
...
+      {
+        throw new IOException(generatedSnapshotFile + ": Can't delete previous version");
+      }
+    }
+
+    try (FileOutputStream out = new FileOutputStream(generatedSnapshotFile))
+    {
+      out.write(stringWriter.toString().getBytes());
+    }
+  }
```
tjni, line 124: This can be replaced by

```
Files.write(generatedSnapshotFile.toPath(), stringWriter.toString().getBytes())
```

As a nitpick, call `getBytes(StandardCharsets.UTF_8)` so as to not rely on the default character encoding of the system.

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/gen/PegasusSchemaSnapshotGenerationCmdLineApp.java
```diff
@@ -0,0 +1,114 @@
...
+          throw new RuntimeException("Output directory '" + outputDir + "' could not be created!");
+        }
+      }
+      if (!outputDir.isDirectory())
+      {
+        throw new RuntimeException("Output directory '" + outputDir + "' is not a directory");
+      }
+      if (!outputDir.canRead() || !outputDir.canWrite())
+      {
+        throw new RuntimeException("Output directory '" + outputDir + "' must be writeable");
```
tjni, line 88: Nit: Must be readable and writeable.

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/gen/PegasusSchemaSnapshotGenerationCmdLineApp.java
```diff
@@ -0,0 +1,114 @@
...
+    if (cliArgs.length != 3)
+    {
+      _logger.error("Invalid arguments");
+      help();
+      System.exit(1);
+    }
+    int i = 0;
+    String resolverPath = RestLiToolsUtils.readArgFromFileIfNeeded(cliArgs[i++]);
+    String inputPath = cliArgs[i++];
+    String outputPath = cliArgs[i];
```
tjni, line 70: For clarity, consider just using `0`, `1`, and `2` directly instead of incrementing `i`.

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/gen/PegasusSchemaSnapshotExporter.java
```diff
@@ -0,0 +1,130 @@
...
+
+    File generatedSnapshotFile = new File(outputDir, fileName + PEGASUS_SCHEMA_SNAPSHOT_SUFFIX);
+
+    if (generatedSnapshotFile.exists())
+    {
+      if (!generatedSnapshotFile.delete())
+      {
+        throw new IOException(generatedSnapshotFile + ": Can't delete previous version");
+      }
+    }
```
tjni, line 118: Is this necessary, given that the way `FileOutputStream` is constructed later is to overwrite any existing files?

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/gen/PegasusSchemaSnapshotExporter.java
```diff
@@ -0,0 +1,130 @@
...
+
+      if (parser.hasError())
+      {
+        throw new RuntimeException(parser.errorMessage());
+      }
+
+      List<DataSchema> topLevelDataSchemas = parser.topLevelDataSchemas();
+      if (topLevelDataSchemas == null || topLevelDataSchemas.isEmpty() || topLevelDataSchemas.size() > 1)
+      {
+        throw new RuntimeException("Could not parse schema : " + inputFile.getAbsolutePath());
```
tjni, line 91: This can probably be a more informative error message.

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/gen/PegasusSchemaSnapshotExporter.java
```diff
@@ -0,0 +1,130 @@
...
+      PegasusSchemaParser parser = AbstractSchemaParser.parserForFileExtension(fileExtension, resolver);
+      parser.parse(new FileInputStream(inputFile));
+
+      if (parser.hasError())
+      {
+        throw new RuntimeException(parser.errorMessage());
+      }
+
+      List<DataSchema> topLevelDataSchemas = parser.topLevelDataSchemas();
+      if (topLevelDataSchemas == null || topLevelDataSchemas.isEmpty() || topLevelDataSchemas.size() > 1)
```
tjni, line 89: Nit: It looks like `topLevelDataSchemas()` never returns `null`, so the condition can be simplified to `topLevelDataSchemas.size() != 1`.

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/gen/PegasusSchemaSnapshotExporter.java
```diff
@@ -0,0 +1,130 @@
...
+      DataSchemaResolver resolver = MultiFormatDataSchemaResolver.withBuiltinFormats(resolverPath);
+      File inputFile = iterator.next();
+      String fileExtension = getFileExtension(inputFile.getName());
+      PegasusSchemaParser parser = AbstractSchemaParser.parserForFileExtension(fileExtension, resolver);
+      parser.parse(new FileInputStream(inputFile));
+
+      if (parser.hasError())
+      {
+        throw new RuntimeException(parser.errorMessage());
+      }
```
tjni, line 86: Can you double check if the parser's error message includes the name of the file? (If it does not, let's either update the parser to include it or include it here.)

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/gen/PegasusSchemaSnapshotExporter.java
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  private static List<DataSchema> parseDataSchema(String resolverPath, String inputPath)
+      throws RuntimeException, FileNotFoundException
+  {
+    Iterator<File> iterator = FileUtils.iterateFiles(new File(inputPath), new String[]{PDL, PDSC}, true);
+    List<DataSchema> schemas = new ArrayList<>();
+    while (iterator.hasNext())
+    {
+      DataSchemaResolver resolver = MultiFormatDataSchemaResolver.withBuiltinFormats(resolverPath);
```
tjni, line 77: Consider implementing this using streams natively, something like:

```
try (Stream<Path> paths = Files.walk(...))
{
  return paths
      .filter(/* files that end in PDL or PDSC */)
      .map(/* convert to DataSchema */)
      .collect(Collectors.toList());
}
```

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/GeneratePegasusSnapshotTask.java
```diff
@@ -0,0 +1,167 @@
...
+    List<String> inputDataSchemaFilenames = StreamSupport.stream(inputDataSchemaFiles.spliterator(), false)
+        .map(File::getPath)
+        .collect(Collectors.toList());
+
+    if (inputDataSchemaFilenames.isEmpty())
+    {
+      getLogger().lifecycle("There are no Pegasus schema input files. Skip generating Pegasus schema snapshots.");
+      return;
+    }
+    getProject().getLogger().info("Generating Pegasus schema snapshot...");
```
tjni, line 73: Instead of using the project's logger, we can use the task's logger as you do above.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -2095,4 +2184,32 @@ private static void configureDataSchemaResourcesRoot(Project project, SourceSet
...
+            task.setEnableArgFile(true);
+          }
+        });
+  }
+
+  private Task publishPegasusSchemaSnapshot(Project project, SourceSet sourceSet, String taskName, Task checkPegasusSnapshotTast,
+      File inputDir, File outputDir)
+  {
+    return project.getTasks().create(sourceSet.getTaskName("publish", taskName),
+        Copy.class, task ->
```
tjni, line 2208: Consider whether the [Sync](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/Sync.html) task is better, since it also deletes anything that has been removed.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -2095,4 +2184,32 @@ private static void configureDataSchemaResourcesRoot(Project project, SourceSet
...
+          task.setClassPath(project.getConfigurations().getByName(PEGASUS_PLUGIN_CONFIGURATION));
+          task.setPegasusSchemaSnapshotDestinationDir(outputDir);
+          if (isPropertyTrue(project, ENABLE_ARG_FILE))
+          {
+            task.setEnableArgFile(true);
+          }
+        });
+  }
+
+  private Task publishPegasusSchemaSnapshot(Project project, SourceSet sourceSet, String taskName, Task checkPegasusSnapshotTast,
```
tjni, line 2204: checkPegasusSnapshotTast => checkPegasusSnapshotTask

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1390,6 +1445,40 @@ protected void configureRestModelGeneration(Project project, SourceSet sourceSet
...
+
+    File pegasusSchemaSnapshotDir = new File(isExtensionSchema ? getPegasusExtensionSchemaSnapshotPath(project, sourceSet)
+        : getPegasusSchemaSnapshotPath(project, sourceSet));
+    pegasusSchemaSnapshotDir.mkdirs();
+
+    Task publishPegasusSchemaSnapshot = publishPegasusSchemaSnapshot(project, sourceSet,
+        isExtensionSchema ? PEGASUS_EXTENSION_SCHEMA_SNAPSHOT: PEGASUS_SCHEMA_SNAPSHOT, checkSchemaSnapshot,
+        publishablePegasusSchemaSnapshotDir, pegasusSchemaSnapshotDir);
+
+    project.getTasks().getByName(ASSEMBLE).dependsOn(publishPegasusSchemaSnapshot);
```
tjni, line 1479: In general, if support for Gradle 4 can be dropped, consider starting to use [task configuration avoidance](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html). This involves using `TaskContainer#register` instead of `TaskContainer#create`, for example, and this line would become:

```
project.getTasks().named(ASSEMBLE,
    assembleTask -> assembleTask.dependsOn(publishPegasusSchemaSnapshot));
```

Feel free also to postpone that until a later time too, as it's not a completely trivial change.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1390,6 +1445,40 @@ protected void configureRestModelGeneration(Project project, SourceSet sourceSet
...
+        isExtensionSchema ? PEGASUS_EXTENSION_SCHEMA_SNAPSHOT: PEGASUS_SCHEMA_SNAPSHOT),
+        CheckPegasusSnapshotTask.class, task ->
+        {
+          task.dependsOn(generatePegasusSchemaSnapshot);
+          // TODO: update CheckPegasusSnapshotTask
+        });
+
+    File pegasusSchemaSnapshotDir = new File(isExtensionSchema ? getPegasusExtensionSchemaSnapshotPath(project, sourceSet)
+        : getPegasusSchemaSnapshotPath(project, sourceSet));
+    pegasusSchemaSnapshotDir.mkdirs();
```
tjni, line 1473: The `into` method on the `Copy` task (and the `Sync` task) configures the destination directory property of the task, which is annotated with `@OutputDirectory`. This means that Gradle will [create the directory for us if it is missing](https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:task_input_output_validation), and this call to `mkdirs()` can be removed.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1390,6 +1445,40 @@ protected void configureRestModelGeneration(Project project, SourceSet sourceSet
...
+  {
+    File schemaDir = isExtensionSchema? project.file(getExtensionSchemaPath(project, sourceSet))
+        : project.file(getDataSchemaPath(project, sourceSet));
+    if (isExtensionSchema && !SharedFileUtils.getSuffixedFiles(project, schemaDir, PDL_FILE_SUFFIX).isEmpty())
+    {
+      project.getLogger().debug("No extension schemas, skip generatePegasusExtensionSchemaSnapshot task");
+      return;
+    }
+    File publishablePegasusSchemaSnapshotDir = project.file(project.getBuildDir().getAbsolutePath()
+        + File.separatorChar + sourceSet.getName() + (isExtensionSchema ? PEGASUS_EXTENSION_SCHEMA_SNAPSHOT: PEGASUS_SCHEMA_SNAPSHOT));
```
tjni, line 1458: It shouldn't be necessary to build a String and pass it through `Project#file`. I would recommend using `java.nio.file.Path` whenever possible, since it allows you to write:

```
project.getBuildDir().toPath().resolve(sourceSet.getName() + ...)
```

Gradle task properties can be `Path` instead of `File` as well, and it is also possible to convert from a `Path` to a `File` using `Path#toFile`.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1390,6 +1445,40 @@ protected void configureRestModelGeneration(Project project, SourceSet sourceSet
...
 
+  protected void configurePegasusSchemaSnapshotGeneration(Project project, SourceSet sourceSet, boolean isExtensionSchema)
+  {
+    File schemaDir = isExtensionSchema? project.file(getExtensionSchemaPath(project, sourceSet))
+        : project.file(getDataSchemaPath(project, sourceSet));
+    if (isExtensionSchema && !SharedFileUtils.getSuffixedFiles(project, schemaDir, PDL_FILE_SUFFIX).isEmpty())
+    {
+      project.getLogger().debug("No extension schemas, skip generatePegasusExtensionSchemaSnapshot task");
+      return;
+    }
```
tjni, line 1456: At first glance, it's not obvious to me why this check only applies to the extension task.

Separately, there is a way to describe this behavior more declaratively that you can consider. Your implementation is pretty close to it already.

If you create a [FileTree](https://docs.gradle.org/current/userguide/working_with_files.html#sec:file_trees) that includes only PDL files, it can be passed to tasks that require them through a task property of type `Iterable<File>` (or a subclass, such as `FileTree` itself) and annotated with `@InputFiles`. Furthermore, the property can be annotated with `@SkipWhenEmpty`, which will cause Gradle to skip the task if no input files were found.

The benefit of this approach is that the task will always exist, so a script that runs, for instance, `./gradlew generatePegasusExtensionSchemaSnapshot` won't fail when no PDL files are found.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -571,6 +573,25 @@
...
+
+  private static final String PEGASUS_SCHEMA_SNAPSHOT_DIR = "pegasusSchemaSnapshot";
+
+  private static final String PEGASUS_EXTENSION_SCHEMA_SNAPSHOT_DIR = "pegasusExtensionSchemaSnapshot";
+
+  private static final String PEGASUS_SCHEMA_SNAPSHOT_DIR_OVERRIDE = "overridePegasusSchemaSnapshotDir";
+
+  private static final String PEGASUS_EXTENSION_SCHEMA_SNAPSHOT_DIR_OVERRIDE = "overridePegasusExtensionSchemaSnapshotDir";
+
+  private static final String ASSEMBLE = "assemble";
```
tjni, line 591: Consider using [LifecycleBasePlugin.ASSEMBLE_TASK_NAME](https://github.com/gradle/gradle/blob/434c8d40bb2776b1513a2e4a3b091f98ff636382/subprojects/platform-base/src/main/java/org/gradle/language/base/plugins/LifecycleBasePlugin.java#L33) constant.

tjni (commented):
Had a little spare time so a few drive-by comments. Feel free to apply or drop them at your preference.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
nickibi: I was thinking that extension task will run by default and extension schemas can only be PDL format, so I added this check.  The regular schema task only runs when configuration property is provided.

Updated GeneratePegasusSnapshotTask based on the suggestion.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
nickibi: That's a good suggestion. I'd prefer to do that later. Thanks.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
nickibi: Used Sync instead

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/gen/PegasusSchemaSnapshotExporter.java
nickibi: Used streams

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/gen/PegasusSchemaSnapshotExporter.java
nickibi: included the name of the file here.

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/gen/PegasusSchemaSnapshotExporter.java
nickibi: Good catch, removed.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
nickibi: Good to know that, removed.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
nickibi: Updated based on suggestion

PR #415, opened 2020-09-17
Change D2 call to async in BackupRequestClient

rachelhanhan:
**Background**
Recently we found some corner cases that is causing D2 Zookeeper call taking a long time and eventually exhausted all threads. After doing some investigation, we found there are 2 places that are making blocking calls to Zookeeper:
1. TimeoutRequestClient
2. BackupRequestClient
We decided to directly deprecate the usage of TimeoutRequestClient, and change the call to be async in BackupRequestClient.

**Changes include**
1. Change the TimeoutRequestClient to be disabled by default (The actual switch is in global config which we will also turn off)
2. Exposed a config to enable the async call in BackupRequestClient.
2. Added the support to make the Zookeeper call to be async in BackupRequestClient.

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/clients/BackupRequestsClient.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/clients/BackupRequestsClient.java
@@ -54,6 +54,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 import org.HdrHistogram.AbstractHistogram;
+import org.checkerframework.checker.nullness.Opt;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -75,6 +76,7 @@ public class BackupRequestsClient extends D2ClientDelegator
   private final ScheduledExecutorService _executorService;
   private final ScheduledThreadPoolExecutor _latenciesNotifierExecutor;
   private final ScheduledFuture<?> _latenciesNotifier;
+  private final boolean _isD2Async;
 
   // serviceName -> operation -> BackupRequestsStrategyFromConfig
   private final Map<String, Map<String, BackupRequestsStrategyFromConfig>> _strategies = new ConcurrentHashMap<>();
@@ -96,7 +98,8 @@ public class BackupRequestsClient extends D2ClientDelegator
   private final Map<String, List<Map<String, Object>>> _configs = new ConcurrentHashMap<>();
 
   public BackupRequestsClient(D2Client d2Client, LoadBalancer loadBalancer, ScheduledExecutorService executorService,
-      BackupRequestsStrategyStatsConsumer statsConsumer, long notifyLatencyInterval, TimeUnit notifyLatencyIntervalUnit)
+      BackupRequestsStrategyStatsConsumer statsConsumer, long notifyLatencyInterval, TimeUnit notifyLatencyIntervalUnit,
+      boolean isD2Async)
   {
     super(d2Client);
     _loadBalancer = loadBalancer;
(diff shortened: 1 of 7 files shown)


d2/src/main/java/com/linkedin/d2/balancer/clients/BackupRequestsClient.java
```diff
@@ -195,24 +247,65 @@ public void restRequest(final RestRequest request, final RequestContext requestC
    */
   private void updateIfNeeded(String serviceName)
   {
-    List<Map<String, Object>> existing = _configs.get(serviceName);
     try
     {
       ServiceProperties serviceProperties = _loadBalancer.getLoadBalancedServiceProperties(serviceName);
```
ssheng, line 252: This is still calling the sync implementation, right?

rachelhanhan: I created another method updatedIfNeeded(serviceName, callback) and introduced a configuration. If async is enabled, this path won't be invoked, the async path will be invoked instead. I did not remove the blocking path in the code, the control is from the config.

d2/src/main/java/com/linkedin/d2/balancer/clients/BackupRequestsClient.java
```diff
@@ -195,24 +247,65 @@ public void restRequest(final RestRequest request, final RequestContext requestC
...
+    {
+      if (existing != serviceProperties.getBackupRequests())
+      { // reference inequality check
+        update(serviceName, serviceProperties.getBackupRequests());
+        _configs.put(serviceName, serviceProperties.getBackupRequests());
+      }
+    }
+  }
+
+  void getStrategy(final String serviceName, final String operation, Callback<Optional<TrackingBackupRequestsStrategy>> callback) {
```
ChaoLinkedIn, line 274: We now have 2 copies of getStrategy (one sync one async) with almost identical functionality. Can we just keep the async one? You can always wait the results if sync operation is needed.

If necessary, just keep requestAsync and let request to call into requestAsync.

d2/src/main/java/com/linkedin/d2/balancer/clients/BackupRequestsClient.java
```diff
@@ -195,24 +247,65 @@ public void restRequest(final RestRequest request, final RequestContext requestC
...
+      @Override
+      public void onSuccess(None result) {
+        callback.onSuccess(getStrategyAfterUpdate(serviceName, operation));
+      }
+    };
+
+    updateIfNeeded(serviceName, updatePropertiesCallback);
+  }
+
+  private void updateIfNeeded(String serviceName, Callback<None> callback)
```
ChaoLinkedIn, line 290: Ditto -- prefer to have one implementation.

d2/src/main/java/com/linkedin/d2/balancer/clients/BackupRequestsClient.java
rachelhanhan: I refactored the code a little bit more for better readability. Yes I have 2 copies of the same logic, because I want to be more cautious and reduce the risk. Most of the common logics have been extracted from the 2 copies though. When we want to clean up the blocking call later, it should be fairly easy to clean them up

ChaoLinkedIn (approved):
As discussed, a simpler workaround is bypass backup request when serviceProperties are not available. It's good though to have a real async implementation.

d2/src/main/java/com/linkedin/d2/balancer/clients/BackupRequestsClient.java
ssheng: Got it.

PR #416, opened 2020-09-21
Improve ParseResult runtime by getting sorted map at the end

JospehChiu:
**Background**
It took 18 seconds originally to generate data templates with 2098 input files. After https://github.com/linkedin/rest.li/pull/376 solving the non-deterministic build output issue, it takes 46 seconds to finish the gradle task generateDataTemplate. It's because DataSchemaParser recursively parsed sources to a TreeMap internally.

**Change Detail**
Change the internal data structure back to LinkedHashMap, and only generates TreeMap on demand to improve the runtime of generateDataTemplate. It will take 24 seconds for 2098 input files.

**Testing**
Covered by unit tests added in https://github.com/linkedin/rest.li/pull/376

### DIFF
--- a/generator/src/main/java/com/linkedin/pegasus/generator/DataSchemaParser.java
+++ b/generator/src/main/java/com/linkedin/pegasus/generator/DataSchemaParser.java
@@ -21,7 +21,6 @@ import com.linkedin.data.schema.DataSchemaLocation;
 import com.linkedin.data.schema.DataSchemaParserFactory;
 import com.linkedin.data.schema.DataSchemaResolver;
 import com.linkedin.data.schema.resolver.AbstractMultiFormatDataSchemaResolver;
-import com.linkedin.data.schema.resolver.ExtensionsDataSchemaResolver;
 import com.linkedin.data.schema.resolver.FileDataSchemaLocation;
 import com.linkedin.data.schema.resolver.InJarFileDataSchemaLocation;
 import com.linkedin.data.schema.resolver.MultiFormatDataSchemaResolver;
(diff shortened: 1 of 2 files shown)


Astro03:
@BrianPin  Does this keep the deterministic outputs still?

generator/src/main/java/com/linkedin/pegasus/generator/DataSchemaParser.java
```diff
@@ -210,8 +210,7 @@ private static ParseResult combine(Collection<ParseResult> parseResults)
   public static class ParseResult
   {
     private static final String EXTENSION_FILENAME_SUFFIX = "Extensions.pdl";
-    // The purpose of the sorting is to keep generated java meta classes consistent accross different input file orders
-    private final Map<DataSchema, DataSchemaLocation> _schemaAndLocations = new TreeMap<>(Comparator.comparing(DataSchema::toString));
+    private final Map<DataSchema, DataSchemaLocation> _schemaAndLocations = new LinkedHashMap<>();
```
junchuanwang, line 213: Add comment explaining why not using TreeMap directly.

junchuanwang (commented):
(1)Another usage in SchemaAnnotationValidatorCmdLineApp.java might also needs to be corrected to prevent behavior change.

(2) Besides please mention in the PR what tests you did

evanw555:
Thanks for the contribution, however I have a few concerns:
- Looking for usages of `ParseResult#getSchemaAndLocations` shows that it's usage in many more places than just the one reference you updated.
- Reverting the behavior of the existing method is backward incompatible, as users may already be relying on that method producing a sorted output.
- By introducing a replacement method, this may cause confusion for users who are using the existing method. They likely won't know this behavior's been moved to another method.

Suggestions:
- Can we just convert the `LinkedHashMap` to a `TreeMap` (as you're doing) on-demand in the existing method? We can improve the performance there by maybe using a field to cache the sorted map.

I'll discuss with you offline about this.

JospehChiu:
The issue will be addressed in #417

PR #417, opened 2020-09-22
Optimize schema parsing by reusing ParseResult instances

evanw555:
This is meant to replace PR #416 

PR #376 made schema parsing deterministic in part by sorting parser
results. When using `DataSchemaParser` to parse data schemas from multiple
sources, a separate `ParseResult` instance is constructed for each source
then ultimately combined into a new singular instance. This introduced a
performance regression when parsing from many sources due to unnecessary
sorting. This change fixes that by allowing `FileFormatDataSchemaParser`
to reuse the root instance rather than constructing a new one.

Also adds more comments for clarity.

### DIFF
--- a/generator/src/main/java/com/linkedin/pegasus/generator/FileFormatDataSchemaParser.java
+++ b/generator/src/main/java/com/linkedin/pegasus/generator/FileFormatDataSchemaParser.java
@@ -48,7 +48,8 @@ import java.util.jar.JarFile;
  * @author Keren Jin
  * @author Joe Betz
  */
-public class FileFormatDataSchemaParser {
+public class FileFormatDataSchemaParser
+{
   static final String SCHEMA_PATH_PREFIX = SchemaDirectoryName.PEGASUS.getName() + "/";
   static final String EXTENSION_PATH_ENTRY = SchemaDirectoryName.EXTENSIONS.getName() + "/";
   private final String _resolverPath;
@@ -62,10 +63,15 @@ public class FileFormatDataSchemaParser {
     _schemaParserFactory = schemaParserFactory;
   }
 
-  public DataSchemaParser.ParseResult parseSources(String sources[]) throws IOException
+  public DataSchemaParser.ParseResult parseSources(String[] sources) throws IOException
   {
     final DataSchemaParser.ParseResult result = new DataSchemaParser.ParseResult();
+    parseSources(sources, result);
+    return result;
+  }
 
+  void parseSources(String[] sources, DataSchemaParser.ParseResult result) throws IOException
+  {
     try
     {
       for (String source : sources)
@@ -132,8 +138,6 @@ public class FileFormatDataSchemaParser {
         final DataSchema schema = _schemaResolver.bindings().get(entry.getKey());
         result.getSchemaAndLocations().put(schema, entry.getValue());
       }
-
-      return result;
     }
     catch (RuntimeException e)
     {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,9 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
-- Include HttpStatus code while throwing IllegalArgumentException
+- Mitigate schema parsing performance regression introduced in `29.5.1` by reusing `ParseResult` instances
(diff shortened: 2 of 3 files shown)


generator/src/main/java/com/linkedin/pegasus/generator/FileFormatDataSchemaParser.java
```diff
@@ -62,10 +63,15 @@ public FileFormatDataSchemaParser(String resolverPath, DataSchemaResolver schema
...
 
-  public DataSchemaParser.ParseResult parseSources(String sources[]) throws IOException
+  public DataSchemaParser.ParseResult parseSources(String[] sources) throws IOException
   {
     final DataSchemaParser.ParseResult result = new DataSchemaParser.ParseResult();
+    parseSources(sources, result);
+    return result;
+  }
 
+  void parseSources(String[] sources, DataSchemaParser.ParseResult result) throws IOException
```
BrianPin, line 73: Should this be `public` since it is meant to be used by others

BrianPin (approved):
LGTM

generator/src/main/java/com/linkedin/pegasus/generator/FileFormatDataSchemaParser.java
nickibi: I think it is only used in DataSchemaParser, so package private looks like a right approach to me.
@BrianPin why do you think we need to make a new API of FileFormatDataSchemaParser?

generator/src/main/java/com/linkedin/pegasus/generator/FileFormatDataSchemaParser.java
```diff
@@ -48,7 +48,8 @@
  * @author Keren Jin
  * @author Joe Betz
  */
-public class FileFormatDataSchemaParser {
+public class FileFormatDataSchemaParser
+{
   static final String SCHEMA_PATH_PREFIX = SchemaDirectoryName.PEGASUS.getName() + "/";
   static final String EXTENSION_PATH_ENTRY = SchemaDirectoryName.EXTENSIONS.getName() + "/";
```
nickibi, line 54: Do you mind removing this unused constant in your PR?

generator/src/main/java/com/linkedin/pegasus/generator/FileFormatDataSchemaParser.java
```diff
@@ -48,7 +48,8 @@
...
   private final String _resolverPath;
```
nickibi, line 55: The _resolverPath is not used anywhere, but It's an input of this class's constructor. I am thinking about shall we create a new constructor which remove the resolverPath as input. We could deprecate the current constructor later. But it is totally out of the scope of this PR, I just came to know it by reading this PR.

generator/src/main/java/com/linkedin/pegasus/generator/FileFormatDataSchemaParser.java
evanw555: It's package-private, so theoretically it can be referenced by a subclass. I can open a separate PR and label it as backward-incompatible.

generator/src/main/java/com/linkedin/pegasus/generator/FileFormatDataSchemaParser.java
evanw555: Interesting... I didn't notice that 🤔  I can make the new constructor.

generator/src/main/java/com/linkedin/pegasus/generator/FileFormatDataSchemaParser.java
evanw555: +1 on Yingjie's comment. It's not meant for public consumption, at least not yet. Keeping things initially restricted makes it easier to make changes later.

generator/src/main/java/com/linkedin/pegasus/generator/FileFormatDataSchemaParser.java
evanw555: Nevermind, deprecations should be done in minor releases. I'll deprecate it in the other PR.

nickibi (approved):
Thanks for fixing the CHANGELOG:)

karthikbalasub (approved):

PR #418, opened 2020-09-22
Deprecate old constructor in FileFormatDataSchemaParser

evanw555:
Follow-up to #417


nickibi (approved):
Thanks for fixing the tech debt.

evanw555:
Looks like the interfaces to remove have been removed since this PR was first opened. This is no longer backward-incompatible, as it just deprecates a constructor.

nickibi:
> Looks like the interfaces to remove have been removed since this PR was first opened. This is no longer backward-incompatible, as it just deprecates a constructor.

Then we can release it now?

PR #419, opened 2020-09-22
Optimize logger in degrader load balancer

rachelhanhan:
There are complaints from service owners that DegraderImpl LoggerFactory.getLogger is taking long time, and the call occurs very frequently that may impact client side performance. After looking into the code, I found that whenever we instantiate a new ImmutableConfig in DegraderImpl, we will call LoggerFactory.getLogger once to get the default value for the logger. This change will create a static variable for the default logger.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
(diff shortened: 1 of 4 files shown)


nizarm (approved):
LGTM, thank you for the quick change

ChaoLinkedIn (approved):
Nice optimization!

PR #421, opened 2020-09-22
Include the given http code in the IllegalArgumentException

nizarm:


ChaoLinkedIn (approved):
LGTM. Please also get it reviewed by ServiceFoundation team. Sounds like _lookup table needs to update.

PR #425, opened 2020-09-24
Add checkPegasusSchemaSnapshotTask

nickibi:
checkPegasusSchemaSnapshotTask is used to check the pegasus schema's compatibility.
The check is based on the snapshots(.pdl files) which are generated in generatePegasusSchemaSnapshotTask.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
+- Add `checkPegasusSchemaSnapshot` task. 
+   - The task will be used to check any pegasus schema compatible and incompatible changes.
+   - The pegasus schema may or may not be part of a rest.li resource.
+   - The task will be triggered at build time, if user provides gradle property: "pegasusPlugin.enablePegasusSchemaCompatibilityCheck=true".
 
 ## [29.7.2] - 2020-09-25
 - Move from lambdas to explicit change listeners since lambda garbage collection is unreliable in Java
--- a/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
+++ b/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
@@ -58,6 +58,7 @@ import org.gradle.api.artifacts.Configuration;
 import org.gradle.api.artifacts.ConfigurationContainer;
 import org.gradle.api.file.FileCollection;
 import org.gradle.api.plugins.JavaBasePlugin;
+import org.gradle.api.plugins.JavaPlatformPlugin;
 import org.gradle.api.plugins.JavaPlugin;
 import org.gradle.api.plugins.JavaPluginConvention;
 import org.gradle.api.tasks.Copy;
@@ -538,6 +539,10 @@ public class PegasusPlugin implements Plugin<Project>
   public static final String SNAPSHOT_FILE_SUFFIX = ".snapshot.json";
   public static final String SNAPSHOT_COMPAT_REQUIREMENT = "rest.model.compatibility";
   public static final String IDL_COMPAT_REQUIREMENT = "rest.idl.compatibility";
+  // Pegasus schema compatibility level configuration, which is used to define the {@link CompatibilityLevel}.
+  public static final String PEGASUS_SCHEMA_SNAPSHOT_REQUIREMENT = "pegasusPlugin.pegasusSchema.compatibility";
+  // CompatibilityOptions Mode configuration, which is used to define the {@link CompatibilityOptions#Mode} in the compatibility checker.
+  private static final String PEGASUS_COMPATIBILITY_MODE = "pegasusPlugin.pegasusSchemaCompatibilityCheckMode";
 
   private static final Pattern TEST_DIR_REGEX = Pattern.compile("^(integ)?[Tt]est");
   private static final String SNAPSHOT_NO_PUBLISH = "rest.model.noPublish";
@@ -591,6 +596,12 @@ public class PegasusPlugin implements Plugin<Project>
   private static final String PEGASUS_EXTENSION_SCHEMA_SNAPSHOT_DIR_OVERRIDE = "overridePegasusExtensionSchemaSnapshotDir";
 
   private static final String SRC = "src";
+
+  private static final String SCHEMA_ANNOTATION_HANDLER_CONFIGURATION = "schemaAnnotationHandler";
+
+  private static final String COMPATIBILITY_OPTIONS_MODE_DATA = "DATA";
+
+  private static final String COMPATIBILITY_LEVEL_BACKWARDS = "BACKWARDS";
 
   @SuppressWarnings("unchecked")
   private Class<? extends Plugin<Project>> _thisPluginType = (Class<? extends Plugin<Project>>)
@@ -750,7 +761,7 @@ public class PegasusPlugin implements Plugin<Project>
     //
     // The configuration contains dependencies to schema annotation handlers which would process schema annotations
     // and validate.
-    Configuration schemaAnnotationHandler = configurations.maybeCreate("schemaAnnotationHandler");
+    Configuration schemaAnnotationHandler = configurations.maybeCreate(SCHEMA_ANNOTATION_HANDLER_CONFIGURATION);
 
     // configuration for publishing jars containing rest idl and generated client builders
     // to the project artifacts for including in the ivy.xml
@@ -883,10 +894,10 @@ public class PegasusPlugin implements Plugin<Project>
                 {
                   task.setInputDir(generateDataTemplatesTask.getInputDir());
                   task.setResolverPath(getDataModelConfig(project, sourceSet)); // same resolver path as generateDataTemplatesTask
-                  task.setClassPath(project.getConfigurations() .getByName("schemaAnnotationHandler")
+                  task.setClassPath(project.getConfigurations() .getByName(SCHEMA_ANNOTATION_HANDLER_CONFIGURATION)
                                            .plus(project.getConfigurations().getByName(PEGASUS_PLUGIN_CONFIGURATION))
-                                           .plus(project.getConfigurations().getByName("runtime")));
-                  task.setHandlerJarPath(project.getConfigurations() .getByName("schemaAnnotationHandler"));
+                                           .plus(project.getConfigurations().getByName(JavaPlatformPlugin.RUNTIME_CONFIGURATION_NAME)));
+                  task.setHandlerJarPath(project.getConfigurations() .getByName(SCHEMA_ANNOTATION_HANDLER_CONFIGURATION));
                   if (isPropertyTrue(project, ENABLE_ARG_FILE))
                   {
                     task.setEnableArgFile(true);
@@ -1449,6 +1460,7 @@ public class PegasusPlugin implements Plugin<Project>
   {
     File schemaDir = isExtensionSchema? project.file(getExtensionSchemaPath(project, sourceSet))
         : project.file(getDataSchemaPath(project, sourceSet));
+
     Path publishablePegasusSchemaSnapshotDir = project.getBuildDir().toPath().resolve(sourceSet.getName() +
             (isExtensionSchema ? PEGASUS_EXTENSION_SCHEMA_SNAPSHOT: PEGASUS_SCHEMA_SNAPSHOT));
 
@@ -1456,16 +1468,34 @@ public class PegasusPlugin implements Plugin<Project>
         isExtensionSchema ? PEGASUS_EXTENSION_SCHEMA_SNAPSHOT: PEGASUS_SCHEMA_SNAPSHOT, schemaDir,
         publishablePegasusSchemaSnapshotDir.toFile(), isExtensionSchema);
 
+    File pegasusSchemaSnapshotDir = project.file(isExtensionSchema ? getPegasusExtensionSchemaSnapshotPath(project, sourceSet)
+        : getPegasusSchemaSnapshotPath(project, sourceSet));
+    pegasusSchemaSnapshotDir.mkdirs();
+
     Task checkSchemaSnapshot = project.getTasks().create(sourceSet.getTaskName("check",
         isExtensionSchema ? PEGASUS_EXTENSION_SCHEMA_SNAPSHOT: PEGASUS_SCHEMA_SNAPSHOT),
         CheckPegasusSnapshotTask.class, task ->
         {
           task.dependsOn(generatePegasusSchemaSnapshot);
-          // TODO: update CheckPegasusSnapshotTask
-        });
+          task.setCurrentSnapshotDirectory(publishablePegasusSchemaSnapshotDir.toFile());
+          task.setPreviousSnapshotDirectory(pegasusSchemaSnapshotDir);
+          task.setCodegenClasspath(project.getConfigurations() .getByName(SCHEMA_ANNOTATION_HANDLER_CONFIGURATION)
+              .plus(project.getConfigurations().getByName(PEGASUS_PLUGIN_CONFIGURATION))
+              .plus(project.getConfigurations().getByName(JavaPlatformPlugin.RUNTIME_CONFIGURATION_NAME)));
+          task.setCompatibilityLevel(isExtensionSchema ? COMPATIBILITY_LEVEL_BACKWARDS
+              :PropertyUtil.findCompatLevel(project, FileCompatibilityType.PEGASUS_SCHEMA_SNAPSHOT));
+          task.setCompatibilityMode(isExtensionSchema ? COMPATIBILITY_OPTIONS_MODE_DATA :
+              PropertyUtil.findCompatMode(project, PEGASUS_COMPATIBILITY_MODE));
+          task.setExtensionSchema(isExtensionSchema);
+          task.setHandlerJarPath(project.getConfigurations() .getByName(SCHEMA_ANNOTATION_HANDLER_CONFIGURATION));
 
-    File pegasusSchemaSnapshotDir = new File(isExtensionSchema ? getPegasusExtensionSchemaSnapshotPath(project, sourceSet)
-        : getPegasusSchemaSnapshotPath(project, sourceSet));
+          task.onlyIf(t ->
+          {
+            String pegasusSnapshotCompatPropertyName = findProperty(FileCompatibilityType.PEGASUS_SCHEMA_SNAPSHOT);
+            return isExtensionSchema || !project.hasProperty(pegasusSnapshotCompatPropertyName) ||
+                !"off".equalsIgnoreCase((String) project.property(pegasusSnapshotCompatPropertyName));
+          });
+        });
 
     Task publishPegasusSchemaSnapshot = publishPegasusSchemaSnapshot(project, sourceSet,
         isExtensionSchema ? PEGASUS_EXTENSION_SCHEMA_SNAPSHOT: PEGASUS_SCHEMA_SNAPSHOT, checkSchemaSnapshot,
@@ -2123,6 +2153,9 @@ public class PegasusPlugin implements Plugin<Project>
       case IDL:
         property = IDL_COMPAT_REQUIREMENT;
         break;
+      case PEGASUS_SCHEMA_SNAPSHOT:
+        property = PEGASUS_SCHEMA_SNAPSHOT_REQUIREMENT;
+        break;
       default:
         throw new GradleException("No property defined for compatibility type " + type);
     }
@@ -2180,7 +2213,8 @@ public class PegasusPlugin implements Plugin<Project>
     });
   }
 
-  private Task generatePegasusSchemaSnapshot(Project project, SourceSet sourceSet, String taskName, File inputDir, File outputDir, boolean isExtensionSchema)
+  private Task generatePegasusSchemaSnapshot(Project project, SourceSet sourceSet, String taskName, File inputDir, File outputDir,
+      boolean isExtensionSchema)
   {
     return project.getTasks().create(sourceSet.getTaskName("generate", taskName),
         GeneratePegasusSnapshotTask.class, task ->
@@ -2189,7 +2223,7 @@ public class PegasusPlugin implements Plugin<Project>
           task.setResolverPath(getDataModelConfig(project, sourceSet).plus(project.files(getDataSchemaPath(project, sourceSet))));
           task.setClassPath(project.getConfigurations().getByName(PEGASUS_PLUGIN_CONFIGURATION));
           task.setPegasusSchemaSnapshotDestinationDir(outputDir);
-          task.setEnableArgFile(isExtensionSchema);
+          task.setExtensionSchema(isExtensionSchema);
           if (isPropertyTrue(project, ENABLE_ARG_FILE))
           {
             task.setEnableArgFile(true);
@@ -2206,6 +2240,7 @@ public class PegasusPlugin implements Plugin<Project>
           task.dependsOn(checkPegasusSnapshotTask);
           task.from(inputDir);
           task.into(outputDir);
+          task.onlyIf(t -> !SharedFileUtils.getSuffixedFiles(project, inputDir, PDL_FILE_SUFFIX).isEmpty());
         });
   }
 }
--- a/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/CheckPegasusSnapshotTask.java
+++ b/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/CheckPegasusSnapshotTask.java
@@ -1,32 +1,176 @@
 /*
-   Copyright (c) 2020 LinkedIn Corp.
-
-   Licensed under the Apache License, Version 2.0 (the "License");
-   you may not use this file except in compliance with the License.
-   You may obtain a copy of the License at
-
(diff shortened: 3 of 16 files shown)


restli-tools/src/test/java/com/linkedin/restli/tools/snapshot/check/TestPegasusSchemaSnapshotCompatibilityChecker.java
```diff
@@ -0,0 +1,83 @@
...
+  private final String FS = File.separator;
+  private String testDir = System.getProperty("testDir", new File("src/test").getAbsolutePath());
+  private String snapshotDir = testDir + FS + "pegasusSchemaSnapshot";
+
+  @Test
+  public void testCompatiblePegasusSchemaSnapshot()
+  {
+    PegasusSchemaSnapshotCompatibilityChecker checker = new PegasusSchemaSnapshotCompatibilityChecker();
+    checker.checkPegasusSchemaCompatibility(snapshotDir + FS + "BirthInfo.pdl", snapshotDir + FS + "compatibleSchemaSnapshot/BirthInfo.pdl",
+        CompatibilityLevel.EQUIVALENT, CompatibilityOptions.Mode.DATA);
```
BrianPin, line 41: Is it worth to also add CompatibilityLevel.OFF and CompatibilityLevel.IGNORE?

restli-tools/src/test/java/com/linkedin/restli/tools/snapshot/check/TestPegasusSchemaSnapshotCompatibilityChecker.java
```diff
@@ -0,0 +1,83 @@
...
+    CompatibilityInfoMap infoMap = checker.getInfoMap();
+    Assert.assertTrue(infoMap.isModelCompatible(CompatibilityLevel.EQUIVALENT));
+  }
+
+  @Test
+  public void testIncompatiblePegasusSchemaSnapshot()
```
BrianPin, line 47: I think it might be better to abstract test input out from the test logic.
One example is the "BirthInfo"

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
```diff
@@ -0,0 +1,297 @@
...
+    String[] cliArgs = cl.getArgs();
+
+    if (cliArgs.length < 2 || cliArgs.length %2 != 0)
+    {
+      _logger.error("Invalid arguments");
+      help();
+      System.exit(1);
+    }
+
+    final String compatLevelValue;
```
BrianPin, line 99: How about just `final String compatableLevel`

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
```diff
@@ -0,0 +1,297 @@
...
+      _logger.error("Cannot read resolver path: " + e.getMessage());
+      System.exit(1);
+    }
+
+    PegasusSchemaSnapshotCompatibilityChecker compatibilityChecker = new PegasusSchemaSnapshotCompatibilityChecker();
+    for (int i = 1; i < cliArgs.length; i += 2)
+    {
+      String prevSnapshot = cliArgs[i-1];
+      String currentSnapshot = cliArgs[i];
+      compatibilityChecker.checkPegasusSchemaCompatibility(prevSnapshot, currentSnapshot, compatLevel, compatMode);
```
BrianPin, line 158: Would it be cleaner if the action `checkPegasusSchemaCompatibility` return the `InfoMap` directly?
I mean : not use the check to have another call to get the info.

The reason being, if we not return info map directly, they are sort of internal state in the checker, that means you will need to get the info map before you can use the check to check again.

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
```diff
@@ -0,0 +1,297 @@
...
+      options.append(compatLevel.name().toLowerCase()).append("|");
+    }
+    options.replace(options.length() - 1, options.length(), ">");
+
+    return options.toString();
+  }
+
+  private static String listCompatModeOptions()
+  {
+    final StringBuilder options = new StringBuilder("<");
```
BrianPin, line 277: We can try to use StringJoiner to replace StringBuilder, that will make this code a bit cleaner because we don't have to add delimiter every time when a element is appended.

You can have it this way
```
StringJoiner joiner = new StringJoiner("|", "<", ">");
Stream.of(CompatibilityOptions.Mode.values()).forEach(StringJoiner::add);
return joiner.toString();
```

nickibi: Updated based on the suggestion

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
nickibi: Updated based on the suggestion

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
nickibi: Fixed

restli-tools/src/test/java/com/linkedin/restli/tools/snapshot/check/TestPegasusSchemaSnapshotCompatibilityChecker.java
nickibi: Used DataProvider instead

restli-tools/src/test/java/com/linkedin/restli/tools/snapshot/check/TestPegasusSchemaSnapshotCompatibilityChecker.java
nickibi: OFF should not do the compatibility check
Add an assertion in incompatibleTest for IGNORE.

BrianPin (approved):

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
```diff
@@ -0,0 +1,282 @@
...
+  {
+    CompatibilityMessage message = new CompatibilityMessage(new Object[]{}, impact, format, args);
+    _infoMap.addModelInfo(message);
+  }
+
+  private DataSchema parseSchema(File schemaFile) throws FileNotFoundException
+  {
+    DataSchemaResolver resolver = MultiFormatDataSchemaResolver.withBuiltinFormats(_resolverPath);
+    PdlSchemaParser parser = new PdlSchemaParser(resolver);
+    parser.parse(new FileInputStream(schemaFile));
```
karthikbalasub, line 239: It seems odd that you need to create a resolver for each file. Do you want to process the same schema multiple times if it is referenced by other schemas? Consider this scenario:

Prev:
A.pdl:
record A {
  b: B
}
B.pdl:
record B {
  f1: int
  f2: float
}

Current:
A.pdl:
record A {
  b: B
}
B.pdl:
record B {
  f1: int
  // f2 deleted
}

Now will the error be reported once for B.pdl or twice, for both A.pdl and B.pdl?
I know we report the error multiple times for resource compatibility checking, but do you think that is the best options for schema checker also?

restli-tools/src/main/java/com/linkedin/restli/tools/compatibility/CompatibilityInfoMap.java
```diff
@@ -166,6 +166,21 @@ public String createSummary()
     return summaryMessage.toString();
   }
 
+  public String createPegasusSchemaSummary()
```
karthikbalasub, line 169: Not used anywhere?

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/CheckPegasusSnapshotTask.java
```diff
@@ -1,32 +1,176 @@
...
+
+  public void setExtensionSchema(boolean isExtensionSchema)
+  {
+    _isExtensionSchema = isExtensionSchema;
+  }
+
+  @Classpath
+  public FileCollection getHandlerJarPath()
+  {
+    return _handlerJarPath;
```
karthikbalasub, line 169: unused?

karthikbalasub (changes requested):

restli-tools/src/main/java/com/linkedin/restli/tools/compatibility/CompatibilityInfoMap.java
nickibi: It is supposed to use to create summary report. When I edited code, we accidentally removed the usage. 
Updated.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/CheckPegasusSnapshotTask.java
nickibi: It will be used to in the following PR. I created it as for a place holder.

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
```diff
@@ -0,0 +1,282 @@
...
+    if (summary.length() > 0)
+    {
+      System.out.println(summary);
+    }
+
+    if (cl.hasOption("report"))
+    {
+      System.out.println(new CompatibilityReport(compatibilityChecker._infoMap, compatLevel).createReport());
+      System.exit(0);
+    }
```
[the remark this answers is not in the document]
nickibi: Umm, printing similar contents of reports is unnecessary. Summary should be good enough for report here.

I used stdout to leverage the existed "CompatibilityLogChecker" class to generate report.

In CheckPegasusSnapshotTask, we write the report into a summary file. Writing into a file here seems like redundant.

Maybe we could remove "report" in the cmd line app?

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
nickibi: Umm, I think based on the way we generated snapshot pdl we have to parse each file.
All the referenced schemas are constructed as inline records in the snapshot.
For example: 
snapshot: A.pdl will be look like following
record A {
  b: record B {
    f1: int 
    f2: float
  }
}

I don't think we could avoid it.

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
```diff
@@ -0,0 +1,282 @@
...
+    {
+      _logger.error("Cannot read resolver path: " + e.getMessage());
+      System.exit(1);
+    }
+
+    PegasusSchemaSnapshotCompatibilityChecker compatibilityChecker = new PegasusSchemaSnapshotCompatibilityChecker();
+    for (int i = 1; i < cliArgs.length; i += 2)
+    {
+      String prevSnapshot = cliArgs[i-1];
+      String currentSnapshot = cliArgs[i];
```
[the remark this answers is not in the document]
nickibi: That's a good point, I will update to use the root directory instead.

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
```diff
@@ -0,0 +1,282 @@
...
+  }
+
+  private static String listCompatModeOptions()
+  {
+    StringJoiner stringJoiner = new StringJoiner("|", "<", ">");
+    Stream.of(CompatibilityOptions.Mode.values()).forEach(e -> stringJoiner.add(e.name()));
+    return stringJoiner.toString();
+  }
+
+  private static void help()
```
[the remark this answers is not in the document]
nickibi: Updated to use directory as input instead of list of files.
Updated the help message accordingly.

restli-tools/src/main/java/com/linkedin/restli/tools/compatibility/CompatibilityInfoMap.java
nickibi: No more usage, deleted.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -538,6 +541,9 @@
   public static final String SNAPSHOT_FILE_SUFFIX = ".snapshot.json";
   public static final String SNAPSHOT_COMPAT_REQUIREMENT = "rest.model.compatibility";
   public static final String IDL_COMPAT_REQUIREMENT = "rest.idl.compatibility";
+  public static final String PEGASUS_EXTENSION_SCHEMA_SNAPSHOT_REQUIREMENT = "pegasusPlugin.pegasusExtensionSchema.compatibility";
+  public static final String PEGASUS_SCHEMA_SNAPSHOT_REQUIREMENT = "pegasusPlugin.pegasusSchema.compatibility";
+  private static final String PEGASUS_COMPATIBILITY_MODE = "pegasusPlugin.pegasusSchemaCompatibilityCheckMode";
```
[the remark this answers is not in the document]
nickibi: Added comments

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
```diff
@@ -0,0 +1,282 @@
...
+   * @param compatibilityLevel compatibility level
+   * @param compatMode compatibilityOptions mode which defines the compatibility check mode.
+   * @return CompatibilityInfoMap which contains information whether the given two files are compatible or not.
+   */
+  public CompatibilityInfoMap checkPegasusSchemaCompatibility(String prevPegasusSchemaPath, String currentPegasusSchemaPath,
+      CompatibilityLevel compatibilityLevel, CompatibilityOptions.Mode compatMode)
+  {
+    _infoMap.isCompatible(compatibilityLevel);
+
+    File preSchemaFile = new File(prevPegasusSchemaPath);
```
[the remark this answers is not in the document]
nickibi: Following we need to parse pdl file to DataSchema, for the DataSchema I called it prevSchema. So I used "prevSchemaFile" to distinguish those two objects. 
I could remove this local variable, just pass 'new File(prevPegasusSchemaPath)' into the parseSchema method to get DataSchema.

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
```diff
@@ -0,0 +1,282 @@
...
+    File currSchemaFile = new File(currentPegasusSchemaPath);
+
+    DataSchema preSchema = null;
+    try
+    {
+      preSchema = parseSchema(preSchemaFile);
+    }
+    catch(FileNotFoundException e)
+    {
+      constructCompatibilityMessage(CompatibilityMessage.Impact.OLD_READER_IGNORES_DATA,
```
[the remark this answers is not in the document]
nickibi: Updated.

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    DataSchema currSchema = null;
+    try
+    {
+      currSchema = parseSchema(currSchemaFile);
+    }
+    catch(FileNotFoundException e)
+    {
+      constructCompatibilityMessage(CompatibilityMessage.Impact.BREAKS_OLD_READER,
```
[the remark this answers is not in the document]
nickibi: Updated.

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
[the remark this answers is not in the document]
nickibi: Updated

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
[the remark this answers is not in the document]
nickibi: Discussed offline.
Write the report into a file. 
In CheckPegasusSchemaSnapshotTask, use CompatibilityLogChecker to check the report file for the compatibility info.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
+- Add `checkPegasusSchemaSnapshot` task.
```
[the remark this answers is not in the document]
nickibi: Update the description with details

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/CheckPegasusSnapshotTask.java
```diff
@@ -1,32 +1,176 @@
...
 @CacheableTask
 public class CheckPegasusSnapshotTask extends DefaultTask
 {
+  private File _currentSnapshotDirectory;
+  private File _previousSnapshotDirectory;
+  private String _compatibilityLevel;
+  private FileCollection _codegenClasspath;
+  private String _compatibilityMode;
+  private FileCollection _handlerJarPath;
+  private boolean _isExtensionSchema = false;
```
aman1309, line 46: since we might use pdl for idl's in 3.0 or for any other future usecase, recommend using 'FileCompatibilityType' instead of boolean flag.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1449,23 +1460,43 @@ protected void configurePegasusSchemaSnapshotGeneration(Project project, SourceS
...
+              PropertyUtil.findCompatLevel(project, FileCompatibilityType.PEGASUS_EXTENSION_SCHEMA_SNAPSHOT)
+              :PropertyUtil.findCompatLevel(project, FileCompatibilityType.PEGASUS_SCHEMA_SNAPSHOT));
+          task.setCompatibilityMode(isExtensionSchema ? COMPATIBILITY_OPTIONS_MODE_DATA :
+              PropertyUtil.findCompatMode(project, PEGASUS_COMPATIBILITY_MODE));
+          task.setExtensionSchema(isExtensionSchema);
+          task.setHandlerJarPath(project.getConfigurations() .getByName(SCHEMA_ANNOTATION_HANDLER_CONFIGURATION));
 
-    File pegasusSchemaSnapshotDir = new File(isExtensionSchema ? getPegasusExtensionSchemaSnapshotPath(project, sourceSet)
-        : getPegasusSchemaSnapshotPath(project, sourceSet));
+          task.onlyIf(t ->
```
aman1309, line 1493: can we unit test for gradle tasks?

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
```diff
@@ -0,0 +1,348 @@
...
+    CompatibilityLevel compatLevel = null;
+    if (cl.hasOption("cl"))
+    {
+      try
+      {
+        compatLevel = CompatibilityLevel.valueOf(cl.getOptionValue("cl").toUpperCase());
+      }
+      catch (IllegalArgumentException e)
+      {
+        _logger.error("Invalid compatibilityLevel");
```
aman1309, line 124: lets print option value and exception trace also.
Same for other places.

restli-tools/src/test/java/com/linkedin/restli/tools/snapshot/check/TestPegasusSchemaSnapshotCompatibilityChecker.java
aman1309: here, CompatibilityLevel.OFF should also not parse the schemas. can we add assert on method calls?

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/CheckPegasusSnapshotTask.java
nickibi: I don't think using Enum instead of a boolean is appropriate here. No matter how many use cases of pdl in the future, it is unrelated to the usage of this boolean flag here. isExtensionSchema boolean not only is used to detect input but also is used for annotation check in the following PR.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -538,6 +539,12 @@
   public static final String SNAPSHOT_FILE_SUFFIX = ".snapshot.json";
   public static final String SNAPSHOT_COMPAT_REQUIREMENT = "rest.model.compatibility";
   public static final String IDL_COMPAT_REQUIREMENT = "rest.idl.compatibility";
+  // Pegasus extension schema compatibility level configuration, which is used to define the {@link CompatibilityLevel}.
+  public static final String PEGASUS_EXTENSION_SCHEMA_SNAPSHOT_REQUIREMENT = "pegasusPlugin.pegasusExtensionSchema.compatibility";
+  // Pegasus schema compatibility level configuration, which is used to define the {@link CompatibilityLevel}.
+  public static final String PEGASUS_SCHEMA_SNAPSHOT_REQUIREMENT = "pegasusPlugin.pegasusSchema.compatibility";
+  // CompatibilityOptions Mode configuration, which is used to define the {@link CompatibilityOptions#Mode} in the compatibility checker.
+  private static final String PEGASUS_COMPATIBILITY_MODE = "pegasusPlugin.pegasusSchemaCompatibilityCheckMode";
```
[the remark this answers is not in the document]
nickibi: All those configurations are optional to provide different configurations. I don't think adding them into a enum is the right solution here.

karthikbalasub (changes requested):

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
```diff
@@ -0,0 +1,345 @@
...
+   */
+  private static List<String> getMatchingPrevAndCurrSnapshotPairs(String prevSnapshotDir, String currSnapshotDir)
+  {
+    Map<String, String> prevFilesMap = createMapFromFiles(prevSnapshotDir);
+    Map<String, String> currFilesMap = createMapFromFiles(currSnapshotDir);
+    List<String> filePairs = new ArrayList<>();
+
+    currFilesMap.forEach((filename, absolutePath) ->
+    {
+      if (currFilesMap.containsKey(filename))
```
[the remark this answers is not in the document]
nickibi: Oops, it should be prevFilesMap.
I would make this method as package private for adding unit test.

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
```diff
@@ -0,0 +1,345 @@
...
+   * @return filesMap Map<String, String>
+   */
+  private static Map<String, String> createMapFromFiles(String snapshotFileDir)
+  {
+    try (Stream<Path> paths = Files.walk(Paths.get(snapshotFileDir)))
+    {
+      return paths
+          .filter(path -> path.toString().endsWith(PDL))
+          .map(path -> path.toFile())
+          .collect(Collectors.toMap(File::getName, File:: getAbsolutePath, (first, second) -> first));
```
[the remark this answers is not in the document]
nickibi: We use FQN as pdl snapshot file name, which includes the namespace.
I will add unit tests.

karthikbalasub (approved):

PR #431, opened 2020-10-01
Fix task caching issue by using the output file to decide if models and spec are equivalent.

karthikbalasub:
Using the property from the task doesn't work when the task is loaded
from cache. The right solution is to read the summary file which
contains the compatibility check report.


gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1374,58 +1373,21 @@ protected void configureRestModelGeneration(Project project, SourceSet sourceSet
...
-                (
-                   !isPropertyTrue(project, SKIP_IDL_CHECK) &&
-                   (
-                      (isTaskSuccessful(checkRestModelTask) && !(checkRestModelTask.isRestSpecEquivalent())) ||
-                      (isTaskSuccessful(checkIdlTask) && !(checkIdlTask.isEquivalent()))
-                   )
-                )
-              );
-            });
+                    !isPropertyTrue(project, IDL_NO_PUBLISH) &&
```
junchuanwang, line 1376: maybe at least logging whether the task is successful?
Personally I prefer to log all `executed`, `skipped` `uptodate` status.

karthikbalasub: Gradle already logs all these details.

junchuanwang (approved):
The changes right now seems to be correct to me.

Though a little difficult to imagine without a description of the original scenario that leads the bug.

nickibi (approved):

PR #433, opened 2020-10-02
Fix collectionMetadata missing link issue when collection count is 0

BrianPin:


restli-server/src/main/java/com/linkedin/restli/internal/server/util/RestUtils.java
```diff
@@ -126,6 +125,17 @@ public static CollectionMetadata buildMetadata(final URI requestUri,
 
       metadata.setLinks(links);
     }
+    else if (pagingContext.getCount() == 0)
+    {
+      // even when we are getting count = 0, we should honor that links
+      // is a required field in the CollectionMetadata record
+      Link selfLink = new Link();
+      selfLink.setRel("self");
```
karthikbalasub, line 133: I thought we already send "self" links when I suggested adding this. 
Since we don't do that for count >0, the consistent solution would be to return an empty links array when count ==0.
Other option is to add "self" for both cases (at the end when count >0). But this might increase the payload size.

BrianPin: Yeah, I looked back the ticket, the empty array is the common expectation

karthikbalasub (approved):

nickibi (approved):

PR #435, opened 2020-10-03
Extend checkPegasusSchemaSnapshot task to be able to check schema annotation compatibility

nickibi:
Extend checkPegasusSchemaSnapshot task to be enable to check schema annotation compatibility, if SchemaAnnotationHandler is provided.

1. Update SchemaAnnotationHandler interface to have new api - annotationCompatibilityCheck, which can be used to check the custom annotation compatibility check
2. Create a new SchemaVisitor - AnnotationCheckResolvedPropertiesVisitor, which is to get a map of node in a schema to it's resolvedProperties. It can be used to check the annotation compatibility for olderResolvedProperties and newerResolvedProperties.
3. AnnotationCompatibilityChecker which leverages the annotation processor to check the compatibilities of annotation changes.

Test:
Unit tests
Manual test in sample repos.

### DIFF
--- a/data/src/main/java/com/linkedin/data/schema/annotation/AnnotationCheckResolvedPropertiesVisitor.java
+++ b/data/src/main/java/com/linkedin/data/schema/annotation/AnnotationCheckResolvedPropertiesVisitor.java
@@ -1,0 +1,105 @@
+/*
+ * Copyright (c) 2020 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.linkedin.data.schema.annotation;
+
+import com.linkedin.data.schema.DataSchema;
+import com.linkedin.data.schema.DataSchemaTraverse;
+import com.linkedin.data.schema.NamedDataSchema;
+import com.linkedin.data.schema.PathSpec;
+import com.linkedin.data.schema.RecordDataSchema;
+import com.linkedin.data.schema.annotation.SchemaAnnotationHandler.CompatibilityCheckContext;
+import java.util.ArrayDeque;
+import java.util.HashMap;
+import java.util.Map;
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
+
+
+/**
+ * This visitor is used to get node in schema to it's resolvedProperties.
+ * The nodesToResolvedPropertiesMap will be used for schema annotation compatibility check
+ *
+ * @author Yingjie Bi
+ */
+public class AnnotationCheckResolvedPropertiesVisitor implements SchemaVisitor
+{
+  private Map<PathSpec, Pair<CompatibilityCheckContext, Map<String, Object>>> _nodeToResolvedPropertiesMap = new HashMap<>();
+
+  @Override
+  public void callbackOnContext(TraverserContext context, DataSchemaTraverse.Order order)
+  {
+    // Skip execute when order is POST_ORDER to avoid traverse the same node twice.
+    // Only execute this callback when order is PRE_ORDER.
+    if (order == DataSchemaTraverse.Order.POST_ORDER)
+    {
+      return;
+    }
+
+    DataSchema currentSchema = context.getCurrentSchema();
+
+    ArrayDeque<String> pathToSchema = context.getSchemaPathSpec().clone();
+    pathToSchema.addFirst(((NamedDataSchema)context.getTopLevelSchema()).getName());
+    PathSpec pathSpec = new PathSpec(pathToSchema);
+
+    CompatibilityCheckContext compatibilityCheckContext = new CompatibilityCheckContext();
+
+    compatibilityCheckContext.setCurrentDataSchema(currentSchema);
+    compatibilityCheckContext.setSchemaField(context.getEnclosingField());
+    compatibilityCheckContext.setUnionMember(context.getEnclosingUnionMember());
+    compatibilityCheckContext.setPathSpecToSchema(pathSpec);
+
+    // If there is no resolvedProperties but properties, used the properties for annotation check.
+    Map<String, Object> properties;
+    if (context.getEnclosingField() != null)
+    {
+      RecordDataSchema.Field field = context.getEnclosingField();
+      properties = chooseProperties(field.getResolvedProperties(), field.getProperties());
+    }
+    else if (context.getEnclosingUnionMember() != null)
+    {
+      // There is no resolvedProperty for unionMember
+      properties = context.getEnclosingUnionMember().getProperties();
+    }
+    else
+    {
+      properties = chooseProperties(currentSchema.getResolvedProperties(), currentSchema.getProperties());
+    }
+
+    _nodeToResolvedPropertiesMap.put(pathSpec, new ImmutablePair<>(compatibilityCheckContext, properties));
+  }
+
+  @Override
+  public VisitorContext getInitialVisitorContext()
+  {
+    return new VisitorContext(){};
+  }
+
+  @Override
+  public SchemaVisitorTraversalResult getSchemaVisitorTraversalResult()
+  {
+    return null;
+  }
+
+  public Map<PathSpec, Pair<CompatibilityCheckContext, Map<String, Object>>>  getNodeToResolvedPropertiesMap()
+  {
+    return _nodeToResolvedPropertiesMap;
+  }
+
+  private Map<String, Object> chooseProperties(Map<String, Object> preferredProperties, Map<String, Object> fallbackProperties)
+  {
+    return preferredProperties.isEmpty() ? fallbackProperties : preferredProperties;
+  }
+}
--- a/data/src/main/java/com/linkedin/data/schema/annotation/IdentitySchemaVisitor.java
+++ b/data/src/main/java/com/linkedin/data/schema/annotation/IdentitySchemaVisitor.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright (c) 2020 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.linkedin.data.schema.annotation;
+
+import com.linkedin.data.schema.DataSchemaTraverse;
+
+
+/**
+ * This SchemaVisitor is used when annotation override is not needed.
+ * It will not traverse the data schema.
+ *
+ * @author Yingjie Bi
+ */
+public class IdentitySchemaVisitor implements SchemaVisitor
+{
+
+  @Override
+  public void callbackOnContext(TraverserContext context, DataSchemaTraverse.Order order) {
+    // skip post order traverse
+    if (order == DataSchemaTraverse.Order.POST_ORDER)
+    {
+      return;
+    }
+
+    // If current schema is a root node, set shouldContinue to false to avoid traverse schema.
+    if (context.getParentSchema() == null)
+    {
+      context.setShouldContinue(false);
+    }
+  }
+
+  @Override
+  public VisitorContext getInitialVisitorContext()
+  {
+    return new VisitorContext(){};
+  }
+
+  @Override
+  public SchemaVisitorTraversalResult getSchemaVisitorTraversalResult()
+  {
+    return new SchemaVisitorTraversalResult();
+  }
+}
--- a/data/src/main/java/com/linkedin/data/schema/annotation/SchemaAnnotationHandler.java
+++ b/data/src/main/java/com/linkedin/data/schema/annotation/SchemaAnnotationHandler.java
@@ -18,6 +18,11 @@ package com.linkedin.data.schema.annotation;
 import com.linkedin.data.message.Message;
 import com.linkedin.data.message.MessageList;
 import com.linkedin.data.schema.DataSchema;
+import com.linkedin.data.schema.PathSpec;
+import com.linkedin.data.schema.RecordDataSchema;
+import com.linkedin.data.schema.UnionDataSchema;
+import com.linkedin.data.schema.compatibility.CompatibilityMessage;
+import com.linkedin.data.schema.compatibility.CompatibilityResult;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -144,6 +149,122 @@ public interface SchemaAnnotationHandler
     return new PathSpecBasedSchemaAnnotationVisitor(this);
   }
 
+  /**
+   * This method is used to indicate whether the schema annotation handler implements the checkCompatibility method.
+   * @return false by default. Subclasses should override this method to return true, if they implement the checkCompatibility method.
+   */
+  default boolean implementsCheckCompatibility()
+  {
+    return false;
+  }
+
+  /**
+   * Check annotations changes are compatible or not.
+   * Also override implementsCheckCompatibility method to return true, after implementing this method.
+   *
+   * @param prevResolvedProperties the previous resolvedProperties.
+   *                               If an annotation is added on an existing field or a new field is added with an annotation,
+   *                               prevResolvedProperties#(annotationSpace) == null.
+   * @param currResolvedProperties the current resolvedProperties.
+   *                               If a field is deleted or the field presents but the annotation is deleted,
+   *                               currResolvedProperties#get(annotationSpace) == null.
+   * @param prevContext the previous annotationCheck context.
+   *                    If an annotation is added on an existing field, preContext will contain pathSpec and field schema information.
+   *                    If a field is added with an annotation, preContext will not contain any schema information(dataSchema, schemaField, pathSpecToSchema etc).
+   * @param currContext the current annotationCheck context
+   *                    if an annotation is deleted, currContext will contain pathSpec and field schema information.
+   *                    if a field with an annotation is deleted, currContext will not contain any schema information(dataSchema, schemaField, pathSpecToSchema etc).
+   * @return {@link AnnotationCompatibilityResult}
+   */
+  default AnnotationCompatibilityResult checkCompatibility(Map<String, Object> prevResolvedProperties, Map<String, Object> currResolvedProperties,
+      CompatibilityCheckContext prevContext, CompatibilityCheckContext currContext)
+  {
+    return new AnnotationCompatibilityResult();
+  }
+
+  /**
+   * CompatibilityCheckContext which contains metadata information:
+   * dataSchema, schemaField, unionMember and pathSpecToSchema.
+   */
+  class CompatibilityCheckContext
+  {
+    DataSchema _currentSchema;
+    RecordDataSchema.Field _schemaField;
+    UnionDataSchema.Member _unionMember;
+    PathSpec _pathSpecToSchema;
+
+    public DataSchema getCurrentDataSchema()
+    {
+      return _currentSchema;
+    }
+
+    public void setCurrentDataSchema(DataSchema currentSchema)
+    {
+      _currentSchema = currentSchema;
+    }
+
+    public RecordDataSchema.Field getSchemaField()
+    {
+      return _schemaField;
+    }
+
+    public void setSchemaField(RecordDataSchema.Field schemaField)
+    {
+      _schemaField = schemaField;
+    }
+
+    public UnionDataSchema.Member getUnionMember()
+    {
+     return _unionMember;
+    }
+
+    public void setUnionMember(UnionDataSchema.Member unionMember)
+    {
+      _unionMember = unionMember;
+    }
+
+    public PathSpec getPathSpecToSchema()
+    {
+      return _pathSpecToSchema;
+    }
+
+    public void setPathSpecToSchema(PathSpec pathSpecToSchema)
+    {
+      _pathSpecToSchema = pathSpecToSchema;
+    }
+  }
+
+  /**
+   * AnnotationCompatibilityResult
+   * it contains a list of {@link CompatibilityMessage}
+   * CompatibilityMessage describes the change is compatible or not
+   */
+  class AnnotationCompatibilityResult implements CompatibilityResult
+  {
+    private final MessageList<CompatibilityMessage> _messages;
+
+    public AnnotationCompatibilityResult()
+    {
+      _messages = new MessageList<CompatibilityMessage>();
+    }
+
+    public void addMessage(CompatibilityMessage message)
+    {
+      _messages.add(message);
+    }
+
+    @Override
+    public Collection<CompatibilityMessage> getMessages()
+    {
+      return _messages;
+    }
+
+    @Override
+    public boolean isError()
+    {
+      return _messages.isError();
+    }
+  }
 
   /**
    * Result the {@link #resolve(List, ResolutionMetaData)} function should return after it is called
--- a/data/src/main/java/com/linkedin/data/schema/annotation/SchemaAnnotationProcessor.java
+++ b/data/src/main/java/com/linkedin/data/schema/annotation/SchemaAnnotationProcessor.java
@@ -72,7 +72,32 @@ public class SchemaAnnotationProcessor
   public static SchemaAnnotationProcessResult process(List<SchemaAnnotationHandler> handlers,
                                                       DataSchema dataSchema, AnnotationProcessOption options)
   {
+    return process(handlers, dataSchema, options, true);
+  }
 
+  /**
+   * This function creates  {@link DataSchemaRichContextTraverser} and use it to wrap {@link SchemaVisitor} to visit the {@link DataSchema}
+   *
+   * Note {@link SchemaAnnotationHandler}'s #resolve() and #validate() function are supposed to be called by {@link SchemaVisitor}
+   *
+   * For the given {@link DataSchema}, it will first invoke each {@link SchemaAnnotationHandler#resolve}
+   * by using the {@link SchemaVisitor} returned by {@link SchemaAnnotationHandler#getVisitor()}
+   *
+   * then it uses {@link SchemaAnnotationValidationVisitor} to invoke each {@link SchemaAnnotationHandler#validate} to validate resolved schema annotation.
+   * users would skip this validation, by passing skipValidateProcessResult as true;
+   *
+   * It will abort in case of unexpected exceptions.
+   * Otherwise will aggregate error messages after all handlers' processing, to the final {@link SchemaAnnotationProcessResult}
+   *
+   * @param handlers the handlers that can resolve the annotation on the dataSchema and validate them
+   * @param dataSchema the dataSchema to be processed
+   * @param options additional options to help schema annotation processing
+   * @param validateProcessResult if it is true, validate {@link SchemaAnnotationProcessResult} .
+   * @return result after process
+   */
+  public static SchemaAnnotationProcessResult process(List<SchemaAnnotationHandler> handlers,
+      DataSchema dataSchema, AnnotationProcessOption options, boolean validateProcessResult)
+  {
     SchemaAnnotationProcessResult processResult = new SchemaAnnotationProcessResult();
     // passed in dataSchema is not changed after processing, this variable stores dynamically constructed dataSchema after each handler.
     processResult.setResultSchema(dataSchema);
@@ -94,7 +119,7 @@ public class SchemaAnnotationProcessor
       catch (Exception e)
       {
         throw new IllegalStateException(String.format("Annotation resolution processing failed at \"%s\" handler",
-                                                      schemaAnnotationHandler.getAnnotationNamespace()), e);
+            schemaAnnotationHandler.getAnnotationNamespace()), e);
       }
       SchemaVisitorTraversalResult handlerTraverseResult = visitor.getSchemaVisitorTraversalResult();
       if (!handlerTraverseResult.isTraversalSuccessful())
@@ -102,7 +127,7 @@ public class SchemaAnnotationProcessor
         hasResolveError = true;
         String errorMsgs = handlerTraverseResult.formatToErrorMessage();
         errorMsgBuilder.append(String.format("Annotation processing encountered errors during resolution in \"%s\" handler. \n",
-                                             schemaAnnotationHandler.getAnnotationNamespace()));
+            schemaAnnotationHandler.getAnnotationNamespace()));
         errorMsgBuilder.append(errorMsgs);
       }
 
@@ -125,36 +150,43 @@ public class SchemaAnnotationProcessor
       return processResult;
     }
 
-    // validate
-    boolean hasValidationError = false;
-    for (SchemaAnnotationHandler schemaAnnotationHandler: handlers)
+    if (validateProcessResult)
     {
-      LOG.debug("DEBUG:  starting validating using \"{}\" handler", schemaAnnotationHandler.getAnnotationNamespace());
-      SchemaAnnotationValidationVisitor validationVisitor = new SchemaAnnotationValidationVisitor(schemaAnnotationHandler);
-      DataSchemaRichContextTraverser traverserBase = new DataSchemaRichContextTraverser(validationVisitor);
-      try {
-        traverserBase.traverse(processResult.getResultSchema());
-      }
-      catch (Exception e)
+      // validate
+      boolean hasValidationError = false;
+      for (SchemaAnnotationHandler schemaAnnotationHandler: handlers)
       {
-        throw new IllegalStateException(String.format("Annotation validation failed in \"%s\" handler.",
-                                                      schemaAnnotationHandler.getAnnotationNamespace()), e);
+        LOG.debug("DEBUG:  starting validating using \"{}\" handler", schemaAnnotationHandler.getAnnotationNamespace());
+        SchemaAnnotationValidationVisitor validationVisitor = new SchemaAnnotationValidationVisitor(schemaAnnotationHandler);
+        DataSchemaRichContextTraverser traverserBase = new DataSchemaRichContextTraverser(validationVisitor);
+        try {
+          traverserBase.traverse(processResult.getResultSchema());
+        }
+        catch (Exception e)
+        {
+          throw new IllegalStateException(String.format("Annotation validation failed in \"%s\" handler.",
+              schemaAnnotationHandler.getAnnotationNamespace()), e);
+        }
+        SchemaVisitorTraversalResult handlerTraverseResult = validationVisitor.getSchemaVisitorTraversalResult();
+        if (!handlerTraverseResult.isTraversalSuccessful())
+        {
+          hasValidationError = true;
+          String errorMsgs = handlerTraverseResult.formatToErrorMessage();
+          errorMsgBuilder.append(String.format("Annotation validation process failed in \"%s\" handler. \n",
+              schemaAnnotationHandler.getAnnotationNamespace()));
+          errorMsgBuilder.append(errorMsgs);
+        }
       }
-      SchemaVisitorTraversalResult handlerTraverseResult = validationVisitor.getSchemaVisitorTraversalResult();
-      if (!handlerTraverseResult.isTraversalSuccessful())
-      {
-        hasValidationError = true;
-        String errorMsgs = handlerTraverseResult.formatToErrorMessage();
-        errorMsgBuilder.append(String.format("Annotation validation process failed in \"%s\" handler. \n",
-                                             schemaAnnotationHandler.getAnnotationNamespace()));
-        errorMsgBuilder.append(errorMsgs);
-      }
+      processResult.setValidationSuccess(!hasValidationError);
+      processResult.setErrorMsgs(errorMsgBuilder.toString());
     }
-    processResult.setValidationSuccess(!hasValidationError);
-    processResult.setErrorMsgs(errorMsgBuilder.toString());
+    else
+    {
+      processResult.setValidationSuccess(true);
+    }
+
     return processResult;
   }
-
 
   /**
    * Util function to get the resolvedProperties of the field specified by the PathSpec from the dataSchema.
--- a/data/src/main/java/com/linkedin/data/schema/compatibility/AnnotationCompatibilityChecker.java
+++ b/data/src/main/java/com/linkedin/data/schema/compatibility/AnnotationCompatibilityChecker.java
@@ -1,0 +1,176 @@
+/*
+ * Copyright (c) 2020 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.linkedin.data.schema.compatibility;
+
+import com.linkedin.data.schema.DataSchema;
+import com.linkedin.data.schema.NamedDataSchema;
+import com.linkedin.data.schema.PathSpec;
+import com.linkedin.data.schema.annotation.DataSchemaRichContextTraverser;
+import com.linkedin.data.schema.annotation.AnnotationCheckResolvedPropertiesVisitor;
+import com.linkedin.data.schema.annotation.SchemaAnnotationHandler;
+import com.linkedin.data.schema.annotation.SchemaAnnotationHandler.CompatibilityCheckContext;
+import com.linkedin.data.schema.annotation.SchemaAnnotationHandler.AnnotationCompatibilityResult;
+import com.linkedin.data.schema.annotation.SchemaAnnotationProcessor;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import org.apache.commons.lang3.tuple.Pair;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ *  The annotation compatibility check is part of the annotation processing framework.
+ *  If users use annotation processor to resolve properties,
+ *  they may also provide a annotation compatibility method to define the way to check annotation's compatibility.
+ *  In this checker, it will call the {@link SchemaAnnotationHandler#checkCompatibility} method.
+ *
+ * @author Yingjie Bi
+ */
+public class AnnotationCompatibilityChecker
+{
+  private static final Logger _log = LoggerFactory.getLogger(AnnotationCompatibilityChecker.class);
+
+  /**
+   * Check the pegasus schema annotation compatibility
+   * Process prevSchema and currSchema in the SchemaAnnotationProcessor to get the resolved result with resolvedProperties.
+   * then using the resolvedProperties to do the annotation compatibility check.
+   * @param prevSchema previous data schema
+   * @param currSchema current data schema
+   * @param handlers SchemaAnnotationHandler list
+   * @return List<AnnotationCompatibilityResult>
+   */
+  public static List<AnnotationCompatibilityResult> checkPegasusSchemaAnnotation(DataSchema prevSchema, DataSchema currSchema,
+      List<SchemaAnnotationHandler> handlers)
+  {
+    // Update handler list to only contain handlers with implementation of checkCompatibility.
+    handlers = handlers
+        .stream()
+        .filter( h -> h.implementsCheckCompatibility())
+        .collect(Collectors.toList());
+    SchemaAnnotationProcessor.SchemaAnnotationProcessResult prevSchemaResult = processSchemaAnnotation(prevSchema, handlers);
+    SchemaAnnotationProcessor.SchemaAnnotationProcessResult currSchemaResult = processSchemaAnnotation(currSchema, handlers);
+    Map<PathSpec, Pair<CompatibilityCheckContext, Map<String, Object>>> prevResolvedPropertiesMap
+        = getNodeToResolvedProperties(prevSchemaResult);
+    Map<PathSpec, Pair<CompatibilityCheckContext, Map<String, Object>>> currResolvedPropertiesMap
+        = getNodeToResolvedProperties(currSchemaResult);
+
+    return getCompatibilityResult(prevResolvedPropertiesMap, currResolvedPropertiesMap, handlers);
+  }
+
+  /**
+   * Iterate the nodeToResolverPropertiesMap, if a node's resolvedProperty contains the same annotationNamespace as SchemaAnnotationHandler,
+   * calling annotationCompatibilityCheck api which is provided in the SchemaAnnotationHandler to do the annotation compatibility check.
+   */
+  private static List<AnnotationCompatibilityResult> getCompatibilityResult(Map<PathSpec, Pair<CompatibilityCheckContext,
+      Map<String, Object>>> prevResolvedPropertiesMap, Map<PathSpec, Pair<CompatibilityCheckContext,
+      Map<String, Object>>> currResolvedPropertiesMap, List<SchemaAnnotationHandler> handlers)
+  {
+    List<AnnotationCompatibilityResult> results = new ArrayList<>();
+
+    prevResolvedPropertiesMap.forEach((pathSpec, prevCheckContextAndResolvedProperty) ->
+    {
+      Map<String, Object> prevResolvedProperties = prevCheckContextAndResolvedProperty.getValue();
+      handlers.forEach(handler ->
+      {
+        String annotationNamespace = handler.getAnnotationNamespace();
+        if (currResolvedPropertiesMap.containsKey(pathSpec))
+        {
+          // If previous schema node and current schema node have the same pathSpec,
+          // they may or may not contain the same annotation namespace as SchemaAnnotationHandler, we need to check further.
+          Pair<CompatibilityCheckContext, Map<String, Object>>
+              currCheckContextAndResolvedProperty = currResolvedPropertiesMap.get(pathSpec);
+          Map<String, Object> currResolvedProperties = currCheckContextAndResolvedProperty.getValue();
+
+          // If prevResolvedProperties or currResolvedProperties contains the same namespace as the provided SchemaAnnotationHandler,
+          // we do the annotation check.
+          if (prevResolvedProperties.containsKey(annotationNamespace) || currResolvedProperties.containsKey(
+              annotationNamespace))
+          {
+            AnnotationCompatibilityResult result =
+                handler.checkCompatibility(prevResolvedProperties, currResolvedProperties,
+                    prevCheckContextAndResolvedProperty.getKey(), currCheckContextAndResolvedProperty.getKey());
+            results.add(result);
+          }
+        }
+        else
+        {
+          if (prevResolvedProperties.containsKey(annotationNamespace))
+          {
+            // prevResolvedPropertiesMap has a pathSpec which the newResolvedPropertiesMap does not have,
+            // it means an existing field is removed.
+            // pass an empty currResolvedPropertiesMap and empty currAnnotationContext to the annotation check.
+            AnnotationCompatibilityResult result =
+                handler.checkCompatibility(prevResolvedProperties, new HashMap<>(),
+                    prevCheckContextAndResolvedProperty.getKey(), new CompatibilityCheckContext());
+            results.add(result);
+          }
+        }
+      });
+      if (currResolvedPropertiesMap.containsKey(pathSpec))
+      {
+        currResolvedPropertiesMap.remove(pathSpec);
+      }
+    });
+
+    currResolvedPropertiesMap.forEach((pathSpec, currCheckContextAndResolvedProperty) ->
+    {
+      handlers.forEach(handler ->
+      {
+        String annotationNamespace = handler.getAnnotationNamespace();
+        Map<String, Object> currResolvedProperties = currCheckContextAndResolvedProperty.getValue();
+        if (currResolvedPropertiesMap.containsKey(annotationNamespace))
+        {
+          // currResolvedPropertiesMap has a PathSpec which the prevResolvedPropertiesMap does not have,
+          // it means there is a new field with new annotations,
+          // pass an empty prevResolvedPropertiesMap and empty prevAnnotationContext to the annotation check.
+          AnnotationCompatibilityResult result = handler.checkCompatibility(new HashMap<>(), currResolvedProperties,
+              new CompatibilityCheckContext(), currCheckContextAndResolvedProperty.getKey());
+          results.add(result);
+        }
+      });
+    });
+    return results;
+  }
+
+  private static Map<PathSpec, Pair<CompatibilityCheckContext, Map<String, Object>>> getNodeToResolvedProperties(
+      SchemaAnnotationProcessor.SchemaAnnotationProcessResult result)
+  {
+    AnnotationCheckResolvedPropertiesVisitor visitor = new AnnotationCheckResolvedPropertiesVisitor();
+    DataSchemaRichContextTraverser traverser = new DataSchemaRichContextTraverser(visitor);
+    traverser.traverse(result.getResultSchema());
+    return visitor.getNodeToResolvedPropertiesMap();
+  }
+
+  private static SchemaAnnotationProcessor.SchemaAnnotationProcessResult processSchemaAnnotation(DataSchema dataSchema,
+      List<SchemaAnnotationHandler> handlers)
+  {
+    SchemaAnnotationProcessor.SchemaAnnotationProcessResult result =
+        SchemaAnnotationProcessor.process(handlers, dataSchema, new SchemaAnnotationProcessor.AnnotationProcessOption(), false);
+    // If any of the nameDataSchema failed to be processed, throw exception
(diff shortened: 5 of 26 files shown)


data/src/main/java/com/linkedin/data/schema/annotation/AnnotationCheckResolvedPropertiesVisitor.java
```diff
@@ -0,0 +1,86 @@
...
+
+    DataSchema currentSchema = context.getCurrentSchema();
+    Map<String, Object> resolvedProperties = currentSchema.getResolvedProperties();
+    if (resolvedProperties.isEmpty())
+    {
+      return;
+    }
+
+    RecordDataSchema.Field schemaField = context.getEnclosingField();
+    UnionDataSchema.Member unionDataMember = context.getEnclosingUnionMember();
```
junchuanwang, line 56: Can you rename this variable so it can be consistent with other naming (I see three naming, enclosingUnionMember, UnionMember, unionDataMember)

data/src/main/java/com/linkedin/data/schema/annotation/AnnotationCheckResolvedPropertiesVisitor.java
```diff
@@ -0,0 +1,86 @@
...
+
+    AnnotationCheckContext annotationCheckContext = new AnnotationCheckContext();
+
+    annotationCheckContext.setDataSchema(currentSchema);
+    annotationCheckContext.setSchemaField(schemaField);
+    annotationCheckContext.setUnionMember(unionDataMember);
+    annotationCheckContext.setPathToSchema(context.getSchemaPathSpec());
```
junchuanwang, line 63: Same here, there are two paths in context, you can rename it to pathSpecToSchema

data/src/main/java/com/linkedin/data/schema/annotation/SchemaAnnotationHandler.java
```diff
@@ -144,6 +149,93 @@ default SchemaVisitor getVisitor()
...
+   * @param olderResolvedProperties the older resolvedProperties
+   * @param newerResolvedProperties the newer resolvedProperties
+   * @param olderContext the older annotationCheck context
+   * @param newerContext the newer annotationCheck context
+   * @return AnnotationCompatibilityResult
+   */
+  default AnnotationCompatibilityResult annotationCompatibilityCheck(Map<String,Object> olderResolvedProperties, Map<String, Object> newerResolvedProperties,
+      AnnotationCheckContext olderContext, AnnotationCheckContext newerContext)
+  {
+    return new AnnotationCompatibilityResult();
```
junchuanwang, line 164: If we make it default method, that means every handler will return a `AnnotationCompatibilityResult`. Is there a way to tell whether or not this check happened? How can we distinguish the two cases

data/src/main/java/com/linkedin/data/schema/annotation/SchemaAnnotationHandler.java
```diff
@@ -144,6 +149,93 @@ default SchemaVisitor getVisitor()
...
+      _pathToSchema = pathToSchema.toArray(new String[0]);
+    }
+  }
+
+  /**
+   * AnnotationCompatibilityResult
+   * it contains a list of {@link CompatibilityMessage}
+   * CompatibilityMessage describes the change is compatible or not
+   */
+  class AnnotationCompatibilityResult implements CompatibilityResult
```
junchuanwang, line 219: Similar to my other comment, maybe we can add a flag to tell whether this check has been conducted or not.

data/src/main/java/com/linkedin/data/schema/compatibility/AnnotationCompatibilityChecker.java
```diff
@@ -0,0 +1,133 @@
...
+      Map<String, Object>>> newerResolvedPropertiesMap, List<SchemaAnnotationHandler> handlers)
+  {
+    List<AnnotationCompatibilityResult> results = new ArrayList<>();
+    for (SchemaAnnotationHandler handler: handlers)
+    {
+      olderResolvedPropertiesMap.entrySet().stream().forEach(e ->
+          {
+            String annotationNamespace = handler.getAnnotationNamespace();
+            PathSpec pathSpec = e.getKey();
+            if (newerResolvedPropertiesMap.containsKey(pathSpec))
```
junchuanwang, line 89: How does the handler handle these Else branch cases?  Maybe add some brief comments

data/src/main/java/com/linkedin/data/schema/compatibility/AnnotationCompatibilityChecker.java
```diff
@@ -0,0 +1,133 @@
...
+            {
+              Pair<SchemaAnnotationHandler.AnnotationCheckContext, Map<String, Object>> olderCheckContextAndResolvedProperty
+                  = olderResolvedPropertiesMap.get(pathSpec);
+              Pair<SchemaAnnotationHandler.AnnotationCheckContext, Map<String, Object>> newerCheckContextAndResolvedProperty
+                  = newerResolvedPropertiesMap.get(pathSpec);
+
+              Map<String, Object> olderResolvedProperties = olderCheckContextAndResolvedProperty.getValue();
+              Map<String, Object> newerResolvedProperties = newerCheckContextAndResolvedProperty.getValue();
+
+              if (olderResolvedProperties.containsKey(annotationNamespace) || newerResolvedProperties.containsKey(annotationNamespace))
```
junchuanwang, line 99: similarly, perhaps add some comments here

data/src/main/java/com/linkedin/data/schema/compatibility/AnnotationCompatibilityChecker.java
```diff
@@ -0,0 +1,133 @@
...
+
+  private static SchemaAnnotationProcessor.SchemaAnnotationProcessResult processSchemaAnnotation(DataSchema dataSchema,
+      List<SchemaAnnotationHandler> handlers)
+  {
+    SchemaAnnotationProcessor.SchemaAnnotationProcessResult result =
+        SchemaAnnotationProcessor.process(handlers, dataSchema, new SchemaAnnotationProcessor.AnnotationProcessOption());
+    // If any of the nameDataSchema failed to be processed, throw exception
+    if (result.hasError())
+    {
+      String schemaName = ((NamedDataSchema) dataSchema).getFullName();
```
junchuanwang, line 128: Add log before throwing exception

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/CheckPegasusSnapshotTask.java
```diff
@@ -173,4 +190,53 @@ public void setHandlerJarPath(FileCollection handlerJarPath)
...
+
+    for (File f : _handlerJarPath)
+    {
+      try
+      {
+        handlerJarPathUrls.add(f.toURI().toURL());
+      }
+      catch (MalformedURLException e)
+      {
+        new GradleException("Annotation compatibility check: could not get schema annotation handler jar " + f.getName() + " to url. " + e.getMessage());
```
junchuanwang, line 214: not throwing exception

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/CheckPegasusSnapshotTask.java
```diff
@@ -173,4 +190,53 @@ public void setHandlerJarPath(FileCollection handlerJarPath)
...
+    ClassLoader classLoader = new URLClassLoader(handlerJarPathUrls.toArray(new URL[handlerJarPathUrls.size()]),
+        getClass().getClassLoader());
+
+    try
+    {
+      _handlerClassNames = SchemaAnnotationHandlerClassUtil.getAnnotationHandlerClassNames(_handlerJarPath, classLoader, getProject());
+    }
+    catch (IOException e)
+    {
+      new GradleException("Annotation compatibility check: could not get annotation handler class name. " + e.getMessage());
```
junchuanwang, line 227: same, not throwing exception

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/CheckPegasusSnapshotTask.java
```diff
@@ -173,4 +190,53 @@ public void setHandlerJarPath(FileCollection handlerJarPath)
...
+    }
+
+    // For now, every schema annotation handler should be in its own module
```
junchuanwang, line 230: The logic is still similar to ValidateSchemaAnnotationTask, do you think we can extract the common part to not repeat yourself? Later if we change the enforcement of expected number of hanlders per module, we don't need to change on two places

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/CheckPegasusSnapshotTask.java
```diff
@@ -76,6 +84,15 @@ public void checkPegasusSnapshot()
       javaExecSpec.args("--report", reportOutput);
       javaExecSpec.args(_previousSnapshotDirectory);
       javaExecSpec.args(_currentSnapshotDirectory);
+      if (_isExtensionSchema)
+      {
+        javaExecSpec.args("--extensionSchema");
+      }
+      else if(hasSchemaAnnotationHandler())
+      {
+        javaExecSpec.args("--jarPath", _handlerJarPath);
```
junchuanwang, line 93: I will recommend to use longer name here, `--handlerJarPath` and `handlerClassName`. We don't need to be concise for this particular use case, but need to be explicit (for example later do a code search would be easier)

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
```diff
@@ -345,4 +402,53 @@ private static void help()
     }
     return null;
   }
+
+  private static List<SchemaAnnotationHandler> getAnnotationHandlers(String handlerJarPaths, String classNames)
```
junchuanwang, line 406: These code is still a lot similar to `SchemaAnnotationValidatorCmdLineApp`, should DRY

restli-tools/src/main/java/com/linkedin/restli/tools/compatibility/CompatibilityReport.java
```diff
@@ -55,11 +55,27 @@ public String createReport()
         .map(it -> "[MD-I]:" + it)
         .collect(Collectors.joining("\n"));
 
+    String annotationCompat = "";
+    String annotationIncompat = "";
+    String annotationIsCompat = "";
+    if(_infoMap.getAnnotationInfo(CompatibilityInfo.Level.INCOMPATIBLE).size() > 0 || _infoMap.getAnnotationInfo(CompatibilityInfo.Level.COMPATIBLE).size() > 0)
```
junchuanwang, line 61: Can you also post example error message to this PR ? like your local test?

Otherwise I was expecting to see it from some unit test.

restli-tools/src/main/java/com/linkedin/restli/tools/idlcheck/CompatibilityInfo.java
```diff
@@ -65,7 +65,9 @@
     PAGING_ADDED(Level.COMPATIBLE, "Method added paging support"),
     PAGING_REMOVED(Level.INCOMPATIBLE, "Method removed paging support"),
     SERVICE_ERROR_ADDED(Level.INCOMPATIBLE, "Service error \"%s\" now applies"),
-    SERVICE_ERROR_REMOVED(Level.COMPATIBLE, "Service error \"%s\" no longer applies");
+    SERVICE_ERROR_REMOVED(Level.COMPATIBLE, "Service error \"%s\" no longer applies"),
+    BREAK_OLD_CLIENTS(Level.INCOMPATIBLE, "Deleting an schema is incompatible change, it breaks old clients"),
+    ANNOTATION_INCOMPATIBLE_CHANGE(Level.INCOMPATIBLE, "%s");
```
junchuanwang, line 70: Just a thought: Unlike other message, Annotation incompatible  message doesn't have a message handler and description, which actually made it more difficult to be code-searched

restli-tools/src/main/java/com/linkedin/restli/tools/compatibility/CompatibilityInfoMap.java
```diff
@@ -296,4 +303,49 @@ public boolean addAll(CompatibilityInfoMap other)
...
+        case BREAKS_NEW_READER:
+          infoType = CompatibilityInfo.Type.TYPE_BREAKS_NEW_READER;
+          break;
+        case BREAKS_OLD_READER:
+          infoType = CompatibilityInfo.Type.TYPE_BREAKS_OLD_READER;
+          break;
+        case BREAKS_NEW_AND_OLD_READERS:
+          infoType = CompatibilityInfo.Type.TYPE_BREAKS_NEW_AND_OLD_READERS;
+          break;
+        case ANNOTATION_INCOMPATIBLE_CHANGE:
```
junchuanwang, line 326: I didn't understand why there is any other error message for annotation other than `ANNOTATION_INCOMPATIBLE_CHANGE` ?

data/src/main/java/com/linkedin/data/schema/compatibility/AnnotationCompatibilityChecker.java
nickibi: log and throw exception is anti-pattern, right ? https://stackoverflow.com/questions/6639963/why-is-log-and-throw-considered-an-anti-pattern

data/src/main/java/com/linkedin/data/schema/annotation/SchemaAnnotationHandler.java
nickibi: We would set up null as default value here to differentiate whether this method is conducted or not. But using null as default value is not ideal, I think adding a flag is a cleaner way to achieve it.

data/src/main/java/com/linkedin/data/schema/annotation/SchemaAnnotationHandler.java
nickibi: I will add a flag in this class.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/CheckPegasusSnapshotTask.java
nickibi: You are right, I will move this part to the SchemaAnnotationHandlerUtil.

restli-tools/src/main/java/com/linkedin/restli/tools/compatibility/CompatibilityInfoMap.java
nickibi: Since we do not provide annotation compatibility rules, customers provide the way how they want to check the annotations. There are some common sense rules like - change breaks new reader, change breaks old reader or change breaks new and old reader. The ANNOTATION_INCOMPATIBLE_CHANGE provide an option to users to define the way how they want to define the compatibility other than above. We provide the framework to support it.

restli-tools/src/main/java/com/linkedin/restli/tools/idlcheck/CompatibilityInfo.java
nickibi: This one is used for customer to define their custom compatibility rules, all the message info will be put here to replace "%s". Do you think we should provide something like - " Annotation incompatible change: "? Or maybe leave it as-is.

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
nickibi: Yes, moved this one to a util class which can be used by both 'SchemaAnnotationValidatorCmdLineApp' and 'PegasusSchemaSnapshotCompatibilityChecker'.

data/src/main/java/com/linkedin/data/schema/annotation/AnnotationCheckResolvedPropertiesVisitor.java
nickibi: In the AnnotationCheckContext, I used String[] _pathToSchema. The reason why I convert pathSpec to String[] paths, because CompatibilityMessage requires String[] paths as input, I do not want users to convert it, which is unnecessary. 
I feel pathToSchema makes more sense here, because it is not PathSpec.

data/src/main/java/com/linkedin/data/schema/annotation/AnnotationCheckResolvedPropertiesVisitor.java
nickibi: enclosingUnionMember name is specifically in the context of traversing schemas. Here to provide some context of schema, I think calling it "unionMember" makes more sense rather than "enclosingUnionMember".
I will update unionDataMember to unionMember.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/CheckPegasusSnapshotTask.java
nickibi: fixed

data/src/main/java/com/linkedin/data/schema/compatibility/AnnotationCompatibilityChecker.java
nickibi: added comments

restli-tools/src/main/java/com/linkedin/restli/tools/compatibility/CompatibilityReport.java
nickibi: Please check the following example:

[checker] [RS-COMPAT]: true

[checker] [MD-COMPAT]: true

[checker] [SCHEMA-ANNOTATION-COMPAT]: false

[checker] [SCHEMA-ANNOTATION-I]:distance: annotation check test, breaks new readers

[checker] [SCHEMA-ANNOTATION-I]:lastName: annotation check test, breaks new readers

**> Task :datarealm-restaurants-models:checkPegasusSchemaSnapshot** FAILED

FAILURE: Build failed with an exception.

- What went wrong:

Execution failed for task ':datarealm-restaurants-models:checkPegasusSchemaSnapshot'.

> There are incompatible changes, find details in /Users/ybi/workspace/datarealm-restaurants_trunk/build/datarealm-restaurants-models/reports/checkPegasusSchema/compatibilityReport.txt

data/src/main/java/com/linkedin/data/schema/compatibility/AnnotationCompatibilityChecker.java
nickibi: Updated the comments and also added logic to handle else branch cases.

data/src/main/java/com/linkedin/data/schema/compatibility/AnnotationCompatibilityChecker.java
nickibi: Added log

data/src/test/java/com/linkedin/data/schema/compatibility/TestAnnotationCompatibilityChecker.java
```diff
@@ -0,0 +1,207 @@
...
+      List<CompatibilityMessage> expectCompatibilityMessage = (List<CompatibilityMessage>) expectedResults.get(i).getMessages();
+      for (int j = 0; j < actualCompatibilityMessage.size(); j++)
+      {
+        Assert.assertEquals(actualCompatibilityMessage.get(j).toString(), expectCompatibilityMessage.get(j).toString());
+      }
+    }
+  }
+
+  @DataProvider
+  private Object[][] annotationCompatibilityCheckTestData()
```
BrianPin, line 59: Sorry for late review, 
A consideration for this function is to make it readable, and reusable.
Would it be possible if we can use some string formatted json to represent the all the datamap or properties.
And use something like DataMapUtil's function to read those json formated string to dataMap objects?

data/src/main/java/com/linkedin/data/schema/annotation/AnnotationCheckResolvedPropertiesVisitor.java
```diff
@@ -0,0 +1,86 @@
...
+public class AnnotationCheckResolvedPropertiesVisitor implements SchemaVisitor
+{
+  private Map<PathSpec, Pair<AnnotationCheckContext, Map<String, Object>>> _nodeToResolvedPropertiesMap = new HashMap<>();
+
+  @Override
+  public void callbackOnContext(TraverserContext context, DataSchemaTraverse.Order order)
+  {
+    if (order == DataSchemaTraverse.Order.POST_ORDER)
+    {
+      return;
```
BrianPin, line 45: Maybe document on why?

BrianPin: Why in terms of not execute the callback on POST_ORDER

data/src/main/java/com/linkedin/data/schema/annotation/AnnotationCheckResolvedPropertiesVisitor.java
```diff
@@ -0,0 +1,86 @@
...
+
+    Pair<AnnotationCheckContext, Map<String, Object>> annotationContextAndResolvedProperties =
+        new ImmutablePair<>(annotationCheckContext, resolvedProperties);
+
+    _nodeToResolvedPropertiesMap.put(
+          new PathSpec(context.getSchemaPathSpec()), annotationContextAndResolvedProperties);
+  }
+
+  @Override
+  public VisitorContext getInitialVisitorContext() {
```
BrianPin, line 73: style convention?

data/src/main/java/com/linkedin/data/schema/annotation/AnnotationCheckResolvedPropertiesVisitor.java
```diff
@@ -0,0 +1,86 @@
...
+    UnionDataSchema.Member unionMember = context.getEnclosingUnionMember();
+
+    AnnotationCheckContext annotationCheckContext = new AnnotationCheckContext();
+
+    annotationCheckContext.setDataSchema(currentSchema);
+    annotationCheckContext.setSchemaField(schemaField);
+    annotationCheckContext.setUnionMember(unionMember);
+    annotationCheckContext.setPathToSchema(context.getSchemaPathSpec());
+
+    Pair<AnnotationCheckContext, Map<String, Object>> annotationContextAndResolvedProperties =
```
BrianPin, line 65: Maybe we can move this statement that generates the annotationContextAndResolvedProperties object into the next line where it is the only place that it is being used?

data/src/main/java/com/linkedin/data/schema/compatibility/AnnotationCompatibilityChecker.java
```diff
@@ -0,0 +1,193 @@
...
+  /**
+   * Check the annotation compatibilities
+   * process olderSchema and newerSchema in the SchemaAnnotationProcessor to get the resolved result with resolvedProperties.
+   * then using the resolvedProperties to do the annotation compatibility check.
+   * @param olderSchema
+   * @param newerSchema
+   * @param handlers SchemaAnnotationHandler list
+   * @return List<AnnotationCompatibilityResult>
+   */
+  public static List<AnnotationCompatibilityResult> checkPegasusSchemaCompatibility(DataSchema olderSchema, DataSchema newerSchema,
```
BrianPin, line 58: This function name seems confusing, in the comment it is meant for annotation check not the schema itself

BrianPin: Please don't increase the length of `checkPegasusSchemaCompatibility` into `checkPegasusSchemaAnnoationaCompatibility`

data/src/main/java/com/linkedin/data/schema/annotation/SchemaAnnotationHandler.java
```diff
@@ -144,6 +149,105 @@ default SchemaVisitor getVisitor()
...
+  /**
+   * Check annotations changes are compatible or not
+   *
+   * @param olderResolvedProperties the older resolvedProperties
+   * @param newerResolvedProperties the newer resolvedProperties
+   * @param olderContext the older annotationCheck context
+   * @param newerContext the newer annotationCheck context
+   * @return AnnotationCompatibilityResult
+   */
+  default AnnotationCompatibilityResult annotationCompatibilityCheck(Map<String,Object> olderResolvedProperties, Map<String, Object> newerResolvedProperties,
```
BrianPin, line 161: Maybe we can have a new interface that specifically deals compatibility check.
My reason being the action to check compatibility maybe not needed for every schema annotation handler, so we create a new interface the inherit schema annotation handler, to add some more specific functions to schema annotation handler.

That way, you don't worry about using default here, and leave original schemaAnnotationHandler some flexibility.

nickibi: We had such discussion during the design phase. The main idea to put it into the SchemaAnnotationHandler is to complete the features of Schema Annotation Processor. In the handler, users provide how to resolve a custom annotation, how to validate a custom annotation. The compatibility check part was missing.

data/src/main/java/com/linkedin/data/schema/compatibility/AnnotationCompatibilityChecker.java
nickibi: Change name to 'checkPegasusSchemaAnnotation'. Later on I will add new method 'checkExtensionSchemaAnnotation' for extension annotation check.

data/src/main/java/com/linkedin/data/schema/annotation/AnnotationCheckResolvedPropertiesVisitor.java
nickibi: Fixed

data/src/main/java/com/linkedin/data/schema/annotation/AnnotationCheckResolvedPropertiesVisitor.java
nickibi: fixed

data/src/main/java/com/linkedin/data/schema/annotation/AnnotationCheckResolvedPropertiesVisitor.java
nickibi: In the DataSchemaRichContextTraverser. The schemaVisitor.callbackOnContext are triggered twice by default. First is PRE_ORDER, then is POST_ORDER. 
For the usecase here, we only need to get the schema context of each node, traversing once is enough. Therefore, when it's POST_ORDER, we skip it.

nickibi: Added comments.

data/src/test/java/com/linkedin/data/schema/compatibility/TestAnnotationCompatibilityChecker.java
nickibi: Discussed offline, update the test cases in a cleaner way.

BrianPin (commented):
LGTM

BrianPin (approved):

data/src/main/java/com/linkedin/data/schema/compatibility/AnnotationCompatibilityChecker.java
```diff
@@ -0,0 +1,193 @@
...
+          {
+            AnnotationCompatibilityResult result =
+                handler.annotationCompatibilityCheck(olderResolvedProperties, newerResolvedProperties,
+                    olderCheckContextAndResolvedProperty.getKey(), newerCheckContextAndResolvedProperty.getKey());
+            if (!result.isCheckAnnotation())
+            {
+              _log.info("There is no implementation of annotationCompatibilityCheck in SchemaAnnotationHandler: "
+                  + handler.getClass().getName());
+              break;
+            }
```
karthikbalasub, line 116: This logic feels odd. Did you consider adding a method on the handler directly?\
 eg,` boolean implementsCompatibilityChecking()`

data/src/main/java/com/linkedin/data/schema/compatibility/AnnotationCompatibilityChecker.java
```diff
@@ -0,0 +1,193 @@
...
+            results.add(result);
+          }
+          newerResolvedPropertiesMap.remove(pathSpec);
+        }
+        else
+        {
+          if (olderResolvedProperties.containsKey(annotationNamespace))
+          {
+            // olderResolvedPropertiesMap has a PathSpec which the newResolvedPropertiesMap does not have,
+            // it could be either newer schema removed annotation or the field is not existed, pass
```
karthikbalasub, line 126: or the field was deleted.

The logic here would be simple if you returned the empty resolveProperties instead if removing it (in AnnotationCheckResolvedPropertiesVisitor) 
So the only case to handle is deleted field. At this point, does it make sense to do annotation compatibility checking? Would the schema compatibility checking be enough?

data/src/main/java/com/linkedin/data/schema/compatibility/AnnotationCompatibilityChecker.java
```diff
@@ -0,0 +1,193 @@
...
+            // an empty newerResolvedPropertiesMap and empty newerAnnotationContext to the annotation check.
+            AnnotationCompatibilityResult result =
+                handler.annotationCompatibilityCheck(olderResolvedProperties, new HashMap<>(),
+                    olderCheckContextAndResolvedProperty.getKey(), new SchemaAnnotationHandler.AnnotationCheckContext());
+            if (!result.isCheckAnnotation())
+            {
+              _log.info("There is no implementation of annotationCompatibilityCheck in SchemaAnnotationHandler: "
+                  + handler.getClass().getName());
+              break;
+            }
```
karthikbalasub, line 136: this duplication can be avoided if we added a method on the handler to check.

data/src/main/java/com/linkedin/data/schema/compatibility/AnnotationCompatibilityChecker.java
```diff
@@ -0,0 +1,193 @@
...
+      for (Map.Entry<PathSpec, Pair<SchemaAnnotationHandler.AnnotationCheckContext, Map<String, Object>>> e : newerResolvedPropertiesMap
+          .entrySet())
+      {
+        Pair<SchemaAnnotationHandler.AnnotationCheckContext, Map<String, Object>>
+            newerCheckContextAndResolvedProperty = newerResolvedPropertiesMap.get(e.getKey());
+        Map<String, Object> newerResolvedProperties = newerCheckContextAndResolvedProperty.getValue();
+        if (newerResolvedProperties.containsKey(annotationNamespace))
+        {
+          // newerResolvedPropertiesMap has a PathSpec which the olderResolvedPropertiesMap does not have,
+          // it could be either newer schema add new annotations or adding new field with annotations, pass
```
karthikbalasub, line 150: either because newer schema aded new annotations or added new field with annotations

data/src/main/java/com/linkedin/data/schema/compatibility/AnnotationCompatibilityChecker.java
```diff
@@ -0,0 +1,193 @@
...
+          // an empty olderResolvedPropertiesMap and empty olderAnnotationContext to the annotation check.
```
karthikbalasub, line 151: How would the compatibility checker differentiate between a new field(compatible) vs adding new annotations(may not be compatible)?

I think we should invoke the checker only for new annotations. And for new fields skip annotation checking? 
Or, send empty context/map for the case where new annotations were added. and use null when a new field was added.

data/src/main/java/com/linkedin/data/schema/annotation/AnnotationCheckResolvedPropertiesVisitor.java
```diff
@@ -0,0 +1,87 @@
...
+    // Skip execute when order is POST_ORDER to avoid traverse same node twice.
+    // Only execute this callback when order is PRE_ORDER.
+    if (order == DataSchemaTraverse.Order.POST_ORDER)
+    {
+      return;
+    }
+
+    DataSchema currentSchema = context.getCurrentSchema();
+    Map<String, Object> resolvedProperties = currentSchema.getResolvedProperties();
+    if (resolvedProperties == null || resolvedProperties.isEmpty())
```
karthikbalasub, line 52: Leave the empty resolvedProperties for fields without any annotation. It would simplify the compatibility checking logic.

data/src/main/java/com/linkedin/data/schema/annotation/AnnotationCheckResolvedPropertiesVisitor.java
karthikbalasub: I think a better solution would be to update CompatibiliyMessage to accept PathSpec and convert to array if needed.
Then you can re-use the PathSpec in the map key below and avoid creating a new array for every node.

data/src/main/java/com/linkedin/data/schema/annotation/SchemaAnnotationHandler.java
```diff
@@ -144,6 +149,105 @@ default SchemaVisitor getVisitor()
...
+    }
+    public void setUnionMember(UnionDataSchema.Member unionMember)
+    {
+      _unionMember = unionMember;
+    }
+    public String[] getPathToSchema()
+    {
+      return _pathToSchema;
+    }
+    public void setPathToSchema(ArrayDeque<String> pathToSchema)
```
karthikbalasub, line 208: add extra newline between methods.

karthikbalasub (changes requested):
Sorry for the late review.

One general comment is to keep the naming consistent, use `prev` and `current` instead of `older` and `newer`

restli-tools/src/main/java/com/linkedin/restli/tools/compatibility/CompatibilityReport.java
junchuanwang: I meant Can you add this to the PR descritpion

data/src/main/java/com/linkedin/data/schema/annotation/AnnotationCheckResolvedPropertiesVisitor.java
junchuanwang: @nickibi CompatibilityMessage's `path` is not well defined (it could be anything, as long as a string array). Here you seems to be really storing the pathspec in the annotationCheckContext

restli-tools/src/main/java/com/linkedin/restli/tools/idlcheck/CompatibilityInfo.java
junchuanwang: I will agree with what you now proposed

restli-tools/src/main/java/com/linkedin/restli/tools/compatibility/CompatibilityInfoMap.java
junchuanwang: then do you think CUSTOM_ANNOTATION_INCOMPATIBLE_CHANGE will be a better definition?

junchuanwang:
One thing I found concerning is that `CompatibilityInfo` already contains types such as `ANNOTATION_CHANGE_BREAKS_NEW_SERVER` which actually become ambiguous after current annotation checking effort, because current change seems to be "schema annotation", to put in a more specific way.

junchuanwang (approved):

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
```diff
@@ -153,6 +178,27 @@ public static void main(String[] args) throws Exception
       compatMode = CompatibilityOptions.Mode.SCHEMA;
     }
 
+    if(cl.hasOption('e'))
```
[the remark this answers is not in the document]
nickibi: There is no value for 'extensionSchema', having this option itself is a flag.
In checkPegasusSnapshotTask:
if (_isExtensionSchema)
{
     javaExecSpec.args("--extensionSchema");
}
Do you thing it provides enough info?

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
```diff
@@ -230,6 +276,24 @@ public CompatibilityInfoMap checkPegasusSchemaCompatibility(String prevPegasusSc
...
+        if (!annotationResult.getMessages().isEmpty())
+        {
+          annotationResult.getMessages().forEach(message -> _infoMap.addAnnotation(message));
+        }
+      }
+    }
+
+    if (_checkExtensionSchemaAnnotation)
+    {
+      //TODO: extension schema annotation check
```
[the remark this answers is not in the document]
nickibi: Actually, I am thinking about using a different method in AnnotationCompatibilityChecker for extension schema check.
the extension schema does not require the annotation processor to process the schema, there is no property resolve/override requirements for extension schemas. We just need a SchemaHandler which provides the compatibility check logic for extension schema. Also we need a schema visitor to get pathSpec and properties. It will traverse the schema once, which reduces the unnecessary schema traverse(process annotation).

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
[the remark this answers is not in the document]
karthikbalasub: Nevermind, I was not sure how "hasOption" worked.

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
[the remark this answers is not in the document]
karthikbalasub: I don't understand how it would be better. You are going to create another branch for extensions which does the. same thing as the annotation checker.

If you have only the extension annotation handler, its visitor would traverse the schema once and get you the same result, right? The path override based traversal will happen only if handlers for annotations that support such overrides are present.

restli-tools/src/main/java/com/linkedin/restli/tools/compatibility/CompatibilityInfoMap.java
nickibi: Discussed offline, we will only have ANNOTATION_INCOMPATIBLE_CHANGE  here.

data/src/main/java/com/linkedin/data/schema/compatibility/AnnotationCompatibilityChecker.java
nickibi: add a method on the handler instead.

data/src/main/java/com/linkedin/data/schema/compatibility/AnnotationCompatibilityChecker.java
nickibi: add a method on the handler.

data/src/main/java/com/linkedin/data/schema/compatibility/AnnotationCompatibilityChecker.java
nickibi: Yes, I agree. 
I will return empty resolvedProperties in AnnotationCheckResolvedPropertiesVistor.
Then, we do not need this else branch check at all.

data/src/main/java/com/linkedin/data/schema/compatibility/AnnotationCompatibilityChecker.java
nickibi: Once we return empty resolvedProperties in AnnotationCheckResolvedPropertiesVistor. We would figure out adding new annotation on an existing field use case. Older schema and newer schema will have the same path specs. It will be handled in the previous check.
 
For this branch, only new field with new annotation use case will fall in to this check. 

My question is for new field with new annotation case, do we need to do the compatibility check?

karthikbalasub: My intuition says we can ignore the new field with new annotation case. Confirm with @junchuanwang as he has better context on annotations.

junchuanwang: Discussed offline with @nickibi, I recommend to add this check for new field with new annotation

data/src/main/java/com/linkedin/data/schema/annotation/SchemaAnnotationHandler.java
```diff
@@ -144,6 +149,105 @@ default SchemaVisitor getVisitor()
...
+   * it contains a list of {@link CompatibilityMessage}
+   * CompatibilityMessage describes the change is compatible or not
+   */
+  class AnnotationCompatibilityResult implements CompatibilityResult
+  {
+    private final MessageList<CompatibilityMessage> _messages;
+
+    private boolean _checkAnnotation = false;
+
+    public boolean isCheckAnnotation()
```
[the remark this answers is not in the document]
nickibi: Removed this method, because add a method to indicate implement compatibility check on the handler.

data/src/main/java/com/linkedin/data/schema/annotation/AnnotationCheckResolvedPropertiesVisitor.java
nickibi: Yes, that's right. 
Updated.

data/src/main/java/com/linkedin/data/schema/annotation/AnnotationCheckResolvedPropertiesVisitor.java
nickibi: Add a new constructor for CompatibiliyMessage to accept PathSpec.

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
```diff
@@ -87,7 +97,22 @@
...
+        .withDescription("path of the jars which contains the annotation handlers")
+        .create("jar"));
+    _options.addOption(OptionBuilder.withArgName("handler-classNames")
+        .withLongOpt("handlerClassName")
+        .hasArgs()
+        .withDescription("class names of the handlers")
+        .create("className"));
+    _options.addOption(OptionBuilder.withArgName("extensionSchema")
+        .withLongOpt("extensionSchema")
+        .withDescription("Check pegasus extension schema annotation")
```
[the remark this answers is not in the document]
nickibi: Updated the description.

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
```diff
@@ -153,6 +178,27 @@ public static void main(String[] args) throws Exception
...
+     try
+     {
+       _handlers = ClassJarPathUtil.getAnnotationHandlers(handlerJarPaths, classNames);
+     }
+     catch (IllegalStateException e)
+     {
+       _logger.error("Error while doing schema compatibility check, could not get SchemaAnnotationHandler classes: " + e.getMessage());
+     }
+
+      _checkPegasusSchemaAnnotation = _handlers.size() > 0;
```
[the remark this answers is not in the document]
nickibi: Updated.

restli-tools/src/main/java/com/linkedin/restli/tools/idlcheck/CompatibilityInfo.java
nickibi: Updated.

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
```diff
@@ -87,7 +97,22 @@
...
         .create("report"));
+    _options.addOption(OptionBuilder.withArgName("annotation_handler_jarPaths")
+        .withLongOpt("handlerJarPath")
+        .hasArgs()
+        .withDescription("path of the jars which contains the annotation handlers")
+        .create("jar"));
+    _options.addOption(OptionBuilder.withArgName("handler-classNames")
+        .withLongOpt("handlerClassName")
+        .hasArgs()
+        .withDescription("class names of the handlers")
```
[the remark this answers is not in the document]
nickibi: Updated

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
```diff
@@ -153,6 +178,27 @@ public static void main(String[] args) throws Exception
...
+    {
+     String handlerJarPaths = cl.getOptionValue("jar");
+     String classNames = cl.getOptionValue("className");
+     try
+     {
+       _handlers = ClassJarPathUtil.getAnnotationHandlers(handlerJarPaths, classNames);
+     }
+     catch (IllegalStateException e)
+     {
+       _logger.error("Error while doing schema compatibility check, could not get SchemaAnnotationHandler classes: " + e.getMessage());
```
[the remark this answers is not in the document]
nickibi: fixed.

restli-tools/src/main/java/com/linkedin/restli/internal/tools/ClassJarPathUtil.java
```diff
@@ -0,0 +1,85 @@
...
+    {
+      try
+      {
+        Class<?> handlerClass = Class.forName(className, false, classLoader);
+        SchemaAnnotationHandler handler = (SchemaAnnotationHandler) handlerClass.newInstance();
+        handlers.add(handler);
+      }
+      catch (Exception e)
+      {
+        throw new IllegalStateException("Error instantiating class: " + className + e.getMessage());
```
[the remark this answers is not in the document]
nickibi: Updated

data/src/main/java/com/linkedin/data/schema/annotation/SchemaAnnotationHandler.java
```diff
@@ -144,6 +149,105 @@ default SchemaVisitor getVisitor()
...
+
+    public AnnotationCompatibilityResult()
+    {
+      _messages = new MessageList<CompatibilityMessage>();
+    }
+
+    @Override
+
+    public Collection<CompatibilityMessage> getMessages() {
+      return _messages;
```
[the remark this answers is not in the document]
nickibi: Add a  addMessage() method.

data/src/main/java/com/linkedin/data/schema/compatibility/AnnotationCompatibilityChecker.java
```diff
@@ -0,0 +1,193 @@
...
+import java.util.List;
+import java.util.Map;
+import org.apache.commons.lang3.tuple.Pair;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ *  Schema Annotation compatibility checker
+ *  The annotation compatibility check is part of the annotation processing framework.
```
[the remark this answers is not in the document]
nickibi: fixed

data/src/main/java/com/linkedin/data/schema/annotation/SchemaAnnotationHandler.java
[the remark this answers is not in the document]
nickibi: fixed

data/src/main/java/com/linkedin/data/schema/compatibility/AnnotationCompatibilityChecker.java
nickibi: Add check for new field with new annotation case.

data/src/main/java/com/linkedin/data/schema/annotation/SchemaAnnotationHandler.java
```diff
@@ -144,6 +149,105 @@ default SchemaVisitor getVisitor()
...
+      AnnotationCheckContext olderContext, AnnotationCheckContext newerContext)
+  {
+    return new AnnotationCompatibilityResult();
+  }
+
+  /**
+   * AnnotationCheckContext which contains metadata information:
+   * dataSchema, schemaField, unionMember and pathToSchema.
+   */
+  class AnnotationCheckContext
```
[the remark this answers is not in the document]
nickibi: Updated

data/src/main/java/com/linkedin/data/schema/compatibility/AnnotationCompatibilityChecker.java
nickibi: I think we should provide such check to the annotation check, if an existing field is removed. Users should decide whether they will do something or not.

data/src/main/java/com/linkedin/data/schema/compatibility/AnnotationCompatibilityChecker.java
nickibi: Updated the logic here.

restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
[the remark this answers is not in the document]
nickibi: Discussed offline.
I introduced a new schema visitor - IdentitySchemaVisitor, which will not traverse the schema. It can be used for extension schema annotation check or other use cases which do not need override properties but would like to do the compatibility check. There will be no extra/unnecessary schema traverse for such case.
Now all the annotation check use cases will fall into the same path: go to annotation processor to process the annotation, then do the annotation compatibility check.

karthikbalasub (changes requested):
Looks much better now, thanks for making the recommended changes.

data/src/main/java/com/linkedin/data/schema/annotation/AnnotationCheckResolvedPropertiesVisitor.java
```diff
@@ -0,0 +1,114 @@
...
+    {
+      RecordDataSchema.Field field = context.getEnclosingField();
+      if (field.getResolvedProperties().isEmpty())
+      {
+        properties = field.getProperties();
+      }
+      else
+      {
+        properties = field.getResolvedProperties();
+      }
```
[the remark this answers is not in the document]
nickibi: Updated to use a utility method.

data/src/main/java/com/linkedin/data/schema/annotation/IdentitySchemaVisitor.java
```diff
@@ -0,0 +1,56 @@
...
+
+  @Override
+  public void callbackOnContext(TraverserContext context, DataSchemaTraverse.Order order) {
+    // skip post order traverse
+    if (order == DataSchemaTraverse.Order.POST_ORDER)
+    {
+      return;
+    }
+
+    if (context.getParentSchema() == null)
```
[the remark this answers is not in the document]
nickibi: Added a comment

data/src/main/java/com/linkedin/data/schema/annotation/IdentitySchemaVisitor.java
```diff
@@ -0,0 +1,56 @@
...
+package com.linkedin.data.schema.annotation;
+
+
+import com.linkedin.data.schema.DataSchema;
+import com.linkedin.data.schema.DataSchemaTraverse;
+
+
+/**
+ * This SchemaVisitor is used to not traverse the data schema.
+ * Users may want to use Schema Annotation Processor framework to do annotation compatibility check,
```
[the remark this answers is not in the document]
nickibi: Updated the doc.

data/src/main/java/com/linkedin/data/schema/annotation/SchemaAnnotationHandler.java
```diff
@@ -144,6 +149,113 @@ default SchemaVisitor getVisitor()
     return new PathSpecBasedSchemaAnnotationVisitor(this);
   }
 
+  /**
+   * This method is used to indicate whether the schema annotation handler implement the checkCompatibility method.
```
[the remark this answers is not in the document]
nickibi: fixed.

data/src/main/java/com/linkedin/data/schema/annotation/SchemaAnnotationHandler.java
```diff
@@ -144,6 +149,113 @@ default SchemaVisitor getVisitor()
...
+   * @return false by default. Users could override this method to return true, if they implement the checkCompatibility method.
```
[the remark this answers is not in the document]
nickibi: fixed

data/src/main/java/com/linkedin/data/schema/compatibility/AnnotationCompatibilityChecker.java
```diff
@@ -0,0 +1,174 @@
...
+          // we do the annotation check.
+          if (prevResolvedProperties.containsKey(annotationNamespace) || currResolvedProperties.containsKey(
+              annotationNamespace))
+          {
+            AnnotationCompatibilityResult result =
+                handler.checkCompatibility(prevResolvedProperties, currResolvedProperties,
+                    prevCheckContextAndResolvedProperty.getKey(), currCheckContextAndResolvedProperty.getKey());
+            results.add(result);
+          }
+          currResolvedPropertiesMap.remove(pathSpec);
```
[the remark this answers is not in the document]
nickibi: Good catch. I updated to use a copyResolvedPropertiesMap for each handler instead.

data/src/main/java/com/linkedin/data/schema/annotation/SchemaAnnotationProcessor.java
```diff
@@ -51,6 +51,9 @@
 {
   private static final Logger LOG = LoggerFactory.getLogger(SchemaAnnotationProcessor.class);
 
+
+  private static boolean _skipValidateProcessResult = false;
```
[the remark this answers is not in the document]
nickibi: Updated.

data/src/main/java/com/linkedin/data/schema/annotation/SchemaAnnotationProcessor.java
```diff
@@ -125,36 +128,71 @@ public static SchemaAnnotationProcessResult process(List<SchemaAnnotationHandler
...
+   * @param handlers the handlers that can resolve the annotation on the dataSchema and validate them
+   * @param dataSchema the dataSchema to be processed
+   * @param options additional options to help schema annotation processing
+   * @param skipValidateProcessResult If it is true, skip validate {@link SchemaAnnotationValidationVisitor} .
+   * @return result after process
+   */
+  public static SchemaAnnotationProcessResult process(List<SchemaAnnotationHandler> handlers,
+      DataSchema dataSchema, AnnotationProcessOption options, boolean skipValidateProcessResult)
+  {
+    _skipValidateProcessResult = skipValidateProcessResult;
```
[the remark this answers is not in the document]
nickibi: Updated based on the recommendation.

data/src/main/java/com/linkedin/data/schema/annotation/SchemaAnnotationHandler.java
```diff
@@ -144,6 +149,113 @@ default SchemaVisitor getVisitor()
...
+    return false;
+  }
+
+  /**
+   * Check annotations changes are compatible or not
+   *
+   * @param prevResolvedProperties the older resolvedProperties
+   * @param currResolvedProperties the newer resolvedProperties
+   * @param prevContext the older annotationCheck context
+   * @param currContext the newer annotationCheck context
```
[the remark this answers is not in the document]
nickibi: Updated

data/src/test/java/com/linkedin/data/schema/compatibility/TestAnnotationCompatibilityChecker.java
```diff
@@ -0,0 +1,193 @@
...
+            {
+              result.getMessages().add(new CompatibilityMessage(prevContext.getPathSpecToSchema(),
+                  CompatibilityMessage.Impact.ANNOTATION_COMPATIBLE_CHANGE, "Updating annotation field \"%s\" value is backward compatible change", ANNOTATION_FIELD_NAME));
+            }
+          }
+        }
+        return result;
+      }
+    };
+    return Collections.singletonList(testHandler);
```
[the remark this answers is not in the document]
nickibi: Added a test with multiple handlers.

data/src/main/java/com/linkedin/data/schema/compatibility/AnnotationCompatibilityChecker.java
```diff
@@ -0,0 +1,176 @@
...
+    {
+      if (!handler.implementsCheckCompatibility())
+      {
+        _log.info("There is no implementation of checkCompatibility method in SchemaAnnotationHandler: "
+            + handler.getClass().getName());
+        continue;
+      }
+      String annotationNamespace = handler.getAnnotationNamespace();
+      Map<PathSpec, Pair<CompatibilityCheckContext,
+          Map<String, Object>>> copyOfCurrResolvedPropertiesMap = new HashMap<>(currResolvedPropertiesMap);
```
karthikbalasub, line 89: This might be a problem for large schemas. Did you consider reversing the loops?

```
remove handlers that don't implement compatibility

foreach entry in prevResolvedProperties:
  foreach hander:
     check(entry)
     remove path from currProperties
foreach entry in currProperties
   foreach handler:
      check(entry)
```

data/src/main/java/com/linkedin/data/schema/annotation/SchemaAnnotationHandler.java
```diff
@@ -144,6 +149,121 @@ default SchemaVisitor getVisitor()
...
+   * This method is used to indicate whether the schema annotation handler implements the checkCompatibility method.
+   * @return false by default. Subclasses should override this method to return true, if they implement the checkCompatibility method.
+   */
+  default boolean implementsCheckCompatibility()
+  {
+    return false;
+  }
+
+  /**
+   * Check annotations changes are compatible or not
```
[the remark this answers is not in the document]
nickibi: Updated doc.

data/src/main/java/com/linkedin/data/schema/compatibility/AnnotationCompatibilityChecker.java
nickibi: Refactored this part based on the recommendation.

karthikbalasub (approved):

PR #437, opened 2020-10-03
Add an option to configure ProtoWriter buffer size. Set the default to 4096 to prevent thrashing

karthikrg:
Use an identity hashset implementation that uses DataComplex#dataComplexHashCode under the hood for better performance when detecting cycles

### DIFF
--- a/data/src/main/java/com/linkedin/data/DataComplexIdentitySet.java
+++ b/data/src/main/java/com/linkedin/data/DataComplexIdentitySet.java
@@ -1,0 +1,131 @@
+/*
+   Copyright (c) 2020 LinkedIn Corp.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package com.linkedin.data;
+
+import java.util.IdentityHashMap;
+
+
+/**
+ * A space and time efficient {@link IdentityHashMap} equivalent optimized for checking {@link DataComplex} by
+ * identity. We don't use {@link IdentityHashMap} as-is since {@link System#identityHashCode(Object)} performs
+ * significantly worse than {@link DataComplex#dataComplexHashCode()} under multi-threaded conditions.
+ *
+ * <p>Instances of this class are typically used for cycle detection.</p>
+ */
+class DataComplexIdentitySet
+{
+  private static final int CAPACITY = 16;
+
+  private final Node[] _table;
+
+  /**
+   * Constructor.
+   */
+  DataComplexIdentitySet()
+  {
+    _table = new Node[CAPACITY];
+  }
+
+  /**
+   * Add the given {@link DataComplex} to the set.
+   *
+   * @param dataComplex The {@link DataComplex} to add.
+   *
+   * @return True if the {@link DataComplex} was already present and not added, false otherwise.
+   */
+  boolean add(DataComplex dataComplex)
+  {
+    final int index = dataComplex.dataComplexHashCode() & (CAPACITY - 1);
+    Node node = _table[index];
+
+    // No entries in bucket, add and return false.
+    if (node == null)
+    {
+      _table[index] = new Node(dataComplex);
+      return false;
+    }
+
+    Node previous = null;
+    while (node != null)
+    {
+      if (node._dataComplex == dataComplex)
+      {
+        // Entry found, return true and bail!
+        return true;
+      }
+
+      previous = node;
+      node = node._next;
+    }
+
+    // No entry found for given data complex. Add it to the end, and return false.
+    previous._next = new Node(dataComplex);
+    return false;
+  }
+
+  /**
+   * Removes the given {@link DataComplex} from the set if it exists.
+   *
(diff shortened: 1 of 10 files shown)


build.gradle
```diff
@@ -52,6 +52,7 @@ project.ext.externalDependency = [
   'codemodel': 'com.sun.codemodel:codemodel:2.2',
   'commonsCli': 'commons-cli:commons-cli:1.0',
   'commonsCodec': 'commons-codec:commons-codec:1.3',
+  'commonsCollections4': 'org.apache.commons:commons-collections4:4.0',
```
tjni, line 55: We haven't traditionally cared about this, but can we copy the HashedMap implementation from Apache Commons if we need it? This is a 500 KB artifact that will be forced on consumers since pegasus:data is pulled in by anything that uses pegasus.

karthikrg: Fixed

tjni:
Thank you for removing the dependency on commons-collection4 and for this work on optimizing traversal. We should commit this in my opinion.

I started thinking about this some more nonetheless, and I think there are additional improvements to be had if this is a hotspot. What we need to do is implement a pointer-based cycle finding algorithm like [Brent's algorithm](https://en.wikipedia.org/wiki/Cycle_detection#Brent's_algorithm). The disadvantage is it will take more steps to realize there is a cycle. The advantage is it uses O(1) memory, if we also change the traversal from recursion to iteration. Since usually there is no cycle, because a cycle is an error, this is a good trade-off.

data/src/main/java/com/linkedin/data/DataComplexIdentitySet.java
```diff
@@ -0,0 +1,131 @@
...
+    Node previous = null;
+    while (node != null)
+    {
+      if (node._dataComplex == dataComplex)
+      {
+        // Entry found, remove it.
+        if (previous == null)
+        {
+          // This is the first node, set table index to null.
+          _table[index] = null;
```
karthikbalasub, line 105: Should be `_table[index] = node.next`

data/src/main/java/com/linkedin/data/DataComplexIdentitySet.java
```diff
@@ -0,0 +1,131 @@
...
+   limitations under the License.
+*/
+
+package com.linkedin.data;
+
+import java.util.IdentityHashMap;
+
+
+/**
+ * A space and time efficient {@link IdentityHashMap} equivalent optimized forr checking {@link DataComplex} by
```
karthikbalasub, line 23: forr -> for

data/src/main/java/com/linkedin/data/DataComplexIdentitySet.java
```diff
@@ -0,0 +1,131 @@
...
+ * identity. We don't use {@link IdentityHashMap} as-is since {@link System#identityHashCode(Object)} performs
+ * significantly worse than {@link DataComplex#dataComplexHashCode()} under multi-threaded conditions.
+ *
+ * <p>Instances of this class are typically used for cycle detection.</p>
+ */
+class DataComplexIdentitySet
```
karthikbalasub, line 29: Please add unittests

karthikbalasub (changes requested):

data/src/main/java/com/linkedin/data/DataComplexIdentitySet.java
karthikrg: Isn't that the same as setting to null since this is the 1st node?

karthikrg:
> Thank you for removing the dependency on commons-collection4 and for this work on optimizing traversal. We should commit this in my opinion.
> 
> I started thinking about this some more nonetheless, and I think there are additional improvements to be had if this is a hotspot. What we need to do is implement a pointer-based cycle finding algorithm like [Brent's algorithm](https://en.wikipedia.org/wiki/Cycle_detection#Brent's_algorithm). The disadvantage is it will take more steps to realize there is a cycle. The advantage is it uses O(1) memory, if we also change the traversal from recursion to iteration. Since usually there is no cycle, because a cycle is an error, this is a good trade-off.

The current hotspot is due to System#identityHashCode. I suggest we profile with this first and optimize more only if there are still issues.

data/src/main/java/com/linkedin/data/DataComplexIdentitySet.java
karthikrg: Fixed!

data/src/test/java/com/linkedin/data/TestDataComplexIdentitySet.java
```diff
@@ -0,0 +1,71 @@
...
+
+    // Ensure that adding the same hashcode map returns false the first time since it doesn't yet exist.
+    Assert.assertFalse(set.add(sameHashCodeMap));
+
+    // Ensure that adding the same hashcod map again returns true.
+    Assert.assertTrue(set.add(sameHashCodeMap));
+
+    // Ensure that the original and cloned maps still exist.
+    Assert.assertTrue(set.add(dataMap));
+    Assert.assertTrue(set.add(clone));
```
karthikbalasub, line 63: Remove dataMap and then add sameHashCodeMap, it should return true, but will return false.

data/src/main/java/com/linkedin/data/DataComplexIdentitySet.java
karthikbalasub: It wouldn't work if the index has more than one element.

_table[i] = A -> B -> null
remove A
_table[i] = null

it should be 
_table[i] = B -> null

karthikbalasub (changes requested):

data/src/test/java/com/linkedin/data/TestDataComplexIdentitySet.java
karthikrg: Got it. Fixed and added this test.

karthikbalasub (approved):

PR #442, opened 2020-10-12
Add getClient async call in warmup load balancer

rachelhanhan:
**Background**
Recently voyager-api found another blocking call that caused a host being shutdown manually after all threads are blocked. After taking a look into the log, we found that there is another blocking call when calling getClient.
In LoadBalancer.java, there are 2 methods in this interface: 1) getClient, 2) getLoadBalancedServiceProperties, both methods have sync and async implementations. Last time we fixed getLoadBalancedServiceProperties by letting the last client going through as async path, however, looks like getClient method is another source that leads to blocked threads.

**How to fix**
After looking into the code, we believe warmUpLoadBalancer is the only load balancer that is making the sync call to get client, the upstream class makes call using the async way, however, because warmUpLoadBalancer and LoadBalancerWithFacilitiesDelegator do not have the async implementation, it defaults to LoadBalancer.java, which is faking the async implementation by waiting for the response. Therefore, as long as we add the async implementation of getClient, the problem should be solved.

Please note, this change is not guarded by any config, since the change is fairly straight forward.

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/util/TogglingLoadBalancer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/util/TogglingLoadBalancer.java
@@ -54,7 +54,7 @@ import java.util.Map;
 
 public class TogglingLoadBalancer implements LoadBalancer, HashRingProvider, ClientFactoryProvider, PartitionInfoProvider, WarmUpService, ClusterInfoProvider
 {
-  private final LoadBalancer _balancer;
+  private final SimpleLoadBalancer _balancer;
   private final WarmUpService _warmUpService;
   private final HashRingProvider _hashRingProvider;
   private final PartitionInfoProvider _partitionInfoProvider;
--- a/d2/src/main/java/com/linkedin/d2/balancer/util/WarmUpLoadBalancer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/util/WarmUpLoadBalancer.java
@@ -288,4 +288,22 @@ public class WarmUpLoadBalancer extends LoadBalancerWithFacilitiesDelegator
     _usedServices.add(serviceName);
     return client;
   }
+
+  @Override
+  public void getClient(Request request, RequestContext requestContext, Callback<TransportClient> clientCallback)
+  {
+    _loadBalancer.getClient(request, requestContext, new Callback<TransportClient>() {
+      @Override
+      public void onError(Throwable e) {
+        clientCallback.onError(e);
+      }
+
+      @Override
+      public void onSuccess(TransportClient result) {
+        String serviceName = LoadBalancerUtil.getServiceNameFromUri(request.getURI());
+        _usedServices.add(serviceName);
+        clientCallback.onSuccess(result);
+      }
+    });
+  }
 }
--- a/d2/src/main/java/com/linkedin/d2/balancer/util/WarmUpLoadBalancer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/util/WarmUpLoadBalancer.java
@@ -306,4 +306,9 @@ public class WarmUpLoadBalancer extends LoadBalancerWithFacilitiesDelegator
       }
     });
   }
+
+  @Override
+  public void getLoadBalancedServiceProperties(String serviceName, Callback<ServiceProperties> clientCallback) {
+    _loadBalancer.getLoadBalancedServiceProperties(serviceName, clientCallback);
+  }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,9 +13,12 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
+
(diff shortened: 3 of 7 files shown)


d2/src/main/java/com/linkedin/d2/balancer/util/TogglingLoadBalancer.java
```diff
@@ -54,7 +54,7 @@
 
 public class TogglingLoadBalancer implements LoadBalancer, HashRingProvider, ClientFactoryProvider, PartitionInfoProvider, WarmUpService, ClusterInfoProvider
 {
-  private final LoadBalancer _balancer;
+  private final SimpleLoadBalancer _balancer;
```
nizarm, line 57: Why do we need to introduce a strong coupling between TogglingLoadBalancer vs SimpleLoadBalancer?

d2/src/main/java/com/linkedin/d2/balancer/util/WarmUpLoadBalancer.java
```diff
@@ -288,4 +288,22 @@ public TransportClient getClient(Request request, RequestContext requestContext)
     _usedServices.add(serviceName);
     return client;
   }
+
+  @Override
+  public void getClient(Request request, RequestContext requestContext, Callback<TransportClient> clientCallback)
```
nizarm, line 293: I see we are enabling the async path by overriding the default implementation here. Does this mean that we already took care of IC Handling for those executions that are done inside PropertyEventBus? If not, this will break those services that are reliant on IC to propagate some contexts

nizarm (changes requested):

d2/src/main/java/com/linkedin/d2/balancer/util/WarmUpLoadBalancer.java
rachelhanhan: My understanding is that once service properties are loaded in memory, we do not need to switch thread to read from d2, only if d2 is not loaded for the first time, this is a problem.

I remember when you debugged an issue last time, without warmup, the first ever call will always fail, but all the following calls pass, so I guess doing the async getClient call isn't a problem for all the existing load balancers, it's only a problem when d2 is not loaded in memory.

I don't think this is making the current situation worse. For the warm up load balancer, if it can load d2 properties in memory in advance, this getClient call doesn't need to be blocking. But I agree it will be a problem if warmup did not finish, then the 1st call will fail, but subsequent calls should work.

nizarm: That is correct, once the properties are loaded into memory this issue won't happen because there is no asynchronous activity involved and the call gets executed in the same calling thread.

If we are depending on Warmup to fix the IC issue, how do we handle a new host deployed to the fleet? How do we handle a new service being added? 

We should look at the IC problem more holistically to make it future proof IMHO. How can we make sure that all the asynchronous executions inside the load balancer are IC friendly so that we can support a true asynchronous load balancer without having to make assumptions?

rachelhanhan: Is it already a problem today for those mps who do not enable warmUp? Will there be a problem when a new host is deployed? If so, I think we should probably put the correct IC restore in place.

I'm surprised that we're depending on warmup load balancer by leveraging the blocking call to make sure IC works. I'm also surprised that the call is never truly async since warmup is globally enabled.

nizarm: > Is it already a problem today for those mps who do not enable warmUp

I dont think any of the Load balancer out there is Asynchrounous since WarmUp is enabled globally and all the D2 calls are effectively is a Sync Call. 

> Will there be a problem when a new host is deployed

This is a danger in trusing Warmup to bring the state to memory to make the IC work. Warmup only works with the previously cached data on disk. For a new host there are no past d2 meta data information to pre-load the load balancer state

d2/src/main/java/com/linkedin/d2/balancer/util/TogglingLoadBalancer.java
rachelhanhan: Because the fact is that we are always passing a SimpleLoadBalancer. This is for dev's benefit to understand how load balancers are wrapped in layers. Otherwise it's so hard to tell which load balancer uses which one, we have so many layers wrapped one by one. When I see this togglingLoadBalancer only accepts SimpleLoadBalancer in its constructor, I just want to make it more explicit.

rachelhanhan:
**Update**
1. I removed the default implementation of the async path since the default implementation is faking it! This is backward compatible change, it will break a small amount of mps - container(unit test only), restli-test-fwk and autortf. Because it does not impact large number of upstreams, I would rather make this backward incompatible change and force everyone to have the truly async implementation.
2. I didn't mark the synchronous path @deprecated because there are so many dependencies on the sync path, it will be a ton of code cleanups.

ChrisCarini:

PR #443, opened 2020-10-13
Add an IntelliJ plugin for PDL.

jplaisted:
This code exists internally at LI. I've done the minimal amount of work to copy it here to open source and get it building.

Note I did need to disable Werror for this one project. We can clean it up later.


BrianPin:
I am somewhat hesitate to put it here, @karthikbalasub  @evanw555  back me up or share your thoughts.
The good side are
1) The parsing of the PDL might get benefit by using pegasus's parser (but I don't know if it is possible to use pegasus parser for pdl intellij plugin

The down side are
1) It is two completely different projects in the repo, the world that uses the pegasus as lib won't not necessary want to build pdl plugin
2) The pdl plugin uses some intellij's gradle plugin and intellij's SDK, that will add more dependency to pegasus and people when facing build issues will argue why intellij plugin or sdk is needed to build pegasus

Seems the downside is a bit more, can we use a seperate repo for it?

jplaisted:
The upside to the same repo is avoiding a very long process to create that repo :) Which is a pretty big win

karthikbalasub (commented):
I agree with Brian, we cannot combine the builds of rest.li and the intellij plugin. Though it might be cheaper now to use the same repo, you'd endup paying the cost of increased build/release time for ever.

I suggest finding a solution where the build and release are separates (if possible with the same repo)

jplaisted:
Also

> It is two completely different projects in the repo, the world that uses the pegasus as lib won't not necessary want to build pdl plugin

I'd argue maybe the bigger issue is just a cognitive one of trying to understand the project. iirc last I built this project it took well over an hour. Building this plugin takes ~14 seconds, assuming `--rerun-tasks` really forces it from scratch.

>you'd endup paying the cost of increased build/release time for ever

Disagree. Code can be moved, just like I'm doing here. So long as the end artifact (the plugin hosted by IntelliJ) stays in the same location.

>I suggest finding a solution where the build and release are separates (if possible with the same repo)

I'll look into that.

jplaisted:
@karthikbalasub how would you feel about including it in the build by default, but not the release? Having a separate set of "main" build files is going to lead to more duplication and maintenance. It seems easy to exclude it from release, however (add it to `privateModules`).

It's also not that bad to exclude it from the build with `-x` to gradle; but you have to remember to do that everywhere. I can also add that to travis, but if someone just clones the project and does `build` it'll still build the plugin.

So question is are we okay adding 14 seconds to an hour long build to avoid maintaining two sets of gradle files? Or would you prefer I silo it as much as possible, even if that introduces extra gradle files?

evanw555:
Agree with Brian and Karthik that this should be in its own repo. The current internal repo doesn't depend on pegasus APIs at all, so there's no good reason for it to be here. Plus, an IntelliJ plugin is of a totally different nature than a standard Java library, so mixing it in wouldn't be clean.

> iirc last I built this project it took well over an hour. Building this plugin takes ~14 seconds

That's with testing. Without testing, it only takes a few minutes. When building the IntelliJ plugin against a new version of IDEA, it takes at least a few minutes to download. That's a significant increase in the build.

> Disagree. Code can be moved, just like I'm doing here

That's an overhead in and of itself, so why don't we just do the overhead now to create it as a new repo? If we're gonna open-source the IDEA plugin, it should be done cleanly from the start.

jplaisted:
> That's with testing. Without testing, it only takes a few minutes. When building the IntelliJ plugin against a new version of IDEA, it takes at least a few minutes to download. That's a significant increase in the build.

Ah, right. `build` isn't just build in gradle. And I have some dependencies cached already. Good points, thanks.

> That's an overhead in and of itself, so why don't we just do the overhead now to create it as a new repo? If we're gonna open-source the IDEA plugin, it should be done cleanly from the start.

Sure, we volunteered for this with some limited capacity dedicated to it. Doing it "the right way" might take significantly longer / more effort, and we may not be able to help with that effort.

I do agree that the point at which the builds are totally separated things start to get quite messy and a new repo would be better. My concerns are the effort around that being much higher and ownership less clear. I'll discuss this with my team some more and we can probably continue this discussion internally.

jplaisted:
I'll close this until we decide how to proceed.

PR #445, opened 2020-10-14
Add ExtensionSchemaAnnotationHandler 

nickibi:
Add ExtensionSchemaAnnotationHandler for extension schema annotation compatibility check.

The new handler implements 'checkCompatibility' method, which introduces specific rules for extension annotation compatibility check.

Test:
unit tests
manual test.

### DIFF
--- a/data/src/main/java/com/linkedin/data/schema/annotation/AnnotationCheckResolvedPropertiesVisitor.java
+++ b/data/src/main/java/com/linkedin/data/schema/annotation/AnnotationCheckResolvedPropertiesVisitor.java
@@ -20,6 +20,8 @@ import com.linkedin.data.schema.DataSchemaTraverse;
 import com.linkedin.data.schema.NamedDataSchema;
 import com.linkedin.data.schema.PathSpec;
 import com.linkedin.data.schema.RecordDataSchema;
+import com.linkedin.data.schema.TyperefDataSchema;
+import com.linkedin.data.schema.UnionDataSchema;
 import com.linkedin.data.schema.annotation.SchemaAnnotationHandler.CompatibilityCheckContext;
 import java.util.ArrayDeque;
 import java.util.HashMap;
@@ -38,6 +40,10 @@ public class AnnotationCheckResolvedPropertiesVisitor implements SchemaVisitor
 {
   private Map<PathSpec, Pair<CompatibilityCheckContext, Map<String, Object>>> _nodeToResolvedPropertiesMap = new HashMap<>();
 
+  private final String FIELD_INDICATOR = "$field";
+  private final String UNION_MEMBER_KEY_INDICATOR = "$unionMemberKey";
+  private final String TYPEREF_INDICATOR = "$typeref";
+
   @Override
   public void callbackOnContext(TraverserContext context, DataSchemaTraverse.Order order)
   {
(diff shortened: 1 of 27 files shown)


data/src/main/java/com/linkedin/data/schema/annotation/AnnotationCheckResolvedPropertiesVisitor.java
```diff
@@ -49,36 +55,47 @@ public void callbackOnContext(TraverserContext context, DataSchemaTraverse.Order
...
+      PathSpec pathSpec = new PathSpec(pathToSchema);
+
+      _nodeToResolvedPropertiesMap.put(pathSpec,
+          new ImmutablePair<>(generateCompatibilityCheckContext(schemaField, unionMember, currentSchema, pathSpec),
+              chooseProperties(schemaField.getResolvedProperties(), schemaField.getProperties())));
+      pathToSchema.removeLast();
     }
-    else
+
+    // Current node is a union member, get the union member key's annotation.
```
BrianPin, line 84: I am wondering if it could be written into a single
if ... else if ... block
instead of two if blocks
Because that way, we will get to know this two logic are meant for output just one thing:
a pathspec to checkerContext map

data/src/main/java/com/linkedin/data/schema/annotation/AnnotationCheckResolvedPropertiesVisitor.java
```diff
@@ -49,36 +55,47 @@ public void callbackOnContext(TraverserContext context, DataSchemaTraverse.Order
...
+      pathToSchema.addLast(UNION_MEMBER_KEY_INDICATOR);
+      PathSpec pathSpec = new PathSpec(pathToSchema);
+
+      _nodeToResolvedPropertiesMap.put(pathSpec,
+              new ImmutablePair<>(generateCompatibilityCheckContext(schemaField, unionMember, currentSchema, pathSpec), unionMember.getProperties()));
+      pathToSchema.removeLast();
     }
 
-    _nodeToResolvedPropertiesMap.put(pathSpec, new ImmutablePair<>(compatibilityCheckContext, properties));
+    // If there is no resolvedProperties but properties, used the properties for annotation check.
```
BrianPin, line 95: is -> are for grammar check

data/src/main/java/com/linkedin/data/schema/annotation/AnnotationCheckResolvedPropertiesVisitor.java
```diff
@@ -49,36 +55,47 @@ public void callbackOnContext(TraverserContext context, DataSchemaTraverse.Order
...
+    Map<String, Object> properties = chooseProperties(currentSchema.getResolvedProperties(), currentSchema.getProperties());
```
BrianPin, line 96: And looks like the last part can be the `else` part of the logic

data/src/main/java/com/linkedin/data/schema/annotation/PegasusExtensionSchemaAnnotationHandler.java
```diff
@@ -0,0 +1,147 @@
...
+import java.util.Map;
+import org.apache.commons.lang3.tuple.Pair;
+
+
+/**
+ * This SchemaAnnotationHandler is used to check extension schema annotation(@extension) compatibility
+ *
+ * @author Yingjie Bi
+ */
+public class PegasusExtensionSchemaAnnotationHandler implements SchemaAnnotationHandler
```
BrianPin, line 31: nit: Wonder how do you feel about make it less lengthy:
PegasusExtensionSchemaAnnotationHandler -> ExtensionSchemaAnnotationHandler

As a code review, I usually hope the code is readable, but also compact.
I think Pegasus is being a bit extra since we know where we are. 
Just a nit suggestion!

data/src/main/java/com/linkedin/data/schema/annotation/PegasusExtensionSchemaAnnotationHandler.java
```diff
@@ -0,0 +1,147 @@
...
+{
+  private final String EXTENSION_ANNOTATION_NAMESPACE = "extension";
```
BrianPin, line 33: There are other definition of `EXTENSION_ANNOTATION_NAMESPACE`, please just make it one, and compact. It can be public, since it is final, nobody is able to change it. Also the lesser the defined string the more efficient memory use.

data/src/main/java/com/linkedin/data/schema/annotation/PegasusExtensionSchemaAnnotationHandler.java
```diff
@@ -0,0 +1,147 @@
...
+
+  @Override
+  public ResolutionResult resolve(List<Pair<String, Object>> propertiesOverrides,
+      ResolutionMetaData resolutionMetadata)
+  {
+    return new ResolutionResult();
```
BrianPin, line 39: The semantic of the function name makes me believe it is actually doing some resolving action, it will be less expected to just return the a default ResolutionResult class instance.
If we want an instance of ResolutionResult, we can use other naming.

data/src/main/java/com/linkedin/data/schema/annotation/PegasusExtensionSchemaAnnotationHandler.java
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  @Override
+  public String getAnnotationNamespace()
+  {
+    return EXTENSION_ANNOTATION_NAMESPACE;
```
BrianPin, line 45: If we make EXTENSION_ANNOTATION_NAMESPACE compact (just one place has it) and public (because it is finalized) maybe we can save this function.

data/src/main/java/com/linkedin/data/schema/annotation/PegasusExtensionSchemaAnnotationHandler.java
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  @Override
+  public AnnotationValidationResult validate(Map<String, Object> resolvedProperties, ValidationMetaData metaData)
+  {
+    return new AnnotationValidationResult();
```
BrianPin, line 51: So I guess right now this class is just a place holder?

data/src/main/java/com/linkedin/data/schema/annotation/PegasusExtensionSchemaAnnotationHandler.java
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  @Override
+  public boolean implementsCheckCompatibility()
+  {
+    return true;
+  }
+
+  @Override
+  public AnnotationCompatibilityResult checkCompatibility(Map<String, Object> prevResolvedProperties, Map<String, Object> currResolvedProperties,
```
BrianPin, line 68: Wonder if we can change `prevResolvedProperties` to `snapshot...properties`, because it is actually from the snapshot.

data/src/main/java/com/linkedin/data/schema/annotation/PegasusExtensionSchemaAnnotationHandler.java
```diff
@@ -0,0 +1,147 @@
...
+      {
+        // Adding a new injected field with extension annotation.
+        appendCompatibilityMessage(result, CompatibilityMessage.Impact.ANNOTATION_COMPATIBLE_CHANGE,
+            "Adding extension annotation on new field: \"%s\" is backward compatible change", currContext.getSchemaField().getName() , currContext.getPathSpecToSchema());
+      }
+    }
+    return result;
+  }
+
+  private void appendCompatibilityMessage(AnnotationCompatibilityResult result, CompatibilityMessage.Impact impact, String message, String context, PathSpec pathSpec)
```
BrianPin, line 142: The above logic only has 1 `ANNOTATION_COMPATIBLE_CHANGE` impact, all the other impact are incompatible.
I wonder how is it better if split into two results (or just one result if you dont care comptaible message, btw why do we need compatible message?)
If we split to two results, and you can add all incompatible msg to a dedicated result. That will make these callers that are saving msg not need to pass the `CompatibilityMessage.Impact.ANNOTATION_INCOMPATIBLE_CHANGE` and less repeating ourselves.

nickibi: We only have ONE result here. result contains a message list which contains all messages no matter it is compatible or incompatible. We will not throw the exception for compatible change, but we do need such information to generate compatibility report. That will be the same user experience as schema compatibility check.

data/src/main/java/com/linkedin/data/schema/annotation/PegasusExtensionSchemaAnnotationHandler.java
nickibi: We need to implement this method to implement SchemaAnnotationHandler. Actually, we do not need this 'validate' method. We could return null here. But it really does not matter, because this method will never be called.

data/src/main/java/com/linkedin/data/schema/annotation/PegasusExtensionSchemaAnnotationHandler.java
nickibi: Same as 'validate' method. We only need checkCompatibility method from SchemaAnnotationHandler. However to implement SchemaAnnotationHandler we have to provide implementation for this resolve method. We could return null here, but it does not matter, because this method will not be called.

data/src/main/java/com/linkedin/data/schema/annotation/PegasusExtensionSchemaAnnotationHandler.java
nickibi: Yes, we could call it 'ExtensionSchemaAnnotationHandler'

data/src/main/java/com/linkedin/data/schema/annotation/PegasusExtensionSchemaAnnotationHandler.java
nickibi: we have converted snapshot.pdl to DataSchema for compatibility check. Technically it is dataschema, not snapshot. So calling it 'snapshot...' is not accurate.

data/src/main/java/com/linkedin/data/schema/annotation/PegasusExtensionSchemaAnnotationHandler.java
nickibi: This function has to be there. No matter where we define this constant. 
While we use this handler, we need to know the namespace of annotation before calling the checkCompatibility method.

BrianPin (approved):
LGTM

PR #446, opened 2020-10-15
Add partition support in relative load balancer inGraph

rachelhanhan:
**Background**
For relative load balancer onboarding, we're trying to find different type of services to onboard, one type is services with partitions. However, one problem with today's inGraph is that we always emit metrics for partition 0, it turns out one of the client that I'm onboarding, they only talk to partition 1, they never talk to partition 0.

To perfectly solve this problem, I think we should fire one inGraph for each partition, however, it's really hard since at inGraph registration time, we do not know how many partitions are there, partitions can be very dynamic. So the trick here is to always emit metrics for the 1st added partition, so that at least we have some meaningful data.

**Changes include**
1. Emit metrics for the first added partition
2. Add a small debug log when health score reduces

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/jmx/RelativeLoadBalancerStrategyJmx.java
+++ b/d2/src/main/java/com/linkedin/d2/jmx/RelativeLoadBalancerStrategyJmx.java
@@ -21,7 +21,6 @@ import com.linkedin.d2.balancer.strategies.LoadBalancerQuarantine;
 import com.linkedin.d2.balancer.strategies.relative.RelativeLoadBalancerStrategy;
 import com.linkedin.d2.balancer.strategies.relative.StateUpdater;
 import com.linkedin.d2.balancer.strategies.relative.TrackerClientState;
-import com.linkedin.d2.balancer.util.partitions.DefaultPartitionAccessor;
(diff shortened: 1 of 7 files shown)


d2/src/main/java/com/linkedin/d2/jmx/RelativeLoadBalancerStrategyJmx.java
```diff
@@ -98,12 +119,17 @@ public double getMaxLatencyRelativeFactor()
   @Override
   public double getNthPercentileLatencyRelativeFactor(double pct)
   {
+    if (isPartitionDataUnavailable())
```
rickzx, line 122: Just to confirm, in a real situation, this function should always return true right?

rachelhanhan: Correct, it will always return false.

rachelhanhan: I just realized it can be a NPE if we always pass partition 0, but when it's using the first valid partition, there shouldn't be NPE anymore, but just to be careful I still added this check.

rickzx (approved):
LGTM

PR #447, opened 2020-10-15
Fix the bug so properties for typeref in unionWithAlias should be pro…

junchuanwang:
…pagated during pegasus to avro translation

The typeref property should be brought to Avro schema translation.

It was not previously working with typeref in Union with Alias since typeref in unionwithalias was handled separately in callbacks. 

This change is bring typeref properties to the union with alias.

### DIFF
--- a/data-avro/src/main/java/com/linkedin/data/avro/PegasusUnionToAvroRecordConvertCallback.java
+++ b/data-avro/src/main/java/com/linkedin/data/avro/PegasusUnionToAvroRecordConvertCallback.java
@@ -63,9 +63,11 @@ class PegasusUnionToAvroRecordConvertCallback implements DataSchemaTraverse.Call
     {
       DataSchema fieldSchema = field.getType().getDereferencedDataSchema();
 
-      Map<String, Object>  propagatedProperties = new HashMap<>();
+      Map<String, Object>  propagatedProperties = fieldSchema.getProperties();
       if (field.getType().getType() == DataSchema.Type.TYPEREF)
       {
+        // If the field schema type is TypeRef,
+        // then use the merged typeref properties instead
         propagatedProperties = ((TyperefDataSchema) field.getType()).getMergedTyperefProperties();
       }
 
@@ -185,6 +187,7 @@ class PegasusUnionToAvroRecordConvertCallback implements DataSchemaTraverse.Call
    * @param field Reference to the field whose schema and default value is being overridden
    * @param modifiedSchema The override schema to use for the specified field
    * @param modifiedDefaultValue The override default value to use for the specified field
+   * @param propagatedProperties The properties value to use for the specified field
    */
   private void overrideUnionFieldSchemaAndDefault(RecordDataSchema.Field field,
       DataSchema modifiedSchema, Object modifiedDefaultValue, Map<String, Object> propagatedProperties)
--- a/data-avro/src/main/java/com/linkedin/data/avro/PegasusUnionToAvroRecordConvertCallback.java
+++ b/data-avro/src/main/java/com/linkedin/data/avro/PegasusUnionToAvroRecordConvertCallback.java
@@ -12,6 +12,7 @@ import com.linkedin.data.schema.EnumDataSchema;
 import com.linkedin.data.schema.MapDataSchema;
 import com.linkedin.data.schema.Name;
 import com.linkedin.data.schema.RecordDataSchema;
+import com.linkedin.data.schema.TyperefDataSchema;
 import com.linkedin.data.schema.UnionDataSchema;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -62,6 +63,12 @@ class PegasusUnionToAvroRecordConvertCallback implements DataSchemaTraverse.Call
     {
       DataSchema fieldSchema = field.getType().getDereferencedDataSchema();
 
+      Map<String, Object>  propagatedProperties = new HashMap<>();
+      if (field.getType().getType() == DataSchema.Type.TYPEREF)
+      {
+        propagatedProperties = ((TyperefDataSchema) field.getType()).getMergedTyperefProperties();
+      }
+
       // The conversion from Pegasus union type to an Avro record is performed when the union appears as either the
       // field's direct type or when the field's type is an array or a map whose (nested) elements is of union type.
       // This conversion ignores the default value when specified on an array or map typed field. Since the elements in
@@ -74,7 +81,7 @@ class PegasusUnionToAvroRecordConvertCallback implements DataSchemaTraverse.Call
 
       if (modifiedSchema != null)
       {
-        overrideUnionFieldSchemaAndDefault(field, modifiedSchema, modifiedDefaultValue);
+        overrideUnionFieldSchemaAndDefault(field, modifiedSchema, modifiedDefaultValue, propagatedProperties);
       }
     }
   }
@@ -180,7 +187,7 @@ class PegasusUnionToAvroRecordConvertCallback implements DataSchemaTraverse.Call
    * @param modifiedDefaultValue The override default value to use for the specified field
    */
   private void overrideUnionFieldSchemaAndDefault(RecordDataSchema.Field field,
-      DataSchema modifiedSchema, Object modifiedDefaultValue)
+      DataSchema modifiedSchema, Object modifiedDefaultValue, Map<String, Object> propagatedProperties)
   {
     // Stash the field's original type and default value, so that we can use this for reverting them back after
     // the schema translation is complete. This is because we don't want the input schema to have any modifications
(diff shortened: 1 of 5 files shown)


data-avro/src/test/java/com/linkedin/data/avro/TestSchemaTranslator.java
```diff
@@ -322,6 +533,20 @@ public void testToAvroSchema_testTypeRefAnnotationPropagation(String schemaBefor
    assertEquals(fieldsPropertiesMap, avroSchemaAsDataMap);
  }
 
+  @Test(dataProvider = "toAvroSchemaData_testTypeRefAnnotationPropagation")
+  public void testToAvroSchema_testTypeRefAnnotationPropagation(String schemaBeforeTranslation,
```
nickibi, line 537: Can you please update the test method name to camelcase to follow the method naming style?
Same as data provider name.

data-avro/src/main/java/com/linkedin/data/avro/PegasusUnionToAvroRecordConvertCallback.java
```diff
@@ -62,6 +63,12 @@ public void callback(List<String> path, DataSchema schema)
     {
       DataSchema fieldSchema = field.getType().getDereferencedDataSchema();
 
+      Map<String, Object>  propagatedProperties = new HashMap<>();
```
nickibi, line 66: Please add comment to explain this specific use case.

nickibi (commented):
Can you please add some description about this bug fix?

This change is overriding the union field property to the propagated property if the union field type is typeref. This is the expected behavior? Could you mind share some context about it?

data-avro/src/main/java/com/linkedin/data/avro/PegasusUnionToAvroRecordConvertCallback.java
```diff
@@ -180,7 +187,7 @@ private DataSchema modifyFieldSchema(RecordDataSchema recordSchema, RecordDataSc
    * @param modifiedDefaultValue The override default value to use for the specified field
    */
```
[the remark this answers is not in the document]
junchuanwang: > Can you please add some description about this bug fix?
> 
> This change is overriding the union field property to the propagated property if the union field type is typeref. This is the expected behavior? Could you mind share some context about it?

Still in initial review. Confirming with customer and once confirmed, I will add more description.

But for what you mentioned:
> This change is overriding the union field property to the propagated property if the union field type is typeref.

This is the desired behavior, but before this fix, it won't work with unionWithAlias type. This fix is to address issue

data-avro/src/test/java/com/linkedin/data/avro/TestSchemaTranslator.java
```diff
@@ -69,6 +69,217 @@ public void testTranslateDefaultBackwardsCompatibility()
 
     assertSame(DataToAvroSchemaTranslationOptions.DEFAULT_OPTIONAL_DEFAULT_MODE, OptionalDefaultMode.TRANSLATE_DEFAULT);
   }
+  @DataProvider
```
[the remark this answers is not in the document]
junchuanwang: Will do

data-avro/src/test/java/com/linkedin/data/avro/TestSchemaTranslator.java
junchuanwang: Will do

data-avro/src/test/java/com/linkedin/data/avro/TestSchemaTranslator.java
```diff
@@ -70,8 +70,220 @@ public void testTranslateDefaultBackwardsCompatibility()
     assertSame(DataToAvroSchemaTranslationOptions.DEFAULT_OPTIONAL_DEFAULT_MODE, OptionalDefaultMode.TRANSLATE_DEFAULT);
   }
 
+  @DataProvider
+  public Object[][] toAvroSchemaDataTestTypeRefAnnotationPropagationUnionWithAlias()
+  {
+    return new Object[][]
+        {
+            {
+                "record test {" +
```
BrianPin, line 79: Suggest to put these test data in a file, and leave the path/file name here for the test function, and then read the file during test, this way, it is easy to see the test data clear and if we want to change it is also easy to change

junchuanwang: I believe new tests in Pegasus are already written using the conversion you mentioned

TBH in order to achieve that in this particular file, maybe it is same as we discussed today: SchemaTranslator need to be rewritten and gradually migrated all the functionality.

data-avro/src/test/java/com/linkedin/data/avro/TestSchemaTranslator.java
```diff
@@ -69,6 +69,217 @@ public void testTranslateDefaultBackwardsCompatibility()
...
   }
+  @DataProvider
+  public Object[][] toAvroSchemaData_testTypeRefAnnotationPropagationUnionWithAlias()
+  {
+    return new Object[][]
+        {
+            {
+                "record test {" +
+                    "  unionTyperef:" +
+                    "  @compliance = {" +
```
aman1309, line 81: nit: would be good to use test annotation instead of compliance here.

aman1309 (approved):

data-avro/src/test/java/com/linkedin/data/avro/TestSchemaTranslator.java
nickibi: It still has '_'

data-avro/src/test/java/com/linkedin/data/avro/TestSchemaTranslator.java
nickibi: I think what Brian suggested was using pdl file(path) as input. You can use TestUtil#dataSchemaFromPdlInputStream(getClass().getResourceAsStream(<fileName>)) to get the DataSchema. It will be better than writing pdl string in the test class. 
However, for your test case, you still need avroschema as expected result. You may still end up writing avro schema string.

data-avro/src/test/java/com/linkedin/data/avro/TestSchemaTranslator.java
```diff
@@ -308,11 +520,11 @@ public void testTranslateDefaultBackwardsCompatibility()
 
  }
 
- @Test(dataProvider = "toAvroSchemaData_testTypeRefAnnotationPropagation")
- public void testToAvroSchema_testTypeRefAnnotationPropagation(String schemaBeforeTranslation,
+ @Test(dataProvider = "toAvroSchemaDataTestTypeRefAnnotationPropagationUnionWithAlias")
+ public void testToAvroSchema_testTypeRefAnnotationPropagationUnionWithAlias(String schemaBeforeTranslation,
```
BrianPin, line 524: Name consistency?  (toAvroSchemaDataTestTypeRefAnnotationPropagationUnionWithAlias)

data-avro/src/main/java/com/linkedin/data/avro/PegasusUnionToAvroRecordConvertCallback.java
```diff
@@ -62,6 +63,14 @@ public void callback(List<String> path, DataSchema schema)
     {
       DataSchema fieldSchema = field.getType().getDereferencedDataSchema();
 
+      Map<String, Object>  propagatedProperties = fieldSchema.getProperties();
```
BrianPin, line 66: I wonder if you can just call `getMergedTyperefProperties`

like 
`Map<String, Object>  propagatedProperties = fieldSchema.getMergedTyperefProperties();`

Because looking into the `getMergedTyperefProperties` function it will do the judge for you if it is a TypeRef

data-avro/src/test/java/com/linkedin/data/avro/TestSchemaTranslator.java
junchuanwang: I understood what Brian meant, but that is really impractical since that file included a lot of (small) tests

data-avro/src/main/java/com/linkedin/data/avro/PegasusUnionToAvroRecordConvertCallback.java
junchuanwang: Hi Brian,
I think you misunderstood. The branch with judge "(getRef().getType() == Type.TYPEREF)" is for a recursive call.

this `getMergedTyperefProperties` is only available in `TyperefDataSchema`

data-avro/src/test/java/com/linkedin/data/avro/TestSchemaTranslator.java
junchuanwang: corrected

nickibi (approved):

data-avro/src/main/java/com/linkedin/data/avro/PegasusUnionToAvroRecordConvertCallback.java
BrianPin: Thanks, You are right, I didnt fint it a typeref specific method

PR #448, opened 2020-10-15
Improve translate schema task performance.

karthikbalasub:
Parse all input sources using the same parser/resolver to avoid loading referenced schemas multiple times.


restli-tools/src/main/java/com/linkedin/restli/tools/data/SchemaFormatTranslator.java
```diff
@@ -207,32 +209,31 @@ private void translateFiles() throws IOException, InterruptedException
...
+      if (schema instanceof NamedDataSchema)
+      {
+        NamedDataSchema namedDataSchema = (NamedDataSchema) schema;
+        String schemaFullname = namedDataSchema.getFullName();
+        LOGGER.debug("Loaded source schema: {}, from location: {}", schemaFullname, location.getSourceFile().getAbsolutePath());
+        topLevelTranslatedSchemas.put(schemaFullname, new SchemaInfo(namedDataSchema, location.getSourceFile(), encode(schema)));
+      }
+      else
+      {
+        LOGGER.error("Parsed a non-named schema as top-level schema.");
```
aman1309, line 234: can we print schema location (or path) with the error?

aman1309 (approved):
Rest LGTM

restli-tools/src/main/java/com/linkedin/restli/tools/data/SchemaFormatTranslator.java
junchuanwang: I believe that was handled during source parsing

restli-tools/src/main/java/com/linkedin/restli/tools/data/SchemaFormatTranslator.java
```diff
@@ -345,46 +346,6 @@ private void writeTranslatedSchemasToDirectory(
...
-          .append(topLevelSchemas.size())
-          .append(" schemas.")
-          .append("\n");
-      return  null;
-    }
-    DataSchema sourceSchema = topLevelSchemas.get(0);
-    if (!(sourceSchema instanceof NamedDataSchema) ||
-        !((NamedDataSchema) sourceSchema).getFullName().equals(schemaFullname))
-    {
-      errorMessages.append("Parsed top-level schema does not match the schema file name. File: ")
```
junchuanwang, line 378: Is this error covered after this change?

restli-tools/src/main/java/com/linkedin/restli/tools/data/SchemaFormatTranslator.java
```diff
@@ -207,32 +209,31 @@ private void translateFiles() throws IOException, InterruptedException
...
-      if (schema != null)
+      DataSchema schema = entry.getKey();
+      DataSchemaLocation location = entry.getValue();
+      if (!location.getSourceFile().getAbsolutePath().endsWith(_sourceFormat))
       {
-        topLevelTranslatedSchemas.put(schemaFullname, new SchemaInfo(schema, sourceFile, encode(schema)));
+        // Skip source files not matching the format specified.
+        continue;
+      }
+      if (schema instanceof NamedDataSchema)
```
nickibi, line 225: Parsing schemas in this way, we may not get expected result. 
DataSchemaParser.ParseResult (parsedSources) also contains embedded name schemas which are unnecessary to be top level schemas.

nickibi (commented):
I don't think this solution will work unless we could find a way to differentiate top level schemas and embedded name schemas in the ParseResult

restli-tools/src/main/java/com/linkedin/restli/tools/data/SchemaFormatTranslator.java
```diff
@@ -207,32 +209,31 @@ private void translateFiles() throws IOException, InterruptedException
...
   {
     Map<String, SchemaInfo> topLevelTranslatedSchemas = new HashMap<>();
+    DataSchemaParser dataSchemaParser = new DataSchemaParser(_resolverPath);
+    DataSchemaParser.ParseResult parsedSources = dataSchemaParser.parseSources(
+        new String[]{_sourceDir.getAbsolutePath()});
 
-    Iterator<File> iter = FileUtils.iterateFiles(_sourceDir, new String[]{_sourceFormat}, true);
-    StringBuilder errorMessages = new StringBuilder();
-    while(iter.hasNext())
+    for(Map.Entry<DataSchema, DataSchemaLocation> entry : parsedSources.getSchemaAndLocations().entrySet())
```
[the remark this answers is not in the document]
BrianPin: Dont we have style check automatically?

restli-tools/src/main/java/com/linkedin/restli/tools/data/SchemaFormatTranslator.java
BrianPin: +1, please consider the side effect of generated non-top level files

restli-tools/src/main/java/com/linkedin/restli/tools/data/SchemaFormatTranslator.java
karthikbalasub: Yes, done here: https://github.com/linkedin/rest.li/blob/master/generator/src/main/java/com/linkedin/pegasus/generator/FileFormatDataSchemaParser.java#L177

restli-tools/src/main/java/com/linkedin/restli/tools/data/SchemaFormatTranslator.java
karthikbalasub: ParseResult will not contain nested schemas. See relevant logic here:
https://github.com/linkedin/rest.li/blob/74121d54a156824aa2c7db15ade34e8852b1ea08/generator/src/main/java/com/linkedin/pegasus/generator/FileFormatDataSchemaParser.java#L259

It only adds top-level schemas to the parse result.

restli-tools/src/main/java/com/linkedin/restli/tools/data/SchemaFormatTranslator.java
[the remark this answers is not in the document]
karthikbalasub: no automated style checks in pegasus

restli-tools/src/main/java/com/linkedin/restli/tools/data/SchemaFormatTranslator.java
karthikbalasub: Yeah, this is just to keep the logic complete. The internal logic only adds Named schemas to parse result. Sadly, changing that API touched too many places, so I didn't include it here.

Added the location to the log anyways.

BrianPin (approved):
LGTM

restli-tools/src/main/java/com/linkedin/restli/tools/data/SchemaFormatTranslator.java
nickibi: Look at here: https://github.com/linkedin/rest.li/blob/894952afb12760a6d3cace8ff5a2ae44ed2149d4/generator/src/main/java/com/linkedin/pegasus/generator/FileFormatDataSchemaParser.java#L139.

restli-tools/src/main/java/com/linkedin/restli/tools/data/SchemaFormatTranslator.java
```diff
@@ -207,32 +209,31 @@ private void translateFiles() throws IOException, InterruptedException
...
+      {
+        NamedDataSchema namedDataSchema = (NamedDataSchema) schema;
+        String schemaFullname = namedDataSchema.getFullName();
+        LOGGER.debug("Loaded source schema: {}, from location: {}", schemaFullname, location.getSourceFile().getAbsolutePath());
+        topLevelTranslatedSchemas.put(schemaFullname, new SchemaInfo(namedDataSchema, location.getSourceFile(), encode(schema)));
+      }
+      else
+      {
+        LOGGER.error("Parsed a non-named schema as top-level schema.");
+        System.exit(1);
```
BrianPin, line 235: Is it worth to just skip the error and keep looping?

nickibi: No, it is a translator. It should have 1 to 1 translation. If there is a non-named schema as a top level schema, it means something wrong. We should throw exception not skip it.

restli-tools/src/main/java/com/linkedin/restli/tools/data/SchemaFormatTranslator.java
BrianPin: I think ybi means in the end of the parsing process, there will still be more non-top level schemas

karthikbalasub: Good catch. Infact the schema translator tests were failing because of that!
Fixed the logic now to check for top level schemas and also ignore schemas loaded from resolver path (not in source dir).

restli-tools/src/main/java/com/linkedin/restli/tools/data/SchemaFormatTranslator.java
karthikbalasub: This check is mainly a failsafe. The DataSchemaParse::parseSources logic already checks for this and includes errors for all schemas, so this would never be hit.

restli-tools/src/main/java/com/linkedin/restli/tools/data/SchemaFormatTranslator.java
nickibi: In the isTopLevelSchema method, we have already checked whether schema is NamedDataSchema, do we still need to do the same check here?

restli-tools/src/main/java/com/linkedin/restli/tools/data/SchemaFormatTranslator.java
```diff
@@ -207,36 +206,67 @@ private void translateFiles() throws IOException, InterruptedException
...
+      // Schema name didn't match.
+      return false;
+    }
+
+    final String parent = path.substring(0, path.lastIndexOf(File.separator));
+    // Finally check if namespace matches.
+    return parent.endsWith(namespace.replace('.', File.separatorChar));
   }
 
+
```
nickibi, line 269: nit: remove extra line

restli-tools/src/main/java/com/linkedin/restli/tools/data/SchemaFormatTranslator.java
karthikbalasub: removed the check, it was redundant.

nickibi (approved):

PR #449, opened 2020-10-15
Clear the generated rest client java directory before generating the files.

karthikbalasub:


nickibi (approved):

junchuanwang (approved):

BrianPin (approved):

evanw555:
Does the task always run, even if there are no changes? If not, then would this affect build performance?

karthikbalasub:
> Does the task always run, even if there are no changes? If not, then would this affect build performance?
The task doesn't run if there are no changes. Not sure how this change would affect performance. The deletion step will also be skipped if the task is skipped.

evanw555:
@karthikbalasub Ah, makes sense. I didn't take that last point into consideration, for some reason I thought that deleting the files would force the task to re-run even without changes. Thanks

PR #451, opened 2020-10-16
Generate specific maps to improve code generated model performance.

karthikrg:
1. Generate specific maps to improve encoding performance and reduce memory usage.
2. Generate specific datacomplex provider to improve decoding performance (Still needs plugging into the infra on client and server, will happen in a follow up PR)
3. Intern schema field names and symbol table symbols to improve symbol table lookup performance.
4. Enhance protobuf to support chunked writers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.10.1] - 2020-11-19
+- Fix bug where records wrapping the same map were not updated when setter was invoked on one record. 
+
 ## [29.10.0] - 2020-11-18
 - Fix relative load balancer log. Bumping the minor version so that it can be picked up by LinkedIn internal services.
 
@@ -4754,7 +4757,8 @@ patch operations can re-use these classes for generating patch messages.
 
 ## [0.14.1]
 
-[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.10.0...master
+[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.10.1...master
+[29.10.1]: https://github.com/linkedin/rest.li/compare/v29.10.0...v29.10.1
 [29.10.0]: https://github.com/linkedin/rest.li/compare/v29.9.2...v29.10.0
 [29.9.2]: https://github.com/linkedin/rest.li/compare/v29.9.1...v29.9.2
 [29.9.1]: https://github.com/linkedin/rest.li/compare/v29.9.0...v29.9.1
--- a/generator-test/src/test/java/com/linkedin/pegasus/generator/override/TestRecord.java
+++ b/generator-test/src/test/java/com/linkedin/pegasus/generator/override/TestRecord.java
@@ -530,4 +530,15 @@ public class TestRecord
     assertEquals(record.getRecordOptionalField().getLocation(), "near");
     assertNull(recordCopy.getRecordOptionalField());
   }
+
+  @Test
+  public void testSetOnRecordWrappingSameMap()
+  {
+    RecordBar bar = new RecordBar();
+    bar.setLocation("some");
+    RecordBar copy = new RecordBar(bar.data());
+    assertEquals(copy.getLocation(), "some");
+    copy.setLocation("other");
+    assertEquals(bar.getLocation(), "other");
+  }
 }
--- a/generator-test/src/test/java/com/linkedin/pegasus/generator/override/TestUnion.java
+++ b/generator-test/src/test/java/com/linkedin/pegasus/generator/override/TestUnion.java
@@ -272,4 +272,15 @@ public class TestUnion
     assertEquals(unionCopy.getRecordBar().getLocation(), "far");
     assertEquals(union.getRecordBar().getLocation(), "near");
   }
+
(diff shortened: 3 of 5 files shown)


li-protobuf/src/main/java/com/linkedin/data/protobuf/OutputStreamWriter.java
```diff
@@ -0,0 +1,359 @@
...
+    {
+      if ((value & ~0x7FL) == 0)
+      {
+        _buffer[_position++] = (byte) value;
+        return;
+      }
+      else
+      {
+        _buffer[_position++] = (byte) (((int) value & 0x7F) | 0x80);
+        value >>>= 7;
```
BrianPin, line 259: If we are trying to copy individual bytes to the _buffer, I think we should shift 8?
0b100000000 = 256
256 >> 8 = 1 seems to be right for this case?

data/src/main/java/com/linkedin/data/DataMap.java
```diff
@@ -110,10 +112,27 @@ public DataMap(int initialCapacity, float loadFactor)
...
   }
 
+  /**
+   * Constructs a {@link DataMap} backed by the given specific map.
+   *
+   * <p>This constructor is meant to be invoked only from code-generated models.</p>
+   *
+   * @param specificMap provides the specific map.
+   *
+   * @see HashMap
```
countmdm, line 122: This probably doesn't make sense anymore.

data/src/main/java/com/linkedin/data/DataMap.java
```diff
@@ -447,5 +554,6 @@ private void instrumentAccess(Object key)
   private boolean _madeReadOnly = false;
   private boolean _instrumented = false;
   private Map<String, Integer> _accessMap;
+  private SpecificMap _specificMap;
```
countmdm, line 557: I think a separate _specificMap is confusing - you cannot have both the _mapp field in superclass and _specificMap set to non-null values, right? The new field takes up some extra memory. Ideally, we need a deeper change that would make _map be able to accommodate a SpecificMap instance.

data/src/main/java/com/linkedin/data/codec/symbol/InMemorySymbolTable.java
```diff
@@ -40,7 +40,7 @@ public InMemorySymbolTable(String symbolTableName, List<String> symbols)
 
     for (int i = 0; i < symbols.size(); i++)
     {
-      String symbol = symbols.get(i);
+      String symbol = symbols.get(i).intern();
```
countmdm, line 43: I am all for getting rid of memory waste due to data duplication, but I also have to warn that string interning should be applied judiciously, i.e. where it really has a potential to get rid of enough duplicate data to guarantee substantial memory savings and offset the CPU performance loss. So, how big are these symbol tables typically are, and how likely are they to contain duplicates?

li-protobuf/src/main/java/com/linkedin/data/protobuf/OutputStreamWriter.java
karthikrg: See how varints are encoded.
https://developers.google.com/protocol-buffers/docs/encoding#varints

data/src/main/java/com/linkedin/data/codec/symbol/InMemorySymbolTable.java
karthikrg: Memory reduction is not the primary goal here.
The primary goal is reducing hashmap lookup for symbol name -> ordinal to instance lookup (==) instead of equals. This is safe because we are doing it only for strings in the schema effectively.

Also discussed this with @karthikbalasub offline.

data/src/main/java/com/linkedin/data/DataList.java
```diff
@@ -239,6 +272,67 @@ public int dataComplexHashCode()
...
+          //
+          // If this is backed by a specific list, delegate traversal to the specific list. Else, use Java 8 forEach
+          // to avoid intermediary object creation.
+          //
+          if (_encapsulatesSpecificList)
+          {
+            ((SpecificList) _list).traverse(callback, cycleChecker);
+          }
+          else {
+            final int[] index = {0};
```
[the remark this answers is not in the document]
karthikrg: This is actually faster per JMH, because .get(i) results in a range check everytime which foreach helps avoid.
The array allocation is on the heap but short lived, so it doesn't impact GC much.

data/src/main/java/com/linkedin/data/DataMap.java
karthikrg: This has been addressed in the latest diff. I moved it to a boolean encapsulatesSpecificMap that is set on init. This is faster than doing an instanceof check at runtime. Also this is safe because the map never changes after being assigned.

data/src/main/java/com/linkedin/data/DataMap.java
karthikrg: Why not? Without this how do we call the placeholder no copy constructor in CheckedMap?

karthikbalasub (commented):
I'm half way through the codegen part. Rest looks okay, you can start creating separate PRs.

PR #452, opened 2020-10-20
Set source compatibility on data template compile task.

karthikbalasub:
Set javac source and target compatibility of dataTemplate compile task to "1.8" as the cogen changes in 29.7.0 is using Java 8 features.


BrianPin (approved):
Seems to solve the problem

nickibi (approved):

evanw555 (approved):
Thanks for the fix

PR #453, opened 2020-10-21
Generate data template classes compatible with Java 1.6

karthikbalasub:
Modify the change listener implementation in generated data templates to  avoid using java 8 features.


nickibi (approved):

BrianPin (approved):
LGTM

karthikrg:
Please do not ship this. We are complicating the solution and reducing performance. I would like us to discuss alternatives to have internal consumers move to Java 1.8.

A lot of our future enhancements here rely on Java 8 support.

PR #454, opened 2020-10-21
[Relative Load Balancer] Improve log

rachelhanhan:
Improve the relative load balancer log:
1. Adding service name, otherwise we do not know which service has changed the degrader score
2. Limiting the size of the recovery map and quarantine map, removed points map. Otherwise for a cluster of hundreds of machines, one entry of log will be too big.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,11 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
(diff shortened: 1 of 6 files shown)


rickzx (approved):
LGTM. The only question is whether we can use String.format() to improve the readability of the toString method. But this is minor issue. If you feel that it's unnecessary, then just go with the current concatenation approach.

PR #458, opened 2020-10-23
Improve wrapped array/map template performance. 

karthikbalasub:
When wrapped objects are added/put in the templates, only the underlying DataComplex is stored before. This forces them to be wrapped again when they are read, nullifying performance improvements made by generating member variables for fields. 

This PR fixes that by caching the wrapped objects when they are added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,11 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.7.16] - 2020-10-23
+- Improve array and map template performance by caching the wrapped types when they are added.
+
 ## [29.7.15] - 2020-10-23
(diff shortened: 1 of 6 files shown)


junchuanwang (approved):

karthikrg:
Discussed offline with @karthikbalasub and we agreed not to commit this. 

This change to not cache on write was intentionally made by me in this [PR](https://github.com/linkedin/rest.li/pull/444). Why?

- There are many application use cases where we only write fields in the template and don't read them (say response formatter)
- Putting data into the hashmap backing the cache and creating the hashmap itself wastes CPU and memory

The one minor flaw of this approach is that the field cache can get invalidated for records. This will be fixed and the backing cache's hashmap be made lazy in this upcoming [PR](https://github.com/linkedin/rest.li/pull/451).

PR #460, opened 2020-10-26
Performance improved: add lazy instantiation of Throwable objects for…

ChaoLinkedIn:
… timeout errors

### DIFF
--- a/r2-core/src/main/java/com/linkedin/r2/transport/http/client/TimeoutCallback.java
+++ b/r2-core/src/main/java/com/linkedin/r2/transport/http/client/TimeoutCallback.java
@@ -21,6 +21,7 @@ import com.linkedin.r2.util.SingleTimeout;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.function.Supplier;
 
 
 /**
@@ -55,7 +56,7 @@ public class TimeoutCallback<T> implements Callback<T>
   public TimeoutCallback(ScheduledExecutorService executor, long timeout, TimeUnit timeoutUnit,
                                final Callback<T> callback, final String timeoutMessage)
   {
-    this(executor, timeout, timeoutUnit, callback, new TimeoutException(
+    this(executor, timeout, timeoutUnit, callback, () -> new TimeoutException(
         "Exceeded request timeout of " + timeoutUnit.toMillis(timeout) + "ms: " + timeoutMessage));
   }
 
@@ -72,6 +73,21 @@ public class TimeoutCallback<T> implements Callback<T>
       final Callback<T> callback, final Throwable timeoutThrowable)
   {
     _timeout = new SingleTimeout<>(executor, timeout, timeoutUnit, callback, (callbackIfTimeout) -> callbackIfTimeout.onError(timeoutThrowable));
+  }
+
+  /**
+   * Construct a new instance.
+   *
+   * @param executor the {@link ScheduledExecutorService} used to schedule the timeout
+   * @param timeout the timeout delay, in the specified {@link TimeUnit}.
+   * @param timeoutUnit the {@link TimeUnit} for the timeout parameter.
+   * @param callback the {@link Callback} to be invoked on success or error.
+   * @param timeoutThrowableSupplier the custom exception supplier that will be used during the timeout
+   */
+  public TimeoutCallback(ScheduledExecutorService executor, long timeout, TimeUnit timeoutUnit,
+      final Callback<T> callback, final Supplier<Throwable> timeoutThrowableSupplier)
+  {
+    _timeout = new SingleTimeout<>(executor, timeout, timeoutUnit, callback, callbackIfTimeout -> callbackIfTimeout.onError(timeoutThrowableSupplier.get()));
   }
 
   @Override
--- a/r2-netty/src/main/java/com/linkedin/r2/netty/client/http2/Http2ChannelLifecycle.java
+++ b/r2-netty/src/main/java/com/linkedin/r2/netty/client/http2/Http2ChannelLifecycle.java
@@ -221,7 +221,7 @@ class Http2ChannelLifecycle implements AsyncPool.Lifecycle<Channel>
           }
         });
       }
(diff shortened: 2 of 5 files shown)


pegasus-common/src/main/java/com/linkedin/common/callback/Callback.java
```diff
@@ -32,4 +35,13 @@
...
   void onError(Throwable e);
+
+  /**
+   * Called with lazy Throwable creation if the asynchronous operation failed with an error
+   * @param throwableSupplier the supplier to create Throwable when needed
+   */
+  default void onError(Supplier<Throwable> throwableSupplier)
+  {
+    onError(throwableSupplier.get());
+  }
```
nizarm, line 46: Do we need this change in the interface? Since this is leverage only inside TimeoutCallback, Why don't we resolve the lambda inside the TimeoutException itself while calling onError?

r2-netty/src/main/java/com/linkedin/r2/netty/client/http2/Http2ChannelLifecycle.java
```diff
@@ -221,7 +221,7 @@ public void onSuccess(Channel channel)
           }
         });
       }
-    }, new ObjectCreationTimeoutException(
+    }, () -> new ObjectCreationTimeoutException(
     "Exceeded creation timeout of " + _channelCreationTimeoutMs + "ms: for HTTP/2 parent channel, remote=" + _address)));
```
nizarm, line 225: Now instead of creating a throwable object, we are creating a lambda. How can we be sure that the cost of a lambda is smaller than the cost of the throwable object?

nizarm (commented):
Overall looks good to me, left some minor comments

pegasus-common/src/main/java/com/linkedin/common/callback/Callback.java
ChaoLinkedIn: Because we want to postpone the Throwable creation until #onError is called. This can save us a lot of time if onError is never called.

r2-netty/src/main/java/com/linkedin/r2/netty/client/http2/Http2ChannelLifecycle.java
ChaoLinkedIn: See pervious comments.

nizarm: I understand that we are trying to lazily create the object instance. But my question is more on the cost of creating a lambda. Since the lambda also involves a closure, wouldnt that create a seperate lambda object for each request. So the question is, did we validate if the cost of lambda object for each request is smaller than the cost of creating the Throwable itself?

nizarm, line 225: Also, note that this code path only gets invoked when there are no spare channels in the pool for the incoming request. This is not invoked for every request!

r2-core/src/main/java/com/linkedin/r2/transport/http/client/TimeoutCallback.java
```diff
@@ -67,13 +68,30 @@ public TimeoutCallback(ScheduledExecutorService executor, long timeout, TimeUnit
...
+   * @param executor the {@link ScheduledExecutorService} used to schedule the timeout
+   * @param timeout the timeout delay, in the specified {@link TimeUnit}.
+   * @param timeoutUnit the {@link TimeUnit} for the timeout parameter.
+   * @param callback the {@link Callback} to be invoked on success or error.
+   * @param timeoutThrowableSupplier the custom exception supplier that will be used during the timeout
+   */
+  public TimeoutCallback(ScheduledExecutorService executor, long timeout, TimeUnit timeoutUnit,
+      final Callback<T> callback, final Supplier<Throwable> timeoutThrowableSupplier)
+  {
+    _timeout = new SingleTimeout<>(executor, timeout, timeoutUnit, callback, callbackIfTimeout -> callbackIfTimeout.onError(timeoutThrowableSupplier));
```
nizarm, line 92: The default method introduced in Callback.java can be avoided if we invoke as follows
`callbackIfTimeout -> callbackIfTimeout.onError(timeoutThrowableSupplier.get())`

nizarm (approved):

pegasus-common/src/main/java/com/linkedin/common/callback/Callback.java
ssheng, line 46: Recommend to avoid this change to a simple interface. I think the suggestion below for calling supplier.get() can help us avoid the need of this change.

ChaoLinkedIn: We can. But the question is wether we want to limit this optimization to TimeoutCallBack or CallBack at all. If you check how many implementation of CallBack, you'll notice that there are a lot of chances of optimization. Of course most of them have limited usage so no need to do anything for now. But should we leave this option open so that when needed, we do not have to come back to change it?

r2-netty/src/main/java/com/linkedin/r2/netty/client/http2/Http2ChannelLifecycle.java
ChaoLinkedIn: According to this (https://stackoverflow.com/questions/2347828/how-expensive-is-thread-getstacktrace/10625065#10625065), the answer is definitely yet.

ChaoLinkedIn: Agree the save here is small. It's just one of the spots that our customer identified.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/TimeoutCallback.java
ChaoLinkedIn: Agree. See my earlier comments.

pegasus-common/src/main/java/com/linkedin/common/callback/Callback.java
ChaoLinkedIn: See my earlier comments.

PR #462, opened 2020-10-27
Set ValidationResult to false if IllegalArgument exceptions are seen during iterative validation using ObjectIterator

junchuanwang:
When we validate insidie "ValidateDataAgainstSchema", it will propagete IllegalArgumentException, this eventaully leads to HTTP 500.

The IllegalArgumentException is thrown by ObjectIterator when data and data schema does not match.

We should consider "IllegalArgumentException" as an validation failure, so we can throw HTTP 422 (invalid entity)


data/src/main/java/com/linkedin/data/schema/validation/ValidateDataAgainstSchema.java
```diff
@@ -173,9 +166,22 @@ protected void validateIterative(DataElement element)
...
+        try
+        {
+          if ((nextElement = it.next()) == null)
+          {
+            break;
+          }
+        }
+        catch (IllegalArgumentException e)
+        {
+          addMessage(nextElement, e.getMessage());
```
karthikbalasub, line 181: Also log the exception for debugging.

karthikbalasub (changes requested):

data/src/main/java/com/linkedin/data/schema/validation/ValidateDataAgainstSchema.java
junchuanwang: There has never been any logs in any Validators, either schema or data validators.

Also I am not sure if need them as theyare caught in the validation result

nickibi (approved):

karthikbalasub (approved):

PR #465, opened 2020-10-29
Support symbolTable requests with suffixes

sangm:
We have a client who adds suffix to the request path.

Given an application named `foo` and a resource with `bar`

```
@RestLiCollection(name = "bar", namespace = "com.foo")
```

they want to support a d2 URL such as

```
d2://foo?action=...
```

the resolved url becomes

```
some-host.com/foo/bar?action=...
```

When we enable protobuf for them, the `RestLiSymbolTableRequestHandler` sees

```
some-host.com/foo/bar/symbolTable
some-host.com/foo/bar/symbolTable/<tableName>
```

In this PR, we are adding support to check the last 2 arguments given a path.

To prevent naming collision and backwards compatibility, we are guarding the
logic behind a header

### DIFF
--- a/restli-server/src/main/java/com/linkedin/restli/server/symbol/RestLiSymbolTableRequestHandler.java
+++ b/restli-server/src/main/java/com/linkedin/restli/server/symbol/RestLiSymbolTableRequestHandler.java
@@ -93,16 +93,8 @@ public class RestLiSymbolTableRequestHandler implements NonResourceRequestHandle
     //
     // When path is service scoped, URI is in the form of /<SERVICE>/symbolTable, else it
     // is in the form of /symbolTable or /symbolTable/<TABLENAME>
-    //
-    boolean isServiceScopedPath = request.getHeaders().containsKey(RestConstants.HEADER_SERVICE_SCOPED_PATH);
-    if (isServiceScopedPath)
-    {
-      return (pathSegments.size() == 3 && pathSegments.get(2).getPath().equals(SYMBOL_TABLE_URI_PATH));
-    }
-    else
-    {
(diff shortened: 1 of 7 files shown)


restli-server/src/main/java/com/linkedin/restli/server/symbol/RestLiSymbolTableRequestHandler.java
```diff
@@ -95,13 +97,22 @@ public boolean shouldHandle(Request request)
...
     //
     boolean isServiceScopedPath = request.getHeaders().containsKey(RestConstants.HEADER_SERVICE_SCOPED_PATH);
+    boolean isSymbolTableHeaderPresent = request.getHeaders().containsKey(SYMBOL_TABLE_REQUEST_HEADER);
+    int pathSize = pathSegments.size();
     if (isServiceScopedPath)
     {
-      return (pathSegments.size() == 3 && pathSegments.get(2).getPath().equals(SYMBOL_TABLE_URI_PATH));
+      return (pathSize == 3 && pathSegments.get(2).getPath().equals(SYMBOL_TABLE_URI_PATH));
+    }
+    else if (isSymbolTableHeaderPresent)
```
karthikrg, line 106: I don't like the idea of adding another header for this. It will complicate the exchange protocol further. Is there some other way we can do this? Like say check routing to see if a resource with that name exists or not?

Or even better add a rule to disallow a rest.li resource from being named symbolTable.

@kbalasub wdyt?

sangm: Having a reserved resource would make sense and seems most straightforward.

karthikrg: @aman1309 can you please work with @sangm on the best path forward here?

aman1309 (approved):

PR #466, opened 2020-10-29
Empty commit to bump minor version

rachelhanhan:
This PR doesn't do anything, it only bump the minor version of pegasus.
Why do we need this?
I need this for LinkedIn HI preparation. For the HI to onboard to relative load balancer, the prerequisites is to have all the mps using a certain container version and a certain pegasus version. To reduce the effort of asking everyone to bump the pegasus version, we want container to give an error when clients downgrade the pegasus version, and we agreed with SF team that the downgrade error will happen on a minor version downgrade. Therefore, I'm upgrading the minor version and will use this version in the container.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.8.0] - 2020-10-29
+- Empty commit to bump pegasus minor version
+
 ## [29.7.15] - 2020-10-23
 Log Streaming Error or Timeout Error in Jetty SyncIOHandler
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.8.0] - 2020-10-29
+- Empty commit to bump pegasus minor version
```
ssheng, line 18: Please add more contexts.

ssheng (approved):

CHANGELOG.md
rachelhanhan: I added a little bit more context, but still want to keep it short in the commit history:
1. Commit history is meant to be short, if they want any more context, they should come to this PR description.
2. Pegasus is open-source, I don't think it's proper to mention all the HI context in this open source.

PR #468, opened 2020-11-05
Update generators to optionally generate files with lower case path.

Astro03:
**Issue**:
Was working on two differnet file systms and noticed that after a rync,
were dupiclate files in similar paths.

case insensitive file system, "\~/com/astro" and "\~/com/ASTRO" point to
the same folder. For a case sensitive file system, "\~/com/astro" and
"\~/com/ASTRO" will be different folders.

 **Example**:

       com/astro/file1
                /file2
       com/aSTRo/file1
                /file2
   3) Case sensitive:
       com/astro/file1
           aSTRo/file2

**Resolution**:
Add abililty for generators to generate paths with lower
case path. Updated the TempalteSpecGenerator,
PegasusDataTemplateGenerator, RestRequestBuilderGenerator, and
JavaRequestBuilderGenerator.

The Gradle Tasks of GenerateDataTemplateTask and
GenerateRestClientTask have been modified to have an optional input property
to generate teh lower case path.

**Testing**:
respectively.


Astro03:
@karthikbalasub @junchuanwang  
Can you both please look at this and see if I missed anything. Thank you.

generator/src/main/java/com/linkedin/pegasus/generator/JavaCodeUtil.java
```diff
@@ -146,7 +157,17 @@ else if (!checker.isPersistent(definedClass))
         }
         else if (definedClass.outer() == null)
         {
-          final File file = new File(targetDirectory, definedClass.fullName().replace('.', File.separatorChar) + ".java");
+          String path;
+          if (generateLowercasePath) {
+            // Create path this way since fullName() has a recursive call.
+            String fullName = definedClass.fullName();
+            String name = definedClass.name();
+            String packageName = fullName.substring(0, fullName.length() - name.length());
```
karthikbalasub, line 165: Use definedClass.getPackage().name() instead?

generator/src/main/java/com/linkedin/pegasus/generator/PegasusDataTemplateGenerator.java
```diff
@@ -49,6 +44,7 @@
    * The system property that specifies whether to generate classes for externally resolved schemas
    */
   public static final String GENERATOR_GENERATE_IMPORTED = "generator.generate.imported";
+  public static final String GENERATOR_GENERATE_PATH_LOWERCASE = "generator.generate.path.lowercase";
```
karthikbalasub, line 47: rename for consistency:
PATH_LOWERCASE -> LOWERCASE_PATH

generator/src/main/java/com/linkedin/pegasus/generator/JavaCodeUtil.java
```diff
@@ -17,26 +17,22 @@
...
 import java.net.URI;
 import java.net.URL;
 import java.net.URLClassLoader;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 import java.util.StringTokenizer;
 
-import com.sun.codemodel.JAnnotationUse;
```
karthikbalasub, line 32: Revert these changes, pegasus codestyle doesn't allow wildcard imports (except for static methods)

generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
```diff
@@ -635,7 +624,8 @@ private ClassTemplateSpec generateUnion(UnionDataSchema schema, TyperefDataSchem
     pushCurrentLocation(_schemaResolver.nameToDataSchemaLocations().get(typerefDataSchema.getFullName()));
 
     final UnionTemplateSpec unionClass = new UnionTemplateSpec(schema);
-    unionClass.setNamespace(typerefDataSchema.getNamespace());
+    String namespace = _lowercaseNamespace ? typerefDataSchema.getNamespace().toLowerCase() : typerefDataSchema.getNamespace();
+    unionClass.setNamespace(namespace);
```
karthikbalasub, line 628: This is changing the namespace on the generate template classes and it is a breaking change. This config should only change the generated directories

restli-common/src/main/java/com/linkedin/restli/restspec/RestSpecCodec.java
```diff
@@ -81,8 +80,27 @@ public RestSpecCodec()
...
+   * @throws IOException
+   */
+  public ResourceSchema readResourceSchema(InputStream inputStream, boolean generateLowercasePath)
+          throws IOException{
     final DataMap data = _dataCodec.readMap(inputStream);
     fixupLegacyRestspec(data);
+    if (generateLowercasePath && data.containsKey("namespace")) {
+      String namespace = data.getString("namespace");
+      if (namespace != null) {
+        data.put("namespace", namespace.toLowerCase());
```
karthikbalasub, line 101: Same here, avoid changing the namespace of the schema

restli-tools/src/main/java/com/linkedin/restli/tools/clientgen/RestRequestBuilderGenerator.java
```diff
@@ -54,6 +47,7 @@
 {
   static final String GENERATOR_REST_GENERATE_DATATEMPLATES = "generator.rest.generate.datatemplates";
   static final String GENERATOR_REST_GENERATE_VERSION = "generator.rest.generate.version";
+  public static final String GENERATOR_REST_GENERATE_PATH_LOWERCASE = "generator.rest.generate.path.lowercase";
```
karthikbalasub, line 50: LOWERCASE_PATH

karthikbalasub (changes requested):

generator/src/main/java/com/linkedin/pegasus/generator/JavaCodeUtil.java
Astro03: Did you read the comment? I'm not usually in this codebase but I saw that there was a recursive call in the way fullName worked and thus did it this way. Obviously creating it with the package and doing tolower there would be faster but i wasn't sure if it would produce the same result. 

thoughts?

Astro03: @karthikbalasub ^

generator/src/main/java/com/linkedin/pegasus/generator/PegasusDataTemplateGenerator.java
Astro03: My original thought was that having grouping of "path" with leaf node "lowercase" made more sense than having a "lowercase" grouping with "path" as a leaf node. 

I will change if this didn't  logic didn't make sense to you.

generator/src/main/java/com/linkedin/pegasus/generator/JavaCodeUtil.java
Astro03: Oh okay. I should do that to all the files. These were done by IntelliJ.  I'll do it for all files. Do we have a linter for this that would catch this issue?

Astro03: reverted. 

I do have concern about needing to change the codestyle since IntelliJ natively did some of these changes without my intervention.

generator/src/main/java/com/linkedin/pegasus/generator/JavaCodeUtil.java
karthikbalasub: Let us keep it this way to be safe.

restli-common/src/main/java/com/linkedin/restli/restspec/RestSpecCodec.java
Astro03: I would love to get on a call and discuss this with you. 

If i remove this, then the the paths will not change for the builders.

Astro03:
Closing since we will go with a different approach. Will change the the file writer class directly -> smarter approach.

Astro03:
Created these PRs with different approaches. 

https://github.com/linkedin/rest.li/pull/473 - all generated paths are lower case
https://github.com/linkedin/rest.li/pull/474 - default generated paths are lower case; use optional flag to make it as it was.

PR #470, opened 2020-11-05
Check whether schemas exist or not before running pegasusSchemaSnapshotCheck task

nickibi:
Check whether schemas exist or not before running pegasusSchemaSnapshotCheck task

### DIFF
--- a/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
+++ b/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
@@ -1423,6 +1423,12 @@ public class PegasusPlugin implements Plugin<Project>
     File schemaDir = isExtensionSchema? project.file(getExtensionSchemaPath(project, sourceSet))
         : project.file(getDataSchemaPath(project, sourceSet));
 
+    if ((isExtensionSchema && !SharedFileUtils.getSuffixedFiles(project, schemaDir, PDL_FILE_SUFFIX).isEmpty()) ||
+            (!isExtensionSchema && !SharedFileUtils.getSuffixedFiles(project, schemaDir, DATA_TEMPLATE_FILE_SUFFIXES).isEmpty()))
+    {
+        return;
+    }
+
     Path publishablePegasusSchemaSnapshotDir = project.getBuildDir().toPath().resolve(sourceSet.getName() +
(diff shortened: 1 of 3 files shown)


gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1441,8 +1447,8 @@ protected void configurePegasusSchemaSnapshotGeneration(Project project, SourceS
           task.dependsOn(generatePegasusSchemaSnapshot);
           task.setCurrentSnapshotDirectory(publishablePegasusSchemaSnapshotDir.toFile());
           task.setPreviousSnapshotDirectory(pegasusSchemaSnapshotDir);
-          task.setCodegenClasspath(project.getConfigurations() .getByName(SCHEMA_ANNOTATION_HANDLER_CONFIGURATION)
-              .plus(project.getConfigurations().getByName(PEGASUS_PLUGIN_CONFIGURATION))
+          task.setCodegenClasspath(project.getConfigurations().getByName(PEGASUS_PLUGIN_CONFIGURATION)
```
evanw555, line 1450: Does changing the ordering here fix the problem we talked about earlier (related to the wrong pegasus version being picked up in the schema annotation handler configuration)?

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -850,7 +850,13 @@ public void apply(Project project)
         configurePegasusSchemaSnapshotGeneration(project, sourceSet, false);
       }
 
-      configurePegasusSchemaSnapshotGeneration(project, sourceSet, true);
+      // Do compatibility check if there are extension schemas.
+      if (!SharedFileUtils.getSuffixedFiles(project,  project.file(getExtensionSchemaPath(project, sourceSet)),
```
evanw555, line 854: Looking at other methods like `configureRestClientConfiguration`, this sort of conditional check is done inside the method. I'd suggest making it consistent.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
nickibi: Yes, it fixes the problem

evanw555 (approved):
LGTM

PR #471, opened 2020-11-07
Fix bug: if there is no input schema, do not run pegasusSchemaSnapshotCheck

nickibi:
Fix bug: 
If there is no input schema, do not run pegasusSchemaSnapshotCheck. The check statement was wrong.


BrianPin (approved):
LGTM

PR #472, opened 2020-11-09
Adding new required record field is considering as backward compatible change in extension schemas

nickibi:
The CompatibilityChecker treats adding new required record field as backward incompatible changes. 
For entity relationship - extension schema, adding new required record field is allowed and it should be considered as backward compatible change. 

Update PegasusSchemaSnapshotCompatibilityChecker to allow such change for extension schemas.


restli-tools/src/main/java/com/linkedin/restli/tools/snapshot/check/PegasusSchemaSnapshotCompatibilityChecker.java
```diff
@@ -266,7 +268,19 @@ public CompatibilityInfoMap checkPegasusSchemaCompatibility(String prevPegasusSc
 
     if (!result.getMessages().isEmpty())
     {
-      result.getMessages().forEach(message -> _infoMap.addModelInfo(message));
+      if (_isExtension)
```
karthikbalasub, line 271: A cleaner solution will be to add a new Mode `EXTENSION` in CompatibilityOptions::Mode and use that to change the behavior of CompatibilityChecker

junchuanwang: Agree that additional mode will be better than an additional flag. we should keep flags as few as possible.

PR #473, opened 2020-11-09
Update generators to generate files with lower case path.

Astro03:
Issue:
Was working on two different file systems and noticed that after a rync,
were duplicate files in similar paths.

case insensitive file system, "~/com/astro" and "~/com/ASTRO" point to
the same folder. For a case sensitive file system, "~/com/astro" and
"~/com/ASTRO" will be different folders.

 Example:

       com/astro/file1
                /file2
       com/aSTRo/file1
                /file2
   3) Case sensitive:
       com/astro/file1
           aSTRo/file2

Resolution:
Force generators to generate paths with lowercase path. Updated the
PegasusDataTemplateGenerator and RestRequestBuilderGenerator.

Testing:
respectively.

### DIFF
--- a/generator/src/main/java/com/linkedin/pegasus/generator/JavaCodeUtil.java
+++ b/generator/src/main/java/com/linkedin/pegasus/generator/JavaCodeUtil.java
@@ -146,7 +146,12 @@ public class JavaCodeUtil
         }
         else if (definedClass.outer() == null)
         {
-          final File file = new File(targetDirectory, definedClass.fullName().replace('.', File.separatorChar) + ".java");
+          // This seems like the safer approach to create path since fullName() has a recursive call.
+          String fullName = definedClass.fullName();
+          String name = definedClass.name();
+          String packageName = fullName.substring(0, fullName.length() - name.length());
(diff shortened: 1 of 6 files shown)


junchuanwang (commented):
is this a reimplementation for https://github.com/linkedin/rest.li/pull/468/files? 

But what is the current way to opt out? (like the system property you set in PegasusDataTemplateGenerator.java in your previous PR)

Astro03:
> is this a reimplementation for https://github.com/linkedin/rest.li/pull/468/files?
> 
> But what is the current way to opt out? (like the system property you set in PegasusDataTemplateGenerator.java in your previous PR)

Here is the one updated with the optional flag: https://github.com/linkedin/rest.li/pull/474

I just think it's cleaner if we force update the paths to be lowercase for generated files. Doing this would be safe since we are talking about generated files.

Astro03:
forced update to trigger a rebuild.

Astro03:
Created https://github.com/linkedin/rest.li/pull/474 to make it optional for path being created older way or with all lower case.

PR #474, opened 2020-11-09
Update generators to optionally generate files with lower case path.

Astro03:
Issue:
Was working on two different file systems and noticed that after a rync,
were duplicate files in similar paths.

case insensitive file system, "~/com/astro" and "~/com/ASTRO" point to
the same folder. For a case sensitive file system, "~/com/astro" and
"~/com/ASTRO" will be different folders.

 Example:

       com/astro/file1
                /file2
       com/aSTRo/file1
                /file2
   3) Case sensitive:
       com/astro/file1
           aSTRo/file2

Resolution:
Add abililty for generators to generate paths with lower
case path. Updated the TempalteSpecGenerator,
PegasusDataTemplateGenerator, RestRequestBuilderGenerator, and
JavaRequestBuilderGenerator.

The Gradle Tasks of GenerateDataTemplateTask and
GenerateRestClientTask have been modified to have an optional input property
to generate teh lower case path.

Testing:
respectively.

### DIFF
--- a/restli-tools/src/test/java/com/linkedin/restli/tools/clientgen/TestRestRequestBuilderGenerator.java
+++ b/restli-tools/src/test/java/com/linkedin/restli/tools/clientgen/TestRestRequestBuilderGenerator.java
@@ -3,6 +3,7 @@ package com.linkedin.restli.tools.clientgen;
 
 import com.github.javaparser.StaticJavaParser;
 import com.github.javaparser.ast.body.MethodDeclaration;
+import com.linkedin.pegasus.generator.GeneratorResult;
 import com.linkedin.restli.internal.common.RestliVersion;
 import com.linkedin.restli.tools.ExporterTestUtils;
 
@@ -11,7 +12,9 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileReader;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.List;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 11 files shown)


junchuanwang (commented):
Thanks for the PR. I am afraid I will ask you to do a quick update to the PR to conform to our  code style regarding "curly brace"

junchuanwang (commented):
remind me please, but I remember last time we talked about we don’t need to change dateTemplate generator? @karthikbalasub  Karthik can you also confirm?

Astro03:
Created https://github.com/linkedin/rest.li/pull/473 to force all generated path creations to be in lower case. 
473 is simplier.

karthikbalasub (approved):

karthikbalasub:
Make sure to update changelog and do a minor version bump for this change.

restli-tools/src/test/java/com/linkedin/restli/tools/clientgen/TestRestRequestBuilderGenerator.java
```diff
@@ -121,6 +134,89 @@ public void testGeneration(RestliVersion version, String ABuildersName, String B
...
+   *       com/astro/file1
+   *           aSTRo/file2
+   *
+   * @param version RestLi version
+   * @param restspec1 First restli spec to generate
+   * @param restspec2 Second restli spec to generate
+   * @param generateLowercasePath True, generate path lowercase; False, generate path as spec specifies.
+   */
+  @Test(dataProvider = "arrayDuplicateDataProvider2")
+  public void testGenerationPathOrder(RestliVersion version, String restspec1, String restspec2, boolean generateLowercasePath) throws Exception
```
junchuanwang, line 164: Can you also add a simple test to show that your change only affect "package" name, but not the root directory (because some user uses custom root directory instead of the sourceSet directory `<module>/src/main`

Astro03: Added new test.

junchuanwang (approved):

PR #479, opened 2020-11-18
Fix log issue in relative load balancer

rachelhanhan:
I just realized that one of my previous change changed log level from info to debug, it should be info level when the state has been changed under relative load balancer. Hence updating it to the correct behavior.


rickzx (approved):
LGTM

PR #480, opened 2020-11-18
remove empty map as invalid union

BrianPin:
Remove the empty union as invalid map.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,9 @@ and what APIs have changed, if applicable.
 ## [Unreleased]
 
 ## [29.11.1] - 2020-11-20
-- Fix the issue when validation on empty union will fail
+- When we do validation on response, in the past empty unions will fail the validation and client will
+  - fail. Now we do not treat empty union as a failure, and just return the empty map as is.
+  - Also, if there are projection, the projection will apply to empty union if it is projected.
 
 ## [29.10.1] - 2020-11-19
 - Fix bug where records wrapping the same map were not updated when setter was invoked on one record. 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
+
+## [29.11.1] - 2020-11-20
 - Fix the issue when validation on empty union will fail
 
 ## [29.10.1] - 2020-11-19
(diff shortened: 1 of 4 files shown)


data/src/main/java/com/linkedin/data/schema/validation/ValidateDataAgainstSchema.java
```diff
@@ -505,11 +505,11 @@ else if (object instanceof DataMap)
       {
         // Pegasus mode
         DataMap map = (DataMap) object;
-        if (map.size() != 1)
+        if (map.size() > 1)
```
nickibi, line 508: Do you mind adding a comment here that we allow empty union

data/src/test/java/com/linkedin/data/schema/validation/TestValidation.java
```diff
@@ -1111,7 +1111,6 @@ public void testUnionNoCoercionValidation() throws IOException
         new Double(1),
         new String(),
         new DataList(),
-        new DataMap(),
```
nickibi, line 1114: shall we move this case to goodObjects?

BrianPin: Added to good object

data/src/main/java/com/linkedin/data/schema/validation/ValidateDataAgainstSchema.java
BrianPin: Added some comment

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
+- Fix the issue when validation on empty union will fail
```
nickibi, line 16: I would recommend to rephrase this description.

BrianPin: I changed it in different way.

nickibi (approved):

data/src/main/java/com/linkedin/data/schema/validation/ValidateDataAgainstSchema.java
```diff
@@ -505,11 +505,12 @@ else if (object instanceof DataMap)
       {
         // Pegasus mode
         DataMap map = (DataMap) object;
-        if (map.size() != 1)
+        // we allow empty union
+        if (map.size() > 1)
```
karthikbalasub, line 509: This will allow empty unions even if they were not caused by a projection. 

I was thinking we should allow only when the union has projection applied. One option is to add a custom property (e.g, "$projectedUnion" = "true" ) when we generate the schema from projection and use that to allow empty union here.
What do you think?

BrianPin: I thought about it when I was working on it (because it was written in the ticket). 
After making the change, I did some manual integration test and when there is a projection, there will only contain the projected field. I think the projection happens after validation. I think the projection still comes to the picture.

I may have probably misunderstood your idea, feel free to comment

BrianPin: The other thought on it was that to introduce projection into validation seems like bound to make it a miss..but that was probably me.

karthikbalasub: AFAIK this is the flow today:

Request -> 1. Validation Filter (create subset schema from projection) -> 2. Resource -> 3. Apply Projection -> 4. Validation Filter (validate using projection schema)

So we can add a flag in the UnionDataSchema created in step 1 using the projection. And then use that flag in the check done in step 4. It is slightly hacky as we are using a property. A cleaner approach would be to add a flag in the union schema to say if empty union data is allowed.

PR #482, opened 2020-11-19
Fix bug where records/unions wrapping the same map were not updated when setting a field via the other record/union

karthikrg:


generator-test/src/test/java/com/linkedin/pegasus/generator/override/TestRecord.java
```diff
@@ -530,4 +530,15 @@ public void testCopyChangeRecordField() throws CloneNotSupportedException
...
+
+  @Test
+  public void testSetOnRecordWrappingSameMap()
+  {
+    RecordBar bar = new RecordBar();
+    bar.setLocation("some");
+    RecordBar copy = new RecordBar(bar.data());
+    assertEquals(copy.getLocation(), "some");
+    copy.setLocation("other");
+    assertEquals(bar.getLocation(), "other");
```
BrianPin, line 542: What is `bar`'s location right now?

generator-test/src/test/java/com/linkedin/pegasus/generator/override/TestRecord.java
```diff
@@ -530,4 +530,15 @@ public void testCopyChangeRecordField() throws CloneNotSupportedException
     assertEquals(record.getRecordOptionalField().getLocation(), "near");
     assertNull(recordCopy.getRecordOptionalField());
   }
+
+  @Test
+  public void testSetOnRecordWrappingSameMap()
+  {
+    RecordBar bar = new RecordBar();
+    bar.setLocation("some");
+    RecordBar copy = new RecordBar(bar.data());
```
BrianPin, line 539: What about deep copy (if we have any)

karthikrg: Those are already tested via other test cases.

BrianPin (approved):
LGTM, please think about wc-test on deco and container.
Approve after asking some questions.
The test seems fine to me, please take few minutes to think more if there is corner cases.

junchuanwang (approved):

PR #484, opened 2020-11-21
Overload Failure Retry Implementation

rickzx:
Overload detectors are developed to recognize CPU overload, near out of memory situations, and request pileups with increased latency. Once overload is detected, some amount of requests will be dropped by the framework before reaching application code and an HTTP/1.1 503 Service Unavailable error will be thrown.

This implementation provides a client retry mechanism that automatically retries certain requests if it is reasonable to do so. 

In the case of overload failure, we must distinguish between two different situations:

_A. Only a single backend host or a small subset of backend hosts in the cluster are overloaded.
B. A large subset of backend hosts in the cluster are overloaded._

For case A, it is very likely that the backend cluster has remaining capacity in other hosts to handle the request. Therefore, it is safe to retry the request immediately by sending the request to a different host in the same cluster.

For case B, the entire backend cluster is degraded. Retry will not only result in another overload failure, but also cause extra burden in the backend cluster. Therefore, the request should not be retried and the overload failure should bubble all the way up to the caller.

Our approach is to make the retry decision on the **server side** by attaching a ServerRetryTracker into the ServerRetryFilter layer. It stores the number of requests categorized by number of retry attempts. It uses the information to estimate a ratio of how many requests are being retried in the cluster. The ratio is then compared with  {@link ServerRetryFilter#_maxRequestRetryRatio} to make a decision on whether or not to retry in the next interval. When calculating the ratio, it looks at the last {@link ServerRetryTracker#_aggregatedIntervalNum} intervals by aggregating the recorded requests.

This simulator for the ServerRetryTracker and ServerRetryFilter can be found here: https://github.com/linkedin/rest.li/pull/483


rachelhanhan (commented):
First pass

Will we add some inGraph monitors in next PR/RB?

r2-core/src/main/java/com/linkedin/r2/filter/transport/ServerRetryFilter.java
```diff
@@ -74,14 +129,132 @@ public void onStreamError(Throwable ex,
...
+        {
+          _aggregatedRetryAttemptsCounter[i] -= intervalToDiscard[i];
+        }
+      }
+
+      // append a new interval
+      _retryAttemptsCounter.addLast(new int[_retryLimit + 1]);
+    }
+
+    public double getRetryRatio()
```
bbarkley, line 239: The concurrency safety of this class is a little hard to follow currently. It looks like things should be safe, because you are using this and `rollOverStats` while holding a lock, and `add` and `rollOverStats` are synchronized and should be safe to use together. It would be nice to have the safety be much clearer though.

One option would be to make this synchronized for clarity. This will introduce contention with calls to `add`, but the method is small and should execute quickly, and not very frequently. Another option would be to consolidate this into a `rolloverStatsAsNeededAndGetRetryRatio` where this class can manage the thread safety and use separate locks to guard the list and the array. Finally you could just document the current behavior with comments.

Whichever way you choose I'd suggest annotating the class with `@ThreadSafe` or `@NotThreadSafe`, and annotate the fields accessed concurrently with `@GuardedBy`.

r2-core/src/main/java/com/linkedin/r2/transport/http/common/HttpConstants.java
```diff
@@ -11,6 +11,11 @@
    */
   public static final String HEADER_RESPONSE_COMPRESSION_THRESHOLD = "X-Response-Compression-Threshold";
 
+  /**
+   * Custom header for the number of retries.
+   */
+  public static final String HEADER_NUMBER_OF_RETRY_ATTEMPTS = "X-Number-Of-Retries-Attempts";
```
bbarkley, line 17: Suggest changing "Retries" to "Retry"

bbarkley (commented):
Great stat here - this is a first pass review but was pretty thorough so hopefully won't find other issues in followups.

d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
```diff
@@ -254,16 +330,33 @@ public void onError(Throwable e)
...
+      for (Throwable throwable: throwables)
+      {
+        if (throwable instanceof RetriableRequestException)
+        {
+          hasRetriableRequestException = true;
+
+          if (((RetriableRequestException) throwable).getDoNotRetryOverride())
+          {
+            return false;
+          }
```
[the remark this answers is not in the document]
rickzx: Fixed

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
```diff
@@ -146,6 +146,8 @@ public D2Client build()
                   _config._executorService,
                   _config.retry,
                   _config.retryLimit,
+                  _config.maxClientRequestRetryRatio,
```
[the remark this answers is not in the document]
rickzx: There's only one D2Client per war making calls to different downstream service endpoints. I think it makes more sense to track retry ratio individually for each downstream endpoint. The latest commit uses a map to track retry info for each extracted service name.

The ratio should be defined by the server in the TransportClientProperties, because service owners have the best knowledge of how much retry traffic is allowed. This is also reflected in the latest commit.

r2-core/src/main/java/com/linkedin/r2/RetriableRequestException.java
```diff
@@ -66,4 +68,14 @@ public RetriableRequestException(Throwable cause)
   {
     super(cause);
   }
+
+  public void setDoNotRetryOverride(boolean doNotRetryOverride)
```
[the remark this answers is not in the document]
rickzx: Added

r2-core/src/main/java/com/linkedin/r2/filter/transport/ServerRetryFilter.java
```diff
@@ -74,14 +129,132 @@ public void onStreamError(Throwable ex,
...
+  }
+
+  private void updateRetryDecision()
+  {
+    long currentTime = _clock.currentTimeMillis();
+
+    synchronized (_lock)
+    {
+      // Check if the current interval is stale
+      if (currentTime >= _lastRollOverTime + _updateIntervalMs)
```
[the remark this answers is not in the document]
rickzx: The lock is to protect two threads from entering the critical section simultaneously. If we put the lock inside the "if" block, then two threads might both pass the staleness check and begin to update the tracker. Anyways, I rewrote the synchronization logic by centralizing it in the tracker. Please take a look.

d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
```diff
@@ -59,15 +62,39 @@
  */
 public class RetryClient extends D2ClientDelegator
 {
+  public static final long DEFAULT_UPDATE_INTERVAL_MS = 5000L;
```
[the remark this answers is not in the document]
rickzx: Fixed

d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
```diff
@@ -154,11 +216,14 @@ public void onSuccess(ByteString result)
...
 
     @Override
-    public boolean doRetryRequest(StreamRequest request, RequestContext context)
+    public boolean doRetryRequest(StreamRequest request, RequestContext context, int numberOfRetryAttempts)
     {
       if (_recorded == true && _content != null)
       {
-        final StreamRequest newRequest = request.builder().build(EntityStreams.newEntityStream(new ByteStringWriter(_content)));
+        final StreamRequest newRequest = request.builder()
+            .addHeaderValue(HttpConstants.HEADER_NUMBER_OF_RETRY_ATTEMPTS, Integer.toString(numberOfRetryAttempts))
```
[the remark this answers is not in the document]
rickzx: Actually, this function will only be called when the request is being retried, so the attempt here is at least 1. But I made another change in `restRequest` and `streamRequest` to still send the header when the attempt is zero. See below

d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
```diff
@@ -234,9 +303,16 @@ public void onError(Throwable e)
             int attempts = exclusionSet.size();
             if (attempts <= _limit)
             {
-              LOG.warn("A retriable exception happens. Going to retry. This is attempt {}. Current exclusion set: ",
-                  attempts, ". Current exclusion set: " + exclusionSet);
-              retry = doRetryRequest(_request, _context);
+              if (isBelowClientRetryRatio())
+              {
+                LOG.warn("A retriable exception happens. Going to retry. This is attempt {}. Current exclusion set: {}",
```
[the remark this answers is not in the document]
rickzx: Fixed

r2-core/src/main/java/com/linkedin/r2/filter/transport/ClientRetryFilter.java
```diff
@@ -73,9 +74,13 @@ public void onStreamError(Throwable ex,
...
       nextFilter.onError(new RetriableRequestException(retryAttr), requestContext, wireAttrs);
     }
-    else
-    {
-      nextFilter.onError(ex, requestContext, wireAttrs);
+    else {
+      Throwable[] throwables = ExceptionUtils.getThrowables(ex);
+      for (Throwable throwable : throwables) {
+        if (throwable instanceof RetriableRequestException) {
+          ((RetriableRequestException) throwable).setDoNotRetryOverride(true);
```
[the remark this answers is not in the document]
rickzx: Fixed

r2-core/src/main/java/com/linkedin/r2/filter/transport/ServerRetryFilter.java
```diff
@@ -45,6 +49,57 @@
 public class ServerRetryFilter implements RestFilter, StreamFilter
 {
   private static final Logger LOG = LoggerFactory.getLogger(ServerRetryFilter.class);
+  private static final int DEFAULT_RETRY_LIMIT = 3;
+  private static final long DEFAULT_UPDATE_INTERVAL_MS = 5000L;
```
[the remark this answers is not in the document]
rickzx: Fixed

r2-core/src/main/java/com/linkedin/r2/filter/transport/ServerRetryFilter.java
```diff
@@ -74,14 +129,132 @@ public void onStreamError(Throwable ex,
...
       cause = cause.getCause();
     }
 
     nextFilter.onError(ex, requestContext, wireAttrs);
   }
+
+  private int getRetryAttempts(Request req)
+  {
+    String retryAttemptsHeader = req.getHeader(HttpConstants.HEADER_NUMBER_OF_RETRY_ATTEMPTS);
+    return retryAttemptsHeader == null ? 0 : Integer.parseInt(retryAttemptsHeader);
```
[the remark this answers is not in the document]
rickzx: Agreed. In the latest commit, `RetryClient` still sends the retry attempts header when the attempt is zero. And the `ServerRetryTracker` only counts the attempts if the header is present. This way, the tracker more accurately represents the retry ratio of the participating clients.

r2-core/src/main/java/com/linkedin/r2/filter/transport/ServerRetryFilter.java
rickzx: I rewrote the thread-safety logic in both `RetryClient` and `ServerRetryFilter`, and annotated them with `@ThreadSafe` and `@GuardedBy`. Hopefully it's cleaner this time.

r2-core/src/main/java/com/linkedin/r2/transport/http/common/HttpConstants.java
rickzx: Fixed

d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
```diff
@@ -59,15 +75,31 @@
...
+  public static final long DEFAULT_UPDATE_INTERVAL_MS = TimeUnit.SECONDS.toMillis(5);
   private static final Logger LOG = LoggerFactory.getLogger(RetryClient.class);
 
+  private final Clock _clock;
+  private final LoadBalancer _balancer;
   private final int _limit;
+  private final long _updateIntervalMs;
 
-  public RetryClient(D2Client d2Client, int limit)
+  Map<String, CallTracker> _retryTrackerMap;
```
rachelhanhan, line 86: I saw that when we invoke CallTracker.startCall, it's doing a lot more than just counting retry count and total call count, it's also counts the latency, outstanding latency and other stuff. Is this a little bit heavy for retry ratio? Especially that the CallTracker is used again in TrackerClient to track each server host's performance, this may defeat the purpose of CallTracker.

r2-core/src/main/java/com/linkedin/r2/filter/transport/ClientRetryFilter.java
```diff
@@ -73,8 +74,13 @@ public void onStreamError(Throwable ex,
     {
       nextFilter.onError(new RetriableRequestException(retryAttr), requestContext, wireAttrs);
     }
-    else
-    {
+    else {
+      Throwable[] throwables = ExceptionUtils.getThrowables(ex);
+      for (Throwable throwable : throwables) {
+        if (throwable instanceof RetriableRequestException) {
+          ((RetriableRequestException) throwable).setDoNotRetryOverride(true);
```
rachelhanhan, line 81: Break this loop if one throwable is retriable?

d2/src/main/java/com/linkedin/d2/balancer/clients/TrackerClientImpl.java
```diff
@@ -307,4 +309,17 @@ else if (throwable instanceof StreamException)
...
     return false;
   }
+
+  private CallCompletion startCall(Request request)
+  {
+    String retryHeader = request.getHeader(HttpConstants.HEADER_NUMBER_OF_RETRY_ATTEMPTS);
+    if (retryHeader != null && Integer.parseInt(retryHeader) > 0)
+    {
+      return _callTracker.startCall(true);
+    }
```
rachelhanhan, line 319: I'm a little bit confused, what's the purpose of tracking # of retries in TrackerClient here? Each trackerClient corresponds to one server host, do we need to track retry count on server host basis?

r2-core/src/main/java/com/linkedin/r2/filter/transport/ServerRetryFilter.java
```diff
@@ -75,13 +116,160 @@ public void onStreamError(Throwable ex,
...
+    {
+      if (numberOfRetryAttempts <= _retryLimit)
+      {
+        synchronized (_counterLock)
+        {
+          _retryAttemptsCounter.getLast()[numberOfRetryAttempts] += 1;
+        }
+      } else
+      {
+        LOG.warn("Unexpected number of retry attempts: " + numberOfRetryAttempts + ", current retry limit: " + _retryLimit);
```
bbarkley, line 195: Is this protecting against malicious clients? Why would we be getting a value that is above the limit? If something is misconfigured and clients are trying more than they should I would think we would want to track this as another entry for the max number of retries that are allowed instead of ignoring it.

bbarkley (commented):
Sorry for taking a while to get through this. Good progress!

d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
```diff
@@ -59,15 +75,31 @@
...
+
+  public RetryClient(D2Client d2Client, LoadBalancer balancer, int limit, long updateIntervalMs, Clock clock)
   {
     super(d2Client);
+    _balancer = balancer;
     _limit = limit;
-    LOG.debug("Retry client created with limit set to: ", _limit);
+    _updateIntervalMs = updateIntervalMs;
+    _clock = clock;
+    _retryTrackerMap = new HashMap<>();
```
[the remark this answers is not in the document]
rickzx: Fixed

d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
```diff
@@ -231,12 +302,34 @@ public void onError(Throwable e)
           }
           else
           {
+            double maxClientRequestRetryRatio;
+            try
+            {
+              Map<String, Object> transportClientProperties =
+                  _balancer.getLoadBalancedServiceProperties(_serviceName).getTransportClientProperties();
```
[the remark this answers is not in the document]
rickzx: I checked the implementation of getLoadBalancedServiceProperties. It will register a listener on ZK properties event bus only if the service has not already been listened on. And after the listener is registered, subsequent call to getLoadBalancedServiceProperties will directly read the properties from LoadBalancerState, which is a local variable. The RequestTimeoutClient is using the same API to get TransportClientProperties: https://github.com/linkedin/rest.li/blob/99999589f89a168a0528aabc4d25c797b755b57b/d2/src/main/java/com/linkedin/d2/balancer/clients/RequestTimeoutClient.java#L150.

The method without a callback is deprecated because we want it to be async. I changed the impl here to use the callback version.

r2-core/src/main/java/com/linkedin/r2/filter/transport/ServerRetryFilter.java
```diff
@@ -75,13 +116,160 @@ public void onStreamError(Throwable ex,
...
+  }
+
+  /**
+   * Stores the number of requests categorized by number of retry attempts. It uses the information to estimate
+   * a ratio of how many requests are being retried in the cluster. The ratio is then compared with
+   * {@link ServerRetryTracker#_maxRequestRetryRatio} to make a decision on whether or not to retry in the
+   * next interval. When calculating the ratio, it looks at the last {@link ServerRetryTracker#_aggregatedIntervalNum}
+   * intervals by aggregating the recorded requests.
+   */
+  private static class ServerRetryTracker
```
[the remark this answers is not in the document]
rickzx: I separate out ServerRetryTracker from ServerRetryFilter into the r2 util package. Added a test suite.

r2-core/src/main/java/com/linkedin/r2/filter/transport/ServerRetryFilter.java
```diff
@@ -75,13 +116,160 @@ public void onStreamError(Throwable ex,
...
+        }
+
+        // append a new interval
+        _retryAttemptsCounter.addLast(new int[_retryLimit + 1]);
+      }
+    }
+
+    public boolean isBelowRetryRatio()
+    {
+      updateRetryDecision();
```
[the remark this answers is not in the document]
rickzx: This is necessary because if there aren't many requests coming in for a period of time, the information in the tracker is stale. The next time we make the retry decision, we have to first rollover the stale windows.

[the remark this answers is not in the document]
bbarkley: Right, but that staleness would only apply if it was an older client not sending the header right? So the calculation at the end of the request can be skipped on requests when it's performed at the beginning.

r2-core/src/main/java/com/linkedin/r2/filter/transport/ServerRetryFilter.java
rickzx: This is to account for a config discrepancy between client and server. Client and server can both configure their own retry limit, so potentially, the client-side config can have a higher value than the server-side config. Do we want to fail the request if the number of attempts is higher than the server-side config value? Or should the server still handle this request? @rachelhanhan @bbarkley @ssheng Any thoughts on this?

I think not adding this request to RetryTracker will help improve efficiency because we can use a fix-sized array instead of a resizable list. And really, requests with high retry attempts should be extremely rare. We can already infer the retry ratio from the requests with lower retry attempts.

bbarkley: I agree we shouldn't support tracking data for any number of retry requests - if the current count is greater than the retry limit we can just treat it as the retry limit max.
Though potentially we should be rejecting the request if the count is higher than what the service wants to allow. Is the server side limit used for anything other than tracking?

d2/src/main/java/com/linkedin/d2/balancer/clients/TrackerClientImpl.java
rickzx: This change is reverted

r2-core/src/main/java/com/linkedin/r2/filter/transport/ClientRetryFilter.java
rickzx: Fixed

d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
rickzx: Addressed. Use a separate light-weight CallTracker for RetryClient. Leave it to container to handle monitoring metrics.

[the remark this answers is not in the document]
rickzx: Addressed. Use a light-weight sliding-window CallTracker for RetryClient. Leave it to container to handle monitoring metrics.

d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
```diff
@@ -231,12 +302,34 @@ public void onError(Throwable e)
...
-              LOG.warn("A retriable exception happens. Going to retry. This is attempt {}. Current exclusion set: ",
-                  attempts, ". Current exclusion set: " + exclusionSet);
-              retry = doRetryRequest(_request, _context);
+              CallTracker callTracker = _retryTrackerMap.get(_serviceName);
+              if (callTracker.getCallStats().getRetryRate() <= maxClientRequestRetryRatio)
+              {
+                LOG.warn("A retriable exception occurred. Going to retry. This is attempt {}. Current exclusion set: {}",
+                    attempts, exclusionSet);
+                _callCompletion.endCall();
+                retry = doRetryRequest(_request, _context, attempts);
```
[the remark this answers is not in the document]
rickzx: Both are removed in the latest commit

degrader/src/main/java/com/linkedin/util/degrader/CallTracker.java
```diff
@@ -115,11 +121,18 @@
...
    * @return an object that can be used to indicate completion of the call
    */
   CallCompletion startCall();
 
+  /**
+   * Indicates the start of a method invocation
+   * @param isRetry whether the call is a retry
+   * @return an object that can be used to indicate completion of the call
+   */
+  CallCompletion startCall(boolean isRetry);
```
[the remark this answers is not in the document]
rickzx: Removed retry-related stuff in default CallTracker

r2-core/src/main/java/com/linkedin/r2/filter/transport/ServerRetryFilter.java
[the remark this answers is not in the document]
rickzx: Right. Removed `updateRetryDecision` here

d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
```diff
@@ -231,12 +296,43 @@ public void onError(Throwable e)
...
+                LOG.warn("Failed to fetch transportClientProperties ", e);
+              }
+
+              @Override
+              public void onSuccess(ServiceProperties result) {
+                _transportClientProperties = result.getTransportClientProperties();
+              }
+            });
+
+            if (_transportClientProperties == null)
```
rachelhanhan, line 312: When you get the _transportClientProperties in async manner, you will not immediately obtain the value. After you invoke _balancer.getLoadBalancedServiceProperties asynchronously, I think when you execute to line 312, for sure it will be null, we should notify a callback instead.

d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
```diff
@@ -254,16 +350,134 @@ public void onError(Throwable e)
...
+    private final Object _counterLock = new Object();
+    private final Object _updateLock = new Object();
+
+    @GuardedBy("_updateLock")
+    private volatile long _lastRollOverTime;
+    private double _currentAggregatedRetryRatio;
+
+    @GuardedBy("_counterLock")
+    private final LinkedList<int[]> _retryCounter;
+    private final int[] _aggregatedRetryCounter;
```
rachelhanhan, line 396: I feel using int[2] to track total count and retry count is a bit implicit without java docs, can we use a new data structure to represent the count?

r2-core/src/main/java/com/linkedin/r2/util/ServerRetryTracker.java
```diff
@@ -0,0 +1,158 @@
...
+    if (numberOfRetryAttempts <= _retryLimit)
+    {
+      synchronized (_counterLock)
+      {
+        _retryAttemptsCounter.getLast()[numberOfRetryAttempts] += 1;
+      }
+    } else
+    {
+      LOG.warn("Unexpected number of retry attempts: " + numberOfRetryAttempts + ", current retry limit: " + _retryLimit);
+      _retryAttemptsCounter.getLast()[_retryLimit] += 1;
```
rachelhanhan, line 80: Wondering why isn't this line under the counterlock?

rachelhanhan (commented):
Overall design looks good now, added some comments on the implementation

d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
rickzx: Fixed. Make isBelowRetryRatio an async call as well.

d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
rickzx: Fixed. Added a RetryCounter impl.

r2-core/src/main/java/com/linkedin/r2/util/ServerRetryTracker.java
rickzx: Good catch. Fixed

d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
```diff
@@ -254,16 +338,201 @@ public void onError(Throwable e)
...
+
+    public void isBelowRetryRatio(SuccessCallback<Boolean> callback)
+    {
+      _balancer.getLoadBalancedServiceProperties(_serviceName, new Callback<ServiceProperties>()
+      {
+        @Override
+        public void onError(Throwable e)
+        {
+          LOG.warn("Failed to fetch transportClientProperties ", e);
+          double maxClientRequestRetryRatio = HttpClientFactory.DEFAULT_MAX_CLIENT_REQUEST_RETRY_RATIO;
```
rachelhanhan, line 443: minor: this line seems redundant

d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
```diff
@@ -254,16 +338,201 @@ public void onError(Throwable e)
...
+    }
+
+    public void rollOverStats()
+    {
+      // rollover the current interval to the aggregated counter
+      synchronized (_counterLock)
+      {
+        RetryCounter intervalToAggregate = _retryCounter.getLast();
+        _aggregatedRetryCounter.addToTotalRequestCount(intervalToAggregate.getTotalRequestCount());
+        _aggregatedRetryCounter.addToRetryRequestCount(intervalToAggregate.getRetryRequestCount());
```
rachelhanhan, line 420: Minor: if you always operate on total and retry count together, maybe you can merge these 2 lines into 2 method? Will leave the decision to you

rachelhanhan (approved):
LGTM
Feel free to open a DMRC ticket if you haven't.

d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
rickzx: Fixed

d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
rickzx: I feel like the current API of RetryCounter is clearer by separating out total request with retry request. Here I need to call the two methods separately https://github.com/linkedin/rest.li/pull/484/files#diff-fb1d5cedc2fd04937e39f9188139915da2a2ae4aa9c736de7503973cc85c4b6eR419

r2-core/src/main/java/com/linkedin/r2/filter/transport/ServerRetryFilter.java
rickzx: Fixed by treating it as the retry limit max. The server side limit is used only for tracking now. One reason for the server to still handle the request above the limit is that it already defines through TransportClientProperties the max ratio of retry request a client could send. I think it's reasonable to let the client decide the per-request retry limit.

rickzx:
To help with retryCountTotal monitoring, adds a listener interface to RetryClient whenever a retry request is sent. A CallTrackerRegistry can implement this interface and increment the retryCountTotal counter when `onRetryRequest` is called.

d2-schemas/src/main/pegasus/com/linkedin/d2/D2TransportClientProperties.pdl
```diff
@@ -141,4 +141,9 @@ record D2TransportClientProperties {
      */
     HTTP_2
   }
+
+  /**
+   * Maximum ratio of retry requests to total requests per client
+   */
+  maxClientRequestRetryRatio: optional double
```
karthikbalasub, line 148: The code is setting a default of 0.2 for this, any reason why it cannot be added in the schema?

d2-schemas/src/main/pegasus/com/linkedin/d2/D2TransportClientProperties.pdl
```diff
@@ -141,4 +141,9 @@ record D2TransportClientProperties {
      */
     HTTP_2
   }
+
+  /**
+   * Maximum ratio of retry requests to total requests per client
```
karthikbalasub, line 146: Is this config applicable only when retry is enabled? If so, please document that here.
Also why are other retry configs (retry, retryLimit, Interval) etc not part of the client properties?

karthikbalasub (changes requested):

d2-schemas/src/main/pegasus/com/linkedin/d2/D2TransportClientProperties.pdl
rickzx: As with other D2 config properties defined in the schema, the default value is defined in two places: in Pegasus factory and in global d2.src config. So the default value can then be easily discovered by D2 config services (e.g. d2-config-service, maestro), and also be tuned easily if needed. We don't wan't to define the it directly in the schema because it creates an additional source of truth of the default value.

d2-schemas/src/main/pegasus/com/linkedin/d2/D2TransportClientProperties.pdl
rickzx: Updated the documentation:
  /**
   * Service desired maximum ratio of retry requests to total requests per client. The retry ratio is applied
   * only when D2 RetryClient is enabled.
   */

Because the other retry configs (retry, retryLimit, updateInterval) are defined on the client-side, in application.cfg. But maxClientRequestRetryRatio is a service-desired property, which is defined on the server-side and pass to the client via zookeeper.

karthikbalasub:
@rickzx Data model changes look good.

r2-core/src/main/java/com/linkedin/r2/util/ServerRetryTracker.java
```diff
@@ -0,0 +1,161 @@
...
+      {
+        _retryAttemptsCounter.getLast()[numberOfRetryAttempts] += 1;
+      }
+    } else
+    {
+      LOG.warn("Unexpected number of retry attempts: " + numberOfRetryAttempts + ", current retry limit: " + _retryLimit);
+      synchronized (_counterLock)
+      {
+        _retryAttemptsCounter.getLast()[_retryLimit] += 1;
+      }
```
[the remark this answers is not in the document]
rickzx: Fixed

r2-core/src/test/java/com/linkedin/r2/util/TestServerRetryTracker.java
```diff
@@ -0,0 +1,106 @@
...
+public class TestServerRetryTracker
+{
+  private ServerRetryTracker _serverRetryTracker;
+  private SettableClock _clock;
+
+  @BeforeMethod
+  public void setUp()
+  {
+    _clock = new SettableClock();
+    _serverRetryTracker = new ServerRetryTracker(ServerRetryFilter.DEFAULT_RETRY_LIMIT,
```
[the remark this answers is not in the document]
rickzx: Added

r2-core/src/test/java/com/linkedin/r2/util/TestServerRetryTracker.java
```diff
@@ -0,0 +1,106 @@
...
+    for (int i = 0; i < 10; i++)
+    {
+      Assert.assertTrue(_serverRetryTracker.isBelowRetryRatio());
+      Assert.assertEquals(_serverRetryTracker.getRetryRatio(),0.0, 0.0001);
+      _clock.addDuration(ServerRetryFilter.DEFAULT_UPDATE_INTERVAL_MS);
+    }
+  }
+
+  @Test
+  public void testServerRetryTracker()
```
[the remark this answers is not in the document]
rickzx: It's kind of hard to separate out the multiple window scenarios because it builds upon the single window case. After separating it out in the latest commit, the test case is still pretty lengthy.

r2-core/src/main/java/com/linkedin/r2/util/ServerRetryTracker.java
```diff
@@ -0,0 +1,161 @@
...
+      synchronized (_counterLock)
+      {
+        _retryAttemptsCounter.getLast()[numberOfRetryAttempts] += 1;
+      }
+    } else
+    {
+      LOG.warn("Unexpected number of retry attempts: " + numberOfRetryAttempts + ", current retry limit: " + _retryLimit);
+      synchronized (_counterLock)
+      {
+        _retryAttemptsCounter.getLast()[_retryLimit] += 1;
```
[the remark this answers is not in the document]
rickzx: Added

d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
```diff
@@ -254,16 +334,200 @@ public void onError(Throwable e)
...
+    private final long _updateIntervalMs;
+    private final Clock _clock;
+    private final String _serviceName;
+
+    private final Object _counterLock = new Object();
+    private final Object _updateLock = new Object();
+
+    @GuardedBy("_updateLock")
+    private volatile long _lastRollOverTime;
+    private double _currentAggregatedRetryRatio;
```
bbarkley, line 374: Minor: if the _updateLock is guarding it there should be a separate annotation for it

d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
```diff
@@ -254,16 +334,200 @@ public void onError(Throwable e)
...
+
+    @GuardedBy("_counterLock")
+    private final LinkedList<RetryCounter> _retryCounter;
+    private final RetryCounter _aggregatedRetryCounter;
```
bbarkley, line 378: Same as above - additional annotation if the lock protects this

d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
```diff
@@ -254,16 +334,200 @@ public void onError(Throwable e)
...
+        }
+
+        // append a new interval
+        _retryCounter.addLast(new RetryCounter());
+      }
+    }
+
+    public void isBelowRetryRatio(SuccessCallback<Boolean> callback)
+    {
+      _balancer.getLoadBalancedServiceProperties(_serviceName, new Callback<ServiceProperties>()
```
bbarkley, line 433: Is this making a call out to something like ZK to get this data? If so doing this on each invocation that could add a lot of latency. Ideally the result would be cached with some TTL (or maybe just fetched once the first time - do we expect this to change in ZK?).

bbarkley (commented):
Really close to getting this wrapped. Great work so far!

d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
```diff
@@ -108,8 +148,21 @@ public void streamRequest(StreamRequest request, Callback<StreamResponse> callba
...
+        .build(request.getEntityStream());
+    ClientRetryTracker retryTracker = updateRetryTracker(newRequest.getURI(), false);
+    final Callback<StreamResponse> transportCallback = new StreamRetryRequestCallback(newRequest, requestContext, callback, retryTracker);
+    _d2Client.streamRequest(newRequest, requestContext, transportCallback);
+  }
+
+  private ClientRetryTracker updateRetryTracker(URI uri, boolean isRetry)
+  {
+    String serviceName = LoadBalancerUtil.getServiceNameFromUri(uri);
+    _retryTrackerMap.putIfAbsent(serviceName, new ClientRetryTracker(_aggregatedIntervalNum, _updateIntervalMs, _clock, serviceName));
```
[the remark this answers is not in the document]
rickzx: Thanks for pointing this out. Good to learn that computeIfAbsent takes a mapping function so the Tracker object will not be created until needed.

d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
rickzx: Fixed

d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
rickzx: Fixed

d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
rickzx: It will only make a call to ZK the first time getting data from a downstream service. At the same time, it will also register a listener to ZK for property updates. The subsequent calls will just read from the local LoadBalancerState.

bbarkley: Thanks for clarifying

bbarkley (approved):
Looks good. Thanks for working through all of the changes here

PR #485, opened 2020-11-21
Enhance request symbol table fetch

karthikrg:
- Return null if uri prefix doesn't match. 
- If the fetch call 404s internally store an empty symbol table and return null. This will avoid repeated invocations to services that are not yet ready to support symbol tables

### DIFF
--- a/restli-tools/src/main/java/com/linkedin/restli/tools/symbol/RestLiSymbolTableProvider.java
+++ b/restli-tools/src/main/java/com/linkedin/restli/tools/symbol/RestLiSymbolTableProvider.java
@@ -20,6 +20,7 @@ import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.linkedin.d2.balancer.util.LoadBalancerUtil;
 import com.linkedin.data.ByteString;
+import com.linkedin.data.codec.symbol.EmptySymbolTable;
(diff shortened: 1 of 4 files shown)


restli-tools/src/main/java/com/linkedin/restli/tools/symbol/RestLiSymbolTableProvider.java
```diff
@@ -220,13 +221,21 @@ public SymbolTable getSymbolTable(String symbolTableName)
   @Override
   public SymbolTable getRequestSymbolTable(URI requestUri)
   {
+    // If the URI prefix doesn't match, return null.
+    if (!requestUri.toString().startsWith(_uriPrefix))
+    {
+      return null;
+    }
```
aman1309, line 228: shouldn't requestUri will always start with uriPrefix?
reason for adding d2 check was for HTTP(s) prefixes adding /serviceName/resources or something else might be different from where symbol table endpoint is hosted since we are using zk serverNodeUri to construct symbol table API endpoint

karthikrg: There is currently no guarantee that request URI has to start with uriPrefix. This check just ensures that.
Yes, practically it is going to be d2 or http(s) but the contract of this class itself makes no assumptions about that. Hence, its safer to code it this way.

karthikbalasub (approved):

PR #486, opened 2020-11-25
Add support for fetching experimental content types announced by the server

karthikrg:
This will help us incrementally support new content types for request bodies without upgrading all the servers.

### DIFF
--- a/restli-client/src/main/java/com/linkedin/restli/client/RestClient.java
+++ b/restli-client/src/main/java/com/linkedin/restli/client/RestClient.java
@@ -26,7 +26,6 @@ import com.linkedin.d2.balancer.KeyMapper;
 import com.linkedin.d2.balancer.ServiceUnavailableException;
 import com.linkedin.d2.balancer.util.URIKeyPair;
 import com.linkedin.d2.balancer.util.URIMappingResult;
-import com.linkedin.data.ByteString;
 import com.linkedin.data.DataMap;
 import com.linkedin.data.template.RecordTemplate;
 import com.linkedin.multipart.MultiPartMIMEUtils;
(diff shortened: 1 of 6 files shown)


restli-client/src/main/java/com/linkedin/restli/client/RestClient.java
```diff
@@ -414,53 +440,59 @@ public void onSuccess(ProtocolVersion protocolVersion)
...
+    }
+
+    try
+    {
+      _client.getMetadata(new URI(_uriPrefix + serviceName), Callbacks.handle(metadata ->
+      {
+        // Use the metadata from the server to extract protocol version only if versionOption is USE_LATEST_IF_AVAILABLE
+        final ProtocolVersion announcedVersion = (versionOption ==
+            ProtocolVersionOption.USE_LATEST_IF_AVAILABLE) ? getAnnouncedVersion(metadata) : null;
+        final Set<String> experimentalContentTypeHeaders = getExperimentalContentTypeHeaders(metadata);
```
karthikbalasub, line 479: Instead of call this experimental content type, can we just say these are server announced content types. Server can list them in the order of preference.

The client can then pick the first one it supports.

restli-client/src/main/java/com/linkedin/restli/client/RestClient.java
```diff
@@ -636,6 +714,11 @@ private ContentType resolveContentType(MessageHeadersBuilder<?> builder, DataMap
         {
           header = contentType.getHeaderKey();
         }
+        else if (_experimentalContentType != null
+            && experimentalContentTypeHeaders.contains(_experimentalContentType.getHeaderKey()))
+        {
+          header = _experimentalContentType.getHeaderKey();
+        }
```
karthikbalasub, line 721: This logic can be:

for header  in serverAnnouncedContentTypes {
  if (ContentType.get(header).isPresent()) {
    return ContentType.getRequestContentType(header, ..)
  }
}

karthikrg:
Based on offline discussions with @ChaoLinkedIn doing this may not be feasible given that zk properties are still controlled at a service level and are not central. Dropping this pr.

PR #492, opened 2020-12-10
Setup the base infra for generating new fluent api client bindings.

karthikbalasub:
New bindings will be generated from templates using velocity template engine.


gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1917,6 +1919,10 @@ protected void configureRestClientGeneration(Project project, SourceSet sourceSe
           {
             task.setEnableArgFile(true);
           }
+          if (isPropertyTrue(project, ENABLE_FLUENT_API))
```
junchuanwang, line 1922: Do you plan to change it later? I thought this would be enabled by default: (fluentAPi constructed by default and users to decide whether to use it)

restli-tools/src/main/java/com/linkedin/restli/tools/clientgen/FluentApiGenerator.java
```diff
@@ -0,0 +1,257 @@
...
+        System.exit(0);
+      }
+      String targetDirectory = cl.getOptionValue('t');
+      String resolverPath = cl.getOptionValue('p');
+      if (ArgumentFileProcessor.isArgFile(resolverPath))
+      {
+        resolverPath = ArgumentFileProcessor.getContentsAsArray(resolverPath)[0];
+      }
+      String[] sources = cl.getArgs();
+      if (sources.length == 1 && ArgumentFileProcessor.isArgFile(sources[0]))
```
junchuanwang, line 111: though I know ArgFile is quite common now and a bit internal, I actually hope this comment can be surfaced somewhere, through help doc or something

junchuanwang (approved):
Quite clean and intuitive. LGTM

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
karthikbalasub: Yes, when we do GA, we'll turn this on by default.

restli-tools/src/main/java/com/linkedin/restli/tools/clientgen/FluentApiGenerator.java
karthikbalasub: Good point. Updated to include help on using arg file.

PR #493, opened 2020-12-14
Fix empty union validation to only allow validation pass when projection in union is present

BrianPin:
Before this fix, the RestLiValidationFilter will allow empty union to pass, i.e if there is a union that is an empty map, when doing the validation. 
That is too wide open and risky. So this fix, change it  restricted only when a projection is specified in the union.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4783,9 +4783,8 @@ patch operations can re-use these classes for generating patch messages.
 ## [0.14.1]
 
 [Unreleased]: https://github.com/linkedin/rest.li/compare/v29.13.1...master
-[29.13.1]: https://github.com/linkedin/rest.li/compare/v29.12.1...v29.13.1
-[29.13.0]: https://github.com/linkedin/rest.li/compare/v29.12.1...v29.13.0
-[29.12.1]: https://github.com/linkedin/rest.li/compare/v29.12.0...v29.12.1
+[29.13.1]: https://github.com/linkedin/rest.li/compare/v29.13.0...v29.13.1
+[29.13.0]: https://github.com/linkedin/rest.li/compare/v29.12.0...v29.13.0
 [29.12.0]: https://github.com/linkedin/rest.li/compare/v29.11.3...v29.12.0
 [29.11.3]: https://github.com/linkedin/rest.li/compare/v29.11.2...v29.11.3
 [29.11.2]: https://github.com/linkedin/rest.li/compare/v29.11.1...v29.11.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and what APIs have changed, if applicable.
 - Bump `javax.mail:mail` dependency from `1.4.1` to `1.4.4` to avoid classloader issues in `javax.activation` code with Java 11.
 - Bump arvo compatibility layer `avroutil` dependency from `0.1.11` to `0.2.17` for Arvo Upgrade HI.
 - Setup the base infra for generating new fluent api client bindings.
+- Fix the restriction of empty union validation from wide open to only allow when there is a projection in the union
 
 ## [29.12.0] - 2020-12-02
 - Add a boolean flag as header for symbol table request to avoid conflict with resource requests.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
+
+## [29.13.1] - 2020-12-14
 - Fix the restriction of empty union validation from wide open to only allow when there is a projection in the union
 
 ## [29.13.0] - 2020-12-12
@@ -4784,7 +4786,8 @@ patch operations can re-use these classes for generating patch messages.
 [Unreleased]: https://github.com/linkedin/rest.li/compare/v29.13.0...master
 [29.13.0]: https://github.com/linkedin/rest.li/compare/v29.12.0...v29.13.0
 =======
-[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.12.1...master
+[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.13.1...master
+[29.13.1]: https://github.com/linkedin/rest.li/compare/v29.12.1...v29.13.1
 [29.12.1]: https://github.com/linkedin/rest.li/compare/v29.12.0...v29.12.1
 >>>>>>> bd549e7ff (bump version)
 [29.12.0]: https://github.com/linkedin/rest.li/compare/v29.11.3...v29.12.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4782,14 +4782,10 @@ patch operations can re-use these classes for generating patch messages.
 
 ## [0.14.1]
 
-<<<<<<< HEAD
-[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.13.0...master
-[29.13.0]: https://github.com/linkedin/rest.li/compare/v29.12.0...v29.13.0
-=======
 [Unreleased]: https://github.com/linkedin/rest.li/compare/v29.13.1...master
 [29.13.1]: https://github.com/linkedin/rest.li/compare/v29.12.1...v29.13.1
+[29.13.0]: https://github.com/linkedin/rest.li/compare/v29.12.1...v29.13.0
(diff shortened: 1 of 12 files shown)


data/src/main/java/com/linkedin/data/schema/UnionDataSchema.java
```diff
@@ -36,6 +36,11 @@
  */
 public final class UnionDataSchema extends ComplexDataSchema
 {
+  /**
+   *
+   */
+  private boolean allowEmptyUnionResponse = false;
```
karthikbalasub, line 42: The schema layer doesn't have the context about it being used in request/response. So using "response" here is not a good idea.

I'd suggest calling this flag "isPartialSchema" to indicate this is a partial schema created from projection. 
Update the doc to say "this flag indicates the union schema is a partial representation with a subset of members" And also say this is for internal use only.

data/src/main/java/com/linkedin/data/schema/UnionDataSchema.java
```diff
@@ -457,4 +462,14 @@ public static String avroUnionMemberKey(DataSchema schema)
   private boolean _membersAliased = false;
 
   private static final Map<String, Integer> _emptyTypesToIndexMap = Collections.emptyMap();
+
+  public void setAllowEmptyUnionResponse(boolean allowEmptyUnionResponse)
+  {
+    this.allowEmptyUnionResponse = allowEmptyUnionResponse;
+  }
+
+  public boolean isAllowEmptyUnionResponse()
```
karthikbalasub, line 471: Comment these are for internal use only.

data/src/main/java/com/linkedin/data/schema/validation/ValidateDataAgainstSchema.java
```diff
@@ -526,6 +526,10 @@ else if (_recursive)
             validate(memberElement, memberSchema, value);
           }
         }
+        else if (!schema.isAllowEmptyUnionResponse())
+        {
+          addMessage(element, "DataMap should have at least one entry for a union type or specify the member in projection");
```
karthikbalasub, line 531: You can remove the part about "specify member in projection". It can cause confusion about this error, which if happens means the resource method didn't populate the union field correctly. Using a projection to avoid the error is not the correct way to handle it.

restli-common/src/main/java/com/linkedin/restli/common/util/ProjectionMaskApplier.java
```diff
@@ -215,6 +215,11 @@ else if (wildcardMask != null)
 
     UnionDataSchema newUnionDataSchema = new UnionDataSchema();
     newUnionDataSchema.setMembers(newUnionMembers, errorMessageBuilder);
+    if (newUnionMembers.size() > 0)
```
karthikbalasub, line 218: This should be 
newMembers.size() < members.size() ?

restli-int-test/src/test/java/com/linkedin/restli/examples/TestEmptyUnionValidation.java
```diff
@@ -0,0 +1,78 @@
...
+    expected.setFoo(new ValidateEmptyUnion.Foo());
+    List<PathSpec> spec = Collections.singletonList(ValidateEmptyUnion.fields().foo().Fuzz());
+    EmptyUnionRequestBuilders requestBuilders = new EmptyUnionRequestBuilders();
+    GetRequest<ValidateEmptyUnion> req = requestBuilders.get().id(1L).fields(spec.toArray(new PathSpec[spec.size()])).build();
+    ValidateEmptyUnion actual = getClient().sendRequest(req).getResponse().getEntity();
+    Assert.assertEquals(actual, expected);
+  }
+
+  @Test(expectedExceptions = RestLiResponseException.class)
+  public void testUnionEmptyWithoutProjection() throws RemoteInvocationException {
```
karthikbalasub, line 72: Add a testcase that projects all union members, it should also fail validation.

karthikbalasub (commented):
The description of this PR should be updated.

data/src/main/java/com/linkedin/data/schema/UnionDataSchema.java
BrianPin: Ok

restli-common/src/main/java/com/linkedin/restli/common/util/ProjectionMaskApplier.java
BrianPin: The `newMembers.size() < members.size()` could be including the case where there is no projection in the union. So I am hesitate to adopt this comment, let me know, Thanks

restli-int-test/src/test/java/com/linkedin/restli/examples/TestEmptyUnionValidation.java
BrianPin: sure, so the criteria will be
`newUnionMembers.size() > 0 && newUnionMembers.size() < members.size()`

restli-common/src/main/java/com/linkedin/restli/common/util/ProjectionMaskApplier.java
```diff
@@ -215,6 +215,11 @@ else if (wildcardMask != null)
 
     UnionDataSchema newUnionDataSchema = new UnionDataSchema();
     newUnionDataSchema.setMembers(newUnionMembers, errorMessageBuilder);
+    if (newUnionMembers.size() > 0 && unionDataSchema.getMembers().size() > newUnionMembers.size())
```
karthikbalasub, line 218: why do you need size > 0 check?

newUnionmember size will be zero only if the mask is empty. Which means none of the members were projected. We should treat that as partial projection.

It is not possible to create that projection using pathspecs, but someone building the mask map directly can do it.

restli-int-test/src/test/java/com/linkedin/restli/examples/TestEmptyUnionValidation.java
```diff
@@ -0,0 +1,91 @@
+/*
+   Copyright (c) 2018 LinkedIn Corp.
```
karthikbalasub, line 2: 2020

karthikbalasub (approved):

restli-common/src/main/java/com/linkedin/restli/common/util/ProjectionMaskApplier.java
BrianPin: For the last sentence, is the following code kind of using pathspecs to create projection?

```
    List<PathSpec> spec = Arrays.asList(
        ValidateEmptyUnion.fields().foo().Fuzz(),
        ValidateEmptyUnion.fields().foo().Bar());

```

restli-common/src/main/java/com/linkedin/restli/common/util/ProjectionMaskApplier.java
BrianPin: Saw your latest comments

CHANGELOG.md
```diff
@@ -4779,8 +4782,10 @@ patch operations can re-use these classes for generating patch messages.
 
 ## [0.14.1]
 
-[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.13.0...master
-[29.13.0]: https://github.com/linkedin/rest.li/compare/v29.12.0...v29.13.0
+[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.13.1...master
+[29.13.1]: https://github.com/linkedin/rest.li/compare/v29.12.1...v29.13.1
+[29.13.0]: https://github.com/linkedin/rest.li/compare/v29.12.1...v29.13.0
+[29.12.1]: https://github.com/linkedin/rest.li/compare/v29.12.0...v29.12.1
```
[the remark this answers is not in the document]
BrianPin: Fixed it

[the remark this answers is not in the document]
BrianPin: Thanks for notice, I should check carefully next time

restli-common/src/main/java/com/linkedin/restli/common/util/ProjectionMaskApplier.java
karthikbalasub: I meant creating an empty projection for a union., with the following mask map:
{ 
   foo: {}
}

using ValidateEmptyUnion.fields().foo() will result in:
{
  foo: 1
 }
 And ValidateEmptyUnion.fields().foo().Fuzz() will result in:
 {
  foo:  {
    Fuzz: 1
  }
 }

PR #500, opened 2021-01-05
Improve support for accessing schema and enum symbol properties.

karthikbalasub:
Add support for accessing schema statically from generated template classes and for getting symbol properties from enum schema properties.

### DIFF
--- a/data/src/main/java/com/linkedin/data/schema/EnumDataSchema.java
+++ b/data/src/main/java/com/linkedin/data/schema/EnumDataSchema.java
@@ -17,6 +17,7 @@
 package com.linkedin.data.schema;
 
 
+import com.linkedin.data.DataMap;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
@@ -24,6 +25,7 @@ import java.util.List;
 import java.util.Map;
 
 import static com.linkedin.data.schema.DataSchemaConstants.ENUM_SYMBOL_PATTERN;
+import static com.linkedin.data.schema.DataSchemaConstants.SYMBOL_PROPERTIES_KEY;
 
 /**
  * {@link DataSchema} for enum.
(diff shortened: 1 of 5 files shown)


nickibi (approved):
LGTM

data/src/main/java/com/linkedin/data/schema/EnumDataSchema.java
```diff
@@ -170,6 +172,25 @@ public boolean contains(String symbol)
...
+  public Map<String, Object> getSymbolProperties(String symbol)
+  {
+    if (!_symbolToIndexMap.containsKey(symbol))
+    {
+      return null;
+    }
+    Object prop = getProperties().get(SYMBOL_PROPERTIES_KEY);
+    if(prop instanceof DataMap)
+    {
+      return ((DataMap) prop).getDataMap(symbol);
```
junchuanwang, line 189: Does this “getDataMap" require assumption that the symbolPropertiesMap always well-formatted?  Do we do any check elsewhere? Is this guaranteed?

generator/src/main/java/com/linkedin/pegasus/generator/JavaDataTemplateGenerator.java
```diff
@@ -1336,6 +1336,11 @@ private JFieldVar generateSchemaField(JDefinedClass templateClass, DataSchema sc
     // Generate the schema field initialization
     schemaField.init(JExpr.cast(getCodeModel()._ref(schema.getClass()), parseSchemaInvocation));
 
+    // Using "dataSchema" as method name since "schema" conflicts with RecordTemplate::schema and "getSchema" conflicts
+    // with TyperefInfo::getSchema
+    final JMethod staticFieldsAccessor = templateClass.method(JMod.PUBLIC | JMod.STATIC, schema.getClass(), "dataSchema");
```
junchuanwang, line 1341: make "dataSchema" as a constant?
But to be honest I would not say I like this new name since it introduced another name!  Don't you think so?

BrianPin (approved):

data/src/main/java/com/linkedin/data/schema/EnumDataSchema.java
karthikbalasub: This is guranteed by the schema parser.

junchuanwang (approved):

PR #501, opened 2021-01-06
Add MaxBatchSize support

nickibi:
This PR is adding max batch size support:
1. Add MaxBatchSize annotation and expose it in the IDL.
2. Add validation based on the max batch size 
3. Add resource compatibility check rules for MaxBatchSize
4. Add related tests

### DIFF
--- a/restli-common/src/main/pegasus/com/linkedin/restli/restspec/RestMethodSchema.pdl
+++ b/restli-common/src/main/pegasus/com/linkedin/restli/restspec/RestMethodSchema.pdl
@@ -29,4 +29,10 @@ record RestMethodSchema includes CustomAnnotationSchema, ServiceErrorsSchema, Su
    * Indicates if this rest method has paging support using the start and count parameters
    */
   pagingSupported: optional boolean
+
+  /**
+   * Specifies the max batch size allowed for this method. It supports BATCH_GET, BATCH_CREATE, BATCH_DELETE,
+   * BATCH_UPDATE and BATCH_PARTIAL_UPDATE methods.
+   */
+  maxBatchSize: optional MaxBatchSizeSchema
 }
--- a/restli-int-test/src/test/java/com/linkedin/restli/examples/TestMaxBatchSize.java
+++ b/restli-int-test/src/test/java/com/linkedin/restli/examples/TestMaxBatchSize.java
@@ -1,0 +1,333 @@
+/*
+   Copyright (c) 2021 LinkedIn Corp.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+ */
+
+package com.linkedin.restli.examples;
+
+import com.linkedin.r2.RemoteInvocationException;
+import com.linkedin.restli.client.BatchGetEntityRequest;
+import com.linkedin.restli.client.Request;
+import com.linkedin.restli.client.Response;
+import com.linkedin.restli.client.RestLiResponseException;
+import com.linkedin.restli.client.response.BatchKVResponse;
+import com.linkedin.restli.client.util.PatchGenerator;
+import com.linkedin.restli.common.BatchCollectionResponse;
+import com.linkedin.restli.common.BatchFinderCriteriaResult;
+import com.linkedin.restli.common.CollectionResponse;
+import com.linkedin.restli.common.CreateStatus;
+import com.linkedin.restli.common.EntityResponse;
+import com.linkedin.restli.common.HttpStatus;
+import com.linkedin.restli.common.PatchRequest;
+import com.linkedin.restli.common.UpdateStatus;
+import com.linkedin.restli.examples.greetings.api.Greeting;
+import com.linkedin.restli.examples.greetings.api.GreetingCriteria;
+import com.linkedin.restli.examples.greetings.api.Tone;
+import com.linkedin.restli.examples.greetings.client.BatchGreetingBuilders;
+import com.linkedin.restli.examples.greetings.client.BatchGreetingRequestBuilders;
+
+import com.linkedin.restli.test.util.RootBuilderWrapper;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.testng.Assert;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+/**
+ * Integration tests to test MaxBatchSize annotation on different batch methods (batchGet, batchCreate, batchUpdate,
+ * batchPartialUpdate, batchDelete, batchFinder).
+ * These integration tests send requests to {@link com.linkedin.restli.examples.greetings.server.BatchGreetingResource}
+ *
+ * @author Yingjie Bi
+ */
+public class TestMaxBatchSize extends RestLiIntegrationTest
+{
+
+  private static final Greeting GREETING_ONE;
+  private static final Greeting GREETING_TWO;
+  private static final Greeting GREETING_THREE;
+  private static final GreetingCriteria CRITERIA_ONE;
+  private static final GreetingCriteria CRITERIA_TWO;
+  private static final GreetingCriteria CRITERIA_THREE;
+
+  static
+  {
+    GREETING_ONE = new Greeting();
+    GREETING_ONE.setTone(Tone.INSULTING);
+    GREETING_ONE.setId(1l);
+    GREETING_ONE.setMessage("Hi");
+
+    GREETING_TWO = new Greeting();
+    GREETING_TWO.setTone(Tone.FRIENDLY);
+    GREETING_TWO.setId(2l);
+    GREETING_TWO.setMessage("Hello");
+
+    GREETING_THREE = new Greeting();
+    GREETING_THREE.setTone(Tone.SINCERE);
+    GREETING_THREE.setId(3l);
+    GREETING_THREE.setMessage("How are you?");
+
+    CRITERIA_ONE = new GreetingCriteria().setId(1l);
+    CRITERIA_TWO = new GreetingCriteria().setId(2l);
+    CRITERIA_THREE = new GreetingCriteria().setId(3l);
+  }
+
+  @BeforeClass
+  public void initClass() throws Exception
+  {
+    super.init();
+  }
+
+  @AfterClass
+  public void shutDown() throws Exception
+  {
+    super.shutdown();
+  }
+
+  @Test(dataProvider = com.linkedin.restli.internal.common.TestConstants.RESTLI_PROTOCOL_1_2_PREFIX + "requestBuilderDataProvider")
+  public void testBatchCreateWithMaxBatchSizeUnderLimitation(RootBuilderWrapper<Long, Greeting> builders) throws RemoteInvocationException
+  {
+    Request<CollectionResponse<CreateStatus>> request = builders.batchCreate()
+        .inputs(Arrays.asList(GREETING_ONE, GREETING_TWO)).build();
+
+    Response<CollectionResponse<CreateStatus>> createResponse =
+        getClient().sendRequest(request).getResponse();
+    Assert.assertEquals(createResponse.getStatus(), 200);
+  }
+
+  @Test(dataProvider = com.linkedin.restli.internal.common.TestConstants.RESTLI_PROTOCOL_1_2_PREFIX + "requestBuilderDataProvider")
+  public void testBatchCreateWithMaxBatchSizeBeyondLimitation(RootBuilderWrapper<Long, Greeting> builders) throws RemoteInvocationException
+  {
+    Request<CollectionResponse<CreateStatus>> request = builders.batchCreate()
+        .inputs(Arrays.asList(GREETING_ONE, GREETING_TWO, GREETING_THREE)).build();
+
+    try
+    {
+      getClient().sendRequest(request).getResponse();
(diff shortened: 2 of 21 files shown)


BrianPin (approved):
LGTM after I been introduce through and read again

restli-server/src/main/java/com/linkedin/restli/server/BaseRestLiServer.java
```diff
@@ -233,4 +237,69 @@ private RestLiArgumentBuilder lookupArgumentBuilder(ResourceMethodDescriptor met
...
+  /**
+   * Method is used to validate the request's batch size is under
+   * the allowed max batch size which is defined in the server resource.
+   *
+   * @throws RestLiServiceException if request's batch size is larger than the allowed max batch size.
+   */
+  private void validateMaxBatchSize(RestLiRequestData requestData, ResourceMethodDescriptor method,
+      ServerResourceContext context) throws RestLiServiceException
+  {
+    MaxBatchSizeSchema maxBatchSizeAnnotation = method.getMaxBatchSize();
```
BrianPin, line 250: nit, I think we can check method first, and return early if the method does not have anything to do with batches.

nickibi: MaxBatchSize annotation is only allowed to add on batch methods. We have such guard in annotation reader. If maxBatchSizeAnnotation == null, it is either a non-batch method or batch method without max batch size annotation. In another word, it includes check the method is batch method or not. We do not need to specify check method itself.  I do not see much value provided if we specifically check the method type first here.

restli-server/src/main/java/com/linkedin/restli/server/BaseRestLiServer.java
```diff
@@ -233,4 +237,69 @@ private RestLiArgumentBuilder lookupArgumentBuilder(ResourceMethodDescriptor met
...
+      Integer criteriaParamIndex = method.getBatchFinderCriteriaParamIndex();
+      Parameter<?> criteriaParam = parameterList.get(criteriaParamIndex);
+      DataList criteriaList = (DataList) context.getStructuredParameter(criteriaParam.getName());
+      return criteriaList.size();
+    }
+    catch (Exception e)
+    {
+      throw new RestLiServiceException(HttpStatus.S_400_BAD_REQUEST,
+          String.format("The batchFinder method: %s does not provide the correct criteria list: %s .",
+              method.getMethod(), e.getMessage()));
```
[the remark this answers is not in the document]
karthikbalasub: On second thought, how about moving this validation after we build the arguments? 
Look for `restLiArgumentBuilder.buildArguments` in RestLiMethodInvoker (line 324). 

Validating input params should happen before we check for batch size. Also, the argument builder has special logic to wrap single element into an array (this can happen for protocol 1.0.0).

What do you think?

restli-tools/src/main/java/com/linkedin/restli/tools/compatibility/ResourceCompatibilityChecker.java
```diff
@@ -1382,4 +1389,73 @@ private void loadResourceLevelServiceErrors(RecordTemplate previousRecord, Recor
...
+    else
+    {
+      int prevValue = prevMaxBatchSize.getValue();
+      int currValue = currMaxBatchSize.getValue();
+      boolean prevValidate = prevMaxBatchSize.isValidate();
+      boolean currValidate = currMaxBatchSize.isValidate();
+      if(prevValue < currValue)
+      {
+        // Increasing max batch size value
+        _infoMap.addRestSpecInfo(CompatibilityInfo.Type.MAX_BATCH_SIZE_VALUE_INCREASED, _infoPath);
```
karthikbalasub, line 1434: This can be incompatible if prevValidate = false and currValidate == true

restli-tools/src/main/java/com/linkedin/restli/tools/compatibility/ResourceCompatibilityChecker.java
```diff
@@ -1382,4 +1389,73 @@ private void loadResourceLevelServiceErrors(RecordTemplate previousRecord, Recor
...
+      else if (prevValue == currValue)
+      {
+        if (prevValidate && !currValidate)
+        {
+          _infoMap.addRestSpecInfo(CompatibilityInfo.Type.MAX_BATCH_SIZE_VALUE_TURN_OFF_VALIDATION, _infoPath);
+        }
+        else if (!prevValidate && currValidate)
+        {
+          _infoMap.addRestSpecInfo(CompatibilityInfo.Type.MAX_BATCH_SIZE_VALUE_TURN_ON_VALIDATION, _infoPath);
+        }
```
karthikbalasub, line 1445: This should be the highest precedence check

restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
```diff
@@ -3201,4 +3216,34 @@ private static String buildExceptionLocationString(Class<?> resourceClass, Metho
...
+    }
+
+    // Only batch methods are allowed to use MaxBatchSize annotation.
+    if (!BATCH_METHODS.contains(resourceMethod))
+    {
+      throw new ResourceConfigException(String.format("The resource method: %s cannot use MaxBatchSize.",
+          resourceMethod.toString()));
+    }
+    MaxBatchSizeSchema maxBatchSizeSchema = new MaxBatchSizeSchema();
+    maxBatchSizeSchema.setValue(maxBatchSizeAnnotation.value());
```
karthikbalasub, line 3245: Check value should be > 0

restli-server/src/main/java/com/linkedin/restli/server/BaseRestLiServer.java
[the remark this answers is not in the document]
nickibi: I agree with that validate input params should happen before we check batch size. 
Are you suggesting we move the whole batch size validation into the RestLiMethodInvoker, before rest.li invokes the resource method(line 326)?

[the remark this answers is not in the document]
karthikbalasub: yes

restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
```diff
@@ -3201,4 +3216,34 @@ private static String buildExceptionLocationString(Class<?> resourceClass, Metho
...
+    final MaxBatchSize maxBatchSizeAnnotation = method.getAnnotation(MaxBatchSize.class);
+    if (maxBatchSizeAnnotation == null)
+    {
+      return;
+    }
+
+    // Only batch methods are allowed to use MaxBatchSize annotation.
+    if (!BATCH_METHODS.contains(resourceMethod))
+    {
+      throw new ResourceConfigException(String.format("The resource method: %s cannot use MaxBatchSize.",
```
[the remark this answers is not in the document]
nickibi: Fixed

restli-server/src/main/java/com/linkedin/restli/server/BaseRestLiServer.java
```diff
@@ -233,4 +237,69 @@ private RestLiArgumentBuilder lookupArgumentBuilder(ResourceMethodDescriptor met
     }
     return argumentBuilder;
   }
+
+  /**
+   * Method is used to validate the request's batch size is under
```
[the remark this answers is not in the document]
nickibi: Fixed

restli-server/src/main/java/com/linkedin/restli/server/BaseRestLiServer.java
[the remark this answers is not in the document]
nickibi: Moved the validation into RestLiMethodInvoker.

restli-tools/src/main/java/com/linkedin/restli/tools/compatibility/ResourceCompatibilityChecker.java
nickibi: That's right, updated the logic here.

restli-tools/src/main/java/com/linkedin/restli/tools/compatibility/ResourceCompatibilityChecker.java
nickibi: Updated

restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
nickibi: Added maxBatchSize value check.

restli-server/src/main/java/com/linkedin/restli/internal/server/RestLiMethodInvoker.java
```diff
@@ -331,6 +338,60 @@ public void invoke(final RestLiRequestData requestData,
...
+        return 0;
+    }
+  }
+
+  private int getBatchFinderCriteriaNumber(ResourceMethodDescriptor method, ServerResourceContext resourceContext)
+  {
+    List<Parameter<?>> parameterList = method.getParameters();
+    Integer criteriaParamIndex = method.getBatchFinderCriteriaParamIndex();
+    Parameter<?> criteriaParam = parameterList.get(criteriaParamIndex);
+    DataList criteriaList = (DataList) resourceContext.getStructuredParameter(criteriaParam.getName());
```
karthikbalasub, line 391: This can be a single entity (not DataList) for protocol 1.0.0. There is a special logic to convert it to *Array type when we build args. It would be cleaner to use the args for the batch finder case.

restli-int-test/src/test/java/com/linkedin/restli/examples/TestMaxBatchSize.java
```diff
@@ -0,0 +1,320 @@
...
+
+  @Test(dataProvider = com.linkedin.restli.internal.common.TestConstants.RESTLI_PROTOCOL_1_2_PREFIX + "requestBuilderDataProvider")
+  public void testBatchFinderWithMaxBatchSizeUnderLimitation(RootBuilderWrapper<Long, Greeting> builders) throws RemoteInvocationException
+  {
+    GreetingCriteria c1 = new GreetingCriteria().setId(1l);
+    GreetingCriteria c2 = new GreetingCriteria().setId(2l);
+
+    Request<BatchCollectionResponse<Greeting>> request = builders.batchFindBy("searchGreetings")
+        .setQueryParam("criteria",
+        Arrays.asList(c1, c2)).build();
```
karthikbalasub, line 276: Add a test with just one criteria.

restli-server/src/main/java/com/linkedin/restli/internal/server/RestLiMethodInvoker.java
nickibi: You are talking about this logic https://github.com/linkedin/rest.li/blob/99999589f89a168a0528aabc4d25c797b755b57b/restli-server/src/main/java/com/linkedin/restli/internal/server/methods/arguments/ArgumentBuilder.java#L505, right? However, for batchFInder case it will never fall in to this statement:https://github.com/linkedin/rest.li/blob/99999589f89a168a0528aabc4d25c797b755b57b/restli-server/src/main/java/com/linkedin/restli/internal/server/methods/arguments/ArgumentBuilder.java#L225. Because batchFinder criteria param is required to use array of record template: https://linkedin.github.io/rest.li/batch_finder_resource_method#method-annotation-and-parameters. It will go to buildRegularArgument path https://github.com/linkedin/rest.li/blob/99999589f89a168a0528aabc4d25c797b755b57b/restli-server/src/main/java/com/linkedin/restli/internal/server/methods/arguments/ArgumentBuilder.java#L232, because the type is array.
I added a test which has one criteria, it works for both protocol 1.0 and 2.0. The request was build through rest.li request builder. 
I also tested build request by a url. It works for this one "/batchGreeting?bq=searchGreetings&criteria[0].id=1".
If I changed the request to "/batchGreeting?bq=searchGreetings&criteria.id=1" the ArgumentBuilder throws exception here: https://github.com/linkedin/rest.li/blob/99999589f89a168a0528aabc4d25c797b755b57b/restli-server/src/main/java/com/linkedin/restli/internal/server/methods/arguments/ArgumentBuilder.java#L340, because a datamap cannot cast to a datalist. 
It means if the code reaches the validateMaxBatchSize, it should be a valid case, it is a DataList. 
I could change it to use args, like (RecordTemplate[])(arg[criteriaParamIndex]).length. But I do not see much value it brings.
Let me know what do you think. Thanks.

restli-int-test/src/test/java/com/linkedin/restli/examples/TestMaxBatchSize.java
nickibi: Added one test with just one criteria.

restli-common/src/main/pegasus/com/linkedin/restli/restspec/RestMethodSchema.pdl
```diff
@@ -29,4 +29,10 @@ record RestMethodSchema includes CustomAnnotationSchema, ServiceErrorsSchema, Su
    * Indicates if this rest method has paging support using the start and count parameters
    */
   pagingSupported: optional boolean
+
+  /**
+   * Specifies the max batch size is allowed for this method. It supports BATCH_GET, BATCH_CREATE, BATCH_DELETE,
```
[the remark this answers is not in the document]
nickibi: fixed

restli-server/src/main/java/com/linkedin/restli/internal/server/RestLiMethodInvoker.java
karthikbalasub: For array of records, we generate FooArray classes that extend from WrappingArrayTemplate, which is a sub class of DataTemplate.

I think your logic works for batchFinder because we parse it always as DataList. 
`paramValue.getClass() != DataList.class` will be false in the first check you linked. 

Why we parse it always as DataList? Could be because of the way the data is represented in the URI (and this is an array of complex types).
`criteria[0].id=1` 
in the URI. See the logic here: https://github.com/linkedin/rest.li/blob/master/restli-common/src/main/java/com/linkedin/restli/internal/common/QueryParamsDataMap.java#L288

So for batch finder, your current logic is fine as criteria should be an array of records.

karthikbalasub (approved):

PR #506, opened 2021-01-07
Detect and resolve naming conflits for unnamed inner types(array, map and union).

karthikbalasub:
Java doesn't allow inner class names to be same as enclosing classes. Detect and resolve such naming conflicts for unnamed inner types(array, map and union).

### DIFF
--- a/generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
+++ b/generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
@@ -924,8 +924,7 @@ public class TemplateSpecGenerator
       if (ancestorClass.getClassName().equals(className))
       {
         className = className + CLASS_NAME_SUFFIX_SEPARATOR + suffix;
-        // Check for name conflict from the immediate enclosing class again.
-        ancestorClass = enclosingClass;
+        break;
       }
       else
       {
--- a/generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
+++ b/generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
@@ -832,7 +832,12 @@ public class TemplateSpecGenerator
           final ClassInfo classInfo = classNameForUnnamedTraverse(enclosingClass, memberName, arraySchema.getItems());
           // Add just the "Array" suffix first. This is to ensure backwards compatibility with the old codegen logic.
           String className = classInfo.name + ARRAY_SUFFIX;
-          className = resolveInnerClassName(enclosingClass, className, ARRAY_SUFFIX);
+          // If this array is for an unnamed inner type (e.g, union) then this will be inner class. So, ensure the Array
+          // class name doesn't conflict with ancestor class names.
+          if (enclosingClass != null && classInfo.namespace.equals(enclosingClass.getFullName()))
+          {
+            className = resolveInnerClassName(enclosingClass, className, ARRAY_SUFFIX);
+          }
           classInfo.name = className;
           return classInfo;
         }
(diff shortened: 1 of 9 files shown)


generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
```diff
@@ -896,6 +906,35 @@ private ClassInfo classNameForUnnamedTraverse(ClassTemplateSpec enclosingClass,
...
+   * @return a class name that doesn't conflict with any of the ancestor classes.
+   */
+  private String resolveInnerClassName(ClassTemplateSpec enclosingClass, String className, String suffix) {
+    ClassTemplateSpec ancestorClass = enclosingClass;
+    while (ancestorClass != null)
+    {
+      if (ancestorClass.getClassName().equals(className))
+      {
+        className = className + CLASS_NAME_SUFFIX_SEPARATOR + suffix;
+        // Check for name conflict from the immediate enclosing class again.
```
mchen07, line 927: why do we need this?

saponniah:
LGTM.

saponniah (approved):

generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
```diff
@@ -896,6 +906,35 @@ private ClassInfo classNameForUnnamedTraverse(ClassTemplateSpec enclosingClass,
...
+   * class name for an inner class and its enclosing class and resolves to a name that doesn't conflict. It does this
+   * by adding a special separator {@link #CLASS_NAME_SUFFIX_SEPARATOR} and the provided suffix to the classname
+   * whenever a conflict is detected.
+   * The special separator is needed to ensure the new name does not conflict with classes generated from other fields.
+   * @param enclosingClass Class enclosing the inner class.
+   * @param className Candidate name for the innerclass
+   * @param suffix Suffix to add to the className if a conflict is found.
+   * @return a class name that doesn't conflict with any of the ancestor classes.
+   */
+  private String resolveInnerClassName(ClassTemplateSpec enclosingClass, String className, String suffix) {
```
mchen07, line 920: where do we guard sibling name conflict?

karthikbalasub: There wouldn't be a conflict because we use the '$' char.

generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
karthikbalasub: Not needed anymore. I added this for the original idea without '$' char.

generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
mchen07: make sense.

mchen07 (approved):
LGTM, thanks for very quick turnaround.

generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
mchen07: Wait a minute, it just occurred to me such a counter example, let's say that domain owners copy and paste, and accidentally use "unionNameConflictArray" as a union field name. Then for the first array field "unionNameConflict", it will use "UnionNameConflictArray" since it is not conflicting with enclosing Class. Now it goes to that union field "unionNameConflictArray", we check that it is not conflicting with enclosing class either, so we just use "UnionNameConflictArray" as inner class name. Now we got sibling conflict, right?

```
record UnionNameConflictRecord {
  // Union's inner class should be UnionNameConflict, the array class will be UnionNameConflictArray$Array to avoid
  // naming conflict with record class.
  unionNameConflict: array[union[int, string]]
  // Union's inner class should be UnionNameConflictArray$Union to avoid naming conflict, and the array class will be
  // UnionNameConflictArray$UnionArray
  unionNameConflictArray: union[int, string]
} 
```

karthikbalasub: I didn't handle this case on purpose as we cannot do it without breaking backwards compatibility. We already add Array and Map suffix for all array/map fields. And this can conflict with other fields. Any solution to fix that would change the generated class names.

junchuanwang:
I think the code logic is reasonably understandable. I don't have issue with the implementation itself.

But what made me curious is what is the problem without your change? The user see build errors and they would try to fix the schema name in their change. Isn't that simpler? Would the rules we introduced hide the real problem?

karthikbalasub:
> I think the code logic is reasonably understandable. I don't have issue with the implementation itself.
> 
> But what made me curious is what is the problem without your change? The user see build errors and they would try to fix the schema name in their change. Isn't that simpler? Would the rules we introduced hide the real problem?

PDLs are also generated for usecases outside of rest.li framework. Some of these usecases publish the PDL and do not generate template classes at that time. So the problem will be detected later. Also, restricting PDL modeling based on limitation in one language (java) is not ideal.

junchuanwang (approved):

mchen07 (approved):

PR #509, opened 2021-01-15
Expose resource method parameters in FilterRequestContext

evanw555:
The interface has a default implementation for now to keep it
compatible, but that should be removed in the next major version.


nickibi (approved):
LGTM

karthikbalasub (approved):

PR #510, opened 2021-01-19
Add Java 11 testing

jessejie7:
This PR configures rest.li's CI builds to run additional tests using Java 11. This is meant to be in addition to the Java 8 build and tests. As in the workflow should go: 

1. Java 8 build
2. Java 8 tests 
3. Java 11 tests

The tests run by Java 8 and Java 11 are the same but they are just run by different jvms.

Using adopt openjdk for both builds. 1.8.0_232 for Java 8 and 11.0.9 for Java 11.

We add the alternateJvmPlugin which allows us to run Tests with a different jvm by specifying the alternateJvm gradle property. These properties can be specified by the -P parameter. 

The recommended way of using this is to run `./gradlew check -PalternateJvm=$JAVA11_HOME/bin/java`. 

Please do not specify this property in the gradle.properties file as it would then make all tests (including the Java 8 build) run with Java 11.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,7 @@ env:
     - GRAVIS="$HOME/gravis"
     # Select the versions of Java 8 and Java 11 that is to be used. A list of available versions can be seen with "jabba ls-remote"
     - JDK8="adopt@1.8.0-232"
+    - JDK=$JDK8
     - JDK11="adopt@1.11.0-9"
 
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: java
 jdk:
+  # JDKs will be maintained by jabba so this entry doesn't do anything
   - oraclejdk8
 
 # Anything after trusty doesn't have Java 8
@@ -10,7 +11,14 @@ dist: trusty
 install: true
 
 # Build commits to master, PRs to master, and release tags
-script: ./scripts/travis/build.sh
+script:
+  # This variable is used by the build script in order to run tests with Java 11
+  - export JAVA11_HOME=$HOME/.jabba/jdk/$JDK11
+  - echo "Java 11 version used for tests:"
+  - $JAVA11_HOME/bin/java -version
+  - echo "Java 8 version used for compilation and tests:"
+  - java -version
+  - ./scripts/travis/build.sh
 branches:
   only:
     - master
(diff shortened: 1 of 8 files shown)


scripts/travis/build.sh
```diff
@@ -53,9 +53,29 @@ fi
...
+  echo "Java 8 build and test failed"
+  exit 1
+fi
+
+# Run tests with Java 11
+# .travis.yml provides this environment variable
+EXIT_CODE_JAVA11=0
+if [ ! -z "$JAVA11_HOME" ]; then
+  echo "Running tests with Java 11"
+  ./gradlew check -PalternateJvm=$JAVA11_HOME/bin/java
```
evanw555, line 70: We don't run tests in all cases. Note how line 53 has `$EXTRA_ARGS`, which can sometimes be `-x test`. This is when we're doing the release-and-publish job. We don't want to run java 11 tests in this case.

jessejie7: Fixed

junchuanwang:
Hi Jesse, what will be the timeline to ask for this PR merged?

jessejie7:
No idea, there are way too many flaky tests. You can see that each build is basically failing at a different step.

jessejie7:
At this point I'm not sure if this PR is usable. I've only succeeded like once out of 20 tries

.travis.yml
```diff
@@ -1,5 +1,6 @@
 language: java
 jdk:
+  # JDKs will be maintained by jabba so this entry doesn't do anything
   - oraclejdk8
```
DPUkyle, line 4: This seems like a lot of complexity. Have you instead tried just tweaking this?
```
jdk:
  - oraclejdk8
  - openjdk11
```

Sources: 
 - https://blog.travis-ci.com/support_for_multiple_jdks (Search for `How do I turn it on already?`)
 - https://docs.travis-ci.com/user/reference/trusty/#jvm-clojure-groovy-java-scala-images

DPUkyle (changes requested):

.travis.yml
jessejie7: I tested that at some point as well but IIRC it didn't install the binaries in accessible locations and/or couldn't change the JDKs on the fly properly.

ChrisCarini:

PR #511, opened 2021-01-20
Fix bug in DataComplexTable that breaks Data::copy if there are hash collisions.

karthikbalasub:
Hashcodes for DataComplex objects are generated using a thread local,
and there can be collisions if multiple threads are used to construct a
DataComplex object.

### DIFF
--- a/data/src/main/java/com/linkedin/data/DataComplexTable.java
+++ b/data/src/main/java/com/linkedin/data/DataComplexTable.java
@@ -18,6 +18,11 @@ package com.linkedin.data;
 
 import java.util.HashMap;
 
+
+/**
+ * Custom hash table when {@link DataComplex} objects are used as keys. This utilizes the custom
+ * {@link DataComplex#dataComplexHashCode()} as the hash for improved performance.
+ */
 class DataComplexTable
 {
   private final HashMap<DataComplexKey, DataComplex> _map;
@@ -57,6 +62,8 @@ class DataComplexTable
     @Override
     public boolean equals(Object other)
     {
+      // "other" is guaranteed to be DataComplex as this class is "private" scoped within DataComplexTable, which only
+      // supports DataComplex objects.
       return _dataObject == ((DataComplexKey) other)._dataObject;
     }
   }
--- a/data/src/main/java/com/linkedin/data/DataComplexTable.java
+++ b/data/src/main/java/com/linkedin/data/DataComplexTable.java
@@ -20,7 +20,7 @@ import java.util.HashMap;
 
 class DataComplexTable
 {
-  private final HashMap<Integer, DataComplex> _map;
+  private final HashMap<DataComplexKey, DataComplex> _map;
 
   DataComplexTable()
   {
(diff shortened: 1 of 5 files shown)


data/src/test/java/com/linkedin/data/TestDataComplexTable.java
```diff
@@ -0,0 +1,111 @@
+package com.linkedin.data;
```
junchuanwang, line 1: nit: Copyright info

data/src/test/java/com/linkedin/data/TestData.java
```diff
@@ -866,6 +866,23 @@ public void testCopy() throws CloneNotSupportedException
...
+    DataMap a = new DataMap();
+    a.put("b", new DataMap());
+    a.put("c", new DataMap());
+
+    // Use the next hashcode to cause collision.
+    a.getDataMap("c")._dataComplexHashCode = DataComplexHashCode.nextHashCode() + 1;
+    DataMap copy = a.copy();
+    assertNotNull(copy.get("b"));
+    assertNotNull(copy.get("c"));
+    assertNotSame(copy.get("b"), copy);
```
junchuanwang, line 882: what is the purpose of test the comparision of  "b"

data/src/test/java/com/linkedin/data/TestDataComplexTable.java
```diff
@@ -0,0 +1,111 @@
...
+    table.put(item1, item1Clone);
+    table.put(item2, item2Clone);
+
+    Assert.assertNotNull(table.get(item1));
+    Assert.assertNotNull(table.get(item2));
+    Assert.assertSame(item1Clone, table.get(item1));
+    Assert.assertSame(item2Clone, table.get(item2));
+  }
+
+  private static class MockDataComplex implements DataComplex
```
junchuanwang, line 31: It might be clearer to throw `UnsupportedOperationException` in methods that not intended to be overridden.

junchuanwang (approved):
Ship with comments

data/src/main/java/com/linkedin/data/DataComplexTable.java
```diff
@@ -29,11 +29,36 @@
...
+    @Override
+    public int hashCode()
+    {
+      return _hashCode;
+    }
+
+    @Override
+    public boolean equals(Object other)
+    {
+      return _dataObject == ((DataComplexKey) other)._dataObject;
```
BrianPin, line 60: Should there be some usual check for the object?
Something like the equals override in this link
https://www.baeldung.com/java-hashcode

data/src/test/java/com/linkedin/data/TestData.java
karthikbalasub: This is to check if the copy is not reusing same datamap object at different levels due to hash collision. (this was happening before the fix).

data/src/main/java/com/linkedin/data/DataComplexTable.java
karthikbalasub: Not needed as the other object will always be DataComplex. This Key type will only be used within DataComplexTable. 
Added a comment to explain this.

BrianPin (approved):

data/src/test/java/com/linkedin/data/TestData.java
junchuanwang: But you didn't seem to create collision for “b"?

karthikbalasub: yes, but still no harm in testing it? :)

junchuanwang: sure! Same thought. Thanks for confirmation

PR #512, opened 2021-01-25
[bugfix] Prevent eager resolution of UNUSED_CONFIGURATIONS values

DPUkyle:
- Invoking Configuration#isEmpty() will force the underlying Configuration to resolve all files - this is bad and can break native IDEA project import
 - Rather, we can see if the Configuration has any first-order dependencies associated with it, which is probably the author's original intent
 - If the Configuration exists and contains any first-order dependencies, only then should we warn.


DPUkyle:
This is just a bugfix which unblocks native IntelliJ import and _may_ have a small performance bump for command-line or CI builds; however I do not think it bears mentioning in the CHANGELOG.

evanw555 (approved):

PR #515, opened 2021-01-26
Sort properties and defaults when encoding PDLs.

tjni:
The PDL encoder already sorts the top-level properties on a field by
their names. This extends this sorting to nested properties.

An example of a nested property that is sorted is:

    @nested = {
      "a" : "a",
      "b" : "b"
    }

When default values are records, their fields are sorted in the order
that the fields are defined in the schema. For example:

    date: record Date {
      month: int
      day: int
      year: int
    } = {
      "month": 1,
      "day": 1,
      "year": 2021
    }

This enables the PDL encoder to be used for autoformatting PDLs, since
the order of map entries will be deterministic.

### DIFF
--- a/data/src/main/java/com/linkedin/data/codec/AbstractJacksonDataCodec.java
+++ b/data/src/main/java/com/linkedin/data/codec/AbstractJacksonDataCodec.java
@@ -67,9 +67,16 @@ public abstract class AbstractJacksonDataCodec implements DataCodec
 
   protected final JsonFactory _factory;
 
+  private boolean _sortKeys;
+
   protected AbstractJacksonDataCodec(JsonFactory factory)
   {
     _factory = factory;
+  }
+
+  public void setSortKeys(boolean sortKeys)
+  {
+    _sortKeys = sortKeys;
   }
 
   @Override
@@ -135,7 +142,7 @@ public abstract class AbstractJacksonDataCodec implements DataCodec
 
   protected Data.TraverseCallback createTraverseCallback(JsonGenerator generator)
   {
-    return createTraverseCallback(generator, false);
+    return createTraverseCallback(generator, _sortKeys);
   }
 
   /**
@@ -167,10 +174,6 @@ public abstract class AbstractJacksonDataCodec implements DataCodec
     {
       return new Parser().parse(jsonParser, expectType);
     }
-    catch (IOException e)
-    {
-      throw e;
-    }
     finally
     {
       DataCodec.closeQuietly(jsonParser);
@@ -196,10 +199,6 @@ public abstract class AbstractJacksonDataCodec implements DataCodec
     try
     {
       return new Parser(true).parse(jsonParser, mesg, locationMap);
-    }
-    catch (IOException e)
-    {
-      throw e;
     }
     finally
     {
--- a/data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
+++ b/data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
@@ -21,7 +21,9 @@ import com.linkedin.data.DataList;
 import com.linkedin.data.DataMap;
 import com.linkedin.data.Null;
 import com.linkedin.data.codec.JacksonDataCodec;
+import com.linkedin.data.template.JacksonDataTemplateCodec;
 import java.io.IOException;
+import java.io.StringWriter;
 import java.io.Writer;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -51,6 +53,7 @@ abstract class PdlBuilder
   ));
   private static final char ESCAPE_CHAR = '`';
   private static final Pattern IDENTIFIER_CHARS = Pattern.compile("[0-9a-zA-Z_-]*");
+  private static final int DEFAULT_JSON_CODEC_BUFFER_SIZE = 4096;
 
   /**
    * Each subclass must define a provider for creating new instances.
@@ -254,13 +257,33 @@ abstract class PdlBuilder
    * Writes an object as raw encoded JSON text.
    * Valid types: DataList, DataMap, String, Int, Long, Float, Double, Boolean, ByteArray
    *
+   * The keys in DataMaps will be sorted in alphabetic order.
+   *
    * @param value JSON object to write
    */
-  abstract PdlBuilder writeJson(Object value) throws IOException;
+  PdlBuilder writeJson(Object value) throws IOException
+  {
+    return writeJson(value, null);
+  }
+
+  /**
+   * Writes an object as raw encoded JSON text.
+   * Valid types: DataList, DataMap, String, Int, Long, Float, Double, Boolean, ByteArray
+   *
+   * When a {@code schema} is present, keys in a DataMap will be sorted in the
+   * order of fields in the corresponding schema. When it is missing, keys in a
+   * DataMap will be sorted in alphabetic order.
+   *
+   * @param value JSON object to write
+   * @param schema an optional schema to use for sorting
+   */
+  abstract PdlBuilder writeJson(Object value, DataSchema schema) throws IOException;
 
   /**
    * Serializes a pegasus Data binding type to JSON.
    * Valid types: DataMap, DataList, String, Number, Boolean, ByteString, Null
+   *
+   * The keys in a DataMap will be sorted in alphabetic order.
    *
    * @param value the value to serialize to JSON.
    * @return a JSON serialized string representation of the data value.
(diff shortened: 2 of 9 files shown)


junchuanwang:
Can you also test if this makes other things sorted when encoding pdl, such as field default value list(or map)?

tjni:
It does appear to also sort the default value when it's a map. Maybe this shouldn't happen, since the fields of the record corresponding to the default value might not be sorted?

junchuanwang:
Sorted order in default value actually seems OK to me but I will confirm with the team

evanw555:
Synced up with @tjni offline. The requirement is just for the JSON objects in the PDL properties to be encoded deterministically. We've determined that sorting them alphabetically is fine, as users probably won't care enough to preserve the original ordering of a JSON object...

As for the default value JSON, we determined that the ideal behavior here is to sort the JSON objects in the record's original field ordering. This would require the JSON encoder to be schema-aware. @tjni is currently looking into this.

aman1309:
> Synced up with @tjni offline. The requirement is just for the JSON objects in the PDL properties to be encoded deterministically. We've determined that sorting them alphabetically is fine, as users probably won't care enough to preserve the original ordering of a JSON object...
> 
> As for the default value JSON, we determined that the ideal behavior here is to sort the JSON objects in the record's original field ordering. This would require the JSON encoder to be schema-aware. @tjni is currently looking into this.

we could have two Json encoder one for properties and another for others. The Codec can not be schema aware.

tjni:
I've updated the code to use a different codec for schema-aware sorting of JSON maps. I found `JacksonDataTemplateCodec`, which I am reusing for this.

data/src/test/java/com/linkedin/data/schema/TestSchemaToPdlEncoder.java
```diff
@@ -50,4 +50,109 @@ public void testEncodeRecordWithEmptyDataMapInProperty() throws IOException
...
+        "      \"c1\" : true,",
+        "      \"c3\" : [ \"one\", \"two\" ]",
+        "    },",
+        "    \"b2\" : 0.05",
+        "  }",
+        "}"), indentedSchema);
+
+    String compactSchema = SchemaToPdlEncoder.schemaToPdl(schema, SchemaToPdlEncoder.EncodingStyle.COMPACT);
+
+    assertEquals(String.join("",
```
junchuanwang, line 137: can you format this as well, thanks!

data/src/test/java/com/linkedin/data/schema/TestSchemaToPdlEncoder.java
```diff
@@ -50,4 +50,109 @@ public void testEncodeRecordWithEmptyDataMapInProperty() throws IOException
     assertEquals(source.getProperties(), encoded.getProperties());
     assertEquals(source, encoded);
   }
+
+  @Test
+  public void testEncodeSortsNestedPropertyMap() throws IOException
```
junchuanwang, line 55: Can you do one more test on nested properties larger than 1 nested layer?

junchuanwang (commented):
No other concerns if the comments of the tests can be fixed.

data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
```diff
@@ -258,6 +261,8 @@ private static String escapeIdentifier(String identifier)
    */
   abstract PdlBuilder writeJson(Object value) throws IOException;
 
+  abstract PdlBuilder writeDefaultValue(Object value, DataSchema schema) throws IOException;
```
evanw555, line 264: I recommend not introducing a method unique to default values, but rather just extend the `writeJson` method to accept a second parameter (of type `DataSchema`). For this parameter, a non-null value indicates a schema-based ordering to use, and `null` indicates to just use alphabetical ordering. This would be cleaner IMO.

data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
```diff
@@ -258,6 +261,8 @@ private static String escapeIdentifier(String identifier)
    */
   abstract PdlBuilder writeJson(Object value) throws IOException;
```
evanw555, line 262: If you agree with my suggestion below, then this method should no longer be abstract and it instead should just delegate to `writeJson(value, null)`

data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
```diff
@@ -51,6 +53,7 @@
   ));
   private static final char ESCAPE_CHAR = '`';
   private static final Pattern IDENTIFIER_CHARS = Pattern.compile("[0-9a-zA-Z_-]*");
+  private static final int DEFAULT_BUFFER_SIZE = 4096;
```
evanw555, line 56: This might cause confusion, since the compact PDL builder has its own concept of a "buffer". I suggest renaming to `DEFAULT_JSON_CODEC_BUFFER` or something.

data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
```diff
@@ -275,7 +280,26 @@ else if (value instanceof DataList)
     {
       return jsonCodec.listToString((DataList) value);
     }
-    else if (value instanceof String)
+
+    return toPrimitiveValue(value);
+  }
+
+  protected String toOrderedBySchemaJson(Object value, DataSchema schema, JacksonDataTemplateCodec jsonCodec)
```
evanw555, line 287: Similar to my other suggestion, I suggest collapsing `toOrderedBySchemaJson` and `toJson` into one unified interface which accepts a `DataSchema` that may be null or non-null (e.g. `toJson(value, jsonCodec, schema)`. For all these methods, the javadoc should clearly state how the ordering is affected by null vs non-null arguments.

evanw555 (commented):
Thanks for the quick changes. Lucky that we already have a schema-aware JSON codec ;)

My comments are generally about keeping the interfaces in these classes clean by avoiding the creation of too many specific methods.

data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
evanw555: If you make this change, document this clearly in the javadoc

data/src/test/java/com/linkedin/data/schema/TestSchemaToPdlEncoder.java
tjni: I added indentation on each line and trim it prior to joining each string. I hope it looks better now.

data/src/test/java/com/linkedin/data/schema/TestSchemaToPdlEncoder.java
tjni: Good idea! Added testEncodeSortsMultiLevelNestedPropertyMap.

data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
tjni: That makes sense to me.

data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
tjni: Done!

data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
tjni: This isn't possible at the moment because the codec that both each code path accepts is different, and there does not appear to be a common method I can call. I've renamed `toOrderedBySchemaJson` to `toJson` though and added documentation for the behavior.

It might be possible to refactor the underlying codecs so that the PdlBuilder can call the same method on both, which would make this possible. Let me know if you think it's worth pursuing.

data/src/main/java/com/linkedin/data/codec/AbstractJacksonDataCodec.java
```diff
@@ -67,11 +67,18 @@
 
   protected final JsonFactory _factory;
 
+  private boolean _sortProperties;
```
evanw555, line 70: Should've commenting on this earlier, but this should be renamed to `_sortKeys`, since "properties" aren't a concept that's inherently known by the JSON codec.

Also, just making sure... this doesn't sort array values as well, right?

data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
```diff
@@ -275,7 +298,36 @@ else if (value instanceof DataList)
...
+    {
+      StringWriter out = new StringWriter(DEFAULT_JSON_CODEC_BUFFER_SIZE);
+      jsonCodec.writeDataTemplate(value, schema, out, true);
+      return out.toString();
+    }
+
+    return toPrimitiveValue(value);
+  }
+
+  private static String toPrimitiveValue(Object value)
```
evanw555, line 328: This naming isn't quite accurate. It's technically not transforming the data "to a primitive", it's transforming "from primitive data to primitive JSON data".

However, I think dividing the logic like this isn't totally necessary. I suggest packing this logic back into `toJson(value, codec)` and having `toJson(value, codec, schema)` just delegate back to that one for the primitive case.

data/src/main/java/com/linkedin/data/codec/AbstractJacksonDataCodec.java
tjni: Good catch. It should be `_sortKeys` and it doesn't sort arrays. Let me make sure I check that in the unit test.

data/src/main/java/com/linkedin/data/schema/PdlBuilder.java
tjni: I missed the fact that `JacksonDataTemplateCodec` extends `JacksonDataCodec`, which makes this possible. Doing this now.

evanw555 (approved):
LGTM! Thanks for the changes. Let's wait until tomorrow to merge this, just to let others who don't work at night leave comments (and to potentially bundle other changes in with this minor version).

junchuanwang (approved):

PR #516, opened 2021-01-26
Implement fluent bindings for collection resource.

karthikbalasub:
Currently support get, batch_get, create and batch_create methods.


restli-tools/src/main/resources/apiVmTemplates/rest.create_id.vm
```diff
@@ -0,0 +1,59 @@
...
+            Response<IdResponse<${spec.keyClassName}>> response = responseTry.get();
+            if (response.hasError()) {
+              result.completeExceptionally(response.getError());
+            } else {
+              result.complete(response.getEntity().getId());
+            }
+          }
+          return Success.of((Void) null);
+        });
+    _engine.run(responseTask);
```
junchuanwang, line 55: I saw you left comment regarding "scheduleAndRun" draft and said start a new plan if no context found.

I think you are talking about this case?
```java
<completionStageImpl>.andThen( () -> {
<resource>.get(); //try to reuse the existing context, if no context found, start a new plan
})
```
But I see here this line (Rest.Li) framework is the right place to run the _engine, therefore I didn't reply to your comment in that draft.

Do you think we can catch the exception and run a plan if no plan found?

restli-tools/src/main/resources/apiVmTemplates/rest.create_return_entity.vm
```diff
@@ -0,0 +1,64 @@
...
+   limitations under the License.
+*#
+
+  #doc($method.schema.doc)
+  public CompletionStage<IdEntityResponse<${spec.keyClassName}, ${spec.entityClassName}>> create(
+      ${spec.entityClassName} entity,
+      boolean returnEntity#if( $method.hasParams()),#end
+      #methodParams($method)
+      ) {
+    Map<String, Object> queryParams = new HashMap<>();
```
junchuanwang, line 23: There were somehow no indention for these lines of code generated in my machine. Is that different on your side?

restli-tools/src/main/java/com/linkedin/restli/tools/clientgen/FluentApiGenerator.java
```diff
@@ -59,23 +55,8 @@
 {
```
junchuanwang, line 55: Since based on my understanding in the future their could be different implementations.

Right now it might be valuable (though not necessary) to mention current implementation is Parseq & Async based.

Same to the "TestGreetingsFluentClient".

junchuanwang (commented):
Will review a second time. 

But first pass review experience is quite truly "fluent". Neat!

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/GenerateRestClientTask.java
```diff
@@ -221,9 +224,9 @@ public void generate()
           }
           javaExecSpec.setClasspath(_pathedCodegenClasspath);
           javaExecSpec.setMain("com.linkedin.restli.tools.clientgen.FluentApiGenerator");
-          javaExecSpec.args("-resolverPath", resolverPathArg);
-          javaExecSpec.args("-rootPath", getProject().getRootDir().getPath());
-          javaExecSpec.args("-targetDir", _destinationDir.getAbsolutePath());
+          javaExecSpec.args("--resolverPath", resolverPathArg);
```
BrianPin, line 227: Does this mean it didnt work when we do just one dash?

restli-tools/src/main/resources/apiVmTemplates/collection.vm
```diff
@@ -1,4 +1,138 @@
...
+  private final ParSeqRestliClient _client;
+  private final Engine _engine;
+
+  #doc( $spec.resource.doc)
+  public ${spec.className}(ParSeqRestliClient client, Engine engine) {
+    _client = client;
+    _engine = engine;
+  }
+
+  // CRUD methods
```
junchuanwang, line 131: There are excessive blank lines generated here as well

restli-tools/src/main/resources/apiVmTemplates/collection.vm
```diff
@@ -1,4 +1,138 @@
...
+    #else
+     import com.linkedin.restli.client.BatchCreateIdRequest;
+     import com.linkedin.restli.common.CreateIdStatus;
+     import com.linkedin.restli.common.BatchCreateIdResponse;
+     import com.linkedin.restli.internal.client.BatchCreateIdDecoder;
+    #end
+  #end
+#end
+@Generated("Generated from $spec.sourceIdlName")
+public class $spec.className {
```
junchuanwang, line 81: Another thing came to my mind is the name of the generated resource

In this example, Greetings, Greeting, Greeting(s)XXXX, those are actually quite confusing terms if we are going to mix another "generated" thing to such similar naming.

I personally suggest we clarify in some way that is is generated client. For example "GreetingsResource" or "GreetingsClient"

restli-tools/src/main/resources/apiVmTemplates/rest.create_id.vm
karthikbalasub: Throwing -> catching an exception during a normal execution flow is not good (it is also less performant). 

We should atleast make scheduleAndRun return a boolean indicating if the task was scheduled for execution. 
After that we can do the check followed by engine.run here. 

However, adding that to engine itself might be useful for other users. And this logic of checking scheduleAndRun followed by engine.run need not be duplicated everywhere. wdyt?

restli-tools/src/main/resources/apiVmTemplates/rest.create_return_entity.vm
karthikbalasub: That is odd. I bumped velocity version specifically for this and the indentation looks okay on my machine. Did you run ./gradlew build?

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/tasks/GenerateRestClientTask.java
karthikbalasub: yeah, this was a recent code, and we never enabled the task before.

restli-tools/src/main/resources/apiVmTemplates/rest.create_id.vm
junchuanwang: the problem is that the task wouldn't have context regarding which engine runs it.

In the https://github.com/linkedin/parseq/pull/291 I added a API to take engine as a parameter.

restli-tools/src/main/resources/apiVmTemplates/rest.create_return_entity.vm
junchuanwang: It is reproduced in at least two machines (I didn't use Mac)
Screen Shot 2021-01-28 at 2 04 25 AM

karthikbalasub: I also reproduced it. This happened after I moved some of the templates to macros. It looks like macros do not keep the source file indentation. 
The only option is to not use macros, but that would mean logic duplication.
I've edited the macros to indroduce some spaces based on where they are used (not perfect). I guess we'd have to live with this oddity..

restli-tools/src/main/resources/apiVmTemplates/rest.create_id.vm
karthikbalasub: What I meant is Engine::scheduleAndRun(task)

restli-tools/src/main/java/com/linkedin/restli/tools/clientgen/FluentApiGenerator.java
karthikbalasub: I think the implementation would be agnostic to the fluent client. Even if we change the implementation, it would just be a different set of templates. The generated API will not change (thus the Test class would remain unchanged)

restli-tools/src/main/resources/apiVmTemplates/collection.vm
karthikbalasub: These seem to be added by the comments in the templates (copyright). Again, it seems like most template engines have such quirks. Since these are not going to be viewed by users, I think we can ignore those.

restli-tools/src/main/resources/apiVmTemplates/collection.vm
karthikbalasub: The current naming is based on the RFC. The idea is to use "Greetings" for the fluent apis and then generate GreetingClient when they are combined into the universal client.

restli-tools/src/main/resources/apiVmTemplates/rest.create_id.vm
junchuanwang: What is the behavior difference between `Engine::scheduleAndRun(task)` and `task.scheduleOrRun(engine)` proposed?

Also I will also be interested to know what is the use case of
Engine::scheduleAndRun(task)

restli-tools/src/main/java/com/linkedin/restli/tools/clientgen/FluentApiGenerator.java
junchuanwang: I know it is OK for now, just want to share that we can make a distinction now.

Because the tests in `TestGreetingsFluentClient` are not trulyi agnostic (you passed in Parseq component for the test.

I said this b ecause tests such as `TestActionResource` did make a distinction in the test name to suggest it is tests using ParSeq.

This is a suggestion and it will be your choice.

restli-tools/src/main/resources/apiVmTemplates/collection.vm
junchuanwang: I also didn't think that a big problem until I found in the code base and realize that how user can easily give simple and sometime random resouce name so things get confusing.( or maybe it is for this integration testing module?)

Also, at least for example
`find . -name "*greetings*java"`
Most of them are pretty self-explaining, non-conflicting whether auto-generated or not.

restli-tools/src/main/resources/apiVmTemplates/rest.create_return_entity.vm
junchuanwang: I have looked up some documentations and I believe it is tunable. But if it is hard to do so properly now or if it is not so important right now, can we mark it somewhere as a **TODO** item?

I tested with the following and it worked on my machine. It seems we need to both (a) add velocity comment indentation in the marco (b) Add indention through velocity comment when use it

Source: https://stackoverflow.com/questions/43304493/how-to-keep-template-whitespace-tabs-formatting-using-apache-velocity-1-7
(Didn't try with the space gobbling option)

Screen Shot 2021-01-28 at 9 02 39 AM

karthikbalasub: Thanks for trying out, I've made the change. This is still not ideal as the indentation is fixed (what if we use the macro at a different indentation level?)
There are definitely ways to do it, for example, you can pass in an $indent arg to the macro. But that is hard to maintain. I think this is a good trade-off for now.

restli-tools/src/main/resources/apiVmTemplates/collection.vm
karthikbalasub: This naming is only for our integ test module. In real world, service names would be meaningful.

restli-tools/src/main/java/com/linkedin/restli/tools/clientgen/FluentApiGenerator.java
karthikbalasub: agree with you, updated the test name.

junchuanwang (approved):

PR #517, opened 2021-01-27
Fix memory leak in CheckedMap when one map is used to create multiple record templates.

karthikbalasub:
Change listener list now clears finalized weak references when it detects any change listener was finalized or when listeners are notified.

### DIFF
--- a/data/src/main/java/com/linkedin/data/collections/CheckedMap.java
+++ b/data/src/main/java/com/linkedin/data/collections/CheckedMap.java
@@ -23,9 +23,6 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Iterator;
-import java.util.LinkedList;
-import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.BiConsumer;
@@ -166,7 +163,7 @@ public class CheckedMap<K,V> implements CommonMap<K,V>, Cloneable
   {
     checkMutability();
     Set<K> keys = null;
-    if (_changeListeners != null)
+    if (_changeListenerHead != null)
     {
       keys = new HashSet<>(keySet());
     }
@@ -184,7 +181,7 @@ public class CheckedMap<K,V> implements CommonMap<K,V>, Cloneable
     CheckedMap<K,V> o = (CheckedMap<K,V>) super.clone();
     o._map = (HashMap<K,V>) _map.clone();
     o._readOnly = false;
-    o._changeListeners = null;
+    o._changeListenerHead = null;
     o._changeListenerReferenceQueue = null;
     return o;
   }
@@ -353,26 +350,18 @@ public class CheckedMap<K,V> implements CommonMap<K,V>, Cloneable
       return;
     }
 
-    if (_changeListeners == null)
+    if (_changeListenerReferenceQueue == null)
     {
-      // Change listeners are mostly used by map wrappers, and we always iterate through them
-      // linearly, so use a linked list.
-      _changeListeners = new LinkedList<>();
       _changeListenerReferenceQueue = new ReferenceQueue<>();
     }
-
+    else
+    {
+      purgeStaleChangeListeners();
+    }
     // Maintain a weak reference to to the listener to avoid leaking the wrapper beyond its
     // lifetime.
-    _changeListeners.add(new WeakReference<>(listener, _changeListenerReferenceQueue));
-    // Clear finalized weak references.
-    if (_changeListenerReferenceQueue.poll() != null)
-    {
-      while (_changeListenerReferenceQueue.poll() != null)
-      {
-        // Do nothing, as we are just clearing the reference queue.
-      }
-      _changeListeners.removeIf(weakReference -> weakReference.get() == null);
-    }
+    _changeListenerHead = new  WeakListNode<>(
+        new WeakReference<>(listener, _changeListenerReferenceQueue), _changeListenerHead);
   }
 
   final private void checkKeyValue(K key, V value)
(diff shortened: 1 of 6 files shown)


aman1309 (approved):

data/src/main/java/com/linkedin/data/collections/CheckedMap.java
```diff
@@ -356,11 +358,21 @@ public final void addChangeListener(ChangeListener<K, V> listener)
       // Change listeners are mostly used by map wrappers, and we always iterate through them
       // linearly, so use a linked list.
       _changeListeners = new LinkedList<>();
+      _changeListenerReferenceQueue = new ReferenceQueue<>();
```
karthikrg, line 361: ReferenceQueue maybe fairly heavy. Have we considered a design of having a static global referencequeue and using a cleanup thread to poll and cleanup references? This is similar to what guava does for it's cache with weak keys.

Also if the queue grows long, the iteration through the linkedlist to remove the node will be an O(N) operation everytime. Have we considered using a custom LinkedList impl where we have access to the Node container so that we can do an O(1) removal?

karthikbalasub: Running a separate thread would locking to ensure no race condition when the change listener list is modified. Also, the change listener list is already being updated in multiple threads when the same datamap is wrapped into records in different threads. So I'm worried the O(1) removal logic would run into concurrency issues. 

I'm not that worried about the list growing too big as the change listeners are mostly short-lived (request scoped).

data/src/main/java/com/linkedin/data/collections/CheckedMap.java
```diff
@@ -541,8 +556,60 @@ private void notifyChangeListenersOnClear(Set<? extends K> keys)
...
   }
 
+  private void purgeStaleChangeListeners()
+  {
+    if (_changeListenerReferenceQueue != null && _changeListenerReferenceQueue.poll() != null)
+    {
+      // Clear finalized weak references.
+      while (_changeListenerReferenceQueue.poll() != null)
+      {
+        // Do nothing, as we are just clearing the reference queue.
```
BrianPin, line 566: Sorry, new to this, but wont this cause infinite looping?

data/src/main/java/com/linkedin/data/collections/CheckedMap.java
```diff
@@ -541,8 +556,60 @@ private void notifyChangeListenersOnClear(Set<? extends K> keys)
...
+      }
+      // Now iterate over change listeners and purge stale references.
+      WeakListNode<ChangeListener<K, V>> node = _changeListenerHead,  prev = null;
+      while (node != null)
+      {
+        if (node._object.get() == null)
+        {
+          if (prev == null)
+          {
+            _changeListenerHead = node._next;
```
BrianPin, line 576: So when the head points to the next node, the original head will be GCed by jvm?

data/src/main/java/com/linkedin/data/collections/CheckedMap.java
karthikbalasub: poll removes the head of the reference queue.

nickibi: It calls queue.poll(), which remove the head of the queue. Once the queue is empty, it will return null, then the loop ends.

data/src/main/java/com/linkedin/data/collections/CheckedMap.java
karthikbalasub: yes.

data/src/main/java/com/linkedin/data/collections/CheckedMap.java
```diff
@@ -481,50 +482,64 @@ private boolean assertCheckMap(Map<? extends K, ? extends V> map)
...
   {
-    if (_changeListeners == null)
+    purgeStaleChangeListeners();
+    if (_changeListenerHead == null)
     {
       return;
     }
-
-    _changeListeners.forEach(listenerRef -> {
+    WeakListNode<ChangeListener<K, V>> node = _changeListenerHead;
```
karthikrg, line 490: Suggest providing a utility iteration method to keep logic simple?

data/src/main/java/com/linkedin/data/schema/validator/DataSchemaAnnotationValidator.java
```diff
@@ -424,6 +424,7 @@ public void callback(List<String> path, DataSchema schema)
         {
           Constructor<? extends Validator> ctor = clazz.getConstructor(DataMap.class);
           DataMap configDataMap = (DataMap) config;
+          configDataMap.makeReadOnly();
```
BrianPin, line 427: Since this is the single line change here, I am out of context, do you mind say why we need this readOnly?

nickibi (approved):

data/src/main/java/com/linkedin/data/collections/CheckedMap.java
karthikrg: Yeah this seems incorrect and also inefficient? Wouldn't it be better to just go over the referencequeue pull elements out one by one and delete them inline rather than iterate through the entire list?

data/src/main/java/com/linkedin/data/schema/validator/DataSchemaAnnotationValidator.java
karthikbalasub: If a DataMap is marked readonly we don't even add change listeners when it is wrapped by record template.
readOnly => will not change => no need for change listener.

So making it readonly here ensures that validator instances do not create a change listener for each field being validated. While the fix in CheckedMap ensures we GC change listeners correctly, this change avoids creating a large list of listeners in the first place.

BrianPin (approved):
My part is OK, LGTM

data/src/main/java/com/linkedin/data/schema/validator/DataSchemaAnnotationValidator.java
karthikbalasub: Added comments

data/src/main/java/com/linkedin/data/collections/CheckedMap.java
karthikbalasub: good point, done.

data/src/main/java/com/linkedin/data/collections/CheckedMap.java
karthikbalasub: To decide between using a doubly linked list and purge as we poll from reference queue vs current impl which traverses the singly linked list separately:

The trade off is extra memory for prev pointer vs additional iteration of the list. And if we use both prev and next pointer, two links are updated for each node during purge.
 So overall the amount of time would be similar: one iteration with two pointers vs two iteration with one pointer.

If we assume the list would be small or most of the items in the list will be purged(this is the expected behavior), the amortized cost of extra iteration is not bad. It gets bad only if a small subset of items are purged.

evanw555 (approved):

PR #522, opened 2021-02-01
Update Netty and depend on individual artifacts.

tjni:
The main objective of this patch is to stop depending on the netty-all
uberjar in favor of the individual netty-* artifacts. This reduces the
weight of the dependencies that rest.li needs (as netty-all weighs in
at a hefty 4.1 MB).

More importantly, this plays much more nicely with other dependencies
that use the individual artifacts. Currently, there are issues inside
LinkedIn where the netty-all artifact shades some but not all of the
netty-* artifacts on the classpath, leading to runtime errors.

For discussion about this problem on the Netty Github, see:

  https://github.com/netty/netty/issues/4671

Since this is a backwards incompatible change, this patch also goes
ahead and updates the Netty version too, cause why not.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -40,6 +40,10 @@ File getEnvironmentScript()
 apply from: environmentScript
 apply from: "${buildScriptDirPath}/configBuildScript.gradle"
 
+def versions = [
+  'netty': '4.1.58.Final'
+]
+
 project.ext.externalDependency = [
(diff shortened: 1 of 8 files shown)


tjni:
After internal discussion, team prefers to pursue shading Netty instead of changing the location of the Netty dependencies.

PR #525, opened 2021-02-04
Exclude old version of velocity dependency.

karthikbalasub:


build.gradle
```diff
@@ -197,6 +197,7 @@ subprojects {
     all*.exclude group: 'javax.jms', module: 'jms'
     all*.exclude group: 'com.sun.jdmk', module: 'jmxtools'
     all*.exclude group: 'com.sun.jmx', module: 'jmxri'
+    all*.exclude group: 'org.apache.velocity', module: 'velocity'
```
junchuanwang, line 200: Can you add comments for our purpose as well?

By the way, the behavior implied by this line is that it will cancel out the pined version brought by this same gradle file to make it work with subprojects like `data-avro`? but  for `restli-tools` since that module explicitly asked for `compile externalDependency.velocity` so won't be affected by this line, is this understanding correct?

junchuanwang (approved):
approve with comments.

build.gradle
karthikbalasub: Added comment.

All sub-modules that depend on velocity will work as we now use "velocity-core-engine" module dependency (for velocity >=2.0). 
The exclude will only affect older versions brought in transitively.

junchuanwang: Thanks for the explanation. Yes you are right, completely using the new version everywhere should solve the issue

(for the record, when I replied, I initially thought the issue was `data-avro` won't work with new version and only works with order version, but it turns out the issue was not this).

nickibi (approved):

BrianPin (approved):

PR #526, opened 2021-02-09
Fix for HTTP 204 returned with content data (Exception stack) during resource throw exception in HTTP2 

BrianPin:
During HTTP2 migration, users find out the behavior of throwing an HTTP 204 with an exception is different. They found when using HTTP2, the response of HTTP 204 returned to the client is actually non-empty. Thus the client side of code is causing issue.

So this PR is to prevent content data from returning to client when the HTTP status code is 204, and we start to do this in exception path only (build error path), because this code has been there for long time, so for existing happy path, we will slow it down do this as long term fix.


restli-server/src/main/java/com/linkedin/restli/internal/server/response/ResponseUtils.java
```diff
@@ -242,7 +242,7 @@ public static RestException buildRestException(RestLiResponseException restLiRes
         .setStatus(restLiResponse.getStatus() == null ? HttpStatus.S_500_INTERNAL_SERVER_ERROR.getCode()
             : restLiResponse.getStatus().getCode());
 
-    if (restLiResponse.hasData())
+    if (restLiResponse.hasData() && restLiResponse.getStatus() != HttpStatus.S_204_NO_CONTENT)
```
BrianPin, line 245: This place meant for the major point of this fix

karthikbalasub: Thanks for the quick fix. One more thing we need to do is to log error if an exception is thrown with success status code. 

I thought doing this in error response builder would help with the logging:

```
diff --git a/restli-server/src/main/java/com/linkedin/restli/internal/server/response/ErrorResponseBuilder.java b/restli-server/src/main/java/com/linkedin/restli/internal/server/response/ErrorResponseBuilder.java
index 71700b6c3..1268439d3 100644
--- a/restli-server/src/main/java/com/linkedin/restli/internal/server/response/ErrorResponseBuilder.java
+++ b/restli-server/src/main/java/com/linkedin/restli/internal/server/response/ErrorResponseBuilder.java
@@ -16,10 +16,9 @@
 
 package com.linkedin.restli.internal.server.response;
 
-import com.linkedin.data.schema.RecordDataSchema;
-import com.linkedin.data.template.RecordTemplate;
 import com.linkedin.restli.common.ErrorDetails;
 import com.linkedin.restli.common.ErrorResponse;
+import com.linkedin.restli.common.HttpStatus;
 import com.linkedin.restli.common.ProtocolVersion;
 import com.linkedin.restli.common.ResourceMethod;
 import com.linkedin.restli.common.RestConstants;
@@ -29,12 +28,13 @@ import com.linkedin.restli.internal.server.RoutingResult;
 import com.linkedin.restli.server.ErrorResponseFormat;
 import com.linkedin.restli.server.RestLiResponseData;
 import com.linkedin.restli.server.RestLiServiceException;
-
 import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.net.HttpCookie;
 import java.util.List;
 import java.util.Map;
+import org.slf4j.Logger;
 
 
 /**
@@ -43,6 +43,7 @@ import java.util.Map;
  */
 public final class ErrorResponseBuilder
 {
+  private static final Logger LOGGER = LoggerFactory.getLogger(ErrorResponseBuilder.class.getName());
   public static final String DEFAULT_INTERNAL_ERROR_MESSAGE = "INTERNAL SERVER ERROR";
   private final ErrorResponseFormat _errorResponseFormat;
 
@@ -72,6 +73,18 @@ public final class ErrorResponseBuilder
 
   public ErrorResponse buildErrorResponse(RestLiServiceException result)
   {
+    if (result.getStatus() != null && result.getStatus().getCode() < 400)
+    {
+      // Invalid to send an error response with success status codes. This should be converted to 500 errors.
+      // Logging an error message now to detect and fix current use cases before we start converting to 500.
+      LOGGER.error("Incorrect use of success status code with error response", result);
+    }
+
+    if (result.getStatus() == HttpStatus.S_204_NO_CONTENT)
+    {
+      // HTTP Spec requires the response body to be empty for HTTP status 204.
+      return null;
+    }
     return buildErrorResponse(result, result.hasOverridingErrorResponseFormat() ? result.getOverridingFormat() : _errorResponseFormat);
   }
```

karthikbalasub: We can still keep your change, it can guard against success responses with 204 status code.

BrianPin: I merged yours within my PR

restli-int-test/src/test/java/com/linkedin/restli/examples/TestGreetingUnstructuredDataResources.java
```diff
@@ -114,4 +112,32 @@ public void testGetInternalServiceException(String resourceURL)
...
+    return new Object[][] {
+        { "/greetingAssociationUnstructuredData/src=exception_204&dest=bar" }
+    };
+  }
+
+  @Test(dataProvider = "exceptionNoContentURLs")
+  public void testGetInternalServiceExceptionWithNoContent(String resourceURL)
+      throws Throwable
+  {
+    sendGet(resourceURL, (conn) -> {
```
nickibi, line 128: nit: format: { should be in the following line
same as following.

restli-int-test-server/src/main/java/com/linkedin/restli/examples/greetings/server/GreetingsResourceImpl.java
```diff
@@ -200,6 +200,10 @@ public CreateResponse create(Greeting entity, @QueryParam("isNullId") @Optional(
   @RestMethod.Get
   public Greeting get(Long key)
   {
+    if (key == 204L)
+    {
+      throw new RestLiServiceException(HttpStatus.S_204_NO_CONTENT);
```
nickibi, line 205: Can you add description for this PR? like explaining what real problem this PR is fixing?
Customers throw RestLiServiceException for httpStatus 204? We want to catch such exception and return null for response content?

restli-int-test/src/test/java/com/linkedin/restli/examples/TestGreetingUnstructuredDataResources.java
```diff
@@ -114,4 +112,32 @@ public void testGetInternalServiceException(String resourceURL)
...
+      assertEquals(serverStatusCode, 204);
+      BufferedReader in = new BufferedReader(new InputStreamReader(conn.getInputStream()));
+      String inputLine;
+      StringBuffer response = new StringBuffer();
+      while ((inputLine = in.readLine()) != null) {
+        response.append(inputLine);
+      }
+      in.close();
+      assertEquals(response.length(), 0);
+      assertTrue(conn.getHeaderField("X-LinkedIn-Error-Response").equalsIgnoreCase("true"));
```
nickibi, line 140: I am kind of confused for this test case. 
204 is not an error code, but in the test resource you wrapped it in a RestLiServiceException. In the assertion, the X-LinkedIn-Error-Response = true, is this the right expectation?

restli-int-test/src/test/java/com/linkedin/restli/examples/TestGreetingUnstructuredDataResources.java
```diff
@@ -16,17 +16,15 @@
...
 import org.testng.annotations.Test;
 
-import static com.linkedin.restli.examples.greetings.server.GreetingUnstructuredDataUtils.UNSTRUCTURED_DATA_BYTES;
-import static com.linkedin.restli.common.RestConstants.HEADER_CONTENT_DISPOSITION;
-import static com.linkedin.restli.examples.greetings.server.GreetingUnstructuredDataUtils.CONTENT_DISPOSITION_VALUE;
-import static com.linkedin.restli.examples.greetings.server.GreetingUnstructuredDataUtils.MIME_TYPE;
-import static org.testng.Assert.assertEquals;
+import static com.linkedin.restli.common.RestConstants.*;
+import static com.linkedin.restli.examples.greetings.server.GreetingUnstructuredDataUtils.*;
+import static org.testng.Assert.*;
```
nickibi, line 27: nit: We should import particular classes rather than using *.

restli-int-test/src/test/java/com/linkedin/restli/examples/TestHttp2With204AndException.java
```diff
@@ -0,0 +1,55 @@
...
+import com.linkedin.restli.test.util.RootBuilderWrapper;
+import java.util.HashMap;
+import java.util.Map;
+import org.testng.Assert;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+
+public class TestHttp2With204AndException extends RestLiIntegrationTest{
```
zizhong, line 19: The issue happens with Jetty + HTTP/2.
However, I'm not sure if it happens with the Test Server + HTTP/2. If not, writing tests separately for HTTP1 and HTTP2 doesn't make much sense. The only thing we need to test IMHO is the response that we are trying to send to the client, with an empty body for this case. Nothing more.

restli-int-test-server/src/main/java/com/linkedin/restli/examples/greetings/server/GreetingsResourceImpl.java
BrianPin: I think I mentioned it in the title. I will provide the empty description a description

BrianPin: Description added

restli-int-test/src/test/java/com/linkedin/restli/examples/TestGreetingUnstructuredDataResources.java
karthikbalasub: It is not, but it is the current behavior. This is why we added the error log, which can help migrate the existing usecases before we start failing for such cases.

restli-int-test/src/test/java/com/linkedin/restli/examples/TestHttp11With204AndException.java
```diff
@@ -0,0 +1,55 @@
...
+          .build();
+
+      Response<Greeting> response = getClient().sendRequest(request).getResponse();
+      Greeting greetingResponse = response.getEntity();
+      Assert.assertEquals(response.getHeaders().get("Content-Length"), "0");
+    }
+    catch (Exception e)
+    {
+      Assert.fail();
+    }
```
karthikbalasub, line 53: No need to do this. Just let the exception go through, the test framework will fail the test. It would also include the stack trace, helping to debug the failure.

restli-int-test/src/test/java/com/linkedin/restli/examples/TestHttp2With204AndException.java
karthikbalasub: The test server actually uses http jetty server provided by r2. 
It might not reproduce the failure if the client behavior is different in the test suite. It does add some value as it guards against change in behavior within the protocol (say jetty HTTP 2.0 decides to add some response body if error response header is true). I'll let you decide if you want to keep it.

restli-int-test/src/test/java/com/linkedin/restli/examples/TestHttp2With204AndException.java
```diff
@@ -0,0 +1,55 @@
...
+          .build();
+
+      Response<Greeting> response = getClient().sendRequest(request).getResponse();
+      Assert.assertEquals(response.getHeaders().get("Content-Length"), "0");
+
+    }
+    catch (Exception e)
+    {
+      Assert.fail();
+    }
```
karthikbalasub, line 53: same here, no need to do try/catch.

karthikbalasub (commented):
Please address the styling issues mentioned by Yingjie.

restli-int-test/src/test/java/com/linkedin/restli/examples/TestGreetingUnstructuredDataResources.java
BrianPin: noticed

restli-int-test/src/test/java/com/linkedin/restli/examples/TestGreetingUnstructuredDataResources.java
BrianPin: I think the lambda in pegasus is doing this way

restli-int-test/src/test/java/com/linkedin/restli/examples/TestGreetingUnstructuredDataResources.java
BrianPin: There might be some question about why this test exists. Along with the other 2 tests this seems redundant.
But actually, the http client in this test is pure, not decorated with HttpBridges and callbacks.
These bridges and callbacks might change an returned error to an RemoteInvokeException, And I am afraid I am still not fully grab the whole picture.

restli-int-test/src/test/java/com/linkedin/restli/examples/TestHttp2With204AndException.java
```diff
@@ -0,0 +1,51 @@
...
+    super.init(false, new RestLiConfig(), transportProperties);
+  }
+
+  @AfterClass
+  public void shutDown() throws Exception
+  {
+    super.shutdown();
+  }
+
+  @Test(expectedExceptions = RemoteInvocationException.class)
```
karthikbalasub, line 39: You should remove expectedExceptions. What this means is you expect the test to throw the exception and it is a failure if no exception was thrown.

restli-int-test/src/test/java/com/linkedin/restli/examples/TestHttp11With204AndException.java
```diff
@@ -0,0 +1,52 @@
...
+    super.init(false, new RestLiConfig(), transportProperties);
+  }
+
+  @AfterClass
+  public void shutDown() throws Exception
+  {
+    super.shutdown();
+  }
+
+  @Test(expectedExceptions = RemoteInvocationException.class)
```
karthikbalasub, line 39: You should remove expectedExceptions. What this means is you expect the test to throw the exception and it is a failure if no exception was thrown.

restli-int-test/src/test/java/com/linkedin/restli/examples/TestGreetingUnstructuredDataResources.java
karthikbalasub: One value this adds is that it checks we still send the error response header when server threw RestLiServiceException with 204 status, as that is the current behavior.

karthikbalasub (commented):
Also update the CHANGELOG

restli-int-test/src/test/java/com/linkedin/restli/examples/TestGreetingUnstructuredDataResources.java
```diff
@@ -114,4 +113,32 @@ public void testGetInternalServiceException(String resourceURL)
...
+      throws Throwable
+  {
+    sendGet(resourceURL, (conn) -> {
+      conn.setDoInput(true);
+      int serverStatusCode = conn.getResponseCode();
+      assertEquals(serverStatusCode, 204);
+      BufferedReader in = new BufferedReader(new InputStreamReader(conn.getInputStream()));
+      String inputLine;
+      StringBuffer response = new StringBuffer();
+      while ((inputLine = in.readLine()) != null) {
```
nickibi, line 136: format

restli-int-test/src/test/java/com/linkedin/restli/examples/TestGreetingUnstructuredDataResources.java
```diff
@@ -16,17 +16,16 @@
...
+import java.io.InputStreamReader;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 
-import static com.linkedin.restli.examples.greetings.server.GreetingUnstructuredDataUtils.UNSTRUCTURED_DATA_BYTES;
-import static com.linkedin.restli.common.RestConstants.HEADER_CONTENT_DISPOSITION;
-import static com.linkedin.restli.examples.greetings.server.GreetingUnstructuredDataUtils.CONTENT_DISPOSITION_VALUE;
-import static com.linkedin.restli.examples.greetings.server.GreetingUnstructuredDataUtils.MIME_TYPE;
+import static com.linkedin.restli.common.RestConstants.*;
+import static com.linkedin.restli.examples.greetings.server.GreetingUnstructuredDataUtils.*;
```
nickibi, line 26: still using * here.

restli-int-test/src/test/java/com/linkedin/restli/examples/TestHttp2With204AndException.java
nickibi: +1

restli-int-test/src/test/java/com/linkedin/restli/examples/TestHttp11With204AndException.java
nickibi: +1
Based on current expected behavior, this test will fail.

restli-int-test/src/test/java/com/linkedin/restli/examples/TestGreetingUnstructuredDataResources.java
nickibi: Thanks for explaining it. @karthikbalasub

restli-int-test/src/test/java/com/linkedin/restli/examples/TestHttp11With204AndException.java
BrianPin: Removed!

restli-int-test/src/test/java/com/linkedin/restli/examples/TestHttp2With204AndException.java
BrianPin: Removed

restli-int-test/src/test/java/com/linkedin/restli/examples/TestHttp2With204AndException.java
BrianPin: Removed

nickibi (approved):
LGTM

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
+- Deal with status code 204, when we see 204 in error path, we will not return data (from data layer only)
```
karthikbalasub, line 16: It might be good to release this soon as it blocks some users

karthikbalasub (approved):

CHANGELOG.md
BrianPin: Updated

PR #530, opened 2021-02-11
Implemention for ParSeqBasedCompletionStage

junchuanwang:


restli-client/src/main/java/com/linkedin/restli/client/ParSeqBasedCompletionStage.java
```diff
@@ -0,0 +1,621 @@
...
+        return Success.of(prevTaskResult.get());
+      }
+    }));
+  }
+
+  @Override
+  public ParSeqBasedCompletionStage<T> whenCompleteAsync(BiConsumer<? super T, ? super Throwable> action,
+      Executor executor)
+  {
+    return nextStageByComposingTask(_task.transformWith("whenCompleteAsync", prevTaskResult -> {
```
karthikbalasub, line 574: Calling transform on the `_task` might fail when `_task` is already completed, right? 

Wouldn't it be cleaner to just add a promise listener on _task?

karthikbalasub (commented):
I think chaining on the running task is not a good idea. Consider the idea of using promise listener on the task. 

Not sure what failures you saw in the tests, but it might be related to this.

restli-client/src/main/java/com/linkedin/restli/client/ParSeqBasedCompletionStage.java
junchuanwang: actually chaining on a task is equivalent to add a promise listner when the task not resolved. If task resolved, then chained task try to get the value directly.

That also implies that with the promise listner approach, when adding the listern, if the task already finished, added listner won't get notified.

Besides, promise can add listners but task doesn't have this interface. It would be hard to add listner

karthikbalasub: Task extends Promise so you have access to add listener. And promise listener will be notified even if the promise is already fulfilled (check the java doc).

I don't think task chaining is equivalent to add listener. Task chaining will try to run the parent task and can fail if it is already completed (what I figured from looking at code).

junchuanwang: > Task chaining will try to run the parent task and can fail if it is already completed (what I figured from looking at code)

That should not be the case. (and it would be weird if that is the case)

Say `TaskA` already finished.
```
TaskB = SthNotStarted();
TaskC = TaskA.chainWith(TaskB);
TaskD = TaskA.andThen( val -> Logic() );
```
Do you mean that TaskC and TaskD will fail and TaskB will not run?  I don't udnerstand why. From what I know, TaskA's result (fail or success) will be propagated.

junchuanwang: btw, the task failure I have previously run through is due to the fact that Travis-CI test machine probably have only 1 core (probably less than 2 CPU ThreadContexts).

In my test, I create two tasks and will complete one with failure after some delay and complete another with success after a longer delay(I used Thread.sleep, which in Java blocks the thread). In my local box it always works but in Travis-CI sometimes it didn't if the longer task starts first (because the shorter one, which supposed to be finished first will run after that)

karthikbalasub: You are correct about chaining on already completed task. However, it might be more performant to use the promise listener? task chaining will schedule the predecessor task to run again. Even if it does nothing, it can still take CPU cycles to execute the runnable etc.
Also we need to see how the trace looks like in both cases. I think task chaining will generate a combined trace. Did you check?

restli-client/src/main/java/com/linkedin/restli/client/ParSeqBasedCompletionStage.java
```diff
@@ -0,0 +1,621 @@
...
+  public ParSeqBasedCompletionStage<T> whenCompleteAsync(BiConsumer<? super T, ? super Throwable> action)
+  {
+    return whenCompleteAsync(action, _asyncExecutor);
+  }
+
+  @Override
+  public CompletableFuture<T> toCompletableFuture()
+  {
+    final CompletableFuture<T> future = new CompletableFuture<T>();
+    _task.toCompletionStage().whenComplete((value, exception) -> {
```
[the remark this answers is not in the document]
junchuanwang: Good catch. I forgot to udpate this.

restli-client/src/main/java/com/linkedin/restli/client/ParSeqBasedCompletionStage.java
```diff
@@ -0,0 +1,621 @@
...
+   * TODO:
+   * (1) Without engine, will use {@link CompletableFuture} default implementation
+   * (2) or use default engine
+   */
+  private ParSeqBasedCompletionStage()
+  {
+  }
+
+  ParSeqBasedCompletionStage(Engine engine, Executor executor)
+  {
```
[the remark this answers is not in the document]
junchuanwang: Agree, moved creation related stuff to the factory.

restli-client/src/main/java/com/linkedin/restli/client/ParSeqBasedCompletionStage.java
junchuanwang: (1) I don't see how addListiner is better, except it is harder to implement IMO. I really want to say that before I choose to chain on the task, I carefully considered the add listner option. You will be absoluitely right about it being less performant.  I think it fine and there is good reason to use the task directly: Promise listerns are lower level APIs. Task compositions use higher level APIs(underneath the hood, it uses "After" object, or InternalUtils.after to register the listners). There are already logic embeded inside builtin tasks such as FusionTask or TupleTasks. If we rewrite everything using PromiseListner, it could produce duplicated code, more work and IMO error-prone. It is like how you will choose high level language over hand-crafted asm.

(2) For the trace, I did some tests, looked at the code, and there is some differences if the task already done (but not major difference), so I don't have concerns. Let me explain:
- There are two kinds of relationship: i.e. successor and parental. For task already done, the "parent_of" relationship will become "potential_parent_of", in the trace that's all. The "successor" relationship would remain the same.
- But the main thing is the relationship would not be messed up, because the relationship is defined whenever the context starts a new task(so it is not related to whether the task is done or not). Also the trace is propagated between sucessing tasks only after the previous task is done (this is similar to how value is propagated), so whether the task is done early has no impact on the propagated trace. For details you only need to check [BaseTask](https://github.com/linkedin/parseq/blob/master/subprojects/parseq/src/main/java/com/linkedin/parseq/BaseTask.java) and [FusionTask](https://github.com/linkedin/parseq/blob/master/subprojects/parseq/src/main/java/com/linkedin/parseq/FusionTask.java)

karthikbalasub: Sounds good, thanks for looking into the details. Resolving this conversation.

restli-client/src/test/java/com/linkedin/restli/client/TestParSeqBasedCompletionStage.java
```diff
@@ -0,0 +1,1363 @@
...
+
+    CompletionStage<String> completionStage = createTestFailedStage(EXCEPTION);
+    finish(completionStage.whenComplete((v, t) -> {}).exceptionally(exceptionallyFunction));
+    verify(exceptionallyFunction, times(1)).apply(verifyCompletionException());
+  }
+
+  @Test
+  public void testWhenCompleteAsync() throws Exception
+  {
+    CountDownLatch waitLatch = new CountDownLatch(1);
```
karthikbalasub, line 1296: You don't need to use a latch. Just convert to completable future and use the get with timeout.

karthikbalasub (approved):

restli-client/src/main/java/com/linkedin/restli/client/ParSeqBasedCompletionStage.java
```diff
@@ -0,0 +1,578 @@
...
+ *
+ *
+ * @param <T> The Type of the value this CompletionStage is holding
+ */
+public class ParSeqBasedCompletionStage<T> implements CompletionStage<T>
+{
+
+  protected Engine _engine = null;
+  protected Task<T> _task = null; // The underlying ParSeq task to acquire the value this completionStage needs
+  protected Executor _asyncExecutor = null;
```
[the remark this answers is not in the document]
junchuanwang: for now, yes, but don't you think there might be class extending this? Like something extended this class but also implements another interface, maybe does the Trace thing we talked about?

restli-client/src/main/java/com/linkedin/restli/client/ParSeqBasedCompletionStage.java
```diff
@@ -0,0 +1,578 @@
...
+ *   Executor _executor;
+ *   Task{@code <String>} task;
+ *   ParSeqBasedCompletionStage{@code <String>} stage =
+ *    new ParSeqBasedCompletionStageFactory(_engine, _executor).buildStageFromTask(task);
+ * </pre></blockquote>
+ *
+ *
+ * @param <T> The Type of the value this CompletionStage is holding
+ */
+public class ParSeqBasedCompletionStage<T> implements CompletionStage<T>
```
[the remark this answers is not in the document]
junchuanwang: we can do it for now (close it for modification/inheritance) but my original thought was this might be inherited (see my other comment)

restli-client/src/main/java/com/linkedin/restli/client/ParSeqBasedCompletionStage.java
```diff
@@ -0,0 +1,578 @@
...
+
+  /**
+   * Not allowing to use the class without an engine and executor
+   * TODO:
+   * (1) Without engine, will use {@link CompletableFuture} default implementation
+   * (2) or use default engine
+   */
+  private ParSeqBasedCompletionStage()
+  {
+  }
```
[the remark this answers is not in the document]
junchuanwang: Then how can we make sure there is no default constructor exposed?

restli-client/src/test/java/com/linkedin/restli/client/TestParSeqBasedCompletionStage.java
junchuanwang: Yes I used this latch extensively in this file and I know some of them are not needed, or to be replaced by your suggestion.

But others are necessary, they are not just for waiting the result: I want to make sure the latch has been counted down, which could prove that the logic actually executed. (You know sometimes the completion stage can completes with failure even without the whole logic executed. ) 

I can do similar thing with checking a counter? But I think latch makes more sense in these tests

restli-client/src/main/java/com/linkedin/restli/client/ParSeqBasedCompletionStage.java
```diff
@@ -0,0 +1,578 @@
...
+  static <U> Task<U> wrapException(Task<U> task)
+  {
+    return task.transform(prevTaskResult -> {
+      if (prevTaskResult.isFailed()) {
+        Throwable t = prevTaskResult.getError();
+        if (t instanceof CompletionException) {
+          return Failure.of(t);
+        } else {
+          CompletionException ex = new CompletionException(prevTaskResult.getError());
+          return Failure.of(ex);
```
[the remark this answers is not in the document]
junchuanwang: Fixed, thanks

restli-client/src/main/java/com/linkedin/restli/client/ParSeqBasedCompletionStage.java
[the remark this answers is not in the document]
karthikbalasub: Let us start with stricter visibility and relax if needed. 
If you think inherited classes can change these variables, then you cannot make assumptions about their state in your code.
You can allow overriding by exposing methods that also ensure these variables are valid when modified by inherited classes

restli-client/src/main/java/com/linkedin/restli/client/ParSeqBasedCompletionStage.java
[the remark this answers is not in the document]
junchuanwang: thanks.I still have cpp rules in my mind (it has defaulted default constructor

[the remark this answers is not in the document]
junchuanwang: you are right. I was wrong. Cpp/java has the same rule regarding default constructor.

PR #531, opened 2021-02-12
Shortcircuit already serialized projection params

karthikrg:


CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.14.5] - 2021-02-11
+
```
mchen07, line 18: Add ChangeLog for this change.

mchen07 (commented):
Looks good to me, except one minor comment.

restli-client/src/test/java/com/linkedin/restli/internal/client/TestQueryParamsUtil.java
```diff
@@ -133,4 +133,21 @@ public void testCustomProjectionDataMapSerializerReturningNull()
     Assert.assertNull(dataMap.getDataMap(RestConstants.PAGING_FIELDS_PARAM));
     Assert.assertNull(dataMap.getDataMap(RestConstants.METADATA_FIELDS_PARAM));
   }
+
+  @Test
+  public void testPreSerializedProjectionParams()
+  {
+    Map<String, Object> queryParams = new HashMap<>();
+    queryParams.put(RestConstants.FIELDS_PARAM, "fields");
```
BrianPin, line 141: I wonder if we could add more cases into testing, 
for example (excerpt from restli doc)
https://linkedin.github.io/rest.li/Projections
```
  map_field:($*:(field1),key1:(field2),key2:(field3))
  array_field:($*:(field1,field2),$start:10,$count:15)
```
and projections like
```
  http://host:port/context/resource/id?fields=person:(firstname,lastname)
```

restli-client/src/main/java/com/linkedin/restli/client/Request.java
```diff
@@ -299,13 +302,29 @@ public UriTemplate getUriTemplate()
...
-    Set<PathSpec> fieldsSet = (Set<PathSpec>) _queryParams.get(RestConstants.FIELDS_PARAM);
-    if (fieldsSet == null)
-    {
+    Object fields = _queryParams.get(RestConstants.FIELDS_PARAM);
+    if (fields == null) {
       return Collections.emptySet();
     }
-    return fieldsSet;
+
+    if (fields instanceof Set)
```
mchen07, line 310: Maybe we can provide an API in `AbstractRequestBuilder` to have such an API for users to directly set fields parameter instead of using `setParam`. Something like

```
public void addFields(String fieldPaths)
public void addFields(PathSpec... fieldPathSpecs)
```

Ideally, we can even expose `fields(String... fieldPaths)` in generated type safe request builders to get this optimization better utilized, but that can be done later. WDYT?

restli-client/src/main/java/com/linkedin/restli/client/Request.java
```diff
@@ -299,13 +302,29 @@ public UriTemplate getUriTemplate()
...
+    {
+      @SuppressWarnings("unchecked")
+      Set<PathSpec> fieldsSet = (Set<PathSpec>) _queryParams.get(RestConstants.FIELDS_PARAM);
+      return fieldsSet;
```
karthikbalasub, line 314: just
return (Set<PathSpec>) fields;

restli-client/src/main/java/com/linkedin/restli/client/Request.java
```diff
@@ -299,13 +302,29 @@ public UriTemplate getUriTemplate()
...
+    }
+    else if (fields instanceof String)
+    {
+      try
+      {
+        MaskTree tree = URIMaskUtil.decodeMaskUriFormat((String) fields);
+        return tree.getOperations().keySet();
+      } catch (IllegalMaskException e) {
+        throw new IllegalArgumentException(e);
```
karthikbalasub, line 323: Please include a message:
"Field param was a string and it did not represent a serialized mask tree"

restli-client/src/test/java/com/linkedin/restli/internal/client/TestQueryParamsUtil.java
karthikbalasub: We should already have tests for MaskTree decoder covering these. Need not test them again here.

karthikbalasub (approved):
Approval with minor comments.

restli-client/src/main/java/com/linkedin/restli/client/Request.java
mchen07: Good point.

restli-client/src/main/java/com/linkedin/restli/client/Request.java
mchen07: As we chatted offline, for typesafe request builders, maybe we don't want to expose application developers such advanced trick to avoid invalid pathSpec string passed in. For infra, this may be a nice feature to adoption. So the suggestion to add API in `AbstractRequestBuilder` may have value there. But this can be done in followup RB with team discussion.

restli-client/src/test/java/com/linkedin/restli/internal/client/TestQueryParamsUtil.java
mchen07: +1 on no testing here.

mchen07 (approved):

restli-client/src/main/java/com/linkedin/restli/client/Request.java
karthikbalasub: As part of Universal client, I have plans to add APIs that accept MaskTree and provide helpers so clients can build it. That would be a good trade-off between typesafety and performance.

restli-client/src/test/java/com/linkedin/restli/internal/client/TestQueryParamsUtil.java
karthikrg: Yep already tested, resolving.

PR #532, opened 2021-02-12
Fix for idl/snapshot task failing with IOException…

aman1309:
Fix for idl/snapshot task failing with IOException when compat summary is not generated/ or is empty.

### DIFF
--- a/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
+++ b/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
@@ -1183,6 +1183,10 @@ public class PegasusPlugin implements Plugin<Project>
     CompatibilityLogChecker logChecker = new CompatibilityLogChecker();
     try
     {
+      if (!compatibilityLogFile.exists())
+      {
+          return true;
+      }
       logChecker.write(Files.readAllBytes(compatibilityLogFile.toPath()));
     }
     catch (IOException e)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.14.6] - 2021-02-12
+- Fix for idl/snapshot task failing with IOException when compat summary is not generated/ or is empty.
+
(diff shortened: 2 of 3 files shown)


gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1183,6 +1183,10 @@ private static boolean isResultEquivalent(File compatibilityLogFile, boolean res
     CompatibilityLogChecker logChecker = new CompatibilityLogChecker();
     try
     {
+      if (!compatibilityLogFile.exists())
```
nickibi, line 1186: isResultEquivalent will be called, only when checkSnapshotTask or checkRestModelTask or checkIdlTask run successfully. If that's the case, how compatibilityLogFile will not exist?

aman1309: if these tasks are skipped then also it's successful

nickibi: If those tasks are skipped, that means it has already executed in the past, no need to rerun, right? I still do not understand why compatibilityLogFile not exist. The change was made in 29.7.3(https://github.com/linkedin/rest.li/pull/431). If compatibilityLogFile not existed is as an expected behavior, we would have received a lot of complains in the past 4 months for this IO exception.

karthikbalasub: I don't think this is the correct fix. The file doesn't exist because the check tasks didn't run and that could happen if the API module was missing.

If this happens in PCL, then the impl project is either missing snapshots or snapshots were not updated correctly.

aman1309: In rare scenario what if there were no resource and first time generated file was empty and if file contents are empty, its not getting created. Also, I did see this changes in #431 and prior to this change isEquivalent in initialized with true. let discuss offline on this.

ChrisCarini:

PR #534, opened 2021-02-17
Compression filter changes

karthikrg:
Always enable client compression filter so that responses can be decompressed. If the request already has an accept encoding header set do not overwrite it

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-version=29.14.5
+version=29.15.0
 group=com.linkedin.pegasus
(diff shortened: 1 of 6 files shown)


gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=29.14.5
+version=29.14.6
```
mchen07, line 1: Since there is behavior change, let's bump major version on this.

karthikrg: Decided to bump minor.

mchen07 (approved):

PR #535, opened 2021-02-18
Fix a bug that rest.idl.skipCheck=true still runs checkSnapshot

JospehChiu:
* When setting rest.idl.skipCheck=true in Gradle, checkSnapshot is still triggered which might be a bug.
* It'd be helpful if gradle properties (IDL_NO_PUBLISH, SKIP_IDL_CHECK, SNAPSHOT_NO_PUBLISH) is public, then another plugin can access to these variables.

### DIFF
--- a/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
+++ b/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
@@ -547,12 +547,12 @@ public class PegasusPlugin implements Plugin<Project>
   private static final String PEGASUS_COMPATIBILITY_MODE = "pegasusPlugin.pegasusSchemaCompatibilityCheckMode";
 
   private static final Pattern TEST_DIR_REGEX = Pattern.compile("^(integ)?[Tt]est");
-  private static final String SNAPSHOT_NO_PUBLISH = "rest.model.noPublish";
-  private static final String IDL_NO_PUBLISH = "rest.idl.noPublish";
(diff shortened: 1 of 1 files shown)


gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1318,7 +1318,7 @@ protected void configureRestModelGeneration(Project project, SourceSet sourceSet
             task.setCodegenClasspath(project.getConfigurations().getByName(PEGASUS_PLUGIN_CONFIGURATION));
             task.setSnapshotCompatLevel(PropertyUtil.findCompatLevel(project, FileCompatibilityType.SNAPSHOT));
 
-            task.onlyIf(t -> isPropertyTrue(project, SKIP_IDL_CHECK));
```
karthikbalasub, line 1321: This is on purpose. 

The plugin has two ways to do compatibility checking:
1. Use current IDL + PDL against prev.snapshot. This is done by two tasks: checkIdl + checkRestModel
2. Use only snapshot files. This is done by checkSnapshot

Normally the plugin uses only option 1. If the skip idl flag is set, it will disable the tasks in option 1 and run the checkSnapshot task instead.

See Lines 1302, 1339, 1358-1366, 1381-1390 for relevant code.

JospehChiu: I see, thanks for the details.

PR #536, opened 2021-02-18
Cleanup compression code to reduce duplication and minimize memcopies

karthikrg:


karthikrg:
@nizarm sorry what is incompatible here?

nizarm:
> @nizarm sorry what is incompatible here?

As discussed offline, there are no issues here. Sorry for the confusion. You can drop this comment

r2-filter-compression/src/main/java/com/linkedin/r2/filter/compression/AbstractCompressor.java
```diff
@@ -0,0 +1,142 @@
...
+    finally
+    {
+      IOUtils.closeQuietly(deflaterStream);
+    }
+
+    return out.toByteArray();
+  }
+
+  @Override
+  public ByteString deflate(ByteString data) throws CompressionException
```
mchen07, line 103: minor: better put some javadoc comment to explain the advantage of this version than the other one.

Otherwise, this looks good to me.

mchen07 (approved):

PR #539, opened 2021-02-24
Prevent RetriableRequestException from cascading to the indirect caller

rickzx:
Suppose A sends request to B, and B sends request to C. When C is overloaded, B will retry its requests on C. If all of the retry attempts failed, the error will be propagated back to A. But we would like to prevent the RetriableRequestException from C from cascading to the indirect caller A, and therefore, we need to disable retry at B in this case.

### DIFF
--- a/r2-core/src/main/java/com/linkedin/r2/filter/transport/ServerRetryFilter.java
+++ b/r2-core/src/main/java/com/linkedin/r2/filter/transport/ServerRetryFilter.java
@@ -115,16 +115,18 @@ public class ServerRetryFilter implements RestFilter, StreamFilter
     {
       if (cause instanceof RetriableRequestException)
       {
-        String message = cause.getMessage();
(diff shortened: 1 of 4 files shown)


r2-core/src/main/java/com/linkedin/r2/filter/transport/ServerRetryFilter.java
```diff
@@ -115,16 +115,18 @@ public void onStreamError(Throwable ex,
     {
       if (cause instanceof RetriableRequestException)
```
rachelhanhan, line 116: This made me feel that we should use 2 different exceptions on the server side and client side. If they are 2 different exceptions, we won't have issue with a client passing the exception to its caller again. We should always let server construct a new retry request from the overload shedder, there shouldn't be much chance that a server can get exactly the same exception converted from its downstream. But since we've been using this, I agree your solution also works.

rachelhanhan (approved):

PR #540, opened 2021-02-26
[rest.li] Fix content type header not set in case of StreamException from rest.li server

maulinpatel007:
This PR is to fix restli-client side decoding of error response data map failure, when restli server returns error data for stream requests.

### DIFF
--- a/restli-server/src/main/java/com/linkedin/restli/internal/server/response/ResponseUtils.java
+++ b/restli-server/src/main/java/com/linkedin/restli/internal/server/response/ResponseUtils.java
@@ -252,16 +252,17 @@ public class ResponseUtils
     return new RestException(restResponse, cause==null ? null : cause.toString(), cause, writableStackTrace);
   }
 
-  public static StreamException buildStreamException(RestLiResponseException restLiResponseException, StreamDataCodec codec)
+  public static StreamException buildStreamException(RestLiResponseException restLiResponseException, ContentType contentType)
   {
     RestLiResponse restLiResponse = restLiResponseException.getRestLiResponse();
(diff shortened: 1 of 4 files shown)


restli-server/src/main/java/com/linkedin/restli/internal/server/response/ResponseUtils.java
```diff
@@ -252,16 +252,17 @@ public static RestException buildRestException(RestLiResponseException restLiRes
     return new RestException(restResponse, cause==null ? null : cause.toString(), cause, writableStackTrace);
   }
 
-  public static StreamException buildStreamException(RestLiResponseException restLiResponseException, StreamDataCodec codec)
+  public static StreamException buildStreamException(RestLiResponseException restLiResponseException, ContentType contentType)
```
karthikrg, line 255: This is a potentially backward incompatible change since it is a public method. I suggest creating an overloaded method and deprecating this method.

aman1309, line 255: since this is a public interface we should not update the interface.
can add the header in RestLiResponseException.restliResponse.headers before building stream exception response

aman1309 (changes requested):
can you also add any unit test for this

restli-server/src/main/java/com/linkedin/restli/internal/server/response/ResponseUtils.java
maulinpatel007: aah.. thanks , let me update the RB

karthikrg (approved):

aman1309 (approved):

PR #546, opened 2021-03-02
Include mime type params when setting the response content-type

klee97:
Rest.li is setting the response `content-type` based on the best matching `accept` mime type. However it only sets the base mime type while ignoring any mime type parameters. If a mime type parameter is malformed, it is silently ignored.

Changes:
* Include mime type params when setting the response content type
* Throw an InvalidMimeTypeException when there are invalid params in the accept header.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-version=29.15.6
+version=29.15.7
 group=com.linkedin.pegasus
 org.gradle.configureondemand=true
 org.gradle.parallel=true
--- a/restli-server/src/main/java/com/linkedin/restli/internal/server/util/MIMEParse.java
+++ b/restli-server/src/main/java/com/linkedin/restli/internal/server/util/MIMEParse.java
@@ -52,6 +52,8 @@ import org.apache.commons.lang.math.NumberUtils;
 public final class MIMEParse
 {
 
+  private static final String QUALITY_PARAM = "q";
+
   /**
    * Parse results container
    */
@@ -67,11 +69,25 @@ public final class MIMEParse
     @Override
     public String toString()
     {
-      StringBuffer s = new StringBuffer("('" + type + "', '" + subType
-                                                + "', {");
-      for (String k : params.keySet())
-        s.append("'" + k + "':'" + params.get(k) + "',");
+      StringBuilder s = new StringBuilder("('").append(type).append("', '").append(subType).append("', {");
+      params.forEach((k, v) -> s.append("'").append(k).append("':'").append(v).append("',"));
       return s.append("})").toString();
+    }
+
+    /**
+     * Build the String for the content type header
+     */
+    String toContentType() {
+      StringBuilder s = new StringBuilder(type).append("/").append(subType);
+      params.forEach((k, v) ->
+      {
+        // Exclude accept type's "q" param from the content type
+        if (!k.equals(QUALITY_PARAM))
+        {
+          s.append("; ").append(k).append("=").append(v);
+        }
+      });
+      return s.toString();
     }
   }
 
@@ -94,7 +110,9 @@ public final class MIMEParse
       String p = parts[i];
       String[] subParts = StringUtils.split(p, '=');
       if (subParts.length == 2)
+      {
         results.params.put(subParts[0].trim(), subParts[1].trim());
+      }
     }
     String fullType = parts[0].trim();
 
@@ -126,10 +144,10 @@ public final class MIMEParse
   protected static ParseResults parseMediaRange(String range)
   {
     ParseResults results = parseMimeType(range);
-    String q = results.params.get("q");
+    String q = results.params.get(QUALITY_PARAM);
     float f = NumberUtils.toFloat(q, 1);
     if (StringUtils.isBlank(q) || f < 0 || f > 1)
-      results.params.put("q", "1");
+      results.params.put(QUALITY_PARAM, "1");
     return results;
   }
 
@@ -178,6 +196,7 @@ public final class MIMEParse
   {
     int bestFitness = -1;
     float bestFitQ = 0;
+    Map<String, String> bestFitParams = Collections.emptyMap();
     ParseResults target = parseMediaRange(mimeType);
 
     for (ParseResults range : parsedRanges)
@@ -191,7 +210,7 @@ public final class MIMEParse
         for (String k : target.params.keySet())
         {
           int paramMatches = 0;
-          if (!k.equals("q") && range.params.containsKey(k)
+          if (!k.equals(QUALITY_PARAM) && range.params.containsKey(k)
                   && target.params.get(k).equals(range.params.get(k)))
           {
             paramMatches++;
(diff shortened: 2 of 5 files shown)


restli-server/src/main/java/com/linkedin/restli/internal/server/util/MIMEParse.java
```diff
@@ -67,12 +67,25 @@
     @Override
     public String toString()
     {
-      StringBuffer s = new StringBuffer("('" + type + "', '" + subType
-                                                + "', {");
+      StringBuilder s = new StringBuilder("('").append(type).append("', '").append(subType).append("', {");
       for (String k : params.keySet())
```
karthikrg, line 71: Doing a keySet and then get(K) is so wasteful. Please use forEach on the map.

restli-server/src/main/java/com/linkedin/restli/internal/server/util/MIMEParse.java
```diff
@@ -67,12 +67,25 @@
...
       return s.append("})").toString();
     }
+
+    /**
+     * Build the String for the content type. "q" param is ignored
+     */
+    String toContentType()
+    {
+      StringBuilder s = new StringBuilder(type).append("/").append(subType);
+      for (String k : params.keySet()) {
```
karthikrg, line 82: Same comment as previous about not using keySet.

restli-server/src/main/java/com/linkedin/restli/internal/server/util/MIMEParse.java
```diff
@@ -67,12 +67,25 @@
...
+        if (!k.equals("q")) {
```
karthikrg, line 83: Please move q to a constant and add a comment explaining why it is ignored.

restli-server/src/main/java/com/linkedin/restli/internal/server/util/MIMEParse.java
```diff
@@ -67,12 +67,25 @@
...
       for (String k : params.keySet())
-        s.append("'" + k + "':'" + params.get(k) + "',");
+        s.append("'").append(k).append("':'").append(params.get(k)).append("',");
       return s.append("})").toString();
     }
+
+    /**
+     * Build the String for the content type. "q" param is ignored
+     */
+    String toContentType()
```
karthikrg, line 79: This has a lot of duplication with the toString() method. Why not parameterize this method adding a boolean param like excludeQuality and invoke accordingly?

restli-server/src/main/java/com/linkedin/restli/internal/server/util/MIMEParse.java
```diff
@@ -93,8 +106,12 @@ protected static ParseResults parseMimeType(String mimeType)
     {
       String p = parts[i];
       String[] subParts = StringUtils.split(p, '=');
-      if (subParts.length == 2)
+      if (subParts.length == 2) {
```
karthikrg, line 109: Why are we doing this manual parsing here instead of using the mimetype parser?

karthikrg (changes requested):

BrianPin:
Just add a general comment that in pegasus/restli the left bracket is on the next new line.

gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=29.15.4
+version=29.15.5
```
karthikbalasub, line 1: This was released today. So you'd have to update here and the changelog.

restli-server/src/main/java/com/linkedin/restli/internal/server/util/MIMEParse.java
karthikbalasub: This MimeParse utility is used to pick the best match based on "q". Not use why we used a custom parser, one difference is that special handling of "*" accept header (lines 118 to 121). This is not handled in jdk MimeType.

junchuanwang:
Is this a backward-compatible change? Cause previously worked header would cause exceptions?

restli-server/src/main/java/com/linkedin/restli/internal/server/util/MIMEParse.java
klee97: `toString` is a representation of the ParseResult, while `toContentType` is the actual mime type, so I think it's better to keep them separate. The string format is a little different, so I don't see much that can be generalized into a helper.

klee97:
> Just add a general comment that in pegasus/restli the left bracket is on the next new line.

Thanks @BrianPin, is there a style command I can run?

klee97:
> Is this a backward-compatible change? Cause previously worked header would cause exceptions?

@junchuanwang I suppose if someone is sending incorrect headers, it would start breaking. I can do a major version bump to be safe

gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=29.15.4
+version=30.15.5
```
karthikbalasub, line 1: This should be 30.0.0

Doing a major version would also mean waiting to merge some other backwards incompatible changes so they can all be bundled in on major version. 
It would add additional time to the release, let us know if that is okay with you.

restli-server/src/main/java/com/linkedin/restli/internal/server/util/MIMEParse.java
```diff
@@ -94,7 +108,14 @@ protected static ParseResults parseMimeType(String mimeType)
...
       String[] subParts = StringUtils.split(p, '=');
       if (subParts.length == 2)
+      {
         results.params.put(subParts[0].trim(), subParts[1].trim());
+      }
+      else
+      {
+        // invalid parameters
+        throw new InvalidMimeTypeException(mimeType);
+      }
```
karthikbalasub, line 118: Is this needed? If existing behavior that ignores invalid params is not blocking you, then we can split the change and move the backwards incompatible change to a separate PR.

restli-server/src/main/java/com/linkedin/restli/internal/server/util/MIMEParse.java
```diff
@@ -67,12 +69,24 @@
...
+    /**
+     * Build the String for the content type header
+     */
+    String toContentType() {
+      StringBuilder s = new StringBuilder(type).append("/").append(subType);
+      params.forEach((k, v) ->
+      {
+        // Exclude accept type's "q" param from the content type
+        if (!k.equals(QUALITY_PARAM))
+          s.append("; ").append(k).append("=").append(v);
```
karthikbalasub, line 86: add {}

gradle.properties
klee97: It's not urgent to fail invalid parameters. We can do as you suggested and split the change.

BrianPin:
> > Just add a general comment that in pegasus/restli the left bracket is on the next new line.
> 
> Thanks @BrianPin, is there a style command I can run?

Nothing I am aware of, I think we probably just need to be manually change this

karthikbalasub (approved):

PR #547, opened 2021-03-04
Fix bug that if the schema is an enum without any symbols, doc gen should handle it rather than throwing exception.

nickibi:
Doc gen did not handle the case that if resource key is an enum type without any symbols. Enum without symbols is a valid use case. This PR is fixing this issue.

Test:
Unit test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.15.6] - 2021-03-04
+- Fix bug that if a schema is an enum without any symbols, doc gen should handle it instead of throwing exception.
+
 ## [29.15.5] - 2021-03-03
 - Fix content type header not set in case of RestliResponseException from non streaming server
 
@@ -4865,7 +4868,8 @@ patch operations can re-use these classes for generating patch messages.
 
 ## [0.14.1]
 
-[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.15.5...master
+[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.15.6...master
+[29.15.6]: https://github.com/linkedin/rest.li/compare/v29.15.5...v29.15.6
 [29.15.5]: https://github.com/linkedin/rest.li/compare/v29.15.4...v29.15.5
 [29.15.4]: https://github.com/linkedin/rest.li/compare/v29.15.3...v29.15.4
 [29.15.3]: https://github.com/linkedin/rest.li/compare/v29.15.2...v29.15.3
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
+- Fix bug that if a schema is an enum without any symbols, doc gen should handle it instead of throwing exception.
```
junchuanwang, line 16: What are the use cases of enum withouth symbols?

data/src/main/java/com/linkedin/data/schema/generator/DefaultSampleDataCallback.java
```diff
@@ -135,7 +135,12 @@ public ByteString getFixed(String fieldName, FixedDataSchema schema)
   public String getEnum(String fieldName, EnumDataSchema enumDataSchema)
   {
     List<String> symbols = enumDataSchema.getSymbols();
-    return symbols.get(nonNegative(symbols.size() - 1));
+    // enum without any symbols is allowed, return "EmptyEnum" as a reference in doc.
+    if (symbols.size() < 1)
+    {
+      return "EmptyEnum";
+    }
+    return symbols.get(symbols.size() - 1);
```
junchuanwang, line 143: Do you want to keep using nonNegative to generate the index?
then I think you can just 
```
    if (symbols.size() < 1)
    {
      symbols = Arrays.asList("EmptyEnum");
    }

```

CHANGELOG.md
nickibi: There are quite a few use cases use empty enum as the resource key. For example, resources only support finder methods.

junchuanwang: (1) I still didn't understand. Can't those enums be replaced by string? What benefit would that bring and why people would choose to use the enum?

(2) I do believe here is a bug to fix but I am not sure if you adding the "EmptyEnum" would be the best solution.

nickibi: 1. I think it could be string, but using Enum, like enum None{} could be explicitly telling people no resource key is needed for this resource. That's my impression of using empty enum as the key, you may talk to those customers to understand their motivation more.
2. I was thinking about just return "". This is rest.li doc which is just used as as an example. For example, if the type is int or string we just return some random value. We could explicitly call "emptyEnum" to tell people this is an empty enum.

data/src/main/java/com/linkedin/data/schema/generator/DefaultSampleDataCallback.java
nickibi: I think we should keep the behavior same as previously, randomly get one symbol from the symbol list. When the symbol list size < 1, I think we should directly return the value to avoid unnecessary randomly generating index step. Updated the code to add nonNegative back.

BrianPin (approved):
Watched demo from nicki offline, I think this LGTM

karthikbalasub (approved):

junchuanwang (approved):

PR #548, opened 2021-03-05
Add separate configuration control for retrying RestRequest and StreamRequest

rickzx:
The current implementation of RetryClient will buffer every single incoming StreamRequest regardless of its entity stream size, which could potentially cause memory issues. Therefore, we would like to disable client retry for stream request by default. We will be adding two separate configs to control restRetry (enable by default) and streamRetry (disable by default).

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
@@ -78,7 +78,7 @@ public class RetryClient extends D2ClientDelegator
 {
   public static final long DEFAULT_UPDATE_INTERVAL_MS = TimeUnit.SECONDS.toMillis(1);
   public static final int DEFAULT_AGGREGATED_INTERVAL_NUM = 5;
-  public static final boolean DEFAULT_REST_RETRY_ENABLED = true;
+  public static final boolean DEFAULT_REST_RETRY_ENABLED = false;
   public static final boolean DEFAULT_STREAM_RETRY_ENABLED = false;
   private static final Logger LOG = LoggerFactory.getLogger(RetryClient.class);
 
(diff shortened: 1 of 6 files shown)


d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
```diff
@@ -78,13 +78,17 @@
 {
   public static final long DEFAULT_UPDATE_INTERVAL_MS = TimeUnit.SECONDS.toMillis(1);
   public static final int DEFAULT_AGGREGATED_INTERVAL_NUM = 5;
+  public static final boolean DEFAULT_REST_RETRY_ENABLED = true;
```
ssheng, line 81: Maybe `false` is a safer default in code. Factory config can default to `true`.

d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
```diff
@@ -94,6 +98,13 @@ public RetryClient(D2Client d2Client, LoadBalancer balancer, int limit)
   }
 
   public RetryClient(D2Client d2Client, LoadBalancer balancer, int limit, long updateIntervalMs, int aggregatedIntervalNum, Clock clock)
```
ssheng, line 100: Let's `@Deprecate` this constructor.

d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
rickzx: Done

d2/src/main/java/com/linkedin/d2/balancer/clients/RetryClient.java
rickzx: Done

ssheng (approved):

PR #559, opened 2021-03-16
Update fluent APIs to include projection mask as optional parameters.

karthikbalasub:


junchuanwang (commented):
Looks quite neat from first pass! Awesome

karthikbalasub:
Rebased with master after merging the mask builder PR #557

tjni:
I apologize I have not been keeping up with your work on fluent builders, but as an eventual consumer, I probably should :)

Take this example from the test:

```
greetings.batchGet(ids,
    new GreetingsFluentClient.BatchGetOptionalParameters().setFields(
        Greeting.createMask().withTone()
    ));
```

To me, this is more unwieldy than:

```
var request = new GreetingsRequestBuilders().batchGet()
    .fields(Greeting.fields.tone())
    .build();

client.sendRequest(request);
````

I feel like the second parameter in the fluent builder needs to have a shorter entry point at least. Maybe for example:

```
greetings.batchGet(ids, request -> request.fields(Greeting.fields.tone()));
```

But also consider:

```
greetings.batchGet(ids)
    .fields(Greeting.fields.tone())
    .execute();
```

which at the expense of needing an `execute()` call flattens everything all at once.

**Comment by karthikbalasub** [comment]
> I apologize I have not been keeping up with your work on fluent builders, but as an eventual consumer, I probably should :)
> 
> Take this example from the test:
> 
> ```
> greetings.batchGet(ids,
>     new GreetingsFluentClient.BatchGetOptionalParameters().setFields(
>         Greeting.createMask().withTone()
>     ));
> ```
> 
> To me, this is more unwieldy than:
> 
> ```
> var request = new GreetingsRequestBuilders().batchGet()
>     .fields(Greeting.fields.tone())
>     .build();
> 
> client.sendRequest(request);
> ```
> 
> I feel like the second parameter in the fluent builder needs to have a shorter entry point at least. Maybe for example:
> 
> ```
> greetings.batchGet(ids, request -> request.fields(Greeting.fields.tone()));
> ```
> 
> But also consider:
> 
> ```
> greetings.batchGet(ids)
>     .fields(Greeting.fields.tone())
>     .execute();
> ```
> 
> which at the expense of needing an `execute()` call flattens everything.

@tjni thanks for the feedback. I've updated it to follow the same structure that the new mask builder uses. It is similar to the first option you suggested.

File: restli-int-test/src/test/java/com/linkedin/restli/examples/TestParseqBasedFluentClientApiWithProjections.java [hunk 1]
```diff
@@ -117,9 +114,7 @@ public void testBatchGetWithProjection() throws Exception
 
     Set<Long> ids = Sets.newHashSet(Arrays.asList(1L, 2L, 3L));
     CompletionStage<Map<Long, EntityResponse<Greeting>>> result = greetings.batchGet(ids,
-        new GreetingsFluentClient.BatchGetOptionalParameters().setFields(
-            Greeting.createMask().withTone()
-        ));
+        optionalParams -> optionalParams.withFields(mask -> mask.withTone()));
```
tjni, line 117: I think it would be better to either be `withMask(mask -> ...)` or `withFields(fields -> ...)`, but this is a minor consideration.

tjni (commented):
Thank you! I like this a lot more, and having a method to set the mask all at once is important for use cases where masks need to be constructed dynamically or shared.

With that said, I wonder, but I am not sure, if for use cases where the projection is determined at the time of building the request, if it would read better to be able to add (or remove) fields directly on the optional parameters.

restli-int-test/src/test/java/com/linkedin/restli/examples/TestParseqBasedFluentClientApiWithProjections.java
junchuanwang: @tjni I actually agree with you, and adding this will be sweet (sugar). This only requires adding another sets of methods, doesn't it?

karthikbalasub: Update the API to withMask.

karthikbalasub:
> Thank you! I like this a lot more, and having a method to set the mask all at once is important for use cases where masks need to be constructed dynamically or shared.
> 
> With that said, I wonder, but I am not sure, if for use cases where the projection is determined at the time of building the request, if it would read better to be able to add (or remove) fields directly on the optional parameters.

The new MaskMap type allows building the mask one field at a time. This was added recently and we generate APIs in datatemplate to set that. 
Users can also create a partial mask, pass that around to add fields and then use that when making the API call. 

For example:
```
Greeting.ProjectionMask mask = Greeting.createMask();
... // Update mask

greetings.get(1L, params -> params.withMask(emptyMask -> mask));
```

junchuanwang (approved):

restli-tools/src/main/java/com/linkedin/restli/tools/clientgen/fluentspec/RestMethodSpec.java
```diff
@@ -124,4 +131,43 @@ public int getQueryParamMapSize()
...
+      case GET_ALL:
+        Set<ProjectionParameterSpec> collectionParts = new HashSet<>();
+        collectionParts.add(new ProjectionParameterSpec(RestConstants.FIELDS_PARAM, FIELDS_MASK_METHOD_NAME, _root.getEntityClass()));
+        if (_schema.getMetadata() != null)
+        {
+          collectionParts.add(new ProjectionParameterSpec(RestConstants.METADATA_FIELDS_PARAM, METADATA_MASK_METHOD_NAME, _root.classToTemplateSpec(_schema.getMetadata().getType())));
+        }
+        if (_schema.hasPagingSupported() && _schema.isPagingSupported())
+        {
+          // TODO: add suppport for paging projection.
```
junchuanwang, line 165: Is this in the roadmap?

karthikbalasub: This will be done along with finder support and we decide how to return paging data.

PR #560, opened 2021-03-16
Regression test to certify legacy Ivy publication behavior

DPUkyle:
Regression test to certify legacy Ivy publication behavior

- Grandparent -> parent -> child pattern certifies that child project can transitively resolve references to schemas contained in grandparent's data-template jar


evanw555:
FYI: Going to chat with @DPUkyle later today about this. Will comment with updates.

evanw555:
This PR is just adding an integration test as a baseline for the current expected behavior of publishing pegasus .ivy artifacts. This changes no actual behavior. Kyle will work on changes in PegasusPlugin to make sure it's Gradle 7 compatible, which will involve publishing .ivy metadata using a newer API. This will alter the generated .ivy metadata, so this PR is helpful in that it will allow us to more easily see what those differences are.

DPUkyle:
I rebased on cbac36d36cdb1e7dd6ad68c21e1a90f44ebb7ccb after adding some javadoc - this is ready to merge.

PR #564, opened 2021-03-20
Add default fill in for custom metadata and collection metadata

BrianPin:
This fix is because we lack of filling in default values in the paging metadata and custom metadata in the collection response for methods like finders and get_all.
Previously we just iterate the elements and filling in the default values, this PR is adding code to check not just elements but also paging metadata and custom metadata

### DIFF
--- a/restli-int-test/src/test/java/com/linkedin/restli/examples/TestFillInDefaultValue.java
+++ b/restli-int-test/src/test/java/com/linkedin/restli/examples/TestFillInDefaultValue.java
@@ -15,7 +15,6 @@
 */
 package com.linkedin.restli.examples;
 
-import com.linkedin.data.Data;
 import com.linkedin.data.DataList;
 import com.linkedin.data.DataMap;
 import com.linkedin.r2.RemoteInvocationException;
@@ -27,13 +26,19 @@ import com.linkedin.restli.client.GetAllRequest;
 import com.linkedin.restli.client.GetRequest;
 import com.linkedin.restli.client.response.BatchKVResponse;
 import com.linkedin.restli.common.BatchFinderCriteriaResult;
+import com.linkedin.restli.common.CollectionMetadata;
+import com.linkedin.restli.common.CollectionResponse;
 import com.linkedin.restli.common.EntityResponse;
+import com.linkedin.restli.common.LinkArray;
 import com.linkedin.restli.common.RestConstants;
 import com.linkedin.restli.examples.defaults.api.FillInDefaultsGetRequestBuilder;
 import com.linkedin.restli.examples.defaults.api.FillInDefaultsRequestBuilders;
 import com.linkedin.restli.examples.defaults.api.HighLevelRecordWithDefault;
+import com.linkedin.restli.examples.defaults.api.LowLevelRecordWithDefault;
 import com.linkedin.restli.examples.defaults.api.RecordCriteria;
+import com.linkedin.restli.server.CollectionResult;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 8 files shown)


junchuanwang (commented):
Should also fix the test in TestResourceSchemaCollection

junchuanwang:
Test failed in TestRseourceSchemaCollection.java

junchuanwang:
What was the behavior before this change? Please add to the description "before versus now"

restli-int-test-api/src/main/pegasus/com/linkedin/restli/examples/defaults/api/FooDefault.pdl
```diff
@@ -0,0 +1,8 @@
+namespace com.linkedin.restli.examples.defaults.api
+
+import com.linkedin.restli.common.CollectionMetadata
+
+record FooDefault {
```
karthikbalasub, line 5: Add a doc to explain which test scenarios this schema is used

karthikbalasub: Though this is a test model, calling it "foo" is not nice. 

Call it: `FieldWithDefaultRecord`  or `HasFieldsWithDefault` ?

restli-int-test-server/src/main/java/com/linkedin/restli/examples/greetings/server/defaults/FooDefaultFillResources.java
```diff
@@ -0,0 +1,48 @@
...
+    {
+      fooDefaults.add(new FooDefault().setId(i));
+    }
+    FooDefault customMetadata = new FooDefault().setId(10);
+
+    return new CollectionResult<FooDefault, FooDefault>(fooDefaults, customMetadata);
+  }
+
+  @Finder("findFooCollection")
+  public CollectionResult<FooDefault, FooDefault> findFooCollection(@QueryParam("count") Integer count)
```
karthikbalasub, line 37: I understand we have other classes that use the same record for metadata. Since we are testing default filling on metadata, I think using a separate record is cleaner.
For example if default filling is broken for non-entity records, the current setup will not catch it.

restli-int-test/src/test/java/com/linkedin/restli/examples/TestFillInDefaultValueFooDefault.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.Collections;
+import java.util.List;
+import org.testng.Assert;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+
+public class TestFillInDefaultValueFooDefault extends RestLiIntegrationTest
```
karthikbalasub, line 24: Rename this to
`TestFillInCollectionMetadataDefaultValue`

restli-int-test/src/test/java/com/linkedin/restli/examples/TestFillInDefaultValueFooDefault.java
```diff
@@ -0,0 +1,128 @@
...
+    expectedCollectionMetadata.setTotal(0);
+
+    FooDefault expectedCustomMetadata = new FooDefault();
+    expectedCustomMetadata.setFooDefaultId(0);
+    expectedCustomMetadata.setId(count);
+    return new Object[][] {{2, expectedElements, expectedCollectionMetadata, expectedCustomMetadata}};
+  }
+
+  @Test(dataProvider = "testFinderData")
+  public void testFinder(int param,
```
karthikbalasub, line 68: `testFinderResponseWithDefaultsFilled`

restli-int-test/src/test/java/com/linkedin/restli/examples/TestFillInDefaultValueFooDefault.java
```diff
@@ -0,0 +1,128 @@
...
+      List<FooDefault> expectedElements,
+      CollectionMetadata expectedCollectionMetadata,
+      FooDefault expectedCustomMetadata) throws RemoteInvocationException
+  {
+    FooDefaultFillingRequestBuilders requestBuilders = new FooDefaultFillingRequestBuilders();
+    FindRequest<FooDefault> request = requestBuilders.findByFindFooCollection()
+        .setParam("count", param)
+        .setParam(RestConstants.FILL_IN_DEFAULTS_PARAM, true)
```
karthikbalasub, line 76: ideally we should test both with true and false and verify the change in behavior. The current tests will not catch the case when server code is directly setting the field and the "fillInDefault" behavior is broken.

restli-int-test-server/src/main/java/com/linkedin/restli/examples/greetings/server/defaults/FooDefaultFillResources.java
```diff
@@ -0,0 +1,48 @@
+package com.linkedin.restli.examples.greetings.server.defaults;
```
evanw555, line 1: Add license to all new Java files.

restli-int-test-api/src/main/pegasus/com/linkedin/restli/examples/defaults/api/FooDefault.pdl
evanw555, line 5: Do we need to add a new record for this use case? We already have a similar record called `LowLevelRecordWithDefault` in the package `com.linkedin.restli.examples.defaults.api. Could we reuse that instead?

restli-int-test-server/src/main/java/com/linkedin/restli/examples/greetings/server/defaults/FooDefaultFillResources.java
```diff
@@ -0,0 +1,48 @@
...
+import com.linkedin.restli.server.CollectionResult;
+import com.linkedin.restli.server.annotations.Finder;
+import com.linkedin.restli.server.annotations.QueryParam;
+import com.linkedin.restli.server.annotations.RestLiCollection;
+import com.linkedin.restli.server.annotations.RestMethod;
+import com.linkedin.restli.server.resources.CollectionResourceTemplate;
+import java.util.ArrayList;
+import java.util.List;
+
+
```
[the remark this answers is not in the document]
evanw555: +1, test resources should have a Javadoc explaining what their purpose is, otherwise it's pretty hard to know what this is for without context.

restli-server/src/main/java/com/linkedin/restli/internal/server/response/CollectionResponseBuilder.java
```diff
@@ -166,8 +172,13 @@ private D buildRestLiResponseData(final Request request,
     final AnyRecord projectedCustomMetadata;
     if (customMetadata != null)
     {
+      DataMap customData = customMetadata.data();
```
evanw555, line 175: nit: this variable name is misleading. I'd rename to `customMetadataMap` or `customMetadataData` to be consistent with the variable `projectedCustomMetadata` below.

restli-int-test-server/src/main/java/com/linkedin/restli/examples/greetings/server/defaults/FooDefaultFillResources.java
```diff
@@ -0,0 +1,48 @@
...
+@RestLiCollection(name = "fooDefaultFilling", namespace = "com.linkedin.restli.examples.defaults.api")
+public class FooDefaultFillResources extends CollectionResourceTemplate<Integer, FooDefault>
```
evanw555, line 15: 1. Similar to Karthik's other comment, this resource should be named according to what it's actually doing. Something like `FillInDefaultsCollectionMetadataResource` ("Resource" should not be pluralized).
2. Also consider whether or not we need this extra resource. Keep in mind that another resource `FieldFillInDefaultResources` exists and it appears to have similar functionality. Can we reuse that?

restli-int-test-server/src/main/java/com/linkedin/restli/examples/greetings/server/defaults/FooDefaultFillResources.java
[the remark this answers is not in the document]
BrianPin: Yes, thanks for the comment, I end up using the other resource class instead

restli-int-test-api/src/main/pegasus/com/linkedin/restli/examples/defaults/api/FooDefault.pdl
BrianPin: Thanks for the comment, end up using the existing default filling purpose schema

restli-int-test-server/src/main/java/com/linkedin/restli/examples/greetings/server/defaults/FooDefaultFillResources.java
BrianPin: Yes, In the final test code, I end up using different schema for CollectionResult generics.

restli-int-test/src/test/java/com/linkedin/restli/examples/TestFillInDefaultValueFooDefault.java
BrianPin: End up augmenting existing test!

restli-int-test/src/test/java/com/linkedin/restli/examples/TestFillInDefaultValueFooDefault.java
BrianPin: Thanks, I added a test for not having the param that set to fill in default values in the test of get_all

restli-int-test-server/src/main/java/com/linkedin/restli/examples/greetings/server/defaults/FooDefaultFillResources.java
BrianPin: End up using existing resource class.

restli-int-test-api/src/main/pegasus/com/linkedin/restli/examples/defaults/api/FooDefault.pdl
BrianPin: Yes, I revise this new schema to use the old one

restli-server/src/main/java/com/linkedin/restli/internal/server/response/CollectionResponseBuilder.java
BrianPin: Changed that in next revision

restli-int-test-server/src/main/java/com/linkedin/restli/examples/greetings/server/defaults/FooDefaultFillResources.java
BrianPin: Yes, end up using existing resources

restli-int-test/src/test/java/com/linkedin/restli/examples/TestFillInDefaultValue.java
```diff
@@ -156,50 +161,96 @@ public void testFillInDefaultBatchGet(Long[] ids, HighLevelRecordWithDefault[] e
...
+  {
+    final int count = 3;
+    List<HighLevelRecordWithDefault> elements = new ArrayList<>();
+    for (int i = 0; i < count; i++)
+    {
+      elements.add(new HighLevelRecordWithDefault(expectedTestData.clone()).setNoDefaultFieldA(i));
+    }
+    CollectionMetadata collectionMetadata = new CollectionMetadata()
+        .setCount(10).setTotal(3).setStart(0).setLinks(new LinkArray());
+    LowLevelRecordWithDefault metadata = new LowLevelRecordWithDefault().setNameWithDefault("i_am_default_name");
```
karthikbalasub, line 174: nit: I'd do this as:
```
LowLevelRecordWithDefault metadata = new LowLevelRecordWithDefault();
metadata.setNameWithDefault(metadata.getNameWithDefault());
```

karthikbalasub (approved):

restli-int-test/src/test/java/com/linkedin/restli/examples/TestFillInDefaultValue.java
BrianPin: gotcha

PR #566, opened 2021-03-23
Bump Gradle wrapper to 6.1

DPUkyle:
- This change does not require the consumer to use Gradle 6.1
 - Forthcoming opt-in features to improve metadata publishing will require Gradle 6.1 (#565)
 - Fixes to #561 require Gradle 5.4 or higher
 - Blocked by #567

PR #568, opened 2021-03-23
Initial support for ivy-publish and feature variants (capabilities)

DPUkyle:
- Additional test coverage for Ivy
 - Continue building with Gradle 5.2.1; requires reflection to access new API added in Gradle 5.3
 - New code block is a no-op unless the consuming project applies the `ivy-publish` plugin
 - Producing [feature variants](https://docs.gradle.org/current/userguide/feature_variants.html) correctly using `ivy-publish` requires Gradle 6.1. Plugin will throw if `ivy-publish` is in use and Gradle version is below 6.1.

--

~Depends on #566~

There's a lot of info to unpack here; I'm preparing a talk to share with the maintainers. And of course I'll squash and rebase before merging.


gradle-plugins/src/integTest/groovy/com/linkedin/pegasus/gradle/PegasusPluginIntegrationTest.groovy
```diff
@@ -22,7 +22,7 @@ class PegasusPluginIntegrationTest extends Specification {
         .withProjectDir(tempDir.root)
         .withPluginClasspath()
         .withArguments('mainDataTemplateJar')
-        .forwardOutput()
+        //.forwardOutput()
```
evanw555, line 25: Context?

gradle-plugins/src/integTest/groovy/com/linkedin/pegasus/gradle/publishing/PegasusPluginIvyPublishIntegrationTest.groovy
```diff
@@ -0,0 +1,340 @@
+package com.linkedin.pegasus.gradle.publishing
```
evanw555, line 1: All new source files should have a license at the top (just copy from any other Java file and update the date).

gradle-plugins/src/integTest/resources/ivy/modern/expectedChildIvyDescriptorContents.txt
```diff
@@ -0,0 +1,30 @@
+  <configurations>
+    <conf name="compile" visibility="public"/>
+    <conf name="dataTemplate" visibility="public" extends="mainGeneratedDataTemplateApiElements,mainGeneratedDataTemplateRuntimeElements"/>
```
evanw555, line 3: This only exposes `dataTemplate`, but not `dataModel`. I'm not sure if this is totally necessary, though. The Pegasus plugin resolves PDL schemas via a `FileCollection` which is basically just the target project's `dataModel` or `testDataModel` configuration. Most use cases add a dependency on external PDL schemas via `dataModel`, but consume the publication's `dataTemplate` configuration.... I think `dataModel` extends `dataTemplate`. Based on your knowledge, is there any risk here?

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -2246,4 +2345,31 @@ private void checkGradleVersion(Project project)
           MIN_SUGGESTED_VERSION));
     }
   }
+
+  /**
+   * Reflection is necessary to obscure types introduced in Gradle 5.3
```
evanw555, line 2350: _General comment for the whole PR:_ Make sure to update the Changelog with whatever is relevant to consumers (e.g. compatibility concerns, new behavior).

gradle-plugins/src/integTest/groovy/com/linkedin/pegasus/gradle/PegasusPluginIntegrationTest.groovy
DPUkyle: Sorry, it's unrelated to this PR. GradleRunner#forwardOutput includes the stdout of the Gradle project being tested in the CI logs. Useful for debugging, but is unnecessarily verbose once things are working as expected. I can omit this line to keep this PR "pure", but this change is also so insignificant I'm not sure it deserves its own PR!

gradle-plugins/src/integTest/groovy/com/linkedin/pegasus/gradle/publishing/PegasusPluginIvyPublishIntegrationTest.groovy
DPUkyle: Even for tests? I'm fine with that, but we haven't been following that well based on existing classes in the `test` and `integTest` sourceSets.

gradle-plugins/src/integTest/resources/ivy/modern/expectedChildIvyDescriptorContents.txt
DPUkyle: tl;dr no risk, but IMO the legacy code exposing `dataModel` was incorrect.

Longer description:

Legacy Gradle was kind of simple, where consumers use the same configuration names to consume and publish dependencies - think of `runtime` and `compile`.

In modern Gradle, things get trickier. Consumers use the `implementation` configuration to specify a dependency, Gradle then internally uses the `compileClasspath` to consume it, and finally publishes a derived `compile` configuration.

My point of making this comparison is that `dataTemplate` and `dataModel` have a similar relationship. From my observations of existing behavior, yes `dataModel` is how consumers specify a set of data templates to consume. When the same project published its metadata, only publishing the `dataTemplate` configuration is necessary. In other words, `dataModel` is an internal implementation detail of a pegasus module, and concealing it from consumers by not enumerating `dataModel` in the Ivy descriptor is correct behavior. The legacy code exposing `dataModel` was incorrect.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
DPUkyle: Will do.

gradle-plugins/src/integTest/groovy/com/linkedin/pegasus/gradle/publishing/PegasusPluginIvyPublishIntegrationTest.groovy
evanw555: We add licenses for all Java tests, yes. If the license wasn't added to the other Gradle integ tests, I'd say this was a mistake. I agree that the necessity of this could be re-evaluated, but for now keeping consistency is preferable.

gradle-plugins/src/integTest/groovy/com/linkedin/pegasus/gradle/PegasusPluginIntegrationTest.groovy
evanw555: Yeah, it's no big deal. I personally like to leave comments about something like this, because it could be ambiguous to someone way down the road as to whether this was intentional or a mistake.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1840,6 +1911,34 @@ protected GenerateDataTemplateTask configureDataTemplateGeneration(Project proje
     return generateDataTemplatesTask;
   }
 
+  private String mapSourceSetToFeatureName(SourceSet sourceSet) {
+    String featureName = "";
+    switch (sourceSet.getName()) {
```
evanw555, line 1916: This appears to be tech debt, since we have to maintain static mappings from what the "natural" configuration names are to what we expect them to be.
1. Is my understanding correct?
2. If so, would there be a benefit in getting users to migrate their usage (e.g. from `restClient` to `mainGeneratedRest`)? Ignoring the cost of migration, of course... I want to understanding if this is "hacky" or totally acceptable in the Gradle world.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1825,6 +1834,68 @@ protected GenerateDataTemplateTask configureDataTemplateGeneration(Project proje
     // FIXME change to #getArchiveFile(); breaks backwards-compatibility before 5.1
     project.getDependencies().add(compileConfigName, project.files(dataTemplateJarTask.getArchivePath()));
 
+    project.getPlugins().withType(IvyPublishPlugin.class, ivyPublish -> {
```
evanw555, line 1837: You already have a lot of documentation which is good, but it would be a good idea IMO to leave a comment here to describe at a high level what this closure is for. The plugin is already hard enough for us to understand, we need all the documentation we can get.

gradle-plugins/src/integTest/groovy/com/linkedin/pegasus/gradle/PegasusPluginIntegrationTest.groovy
DPUkyle: Changing this line isn't worth the trouble; reverting.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
DPUkyle: Sure, I'll add extra docs.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
DPUkyle: This... is hard to explain. The extra complexity is related to the legacy dataModel/dataTemplate/dataTemplate configurations and preserving compatibility.

The first step is to please read and understand Gradle's concept of a "feature variant". This [doc](https://docs.gradle.org/6.1/userguide/feature_variants.html) is crucial. The pegasus data-template jar is a perfect example of a feature variant. In my own words, it's an "alternate" artifact addressable at the same coordinates as the "primary" artifact, and having a parallel set of its own dependencies.

Now, when registering a "feature variant" Gradle generates its own configurations which can represent the same concepts as dataModel/dataTemplate. The key difference is that these are derived from the sourceSet containing the "feature", whereas legacy pegasus does much extra work to associate the `mainGeneratedDataTemplate` sourceset to `dataTemplate` and `dataModel` configurations.

When doing it the Gradle way, a sourceset called `mainGeneratedDataTemplate` will spawn configurations named `mainGeneratedDataTemplateApiElements` and `mainGeneratedDataTemplateRuntimeElements`.In theory, these are what consumers should be modifying instead

Long-term, my design recommendation is to rename the `mainGeneratedDataTemplate` SourceSet to simply `dataTemplate`. This will spawn `dataTemplateApi` and `dataTemplateImplementation` configurations to the project applying the pegasus plugin, and consumers can decide how they want to manipulate the dependencies.

evanw555: Got it, thanks for the explanation. I should document this somewhere 😃

evanw555 (approved):

PR #575, opened 2021-03-29
Update ParseqBasedCompletionStage to get task from instances of same class for completionStage composition

junchuanwang:
Previously we just treat another ParseqBasedCompletionStage as a competionStage and convert it into a task.

The idea of the change now is to try to add tasks from different ParseqBasedCompletionStage into same plan by best effort and the trace/task from previous ParseqBasedCompletionStage is reused.


restli-client/src/main/java/com/linkedin/restli/client/ParSeqBasedCompletionStageFactory.java
```diff
@@ -60,15 +60,39 @@ private void checkEngine()
...
   public ParSeqBasedCompletionStage<T> buildStageFromTask(Task<T> task)
   {
     checkEngine();
     return new ParSeqBasedCompletionStage<T>(_engine, _asyncExecutor,task);
   }
 
+  /**
+   * Convenient method to create a {@link ParSeqBasedCompletionStage} with a {@link Task} has not started.
+   *
+   * Will build {@link ParSeqBasedCompletionStage} by using a {@link Task} that has not started running, which will
```
karthikbalasub, line 75: This line is repeating same statement as above. Can we merge them?

restli-client/src/main/java/com/linkedin/restli/client/ParSeqBasedCompletionStageFactory.java
```diff
@@ -60,15 +60,39 @@ private void checkEngine()
...
+   *    be run by the engine provided in this factory.
+   *
+   *
+   *  Note0: In {@link ParSeqBasedCompletionStage}, all {@link Task} used was assumed to have started running and will
+   *         produce a value in the future.
+   *  Note1: Use {@link ParSeqBasedCompletionStage##buildStageFromTask(Task)} if the Task has started
+   *  Note2: Difference between this method and {@link ParSeqBasedCompletionStage##buildStageFromCompletionStage(CompletionStage)}
+   *         :If {@link Task} produces an exception, this method will have the very exception, while {@link ##buildStageFromCompletionStage(CompletionStage)}
```
karthikbalasub, line 83: have the same exception

karthikbalasub (approved):

restli-client/src/test/java/com/linkedin/restli/client/TestParSeqBasedCompletionStage.java
```diff
@@ -1302,6 +1325,75 @@ public void testWhenCompleteAsync() throws Exception
     assertTrue(waitLatch.await(1000, TimeUnit.MILLISECONDS));
   }
 
+  private void testWithcomposableApi(CompletionStage stage, List<CountDownLatch> latches) throws Exception {
```
BrianPin, line 1328: composable to Composable?

restli-client/src/test/java/com/linkedin/restli/client/TestParSeqBasedCompletionStage.java
```diff
@@ -1302,6 +1325,75 @@ public void testWhenCompleteAsync() throws Exception
...
+    if (latches == null)
+    {
+      // synchronization not needed, then wait blockingly
+      finish(completionStage.thenCompose(r -> stage).thenAccept(consumer));
+      verify(consumer, times(1)).accept(TESTVALUE2);
+    }
+    else
+    {
+      new Thread(() ->{
+        try {
```
BrianPin, line 1340: although it is tiny, but the code format seems a little bit inconsistent.
for line 1338, it is a bracket in new line, but in line 1339 1340 it is bracket in same line

restli-client/src/main/java/com/linkedin/restli/client/ParSeqBasedCompletionStage.java
```diff
@@ -561,10 +561,27 @@
...
     return _task;
   }
+
+  /**
+   * Special treatment to ParSeqBasedCompletionStage
+   * Note that there is no assumption that generated Task has been started
+   * The code which uses this method should consider that
+   */
+  @SuppressWarnings({"unchecked"})
+  protected <U> Task<U> getOrGenerateTaskFromStage(CompletionStage<? extends U> stage)
```
BrianPin, line 575: Just a 2$, maybe I am just being naive.
Here is what I think, can we give flexibility to user in terms of what they want in the condition of the `else` part?
I am assuming this function here is act very similar to what we usually see in other frameworks: `getOrElse`
Just as this example https://medium.com/@Shehaaz/scala-options-explained-11d25fd67268 

Just to verify my understanding here, this `getOrGenerateTaskFromStage`  functionis trying to get a `Task` from a `CompletionStage`, but if it is not Parseq CompletionStage, then we generate a `Task` from whatever stage it is.

So I was thinking maybe we could let the caller provide a default Task?

BrianPin: We can name it like `getTaskOrElse`, and the function argument can take a default `Task` from user's input?

restli-client/src/main/java/com/linkedin/restli/client/ParSeqBasedCompletionStage.java
```diff
@@ -561,10 +561,27 @@
...
-  Task<T> getTask()
+  protected Task<T> getTask()
   {
     return _task;
   }
+
+  /**
+   * Special treatment to ParSeqBasedCompletionStage
+   * Note that there is no assumption that generated Task has been started
+   * The code which uses this method should consider that
```
BrianPin, line 572: Hope to have a clarity here, if it is a Task from `ParseqBasedCompletionStage` it is not guaranteed it is not started yet, but for the default generated task from the any other stage, is it still not guaranteed it may or may not started yet?

restli-client/src/main/java/com/linkedin/restli/client/ParSeqBasedCompletionStage.java
junchuanwang: no, as the name suggest, "getOrGenerate", there is no "else" (since else means generate).

restli-client/src/main/java/com/linkedin/restli/client/ParSeqBasedCompletionStage.java
junchuanwang: > is it still not guaranteed it may or may not started yet?
no guarantee, that is correct. But if from other stage, most likely they have started (they are probably called by `##ensureFuture` during transition of stages).

Also don't worry about they don't get started, just call `##ensureFuture` again.

But I also mentioned somewhere in the ParseqBasedCompletionStageFactory, if one build ParSeqBasedCompletionStage using a task,  one should be careful about starting that task himself, that is because the input task usually need some special treatment.

So to conclude, usually we should try to keep consistent to think that task in ParSeqBasedCompletionStage have all started, if not certain, should call `##ensureFuture`

restli-client/src/test/java/com/linkedin/restli/client/TestParSeqBasedCompletionStage.java
junchuanwang: resolved

restli-client/src/main/java/com/linkedin/restli/client/ParSeqBasedCompletionStageFactory.java
junchuanwang: change. But "very" is also an adjective...

PR #580, opened 2021-03-31
FluentAPI implementation for action set and action methods

junchuanwang:
(1) Added action sets and action methods
(2) cleanred param class display a little bit.
(3) Basic tetss

Now the action methods have two sets of API
(1) Generic one:
`action(Function<ActionParamClass, ActionParamClass> paramProvider)`
(2) If the action method has required params. 
`action(...requiredParam..., Function<ActionOptionalParamClass, ActionOptionalParamClass> optionalParamProvider)`


restli-client/src/main/java/com/linkedin/restli/client/util/FluentClientUtils.java
```diff
@@ -61,4 +66,22 @@ private FluentClientUtils()
     }
     return input;
   }
+
+  public static Map<FieldDef<?>, Object> buildReadOnlyActionParameters(Map<FieldDef<?>, Object> actionParams)
+  {
+    return ActionRequestBuilder.buildReadOnlyActionParameters(actionParams);
```
karthikbalasub, line 72: We haven't been building readonly parameters in the fluent client. Since we don't create an intermediate request builder, there is really no need to build them.

restli-int-test/src/test/java/com/linkedin/restli/examples/TestGreetingsClient.java
```diff
@@ -338,6 +338,7 @@ public void testUpdateToneAction(RootBuilderWrapper<Long, Greeting> builders) th
             .setActionParam("DelOld", false)
             .build();
     ResponseFuture<Greeting> responseFuture = getClient().sendRequest(request);
+//    Response resp = responseFuture.getResponse();
```
karthikbalasub, line 341: revert

restli-tools/src/main/java/com/linkedin/restli/tools/clientgen/fluentspec/BaseResourceSpec.java
```diff
@@ -225,8 +250,7 @@ else if ((spec instanceof CollectionResourceSpec) && ((CollectionResourceSpec) s
             imports.add(CreateIdRequest.class.getName());
             imports.add(IdResponse.class.getName());
             imports.add(IdResponseDecoder.class.getName());
-            if (methodSpec.returnsEntity())
-            {
+            if (methodSpec.returnsEntity()) {
```
karthikbalasub, line 253: code style in this file should be fixed.

restli-tools/src/main/resources/apiVmTemplates/action.vm
```diff
@@ -0,0 +1,114 @@
...
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*#
+
+  #foreach($method in $spec.actions)
+    #set($actionParamClassName = "${util.nameCapsCase($method.name)}ActionParameters")
+    #doc($method.schema.doc)
+
+    ## methods are named as "action<ActionName>" so it is consistent with current RequestBuilder naming
```
karthikbalasub, line 21: I don't think we need the "action" prefix. 

Actions are usually  "verbs" and so client.<actionName> would be easier to read. For example,
greetings.purge(123)
vs
greetings.actionPurge(123)

restli-tools/src/main/resources/macros/library.vm
```diff
@@ -30,29 +29,57 @@
...
+          this.$param.paramName = $param.paramName;
+          _actionParams.put(_resourceSpec.getRequestMetadata("${method.name}").getFieldDef("$param.paramName"), $param.paramName);
+          return this;
+        }
+      #end
+
+      public Map<FieldDef<?>, Object> buildParametersMap()
+      {
+        return FluentClientUtils.buildReadOnlyActionParameters(_actionParams);
+      }
```
karthikbalasub, line 74: As I mentioned above, we don't have to build read only parameters.

restli-tools/src/main/resources/apiVmTemplates/action.vm
```diff
@@ -0,0 +1,114 @@
...
+    @SuppressWarnings("unchecked")
+    public CompletionStage<${method.valueClassDisplayName}> action${util.nameCapsCase(${method.name})}(
+      #if(${method.isEntityAction()})
+      $spec.keyClassName $spec.idName,
+      #end
+        #set($paramLen = ${method.getParameters().size()})
+        #set($lastParam = ${method.getParameters().get($paramLen - 1)})
+        #if($paramLen > 1)
+          #foreach($idx in [2..$paramLen])
+            #set($param = ${method.getParameters().get($idx - 2)})
```
karthikbalasub, line 32: To be consistent with Rest method behavior, we would have to add only the required parameters here.

junchuanwang: Sorry I don't understand this comment. I am only adding the "actionParam" here. What do you mean by 
"add only the required parameters here"

karthikbalasub: ActionParams can be optional or required. In the fluent api, only the required parameters are directly specified. All optional parameters go into a xxxOptionalParams class. This is so we can add new optional parameters to action methods without breaking fluent APIs. 

This is already done for REST methods.

restli-tools/src/main/java/com/linkedin/restli/tools/clientgen/fluentspec/BaseResourceSpec.java
karthikbalasub: Still needs fixing :(

restli-tools/src/main/resources/apiVmTemplates/action.vm
```diff
@@ -0,0 +1,129 @@
...
+    public CompletionStage<${method.valueClassDisplayName}> ${util.nameCamelCase(${method.name})}(
+      #if(${method.isEntityAction()})
+      $spec.keyClassName $spec.idName,
+      #end
+        #set($paramLen = ${method.getRequiredParameters().size()})
+        #if($paramLen > 0)
+          #set($lastParam = ${method.getRequiredParameters().get($paramLen - 1)})
+          #if($paramLen > 1)
+            #foreach($idx in [2..$paramLen])
+              #set($param = ${method.getRequiredParameters().get($idx - 2)})
```
karthikbalasub, line 33: Can you do `ids in [0..$paramLen-2]` instead?

karthikbalasub (approved):

restli-tools/src/main/resources/apiVmTemplates/action.vm
junchuanwang: this is what I wrote initially. But the velocity template doesn't support this syntax(I will have to define an additional variable to make it work).

PR #584, opened 2021-04-02
Fix default value resolution logic in avro translator to match PDL.

karthikbalasub:
Avro schema translator should use the default value from the nested field if the parent record level default doesn't provide a value for the field.
This behavior is consistent with how PDL schema defaults are handled.

### DIFF
--- a/data-avro/src/main/java/com/linkedin/data/avro/DefaultDataToAvroConvertCallback.java
+++ b/data-avro/src/main/java/com/linkedin/data/avro/DefaultDataToAvroConvertCallback.java
@@ -215,6 +215,17 @@ class DefaultDataToAvroConvertCallback extends AbstractDefaultDataTranslator imp
           (_options.getOptionalDefaultMode() == OptionalDefaultMode.TRANSLATE_TO_NULL && isTranslatedUnionMember) ||
           (fieldValue == Data.NULL));
     }
+    else if (fieldValue == null)
+    {
+      // If the default specified at parent level doesn't specify a value for the field, use the default specified at
+      // field level.
+      fieldValue = field.getDefault();
+      if (fieldValue == null)
+      {
+        throw new IllegalArgumentException(
+            message(path, "Cannot translate required field without default."));
+      }
+    }
     Object resultFieldValue = translate(path, fieldValue, fieldDataSchema);
     _newDefaultSchema = fieldDataSchema;
     return resultFieldValue;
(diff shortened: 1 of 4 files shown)


data-avro/src/test/java/com/linkedin/data/avro/TestDataTranslator.java
```diff
@@ -1710,6 +1710,32 @@ private void testDataTranslation(String schemaText, String[][] row) throws IOExc
...
+            "{\"bar\":{\"barInt\":42}}",
+            PegasusToAvroDefaultFieldTranslationMode.TRANSLATE,
+            "{\"bar\":{\"barInt\":42}}",
+            false,
+            "",
+        },
+
 
         // Below are test case example for invalid option combination:
         // If the field in pegasus schema has been translated as required, its data cannot be translated as optional
```
nickibi, line 1741: Shall we also add a negative test case which will fall into the exception path: throw new IllegalArgumentException(
            message(path, "Cannot translate required field without default."))?

karthikbalasub: It is not possible to test that case as PDL schema validation will fail for the same reason (I tried it).

nickibi (approved):
LGTM

data-avro/src/main/java/com/linkedin/data/avro/DefaultDataToAvroConvertCallback.java
```diff
@@ -215,6 +215,17 @@ protected Object translateField(List<Object> path, Object fieldValue, RecordData
           (_options.getOptionalDefaultMode() == OptionalDefaultMode.TRANSLATE_TO_NULL && isTranslatedUnionMember) ||
           (fieldValue == Data.NULL));
     }
+    else if (fieldValue == null)
+    {
+      // If the default specified at parent level doesn't specify a value for the field, use the default specified at
+      // field level.
+      fieldValue = field.getDefault();
+      if (fieldValue == null)
```
junchuanwang, line 223: if this field is a record,  and this field doesn't have the defaults. So `fieldValue` will be null.

But what is each field of this record it self have defaults?  Does that mean this record field implictly have a default?

Are we allowing that?

karthikbalasub: If this field is a record without default. Then the parent should've specified a default value for this field (even if all its nested fields have default). So there is no concept of implicit default. 
Think of it this way: The default value specified at any level should be a valid representation of the schema. So there should be an entry for all required fields at each level (either from parent or at the field level). No assumptions can be made about the nested levels.

junchuanwang (approved):

PR #586, opened 2021-04-02
Add assocKey each part key binding methods for fluent api

junchuanwang:
This will be applied on 

Get
Delete
Update
PartialUpdate

### DIFF
--- a/restli-tools/src/main/resources/apiVmTemplates/resource.vm
+++ b/restli-tools/src/main/resources/apiVmTemplates/resource.vm
@@ -186,6 +186,31 @@ public class ${spec.className}${class_name_suffix} implements ${spec.toImplement
   }
   #end
 
+  ## For Get, Delete, Update, PartialUpdates for association key, will generate additional methods
+  ##     to use each part key as parameters.
+  #if (${spec.getResource().hasAssociation()})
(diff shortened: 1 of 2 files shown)


restli-tools/src/main/resources/apiVmTemplates/rest.get.vm
```diff
@@ -39,6 +40,34 @@
...
+      return get(${spec.idName},
+          #**##optionalMethodCallArgs($method)##
+      );
+    }
+
+
+    #doc($method.schema.doc)
+    public CompletionStage<${spec.entityClassName}> get(
+$ScatteredKey
+        ExecutionGroup executionGroup#if($method.requiredParams.size() > 0),#end
```
karthikbalasub, line 61: Looks like this has dependency on execution group?

restli-tools/src/main/resources/apiVmTemplates/resource.vm
```diff
@@ -186,6 +186,31 @@ public class ${spec.className}${class_name_suffix} implements ${spec.toImplement
   }
   #end
 
+  ## For Get, Delete, Update, PartialUpdates for association key, will generate additional methods
```
karthikbalasub, line 189: Templates for delete, update etc should also be updated?

Issue #587, opened 2021-04-02
TimingKey Memory Leak


zizhong:
Currently, TimingKey is implemented with a concurrent hashmap. TimingKey currently only provides registerKey to add an entry to the hashmap.
Each HTTPClient will create several different filters. Each filter will trigger registerKey several times.
There is no mechanism to reclaim the memory after the filters are unreachable.

If an app keeps creates new HTTPClients, the issue becomes more obvious.

ChrisCarini:
The [linkedin/rest.li](https://github.com/linkedin/rest.li) repository is being archived - see https://github.com/linkedin/rest.li/issues/1178 for details.

Ahead of it's archiving, we are closing all issues and PRs.

PR #588, opened 2021-04-02
Fix TimingKey Memory Leak #587

zizhong:
Details can be found in https://github.com/linkedin/rest.li/issues/587.

### DIFF
--- a/r2-core/src/main/java/com/linkedin/r2/filter/TimedStreamFilter.java
+++ b/r2-core/src/main/java/com/linkedin/r2/filter/TimedStreamFilter.java
@@ -34,12 +34,13 @@ import static com.linkedin.r2.filter.TimedRestFilter.ON_RESPONSE_SUFFIX;
  *
  * @author Xialin Zhu
  */
-/* package private */ class TimedStreamFilter implements StreamFilter
+public class TimedStreamFilter implements StreamFilter
 {
   private final StreamFilter _streamFilter;
   private final TimingKey _onRequestTimingKey;
   private final TimingKey _onResponseTimingKey;
   private final TimingKey _onErrorTimingKey;
+  private boolean _shared;
 
   /**
    * Registers {@link TimingKey}s for {@link com.linkedin.r2.message.timing.TimingNameConstants#TIMED_STREAM_FILTER}.
@@ -60,6 +61,7 @@ import static com.linkedin.r2.filter.TimedRestFilter.ON_RESPONSE_SUFFIX;
         filterClassName, TimingImportance.LOW);
     _onErrorTimingKey = TimingKey.registerNewKey(timingKeyPrefix + ON_ERROR_SUFFIX + timingKeyPostfix,
         filterClassName, TimingImportance.LOW);
+    _shared = false;
   }
 
   @Override
@@ -90,5 +92,17 @@ import static com.linkedin.r2.filter.TimedRestFilter.ON_RESPONSE_SUFFIX;
   {
     TimingContextUtil.markTiming(requestContext, _onErrorTimingKey);
     _streamFilter.onStreamError(ex, requestContext, wireAttrs, new TimedNextFilter<>(_onErrorTimingKey, nextFilter));
+  }
+
+  public void setShared() {
+    _shared = true;
+  }
+
+  public void onShutdown() {
+    if (!_shared) {
+      TimingKey.unregisterKey(_onErrorTimingKey);
+      TimingKey.unregisterKey(_onRequestTimingKey);
+      TimingKey.unregisterKey(_onResponseTimingKey);
+    }
   }
 }
--- a/r2-core/src/main/java/com/linkedin/r2/filter/transport/FilterChainClient.java
+++ b/r2-core/src/main/java/com/linkedin/r2/filter/transport/FilterChainClient.java
@@ -17,10 +17,13 @@
 /* $Id$ */
 package com.linkedin.r2.filter.transport;
 
-
 import com.linkedin.common.callback.Callback;
 import com.linkedin.common.util.None;
 import com.linkedin.r2.filter.FilterChain;
+import com.linkedin.r2.filter.TimedRestFilter;
+import com.linkedin.r2.filter.TimedStreamFilter;
+import com.linkedin.r2.filter.message.rest.RestFilter;
+import com.linkedin.r2.filter.message.stream.StreamFilter;
 import com.linkedin.r2.message.RequestContext;
 import com.linkedin.r2.message.Response;
 import com.linkedin.r2.message.rest.RestRequest;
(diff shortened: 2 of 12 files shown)


zizhong:
@nizarm Please help to review when you have time! Thanks!

r2-core/src/main/java/com/linkedin/r2/filter/transport/FilterChainClient.java
```diff
@@ -46,15 +52,17 @@
...
   /**
    * Construct a new instance by composing the specified {@link TransportClient}
    * and {@link FilterChain}.
    *
    * @param client the {@link TransportClient} to be composed.
    * @param filters the {@link FilterChain} to be composed.
+   * @param sharedFilters the {@link FilterChain} can be used by other clients.
    */
-  public FilterChainClient(TransportClient client, FilterChain filters)
+  public FilterChainClient(TransportClient client, FilterChain filters, FilterChain sharedFilters)
```
nizarm, line 65: This is a backward-incompatible change, can we introduce a new overloaded constructor and deprecate the usage of the existing constructor?

r2-core/src/main/java/com/linkedin/r2/filter/TimedStreamFilter.java
```diff
@@ -34,7 +36,7 @@
  *
  * @author Xialin Zhu
  */
-/* package private */ class TimedStreamFilter implements StreamFilter
+public class TimedStreamFilter implements StreamFilter
```
nizarm, line 39: Curios, why did we made this public?

r2-core/src/main/java/com/linkedin/r2/message/timing/TimingKey.java
```diff
@@ -32,6 +35,7 @@
 public class TimingKey
 {
   private static final Map<String, TimingKey> _pool = new ConcurrentHashMap<>();
+  private static final ExecutorService _unregisterExecutor = Executors.newFixedThreadPool(1);
```
nizarm, line 38: Instead of creating a new thread pool at the `TimingKey` level, can we pass the executor to the `unregisterKey` function and leverage the standard executor passed around?

One option is to pass down the standard executor `_shutdownCallbackExecutor` from `HttpClientFactory`

r2-core/src/main/java/com/linkedin/r2/filter/transport/FilterChainClient.java
```diff
@@ -94,6 +103,27 @@ public void streamRequest(StreamRequest request,
...
+      .flatMap(Collection::stream)
+        .forEach(TimingKey::unregisterKey);
+
+    restFilters.stream()
+      .filter(filter -> !sharedRestFilters.contains(filter))
+      .filter(TimedRestFilter.class::isInstance)
+      .map(TimedRestFilter.class::cast)
+      .map(TimedRestFilter::getTimingKeyList)
+      .flatMap(Collection::stream)
+        .forEach(TimingKey::unregisterKey);
```
nizarm, line 126: I think the usage of sharedStreamFilter is an overKill here. Probably we just need some kind of flag on the existing filters to identify if it is a global filter vs a local filter.

r2-netty/src/main/java/com/linkedin/r2/transport/http/client/HttpClientFactory.java
```diff
@@ -1121,7 +1120,7 @@ private TransportClient getClient(Map<String, ? extends Object> properties,
     filters = filters.addLastRest(disruptFilter);
     filters = filters.addLast(disruptFilter);
 
-    client = new FilterChainClient(client, filters);
+    client = new FilterChainClient(client, filters, _filters);
```
nizarm, line 1123: As mentioned in the FilterChainClient file, instead of passing down the common _filters down, it would be more clear if we introduce some kind of flag in the existing filter to identify global vs local filters

nizarm (changes requested):

r2-netty/src/main/java/com/linkedin/r2/transport/http/client/HttpClientFactory.java
zizhong: Sure, we can.

r2-core/src/main/java/com/linkedin/r2/message/timing/TimingKey.java
zizhong: `_shutdownCallbackExecutor` is a boolean in HttpClientFactory.
TimingKey may be used more wildly than in `HttpClientFactory`, so that a separate thread provides better usability for other clients as well.
Any specific concerns against a new thread?

r2-core/src/main/java/com/linkedin/r2/filter/transport/FilterChainClient.java
zizhong: Fixed with a flag in the filter instead.

r2-core/src/main/java/com/linkedin/r2/filter/TimedStreamFilter.java
zizhong: We need to call the methods in another package.

r2-core/src/main/java/com/linkedin/r2/filter/transport/FilterChainClient.java
zizhong: Fixed

r2-core/src/main/java/com/linkedin/r2/message/timing/TimingKey.java
nizarm: If there are 100s of clients in service, that will end up creating 100s of threads. Having the executor run this make sure there is an upper bound on the total number of threads created.

Also, it is not applicable in this case - it is always encouraged to use executor so that other SI infra works as expected like IC context

r2-core/src/main/java/com/linkedin/r2/filter/transport/FilterChainClient.java
```diff
@@ -53,6 +59,7 @@
    *
    * @param client the {@link TransportClient} to be composed.
    * @param filters the {@link FilterChain} to be composed.
+   * @param sharedFilters the {@link FilterChain} can be used by other clients.
```
nizarm, line 62: This java doc can be removed now right?

nizarm (approved):
LGTM

r2-core/src/main/java/com/linkedin/r2/message/timing/TimingKey.java
zizhong: Hmm, the ConcurrentHashMap is static and shared between clients, so that I made the executor static and shared between clients.

zizhong:
@nizarm Thank you for the insightful suggestions and careful reviews!

PR #589, opened 2021-04-13
Set up CI publishing for JFrog Artifactory

evanw555:
This is part of the effort to migrate from Bintray to JFrog. This commit
does not remove the Bintray publishing step, it only sets up JFrog
publishing to a dummy test repo in parallel. A follow-up change will
remove the Bintray process altogether.


build_script/jfrog.gradle
```diff
@@ -0,0 +1,61 @@
...
+    }
+  }
+}
+
+// Utility task to ensure that we aren't attempting a publish without providing JFrog credentials.
+task assertJFrogCredentialsExist() {
+  doLast {
+    if (!jfrogUser || !jfrogKey) {
+      throw new GradleException(
+          "Cannot perform JFrog upload. Missing '${jfrogUserEnv}' or '${jfrogKeyEnv}' environment variable.")
```
BrianPin, line 55: When printed, will we see `${jfrogUserEnv}` or `JFROG_USER` or <<...empty...>>?
And if this message does come up, how do we get to the environment and set these variables?

scripts/travis/publish-tag.sh
```diff
@@ -65,8 +65,18 @@ fi
 while sleep 9m; do echo "[Ping] Keeping Travis job alive ($((SECONDS / 60)) minutes)"; done &
 WAITER_PID=$!
 
-# Build and publish to Bintray
-echo "All checks passed, attempting to publish Rest.li $VERSION to Bintray..."
+# TODO: This just publishes to the test-repo for now, delete the Bintray publish step once no longer testing
+# Publish to JFrog Artifactory
+echo "All checks passed, attempting to publish Rest.li $VERSION to JFrog Artifactory..."
+./gradlew artifactoryPublish
```
BrianPin, line 71: So looks like there is a task publishing to bintray and this new task publish to jfrog, are we going to have 2 for temp and eventually publish to jfrog (sounds like a migration process)

BrianPin (approved):
LGTM

build_script/jfrog.gradle
evanw555: This just prints the name of the variable (e.g. `JFROG_USER`). This can be resolved via `export JFROG_USER=foo` in bash before running the Gradle task. However, this is normally done in CI so the env variables are set in the Travis config.

scripts/travis/publish-tag.sh
evanw555: Yeah, after this change, the process will be (1) publish to Bintray as usual and (2) publish to a test repo in JFrog.
I will do a follow-up change, after which the process will be (1) don't publish to Bintray at all and (2) publish to JFrog.

The reason for doing this in 2 steps is to confirm that the process works before cutting off Bintray (since that would block our release process).

build_script/jfrog.gradle
karthikbalasub: I believe the user/key is configured in Travis for all projects under linkedin. Is that right? Might be a good idea to add a comment here to explain how they are configured.

karthikbalasub (approved):

evanw555:
> I believe the user/key is configured in Travis for all projects under linkedin. Is that right? Might be a good idea to add a comment here to explain how they are configured.

@karthikbalasub Actually no, the current best practice is to just use personal credentials to publish to JFrog. This is currently set in the Travis config, not Travis itself. I can update the message though

PR #593, opened 2021-04-15
Remove deprecated HttpClientFactory constructors.

tjni:
Also moved some of the logic in the build() method into the constructor
so that we can apply the standard pattern of passing the builder into a
private constructor instead of each individual parameter.

I noticed this while starting to work on shading Netty and think it would be
a good idea to bundle this backwards incompatible change with the next
major version.


evanw555:
Someone from SI-Core should take a look at this (GitHub suggests @nizarm ?), but yeah if it's approved then we'll bundle it with the next major release.

tjni:
Doing some cleaning up of my stale PRs so closing this.

PR #594, opened 2021-04-15
Try to tune parameter in tests so it can be stable in Travis-CI's environment

junchuanwang:
It became flaky in Travis-CI because Travis-CI provides so few threads in the common fork-join pool.


evanw555 (approved):
Thanks for addressing this so fast

PR #595, opened 2021-04-16
Shade Netty.

tjni:
Per internal discussions, we'd like to shade Netty usages in rest.li,
so this change relocates io.netty to com.linkedin.pegasus.io.netty.

A new module named pegasus-shaded-netty is introduced which produces
an artifact in the 'shadow' configuration that includes the Netty code
relocated, and all references to io.netty in rest.li are changed.

It was necessary to create a new module instead of shading r2-netty
directly because there does not appear to be a straightforward way in
Gradle that I could find to replace the main artifact published with the
shaded artifact. The best I could figure out is how to generate the
artifact on a custom configuration, so I think it is a good idea to hide
that detail as a dependency of r2-netty.

Because these shaded classes are exposed in the public API of r2-netty,
this is a backwards incompatible change.


r2-netty/build.gradle
```diff
@@ -1,12 +1,11 @@
 dependencies {
-  compile project(':pegasus-common')
   compile project(':data')
+  compile project(':pegasus-common')
+  compile project(path: ':pegasus-shaded-netty', configuration: 'shadow')
```
evanw555, line 4: Relying on special configurations from other projects is ok for test sources but I believe this will cause problems for main sources. Rest.li is published as a Maven (.pom) publication which doesn't support special configurations. For any consumers of `r2-netty`, the transitive dependency on `pegasus-shaded-netty` will likely be broken.

settings.gradle
```diff
@@ -19,6 +19,7 @@ include 'degrader'
 include 'entity-stream'
 include 'li-jersey-uri'
 include 'li-protobuf'
+include 'pegasus-shaded-netty'
```
evanw555, line 22: I suggest renaming this subproject to `li-netty` for consistency (since this is close to what is done for `li-jersey-uri` and `li-protobuf` though without the shading)

build.gradle
```diff
@@ -40,6 +40,10 @@ File getEnvironmentScript()
 apply from: environmentScript
 apply from: "${buildScriptDirPath}/configBuildScript.gradle"
 
+def versions = [
```
evanw555, line 43: Could you do us a favor and group-lock other dependencies while you're at it? (e.g. spring, parseq, guice)

pegasus-shaded-netty/build.gradle
```diff
@@ -0,0 +1,17 @@
...
+  implementation externalDependency.nettyCodec
+  implementation externalDependency.nettyCodecHttp
+  implementation externalDependency.nettyCodecHttp2
+  implementation externalDependency.nettyCommon
+  implementation externalDependency.nettyHandler
+  implementation externalDependency.nettyTransport
+}
+
+tasks.named('shadowJar') {
+  relocate 'io.netty', 'com.linkedin.pegasus.io.netty'
```
FranklinYinanDing, line 16: I'm wondering if Netty client uses any file under `META-INF`. I know Netty server does. If the case is the same, more relocate is needed for those files.

settings.gradle
zizhong: +1

build.gradle
```diff
@@ -40,6 +40,10 @@ File getEnvironmentScript()
...
+  'netty': '4.1.58.Final'
```
zizhong, line 44: From https://netty.io/downloads.html, netty-4.1.63.Final is recommended.

r2-netty/build.gradle
zizhong: Interesting, if true, we may need an alternative solution, @tjni Is there a way to test this part?

pegasus-shaded-netty/build.gradle
tjni: Great catch, Yinan. I'm not certain whether the following files under META-INF are used:

1. MANIFEST.MF
2. INDEX.LIST
3. io.netty.versions.properties
4. maven/io.netty/netty-all/pom.xml
5. maven/io.netty/netty-all/pom.properties

However, none of them look necessary, so I would be comfortable excluding them and only looking into them if an issue comes up (your mileage may vary).

There are two other sets of files. One set of files are only used by GraalVM's native-image tool and can be safely excluded.

The last set of files are native libraries for Linux and OSX:

1. native/libnetty_resolver_dns_native_macos_x86_64.jnilib
2. native/libnetty_transport_native_kqueue_x86_64.jnilib
3. native/libnetty_transport_native_epoll_x86_64.so
4. native/libnetty_transport_native_epoll_aarch_64.so

The last three are special libraries to either optimize or provide more features in the Netty transport layer (kqueue for OSX and epoll for Linux). To use them, the code must use classes that are prefixed with KQueue or Epoll, which rest.li does not. So these can be safely excluded.

The first one seems to be a better DNS resolver for OSX. I am not certain under what conditions it is used, but I can include the dependency explicitly and relocate it (Netty supports loading native libraries that are relocated).

FranklinYinanDing: Thanks for looking into it! 👍

r2-netty/build.gradle
tjni: I updated `publications.gradle` to upload the shaded artifact from li-netty. The POM that I inspected looks reasonable to me, but I'm not an expert. If either of you have suggestions for good steps to test that the artifact can be resolved, please let me know. Please feel free also to pull these changes down locally and inspect if you have some experience here.

settings.gradle
tjni: Good idea. I've renamed it to li-netty.

build.gradle
tjni: Absolutely, done!

pegasus-shaded-netty/build.gradle
```diff
@@ -0,0 +1,17 @@
+plugins {
```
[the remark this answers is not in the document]
tjni: As I was doing this, I realized that the build logic needed to properly shade was already a bit subtle and complicated, so I felt it was better to isolate it to its own subproject, which only has the single purpose of shading Netty properly. Let me know though if you feel strongly about this, though, because with some additional work it should be technically possible to shade this directly into the r2-netty subproject.

pegasus-shaded-netty/build.gradle
tjni: I have added the appropriate native library dependencies and relocation rules.

build.gradle
tjni: Done!

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.2.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.4-bin.zip
```
aman1309, line 3: is this necessary? we don't wanna be using new features which are incompatible with lower versions.

tjni: I did this initially because I couldn't get the shadow plugin working and didn't know if it was because of issues between Gradle 5.2.1 and 5.6.4, and I didn't bother spending the time to try it again with the older Gradle version. Since this seems important to you, I'll downgrade Gradle and run it through the CI again.

r2-netty/build.gradle
zizhong: I also checked locally and the release looks good.

zizhong (approved):
Thanks for taking care of this!

zizhong (changes requested):
Please try on a few MPs to verify the correctness as some folks concern due to the large code changes.

tjni:
Doing some cleaning up of my stale PRs so closing this.

PR #597, opened 2021-04-16
Use runtimeClasspath instead of runtime config.

tjni:
This is needed for compatibility with Gradle 7, which removes the long
deprecated "runtime" and "compile" configurations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.18.4] - 2021-05-06
+- Replace runtime configuration with runtimeClasspath configuration in plugin for compatibility with Gradle 7.
+
 ## [29.18.3] - 2021-05-03
 - Strictly enforce Gradle version compatibility in the `pegasus` Gradle plugin.
   - Minimum required Gradle version is now `1.0` (effectively backward-compatible).
@@ -4928,7 +4931,8 @@ patch operations can re-use these classes for generating patch messages.
 
 ## [0.14.1]
 
-[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.18.3...master
+[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.18.4...master
+[29.18.4]: https://github.com/linkedin/rest.li/compare/v29.18.3...v29.18.4
 [29.18.3]: https://github.com/linkedin/rest.li/compare/v29.18.2...v29.18.3
 [29.18.2]: https://github.com/linkedin/rest.li/compare/v29.18.1...v29.18.2
 [29.18.1]: https://github.com/linkedin/rest.li/compare/v29.18.0...v29.18.1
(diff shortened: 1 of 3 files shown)


gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -900,7 +899,7 @@ protected void configureSchemaAnnotationValidation(Project project,
                   task.setResolverPath(getDataModelConfig(project, sourceSet)); // same resolver path as generateDataTemplatesTask
                   task.setClassPath(project.getConfigurations() .getByName(SCHEMA_ANNOTATION_HANDLER_CONFIGURATION)
                                            .plus(project.getConfigurations().getByName(PEGASUS_PLUGIN_CONFIGURATION))
-                                           .plus(project.getConfigurations().getByName(JavaPlatformPlugin.RUNTIME_CONFIGURATION_NAME)));
+                                           .plus(project.getConfigurations().getByName(JavaPlugin.RUNTIME_CLASSPATH_CONFIGURATION_NAME)));
```
BrianPin, line 902: Hi Teddy, have a question, this from behind the scene change to use "runtime" to "runtimeClasspath" right?
Are they getting the same configuration under the hood?
I just to know if it will change anything.

tjni: In theory, it is not exactly the same. The `runtimeClasspath` configuration extends from the `runtime` configuration, but it also does from the newer `runtimeOnly` and `implementation` configurations. This is documented [here](https://docs.gradle.org/6.8.3/userguide/java_plugin.html#tab:configurations).

In practice, I don't expect it to change anything because very few people internally use `runtimeOnly` and `implementation`, and because `runtimeClasspath` is strictly larger than `runtime`, and enlarging the classpath in my experience has a lower chance of leading to an issue that shrinking the classpath.

CHANGELOG.md
```diff
@@ -4921,10 +4924,21 @@ patch operations can re-use these classes for generating patch messages.
...
+[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.18.1...master
+>>>>>>> 5cd01279a (Use runtimeClasspath instead of runtime config.)
 [29.18.1]: https://github.com/linkedin/rest.li/compare/v29.18.0...v29.18.1
 [29.18.0]: https://github.com/linkedin/rest.li/compare/v29.17.4...v29.18.0
+=======
+[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.18.3...master
+[29.18.3]: https://github.com/linkedin/rest.li/compare/v29.18.2...v29.18.3
+[29.18.2]: https://github.com/linkedin/rest.li/compare/v29.17.5...v29.18.2
+[29.17.5]: https://github.com/linkedin/rest.li/compare/v29.17.4...v29.17.5
+>>>>>>> e4749f625 (Use runtimeClasspath instead of runtime config.)
```
nickibi, line 4941: Looks like there are some conflicts which need to be fixed.

tjni: Good catch! Fixed.

nickibi (approved):

BrianPin (approved):

PR #607, opened 2021-04-26
Add LI JFrog Artifactory to buildscript repos

evanw555:
Bintray is doing intentional service blackouts right now in order to get
the attention of use cases that are still reading artifacts from
Bintray (of which pegasus is one...). This adds LI's JFrog Artifactory
instance as a maven repo and bumps parseq to the version that's
currently hosted on JFrog. The Bintray address can be removed once
that's down for good, just to be sure.


junchuanwang (approved):
thanks for addressing this

karthikbalasub (approved):

PR #608, opened 2021-04-27
Add debug log for cluster subsetting updates

rickzx:


d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerState.java
```diff
@@ -715,6 +716,15 @@ public void setDelayedExecution(long delayedExecution)
       _weightedSubsetsCache.computeIfAbsent(serviceName, k -> new ConcurrentHashMap<>());
       _weightedSubsetsCache.get(serviceName).put(partitionId, subsetClients);
 
+      debug(_log, "cluster subset updated for service ", serviceName, ": [",
+          subsetClients.values().stream()
+            .limit(LOG_SUBSET_MAX_SIZE)
+            .map(client -> client.getUri() + ":" + client.getSubsetWeight(partitionId))
+            .collect(Collectors.joining(",")),
+          (subsetClients.size() > LOG_SUBSET_MAX_SIZE ? "...(total " + subsetClients.size() + ")" : ""),
```
rachelhanhan, line 724: Should we alway print the subset size? Otherwise if it prints fewer than 20 clients, we have to manually count the client in order to understand the subsetting size.

d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerState.java
```diff
@@ -679,6 +679,7 @@ public void setDelayedExecution(long delayedExecution)
                                                   Map<URI, TrackerClient> potentialClients)
   {
     SubsettingStrategy<URI> subsettingStrategy = _subsettingStrategyFactory.get(serviceName, minClusterSubsetSize, partitionId);
+    final int LOG_SUBSET_MAX_SIZE = 20;
```
rachelhanhan, line 682: This can be static.

BTW how do we decide the size of 20, if the default subset size is 10, I would imagine something around 10 would cover many use cases.

d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerState.java
rickzx: Good point. I think always print out the subset size makes sense

d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerState.java
rickzx: Changed to static. 

Since this is a debug level log, leaving a little room may be useful for experimenting and tuning this config. We don't expect this to be in the actual production log.

rachelhanhan (approved):

PR #611, opened 2021-04-28
Fix bug in generated fluent client APIs when typerefs are used as association key params

karthikbalasub:
Fix bug in generated fluent client APIs when typerefs are used as association key params


junchuanwang (approved):
Thanks for the fix

PR #612, opened 2021-04-28
Add support for CONSTANT_QPS dark canary cluster strategy

lesterhaynes:
This implementation of the CONSTANT_QPS dark canary cluster strategy leverages a
circular buffer and rate limiter to produce a guaranteed rate of traffic matching
the configured value.


darkcluster/src/main/java/com/linkedin/darkcluster/util/ExpiringCircularBuffer.java
```diff
@@ -0,0 +1,143 @@
...
+    bumpWriterPosition();
+    return true;
+  }
+
+  /**
+   * TODO: javadocs
+   * @return
+   */
+  public synchronized E element() {
+    LOGGER.info("Trying a get. Reader position is {}", _readerPosition);
```
lesterhaynes, line 62: verbose logging like this will be removed in actual release candidate

darkcluster/src/main/java/com/linkedin/darkcluster/impl/ConstantQpsDarkClusterStrategy.java
```diff
@@ -0,0 +1,132 @@
...
+    _baseDarkClusterDispatcher = baseDarkClusterDispatcher;
+    _notifier = notifier;
+    _clusterInfoProvider = clusterInfoProvider;
+    _rateLimiter = rateLimiter;
+  }
+
+  @Override
+  public boolean handleRequest(RestRequest originalRequest, RestRequest darkRequest, RequestContext requestContext) {
+    float sendRate = getPerHostSendRate();
+    _rateLimiter.setRate(sendRate, 1000, 1);
```
rickzx, line 65: Let's avoid using magic numbers like 1000 and 1 here. For 1000, it'd better to use something like `private static final long ONE_SECOND_PERIOD = TimeUnit.SECONDS.toMillis(1);`. For the burst parameter, I think we might want to use unlimited burst here, something like, `private static final int UNLIMITED_BURST = Integer.MAX_VALUE;`

darkcluster/src/main/java/com/linkedin/darkcluster/impl/ConstantQpsDarkClusterStrategy.java
```diff
@@ -0,0 +1,132 @@
...
+  {
+    _rateLimiter.submit(new Callback<None>() {
+      @Override
+      public void onError(Throwable e) {
+        //
+      }
+
+      @Override
+      public void onSuccess(None result) {
+        _baseDarkClusterDispatcher.sendRequest(originalRequest, darkRequest, requestContext, 1);
```
rickzx, line 127: Ditto for magic number here

darkcluster/src/main/java/com/linkedin/darkcluster/impl/ConstantQpsDarkClusterStrategy.java
```diff
@@ -0,0 +1,132 @@
...
+  private final String _originalClusterName;
+  private final String _darkClusterName;
+  private final Integer _darkClusterPerHostInboundTargetRate;
+  private final BaseDarkClusterDispatcher _baseDarkClusterDispatcher;
+  private final Notifier _notifier;
+  private final ClusterInfoProvider _clusterInfoProvider;
+  private final GuaranteedRateLimiter _rateLimiter;
+
+  public ConstantQpsDarkClusterStrategy(@Nonnull String originalClusterName, @Nonnull String darkClusterName,
+      @Nonnull Integer darkClusterPerHostInboundTargetRate, @Nonnull BaseDarkClusterDispatcher baseDarkClusterDispatcher,
```
rickzx, line 51: Minor: Do you think darkClusterPerHostQPS is a more informative name than inboundTargetRate?

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
```diff
@@ -63,13 +71,19 @@
...
   public DarkClusterStrategyFactoryImpl(@Nonnull Facilities facilities,
                                         @Nonnull String sourceClusterName,
                                         @Nonnull DarkClusterDispatcher darkClusterDispatcher,
                                         @Nonnull Notifier notifier,
                                         @Nonnull Random random,
-                                        @Nonnull DarkClusterVerifierManager verifierManager)
+                                        @Nonnull DarkClusterVerifierManager verifierManager,
+                                        @Nonnull ScheduledExecutorService rateLimiterScheduler,
+                                        @Nonnull ExecutorService rateLimiterExecutor,
+                                        @Nonnull ExpiringCircularBuffer<Callback<None>> rateLimiterStoredCallbacks)
```
rickzx, line 86: Instead of passing all three parameters to construct a RateLimiter, can we simply pass in a RateLimiter instance instead?

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
```diff
@@ -122,34 +139,63 @@ private DarkClusterStrategy createStrategy(String darkClusterName, DarkClusterCo
       DarkClusterStrategyNameArray strategyList = darkClusterConfig.getDarkClusterStrategyPrioritizedList();
       for (com.linkedin.d2.DarkClusterStrategyName darkClusterStrategyName : strategyList)
       {
-        switch(darkClusterStrategyName)
```
rickzx, line 125: Let's make sure to restore the switch() statements for public release

darkcluster/src/main/java/com/linkedin/darkcluster/util/ExpiringCircularBuffer.java
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  public synchronized void setCapacity(int capacity) {
+    if (capacity < 1) {
+      throw new IllegalArgumentException("capacity can't be less than 1.");
+    }
+    super.clear();
+    super.addAll(Collections.nCopies(capacity, null));
+    _ttlBuffer = new ArrayList<>(Collections.nCopies(capacity, null));
+  }
```
rickzx, line 131: Since the buffer does not allow dynamic resizing, do we really need a setCapacity() public method? I think it suffices to initialize an ArrayList in the constructor.

darkcluster/src/main/java/com/linkedin/darkcluster/util/ExpiringCircularBuffer.java
```diff
@@ -0,0 +1,143 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * TODO: javadocs
+ * @param <E>
+ */
+@SuppressWarnings("serial")
+public class ExpiringCircularBuffer<E> extends ArrayList<E> implements Queue<E> {
```
rickzx, line 37: IMO, since this is a util class designed primarily for dark cluster use case, it might be better to make this class self-contained so we don't have to implement a lot of unused/unsupported methods.

We probably only need add(), peek(), isFull() and size() for this class.

darkcluster/src/main/java/com/linkedin/darkcluster/util/ExpiringCircularBuffer.java
```diff
@@ -0,0 +1,143 @@
...
+    _ttlBuffer.set(_writerPosition, Instant.now());
+    bumpWriterPosition();
+    return true;
+  }
+
+  /**
+   * TODO: javadocs
+   * @return
+   */
+  public synchronized E element() {
```
rickzx, line 61: Can we have a more granularized synchronization control using a private lock object, instead of method-level synchronization? You can take a look at: https://www.baeldung.com/java-synchronization-bad-practices. I would imagine for most parts of the read operation, we don't need to obtain the lock.

darkcluster/src/main/java/com/linkedin/darkcluster/util/ExpiringCircularBuffer.java
```diff
@@ -0,0 +1,143 @@
...
+  public synchronized void setCapacity(int capacity) {
+    if (capacity < 1) {
+      throw new IllegalArgumentException("capacity can't be less than 1.");
+    }
+    super.clear();
+    super.addAll(Collections.nCopies(capacity, null));
+    _ttlBuffer = new ArrayList<>(Collections.nCopies(capacity, null));
+  }
+
+  public void setTtl(int ttl, ChronoUnit ttlUnit)
```
rickzx, line 133: Ditto. Maybe just set the ttl in the constructor.

darkcluster/src/main/java/com/linkedin/darkcluster/util/ExpiringCircularBuffer.java
```diff
@@ -0,0 +1,143 @@
...
+ */
+@SuppressWarnings("serial")
+public class ExpiringCircularBuffer<E> extends ArrayList<E> implements Queue<E> {
+  private static final Logger LOGGER = LoggerFactory.getLogger(ExpiringCircularBuffer.class);
+  private Duration _ttl;
+  private ArrayList<Instant> _ttlBuffer;
+  private volatile int _readerPosition;
+  private volatile int _writerPosition;
+
+  public ExpiringCircularBuffer(int capacity, int ttl, ChronoUnit ttlUnit) {
```
rickzx, line 44: The name sounds a little bit confusing to me. Maybe `EvictingCircularBuffer` or `CircularBufferWithTimedExpiry`

darkcluster/src/main/java/com/linkedin/darkcluster/util/ExpiringCircularBuffer.java
```diff
@@ -0,0 +1,143 @@
...
+   */
+  public synchronized E element() {
+    LOGGER.info("Trying a get. Reader position is {}", _readerPosition);
+    E e = this.get(_readerPosition);
+    Instant ttl = _ttlBuffer.get(_readerPosition);
+    if (e == null) {
+      if (_readerPosition == 0) {
+        throw new NoSuchElementException("Buffer is empty");
+      } else {
+        LOGGER.info("Hit a hole in the buffer, bump readerPosition and retry.");
```
rickzx, line 69: There are a lot of info level logs here. We don't really want to expose this level of detailed information to the user. If this is just for debugging purpose, let's remove it for the public release.

darkcluster/src/main/java/com/linkedin/darkcluster/util/GuaranteedRateLimiter.java
```diff
@@ -0,0 +1,279 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+import javax.annotation.Nonnull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * A simple {@link AsyncRateLimiter} implementation that re-issues permits at every specified period of time.
+ * A submitted callback's #onError is invoked with {@link RejectedExecutionException} if the current buffered
+ * callbacks exceeds to the maximum allowed by the implementation. Based heavily on SmoothRateLimiter by Sean Sheng.
```
rickzx, line 39: Let's also add a @link to the SmoothRateLimiter. Not needed to mention the author here IMO.

darkcluster/src/main/java/com/linkedin/darkcluster/util/GuaranteedRateLimiter.java
```diff
@@ -0,0 +1,279 @@
...
+ *
+ * @author Lester Haynes
+ * @author Sean Sheng
+ * @author Francesco Capponi (fcapponi@linkedin.com)
+ */
+public class GuaranteedRateLimiter implements AsyncRateLimiter
```
rickzx, line 45: Would it be possible to extend the SmoothRateLimiter to avoid a lot of code duplication? Or maybe pull out the EventLoop class from the inner class, so both SmoothRateLimiter and GuaranteedRateLimiter can use it.

darkcluster/src/main/java/com/linkedin/darkcluster/util/GuaranteedRateLimiter.java
```diff
@@ -0,0 +1,279 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+import javax.annotation.Nonnull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * A simple {@link AsyncRateLimiter} implementation that re-issues permits at every specified period of time.
```
rickzx, line 37: Probably also mention its behavior: elements are not removed after read, and it uses an underlying circular buffer.

rickzx (changes requested):
Thanks for the great work! Haven't went through all the files yet. Left some initial reviews.

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
lesterhaynes: It's passed this way in order to give access to setCapacity and setTtl on _rateLimiterStoredCallbacks. Could probably pass just the GuaranteedRateLimiter if I added public methods to GuaranteedRateLimiter that forward to the underlying setCapacity and setTtl methods on GuaranteedRateLimiter's ExpiringCircularBuffer. I'll draft that up

darkcluster/src/main/java/com/linkedin/darkcluster/impl/ConstantQpsDarkClusterStrategy.java
lesterhaynes: sounds good

darkcluster/src/main/java/com/linkedin/darkcluster/impl/ConstantQpsDarkClusterStrategy.java
lesterhaynes: +1

darkcluster/src/main/java/com/linkedin/darkcluster/util/ExpiringCircularBuffer.java
lesterhaynes: This is tricky. We need to be able to allow upstream users the ability to supply their own buffer (LI has its own reasons for wanting to do this for example). After it's supplied though, there exists the need to change the size of that buffer through configuration while the application is running. If the upstream users didn't need to supply their own implementation, we could just re-instantiate a rest.li-only component here with updated constructor params.

darkcluster/src/main/java/com/linkedin/darkcluster/util/GuaranteedRateLimiter.java
lesterhaynes: Agreed, will want to deduplicate a lot of this once the overall design is approved

darkcluster/src/main/java/com/linkedin/darkcluster/util/ExpiringCircularBuffer.java
lesterhaynes: +1

darkcluster/src/main/java/com/linkedin/darkcluster/util/GuaranteedRateLimiter.java
lesterhaynes: Sounds good

darkcluster/src/main/java/com/linkedin/darkcluster/util/GuaranteedRateLimiter.java
lesterhaynes: Good idea

darkcluster/src/main/java/com/linkedin/darkcluster/util/ExpiringCircularBuffer.java
lesterhaynes: Evicting is probably better than Expiring. We'll go with EvictingCircularBuffer

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
lesterhaynes: +1

darkcluster/src/main/java/com/linkedin/darkcluster/util/ExpiringCircularBuffer.java
lesterhaynes: Definitely merits to that. I think it depends on how we end up deduplicating code between SmoothRateLimiter and GuaranteedRateLimiter. Thinking it may be easier to do that if we extend from a common superclass here.

darkcluster/src/main/java/com/linkedin/darkcluster/impl/ConstantQpsDarkClusterStrategy.java
rickzx: Actually, on a second thought, we might want to spread out the requests during the 1 second. Given how burst is calculated (https://github.com/linkedin/rest.li/blob/7372bc76f7ce78007ecb24ca80f17399e5246e13/r2-core/src/main/java/com/linkedin/r2/transport/http/client/ratelimiter/Rate.java#L41), we might want to use the formula `burst = Math.ceil(sendRate / 1000)`. So for example, when QPS is < 1000, burst = 1. So it will be able to spread each request evenly during the 1 second. But when QPS is 2000, burst = 2, meaning that 2 requests will be issued every 1 ms.

darkcluster/src/main/java/com/linkedin/darkcluster/util/ExpiringCircularBuffer.java
rickzx: The other benefit of not extending ArrayList is that we can have a more granular concurrency control. Currently, we are relying on the parent class implementation of the data structure, so we have to put synchronized key word everywhere.

rickzx (changes requested):

darkcluster/src/main/java/com/linkedin/darkcluster/util/ExpiringCircularBuffer.java
lesterhaynes: Took your advice here. Next draft will see EvictingCircularBuffer implement a new interface that describes put() and get() methods only, since that's all we use in SmoothRateLimiter as is.

darkcluster/src/main/java/com/linkedin/darkcluster/impl/ConstantQpsDarkClusterStrategy.java
lesterhaynes: Yeah, I like darkClusterPerHostQps better

darkcluster/src/main/java/com/linkedin/darkcluster/impl/ConstantQpsDarkClusterStrategy.java
lesterhaynes: Next draft will include this

darkcluster/src/main/java/com/linkedin/darkcluster/impl/ConstantQpsDarkClusterStrategy.java
lesterhaynes: private static final int NUM_REQUESTS_TO_SEND_PER_RATE_LIMITER_CYCLE = 1;

darkcluster/src/main/java/com/linkedin/darkcluster/util/ExpiringCircularBuffer.java
lesterhaynes: Yeah, I can do better ;)

darkcluster/src/main/java/com/linkedin/darkcluster/util/ExpiringCircularBuffer.java
```diff
@@ -0,0 +1,143 @@
...
+import java.util.NoSuchElementException;
+import java.util.Objects;
+import java.util.Queue;
+import org.apache.commons.lang3.NotImplementedException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * TODO: javadocs
```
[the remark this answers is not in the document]
lesterhaynes: done

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
```diff
@@ -63,13 +66,15 @@
...
+  private final ConstantQpsRateLimiter _rateLimiter;
 
   public DarkClusterStrategyFactoryImpl(@Nonnull Facilities facilities,
                                         @Nonnull String sourceClusterName,
                                         @Nonnull DarkClusterDispatcher darkClusterDispatcher,
                                         @Nonnull Notifier notifier,
                                         @Nonnull Random random,
-                                        @Nonnull DarkClusterVerifierManager verifierManager)
+                                        @Nonnull DarkClusterVerifierManager verifierManager,
+                                        @Nonnull ConstantQpsRateLimiter rateLimiter)
```
rickzx, line 77: It's better to make the constructor backward compatible, so other external MP's based on this library won't break. One option is to keep the old constructor, and call `this(facilities, ..., verifierManager, null)`. And when rateLimiter is `null`, we don't need to support constant QPS strategy

rickzx (commented):
First pass. Will need tot take a look at the SmoothRateLimiter later

d2-schemas/src/main/pegasus/com/linkedin/d2/DarkClusterConfig.pdl
```diff
@@ -11,14 +11,19 @@ record DarkClusterConfig {
...
-   * Max rate dispatcher can send to dark canary. Measured in qps. Will act as upper bound to protect canaries in case of traffic spikes
+   * Number of requests to store in the circular buffer used for asynchronous dispatching by the CONSTANT_QPS strategy.
    */
-  dispatcherOutboundMaxRate: int = 2147483647
+  dispatcherMaxRequestsToBuffer: int = 1
+
+  /**
+   * Amount of time in seconds that a request is eligible for asynchronous dispatch once it is added to the circular buffer by the CONSTANT_QPS strategy.
+   */
+  dispatcherOldestRequestAge: int = 1
```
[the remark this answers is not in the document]
lesterhaynes: Works for me.

d2-schemas/src/main/pegasus/com/linkedin/d2/DarkClusterStrategyName.pdl
```diff
@@ -16,7 +16,11 @@ enum DarkClusterStrategyName {
   RELATIVE_TRAFFIC
 
   /**
-   * This strategy will try to maintain a certain queries per second to the entire dark cluster. Configured with "dispatcherOutboundTargetRate".
+   * This strategy will maintain a certain number of queries per second to the entire dark cluster. It does so by adding every inbound
+   * request to a circular buffer which is consumed asynchronously by a rate-limited event loop. Configured with:
+   *   "dispatcherOutboundTargetRate": the target rate which each dark cluster host should receive.
+   *   "dispatcherMaxRequestsToBuffer": the number of requests to store in the circular buffer.
+   *   "dispatcherOldestRequestAge": time in seconds that a request is eligible for dispatch once it is added to the circular buffer.
```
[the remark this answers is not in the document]
lesterhaynes: good!

darkcluster/src/main/java/com/linkedin/darkcluster/impl/ConstantQpsDarkClusterStrategy.java
```diff
@@ -0,0 +1,162 @@
...
+ * and uses that to calculate the number of requests to send in order to make the QPS per dark cluster host constant and equal
+ * to a specified value, assuming all hosts in the source cluster send traffic.
+ *
+ * This strategy differs from the RELATIVE_TRAFFIC and IDENTICAL_TRAFFIC strategies in that requests are dispatched by a
+ * rate-limited event loop after being stored in a circular buffer. This provides a steady stream of outbound traffic that
+ * only duplicates requests when the inbound rate of traffic is less than the outbound rate. With the other strategies,
+ * requests are randomly selected based on a multiplier. With this strategy, all requests are submitted to the rate-limiter,
+ * which dispatches and evicts stored requests based on its configuration.
+ */
+public class ConstantQpsDarkClusterStrategy implements DarkClusterStrategy {
```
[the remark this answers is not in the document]
lesterhaynes: Trying to be consistent with this. Let me know if you find any more.

darkcluster/src/main/java/com/linkedin/darkcluster/impl/ConstantQpsDarkClusterStrategy.java
```diff
@@ -0,0 +1,162 @@
...
+   * It uses the {@link ClusterInfoProvider} to make the following calculation:
+   *
+   * RequestsPerSecond = ((# instances in dark cluster) * darkClusterPerHostQps) / (# instances in source cluster)
+   *
+   * For example, if there are 2 dark instances, and 10 instances in the source cluster, with a darkClusterPerHostQps of 50, we get:
+   * RequestsPerSecond = (2 * 50)/10 = 10.
+   *
+   * another example:
+   * 1 dark instance, 7 source instances, darkClusterPerHostQps = 75.
+   * #equestsPerSecond = (1 * 75)/7 = 10.71429.
```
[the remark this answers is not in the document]
lesterhaynes: good

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
```diff
@@ -51,6 +53,7 @@
 {
   private static final Logger LOG = LoggerFactory.getLogger(DarkClusterStrategyFactoryImpl.class);
   public static final DarkClusterStrategy NO_OP_DARK_CLUSTER_STRATEGY = new NoOpDarkClusterStrategy();
+  private static final String RATE_LIMITER_NAME = "DarkClusterRateLimiter";
```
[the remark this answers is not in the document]
lesterhaynes: Absolutely

r2-core/src/main/java/com/linkedin/r2/transport/http/client/ConstantQpsRateLimiter.java
```diff
@@ -0,0 +1,100 @@
...
+ * demands a constant rate of callback execution, and it's not important that all callbacks are executed, or executed only once.
+ *
+ * Rest.li's original use case for this rate-limiter is to supply Dark Clusters with a steady stream of request volume for
+ * testing purposes under a given load.
+ */
+public class ConstantQpsRateLimiter extends SmoothRateLimiter {
+  private final EvictingCircularBuffer _evictingCircularBuffer;
+
+  public ConstantQpsRateLimiter(
+      ScheduledExecutorService scheduler, Executor executor, Clock clock, EvictingCircularBuffer storedCallbacks)
```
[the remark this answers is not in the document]
lesterhaynes: Don't see why not

r2-core/src/main/java/com/linkedin/r2/transport/http/client/EvictingCircularBuffer.java
```diff
@@ -0,0 +1,245 @@
...
+  private final AtomicInteger _writerPosition = new AtomicInteger();
+  private final Clock _clock;
+
+
+  public EvictingCircularBuffer(int capacity, int ttl, ChronoUnit ttlUnit)
+  {
+    this(capacity, ttl, ttlUnit, SystemClock.instance());
+  }
+
+  EvictingCircularBuffer(int capacity, int ttl, ChronoUnit ttlUnit, Clock clock)
```
rickzx, line 65: Also make this public? Can we add a javadoc here?

r2-core/src/main/java/com/linkedin/r2/transport/http/client/EvictingCircularBuffer.java
```diff
@@ -0,0 +1,245 @@
...
+      ttl = _ttlBuffer.get(thisReaderPosition);
+    }
+    finally
+    {
+      thisLock.readLock().unlock();
+    }
+
+    if (callback == null)
+    {
+      if (thisReaderPosition == 0)
```
rickzx, line 123: Reaching the position 0 doesn't necessarily mean that the buffer is empty. It could also be the case that the element at position 0 has expired (but the element at position 1 is not necessarily expired). What we can do is probably to add an atomic counter of the number of items in the buffer. When put, increment the counter; when expire, decrement the counter.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/EvictingCircularBuffer.java
```diff
@@ -0,0 +1,245 @@
...
+        return get();
+      }
+    }
+    else if (Duration.between(ttl, Instant.ofEpochMilli(_clock.currentTimeMillis())).compareTo(_ttl) > 0)
+    {
+      if (thisReaderPosition != 0 || this.size() == 1)
+      {
+        thisLock.writeLock().lock();
+        try {
+          if (callback == _callbacks.get(thisReaderPosition))
```
rickzx, line 138: What's the purpose of this check?

r2-core/src/main/java/com/linkedin/r2/transport/http/client/EvictingCircularBuffer.java
```diff
@@ -0,0 +1,245 @@
...
+  }
+
+  private int getAndBumpReaderPosition()
+  {
+    return (_readerPosition.getAndUpdate(x -> (x + 1) % _callbacks.size()));
+  }
+
+  private int size()
+  {
+    return (int) _callbacks.stream().filter(Objects::nonNull).count();
```
rickzx, line 243: If we can keep an atomic counter of the size, we don't need to go through the entire array each time we call size()

r2-core/src/main/java/com/linkedin/r2/transport/http/client/EvictingCircularBuffer.java
```diff
@@ -0,0 +1,245 @@
...
+    finally
+    {
+      thisLock.readLock().unlock();
+    }
+
+    if (callback == null)
+    {
+      if (thisReaderPosition == 0)
+      {
+        throw new NoSuchElementException("Buffer is empty");
```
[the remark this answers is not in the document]
lesterhaynes: Addressed this by calling Queue.remove() instead of Queue.poll() in QueueBasedCallbackStore.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/EvictingCircularBuffer.java
```diff
@@ -0,0 +1,245 @@
...
+   * to ensure all reads and writes are drained coinciding with the resizing of the buffer.
+   * @param capacity
+   */
+  void setCapacity(int capacity)
+  {
+    if (capacity < 1)
+    {
+      throw new IllegalArgumentException("capacity can't be less than 1");
+    }
+    // acquire write lock for all elements in the buffer to prevent reads while the buffer is re-created,
```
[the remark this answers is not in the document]
lesterhaynes: Spoke about this on Zoom. In summary, while we may not love these "change after instantiate" methods, they appear necessary in order to allow customers to change them through darkclusterconfig. If we wanted to get rid of them, the alternative would be passing the ratelimiter's components into the constructor of DarkClusterStrategyFactoryImpl instead of the ratelimiter itself, which is less clean than having these setters.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/EvictingCircularBuffer.java
lesterhaynes: Opted to remove the constructor that does not have Clock as a parameter. Given that we already need a Clock for SmoothRateLimiter (and ConstantQpsRateLimiter), it makes sense to re-use the Clock passed into there.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/EvictingCircularBuffer.java
lesterhaynes: Added comment above the check in the code. Since we release the read lock prior to this point, it's possible a write could have changed the value from underneath us. This check is to ensure that, once we have the write lock, the value is still the same as what we thought it was.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/EvictingCircularBuffer.java
lesterhaynes: Chatted about this on Zoom. Opted to implement a much cleaner solution where we iterate over a while loop that increments an int until it reaches the size of the buffer.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/EvictingCircularBuffer.java
lesterhaynes: size() method has been removed after implementing while-loop-based iteration on the get() call.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/EvictingCircularBuffer.java
```diff
@@ -0,0 +1,245 @@
...
+ * - Must always accept submission of new Callbacks, replacing oldest Callbacks when buffer is exhausted
+ * - Must provide a unique Callback between subsequent get() calls
+ * - Must be able to honor get requests in excess of put requests, returning previously returned Callbacks is acceptable.
+ * - During periods of insufficient write throughput, must prune stale Callbacks, ultimately throwing NoSuchElementException
+ *   to upstream callers when inbound write throughput has dropped to zero.
+ * - Must do the above with high performance, without adding meaningful latency to reader/writer threads.
+ *
+ * This class is thread-safe, achieving performance through granular locking of each element of the underlying circular buffer.
+ */
+@SuppressWarnings("serial")
```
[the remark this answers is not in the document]
lesterhaynes: Good call-out. I don't think this is needed anymore now that I'm not having EvictingCircularBuffer implement Queue.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/SmoothRateLimiter.java
```diff
@@ -87,18 +94,23 @@
    */
   public SmoothRateLimiter(ScheduledExecutorService scheduler, Executor executor, Clock clock, Queue<Callback<None>> pendingCallbacks,
                            int maxBuffered, BufferOverflowMode bufferOverflowMode, String rateLimiterName)
+  {
+    this(scheduler, executor, clock, new QueueBasedCallbackStore(pendingCallbacks), bufferOverflowMode, rateLimiterName, new LimitedExecutionTracker(maxBuffered));
+  }
+
```
rickzx, line 100: Let's move the JavaDoc down here and add a @param for executionTracker

r2-core/src/main/java/com/linkedin/r2/transport/http/client/SmoothRateLimiter.java
```diff
@@ -350,4 +364,48 @@ public void run()
       }
     }
   }
+
+  private static class LimitedExecutionTracker implements RateLimiterExecutionTracker
```
rickzx, line 368: LimitedExecutionTracker -> BoundedRateLimiterExecutionTracker
UnlimitedExecutionTracker -> UnboundedRateLimiterExecutionTracker?

r2-core/src/main/java/com/linkedin/r2/transport/http/client/ratelimiter/CallbackStore.java
```diff
@@ -0,0 +1,43 @@
...
+
+import com.linkedin.common.callback.Callback;
+import com.linkedin.common.util.None;
+import java.util.NoSuchElementException;
+
+
+/**
+ * A lightweight queue-like interface specifically for Callbacks
+ */
+public interface CallbackStore
```
rickzx, line 27: CallbackStore -> CallbackBuffer?

r2-core/src/main/java/com/linkedin/r2/transport/http/client/ratelimiter/QueueBasedCallbackStore.java
```diff
@@ -0,0 +1,48 @@
...
+import com.linkedin.common.util.None;
+import com.linkedin.util.ArgumentUtil;
+import java.util.NoSuchElementException;
+import java.util.Queue;
+
+
+/**
+ * A simple CallbackStore implementation that delegates to the provided Queue
+ */
+public class QueueBasedCallbackStore implements CallbackStore
```
rickzx, line 29: QueueBasedCallbackStore  -> SimpleCallbackBuffer?

rickzx (changes requested):
LGTM. Some final thoughts on minor issues. After this round of review, I think we are ready to check it in. Thanks for making the changes

r2-core/src/main/java/com/linkedin/r2/transport/http/client/EvictingCircularBuffer.java
```diff
@@ -0,0 +1,238 @@
...
+   * Calls to get() will always succeed as long as calls to put() continue at a cadence within the ttl duration. When write
+   * throughput has dropped to zero, get() will eventually throw NoSuchElementException once the circular buffer has become
+   * fully pruned through expired ttl.
+   * @return Callback
+   * @throws NoSuchElementException if internal circular buffer is empty
+   */
+  public Callback<None> get() throws NoSuchElementException
+  {
+    int getAttemptIteration = 0;
+    while (getAttemptIteration++ <= getCapacity())
```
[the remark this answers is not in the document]
lesterhaynes: for loop here is definitely more standard

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.19.1] - 2021-05-14
```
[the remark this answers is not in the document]
lesterhaynes: Bumped the date to today. 29.19.1 is still a good version to use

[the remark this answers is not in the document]
lesterhaynes: Chatted briefly about this. Copied the changelog from head and ran the script to stamp the version.

rickzx (changes requested):
Let's address the test failing issues in the travis build: https://travis-ci.com/github/linkedin/rest.li/builds/228528858

rickzx (approved):
LGTM!

PR #613, opened 2021-04-28
Add APIs with assock key param fluentapi

junchuanwang:
Current FluentAPI uses API such as
`get(CompoundKey)`

this change another set of API which use spreaded version of the compoundKey
e.g., if CompoundKey consists of Long and String, it would be
`get(Long id, String message)`

### DIFF
--- a/restli-tools/src/main/resources/apiVmTemplates/action.vm
+++ b/restli-tools/src/main/resources/apiVmTemplates/action.vm
@@ -48,7 +48,7 @@ Copyright (c) 2021 LinkedIn Corp.
         );
 
         #end ## end if showTemplate
-      #end ## foreach checkAssoc
+      #end ## foreach flattenAssocKey
     #end ## end withEG
     #if(${method.hasActionParams()})
       #actionAllParamClass($method)
@@ -162,7 +162,7 @@ Copyright (c) 2021 LinkedIn Corp.
               )##
           }
         #end ## end if show template
-      #end ## foreach checkAssoc
+      #end ## foreach flattenAssocKey
     #end ## foreach method
   #end ## end withEG
 #end ## is_interface
--- a/restli-tools/src/main/resources/apiVmTemplates/action.vm
+++ b/restli-tools/src/main/resources/apiVmTemplates/action.vm
@@ -16,27 +16,39 @@ Copyright (c) 2021 LinkedIn Corp.
 #if($is_interface)
   #foreach($method in $spec.actions)
     #foreach($withEG in [true, false])
-      #set($actionParamClassName = "${util.nameCapsCase($method.name)}ActionParameters")
-      #set($actionOptionalParamClassName = "${util.nameCapsCase($method.name)}ActionOptionalParameters")
-      #doc($method.schema.doc)
+      #foreach($checkAssoc in [true, false]) ##  association key colleciton will have API with assocKeyParamsWithOptAndEg for Get, Delete, Update, PartialUpdate, Action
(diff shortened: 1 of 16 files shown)


restli-tools/src/main/resources/apiVmTemplates/action.vm
```diff
@@ -48,94 +60,109 @@ Copyright (c) 2021 LinkedIn Corp.
...
+                                      );
+          #**##makeRequestAndReturn(
+          ${method.valueClassDisplayName},
+          ${method.valueClassDisplayName},
+          "resp.getEntity()",
+                $withEG
+              )##
+          }
+        #end ## end if show template
+      #end ## foreach checkAssoc
```
karthikbalasub, line 165: flattenAssocKey

restli-tools/src/main/resources/apiVmTemplates/rest.delete.vm
```diff
@@ -15,70 +15,112 @@
 *#
 #if($is_interface)
   #foreach($withEG in [true, false])
```
karthikbalasub, line 17: If we switch the order (do withEg and flattenAssocKey outside the is_interface check), then the common logic to compute `keyStubNoOptional` `keyStubWithOptional` etc (lines 18 - 45) can be de duplicated.

foreach withEg
  foreach flattenAssocKey
   // Common logic: showTemplate, keyStubNoOptional, keyStubWithOptional
   if is_interface
      // interface
   else
      // impl

junchuanwang: Not really. We defined some class in `is_interface`.

Also I intend to use "is_interface" in the top level. Because initially we really want to use separate files for interface but later I choose to keep them in the same file. That is a deliberate decision.

karthikbalasub: Extracting the logic for the different method variations like EG and flattenAssocKey outside doesn't conflict with the points you mentioned.

If you think about it, the method variants are common between interface and impl.

junchuanwang: I undertsand you are saying the "showTeplate" is redundant. But  according to you, the interface checking will be redundant. 
```
if is_interface
// define some class
endif


foreach withEg
foreach flattenAssocKey
// Common logic: showTemplate, keyStubNoOptional, keyStubWithOptional
if is_interface
// interface
else
// impl
endforeach
endforeach
```

My preference is to keep some duplicates right now, so the interface/impl can be two separate logic groups.

karthikbalasub:
```
foreach withEg
  foreach flattenAssocKey
    // Common logic: showTemplate, keyStubNoOptional, keyStubWithOptional
    if is_interface
      // interface
      // define some class
    else
      // impl
    endif
  endforeach
endforeach
```

This is what I was proposing. You are still keeping interface/impl as two logic groups, while also de-duplicating the common logic between them.

junchuanwang: in your code, `define some class` needs guarding variables (which will make the code messy), otherwise there will be multiple definition of same class. That is the biggest problem.

Plus if we put "is_interface" inside foreach withEg, than better change every file to make it consitent.

karthikbalasub: Okay, got it. Thanks for the clarification. We can keep it as it is now.

karthikbalasub (approved):

PR #614, opened 2021-04-30
Check Gradle version against min required/suggested versions

evanw555:
During application of the Pegasus Gradle plugin, check the current project version agains the min required/suggested versions and either fail or warn with an actionable message.

For now, only enforce Gradle `1.0` to keep compatibility. I will follow this up with a backward-incompatible PR for the next major release to enforce `5.2.1`.

The recommended version should be what the next enforced version will be. For now, that's `5.2.1` (see above). Then, it'll be `5.3`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,22 +13,25 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
+- Strictly enforce Gradle version compatibility in the `pegasus` Gradle plugin.
+  - Minimum required Gradle version is now `1.0` (effectively backward-compatible).
+  - Minimum suggested Gradle version is now `5.2.1`
 
 ## [29.18.2] - 2021-04-28
 - Fix bug in generated fluent client APIs when typerefs are used as association key params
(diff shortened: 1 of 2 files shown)


BrianPin:
This LGTM, but do you consider issue warning first and then change to throw exceptions some other time?
I don't know if anyone would be noticed if it is warning though.

karthikbalasub (approved):

evanw555:
@BrianPin To be honest, I doubt that logging warnings will be very effective. I'm logging the "suggested" version just to make sure that we're at least explicit about how we intend to change our Gradle compatibility.

> do you consider issue warning first and then change to throw exceptions some other time?

I think this change effectively does that. This PR logs a warning for our "suggested" version, which will later become the "required" version and result in an error. In our next major version, the warning for 5.2.1 will become an error. Seems to be equivalent to what you're saying.

nickibi (approved):

junchuanwang:
+1 to Brain's point. can you add some logging for task in Info level

junchuanwang:
lol. I smiled after seeing your change for changelog. Ship

junchuanwang (approved):

PR #616, opened 2021-05-02
Fix bottlenecks in DataSchemaParser

cckroets:
Both of these optimizations combined, reduce the entirety of client code
generation from 2 minutes to 20 seconds on a large codebase. These 
bottlenecks were discovered by using a flame-graph attached below. The 
sections highlighted in purple essentially don't exist after these changes.

1) Remove usage of `TreeMap`. Using a `TreeMap` with a `DataSchema` key is
very expensive. Every single comparison was calling `toString()` which
forces both schemas to be serialized to JSON, then compared.
The point of this `TreeMap` was just to have deterministic
ordering, when the source file ordering has changed. To maintain this
behavior we can just sort the input up front, and maintain a
`LinkedHashMap` for consistent ordering.

2) Replace `binding()` call in loop with `existingDataSchema()`. The
implementation of `AbstractMultiFormatDataSchemaResolver.bindings()`
is very expensive. It creates a new aggregated map on each invocation. 
So calling this once for a single lookup is actually more expensive than 
a linear search would have been. Replaced this call with 
`existingDataSchema()` which is an equivalent and much cheaper 
operation. Ideally, the `DataSchemaResolver` interface wouldn't contain
methods that expose `Map` objects directly, but that would require a
major backwards-incompatible change.

Screen Shot 2021-05-02 at 11 31 57 AM

### DIFF
--- a/generator/src/main/java/com/linkedin/pegasus/generator/DataSchemaParser.java
+++ b/generator/src/main/java/com/linkedin/pegasus/generator/DataSchemaParser.java
@@ -30,13 +30,13 @@ import java.io.File;
 import java.io.FileFilter;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Comparator;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.TreeMap;
 import java.util.stream.Collectors;
 import org.apache.commons.io.FilenameUtils;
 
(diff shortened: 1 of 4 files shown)


generator/src/main/java/com/linkedin/pegasus/generator/DataSchemaParser.java
```diff
@@ -205,8 +208,8 @@ private void init(AbstractMultiFormatDataSchemaResolver resolver,
   public static class ParseResult
   {
     private static final String EXTENSION_FILENAME_SUFFIX = "Extensions.pdl";
-    // Sort the results to ensure that the output is deterministic for a given set of source inputs
-    private final Map<DataSchema, DataSchemaLocation> _schemaAndLocations = new TreeMap<>(Comparator.comparing(DataSchema::toString));
+    // Store the results in a LinkedHashMap to ensure ordering is deterministic for a given set of source inputs
```
zizhong, line 211: Is storing results in a hashmap giving the deterministic ordering? IMHO, the sortedSources gives the deterministic ordering.

zizhong:
@cckroets Please rebase your branch as the current is out-of-date.

generator/src/main/java/com/linkedin/pegasus/generator/DataSchemaParser.java
cckroets: Correct, this map "maintains" the ordering from the original sort. (Since this map is later iterated upon)

BrianPin (approved):
Sorting on the sources might be a good idea I agree. 
It was LinkedHashMap before, then it became TreeMap, then now you changed it back to LinkedHashMap but with sorting of the sources.
Just to keep an eye on where the unit test happens and likely you won't need to add cases but just check if the existing test cases are enough for your changes.

BrianPin:
Also please remember version bump and downstream test stuff.

zizhong (approved):
LG

karthikbalasub (approved):
Thanks for the contribution!

PR #617, opened 2021-05-07
D2 client side implementation of server-reported health

rachelhanhan:
Implemented the following components in the D2 client side:
1. Read the server-reported load from the TrackerClient through response wire attributes. If it's not populated, we use -1 as the default.
2. Added 2 relative threshold for the server load.
3. Added one client-side config to enabled the client-side load balancing using server reported load.

### DIFF
--- a/d2-schemas/src/main/pegasus/com/linkedin/d2/D2RelativeStrategyProperties.pdl
+++ b/d2-schemas/src/main/pegasus/com/linkedin/d2/D2RelativeStrategyProperties.pdl
@@ -49,6 +49,18 @@ record D2RelativeStrategyProperties {
   lowErrorRate: optional double
 
   /**
+   * If the latest server reported load score is above this specified factor of the cluster average,
+   * we will decrease the health score by downStep.
+   */
+  relativeLoadHighThresholdFactor: optional double
+
+  /**
+   * If the latest server reported load score is under this specified factor of the cluster average,
+   * we will increase the health score by upStep.
+   */
+  relativeLoadLowThresholdFactor: optional double
+
+  /**
    * The health score for a server will not be calculated unless the number of calls to it in the interval
    * meets or exceeds the minimum call count.
    */
--- a/d2-schemas/src/main/pegasus/com/linkedin/d2/D2RelativeStrategyProperties.pdl
+++ b/d2-schemas/src/main/pegasus/com/linkedin/d2/D2RelativeStrategyProperties.pdl
@@ -49,18 +49,6 @@ record D2RelativeStrategyProperties {
   lowErrorRate: optional double
 
   /**
-   * If the latest server reported load score is above this specified factor of the cluster average,
-   * we will decrease the health score by downStep.
-   */
-  relativeLoadHighThresholdFactor: optional double
-
-  /**
-   * If the latest server reported load score is under this specified factor of the cluster average,
-   * we will increase the health score by upStep.
-   */
-  relativeLoadLowThresholdFactor: optional double
-
-  /**
    * The health score for a server will not be calculated unless the number of calls to it in the interval
    * meets or exceeds the minimum call count.
    */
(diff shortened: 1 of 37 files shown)


d2-schemas/src/main/pegasus/com/linkedin/d2/D2RelativeStrategyProperties.pdl
```diff
@@ -48,6 +48,18 @@ record D2RelativeStrategyProperties {
...
 
+  /**
+   * If the latest server reported load score is above this specified factor of the cluster average,
+   * we will decrease the health score by downStep.
+   */
+  relativeLoadHighThresholdFactor: optional double
+
+  /**
+   * If the latest server reported load score is under this specified factor of the cluster average,
+   * we will increase the health score by upStep.
```
rickzx, line 59: I think we can have more discussion on up/downStep model vs. P2C. Sticky routing is currently best effort. It's not guaranteed to reach the same host when the pointsMap changes. I still think that 5s maybe too long for server reported load to be effective.

degrader/src/main/java/com/linkedin/util/degrader/CallTracker.java
```diff
@@ -231,6 +231,12 @@
      */
     int getOutstandingCount();
 
+    /**
+     * Returns the latest load score reported by the server at the end of the interval.
+     * @return the latest load score reported by the server at the end of the interval.
+     */
+    int getServerLoadScore();
```
rickzx, line 238: getServerReportedLoad() ?

d2/src/main/java/com/linkedin/d2/balancer/clients/TrackerClientImpl.java
```diff
@@ -192,11 +192,11 @@ public void onResponse(TransportResponse<RestResponse> response)
       if (response.hasError())
       {
         Throwable throwable = response.getError();
-        handleError(_callCompletion, throwable);
+        handleError(_callCompletion, throwable, response.getWireAttributes());
```
rickzx, line 195: I think we don't need to parse server load from wireAttributes when `enableServerReportedLoad` is set to false.

rickzx (changes requested):
Great progress! Did a first pass review.  I think we can discuss more on up/downStep model vs. P2C. Sticky routing is currently best effort. It's not guaranteed to reach the same host when the pointsMap changes anyways. I still think that 5s maybe too long for server reported load to be effective.

d2/src/main/java/com/linkedin/d2/balancer/clients/TrackerClientImpl.java
rachelhanhan: There are 3 places that we can potentially control the behavior on the client side:
1. TrackerClient - whether to pass wire attributes
2. CallTracker - whether to update the server-reported load in each interval
3. load balancer strategy - whether to take the server-reported load into load balancing decision.

When I designed this I also debated where I should put the control, in the end I only put one control in load balancer strategy to make the control logic all in one place.

Now TrackerClient and CallTracker just update the server reported load whenever it's ready, and in the load balancing strategy there are 2 cases we need to watch out:
1. Server enables reporting, but client did not enable, which we respect the client side config flag
2. Server does not enable yet, but client enables, which means the reported load will always be -1. Either case, we control the RLB to not consider server reported load.

There is definitely a way to add control in TrackerClient, where we can always report -1 as the score. There will be more code we need to cleanup later if we add a control here. What do you think?

rachelhanhan: I just changed this logic in my latest commit. I guarded the code with the config on both TrackerClient and the hash ring selector logic.

d2-schemas/src/main/pegasus/com/linkedin/d2/D2RelativeStrategyProperties.pdl
```diff
@@ -48,6 +48,18 @@ record D2RelativeStrategyProperties {
    */
   lowErrorRate: optional double
 
+  /**
+   * If the latest server reported load score is above this specified factor of the cluster average,
+   * we will decrease the health score by downStep.
+   */
+  relativeLoadHighThresholdFactor: optional double
```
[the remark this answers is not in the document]
rachelhanhan: Changed all of them to be consistent

ChrisCarini:

PR #618, opened 2021-05-12
Remove leftover references to deprecated configs.

tjni:
A few remaining references to the deprecated "runtime", "compile", and
"testCompile" configurations need to be replaced to work with Gradle 7.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.18.12] - 2021-05-26
+- Remove leftover references to deprecated configurations to work with Gradle 7.
+
 ## [29.18.11] - 2021-05-24
 - Add support for returning location of schema elements from the PDL schema parser.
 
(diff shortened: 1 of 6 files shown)


karthikbalasub (approved):

CHANGELOG.md
```diff
@@ -4937,8 +4940,13 @@ patch operations can re-use these classes for generating patch messages.
 
 ## [0.14.1]
 
+<<<<<<< HEAD
```
karthikbalasub, line 4943: please resolve these conflicts

tjni:
I have changed the plugin to apply the java-library plugin instead of the java plugin, which creates the "api" configuration. The plugin adds to the dependencies of the main Java artifact all of the dependencies of the data-template artifact, as well as the data-template artifact itself, which was declared on the "compile" configuration but now needs to be declared on the "api" configuration.

I have run a compatibility test of this internally using container and deco products, and none of the failures look related to this change. I will add the links when I am on my work computer.

DPUkyle:
> I have changed the plugin to apply the java-library plugin instead of the java plugin, which creates the "api" configuration. The plugin adds to the dependencies of the main Java artifact all of the dependencies of the data-template artifact, as well as the data-template artifact itself, which was declared on the "compile" configuration but now needs to be declared on the "api" configuration.
> 
> I have run a compatibility test of this internally using container and deco products, and none of the failures look related to this change. I will add the links when I am on my work computer.

Let's pause and discuss this - adding `java-library` seems like a big change, and my tests have shown that api/implementation separation is ineffective with the legacy publishing process at LinkedIn: it simply doesn't work and _all_ dependencies, including runtime, end up on consumers' compile-time classpaths.

Instead I think we should focus on moving to `ivy-publish`, which is a prerequisite for moving to Gradle 7.

PR #619, opened 2021-05-12
Expose getResourceClass in ResourceDefinition

mchen07:
This PR exposes `getResourceClass` in `ResourceDefinition` so that `ResourceDefinitionListener` can code using that information. Current ResourceModel (the implementation of ResourceDefinitionListener) already has that method implemented.

### DIFF
--- a/restli-server/src/main/java/com/linkedin/restli/server/ResourceDefinition.java
+++ b/restli-server/src/main/java/com/linkedin/restli/server/ResourceDefinition.java
@@ -43,19 +43,11 @@ public interface ResourceDefinition
 
   /**
    * Gets the rest.li resource java class.
-   * By default we assume java resource class name is the same as annotated namespace name,
-   * if that class cannot be found, return null.
    *
    * @return java class for this rest.li resource.
    */
   default Class<?> getResourceClass() {
-
-    String clazzName = getNamespace() + "." + getName();
-    try {
-      return Class.forName(clazzName);
-    } catch (ClassNotFoundException e) {
-      return null;
-    }
+    throw new UnsupportedOperationException();
   }
 
   /**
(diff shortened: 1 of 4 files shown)


restli-server/src/main/java/com/linkedin/restli/server/ResourceDefinition.java
```diff
@@ -41,6 +41,13 @@
    */
   String getNamespace();
 
+  /**
+   * Gets the rest.li resource java class.
+   *
+   * @return java class for this rest.li resource.
+   */
+  Class<?> getResourceClass();
```
karthikbalasub, line 49: Add a default implementation as this is changing the interface.

Also add @Override annotation on the ResourceModel class for this method.

mchen07: I thought of add @override for `ResourceModel`, then found that all other methods there didn't add @override, I will change all of them then to be consistent.

Will add default implementation to return namespace version, although this is not breaking internal, you are right that this may break open source consumers.

restli-server/src/main/java/com/linkedin/restli/server/ResourceDefinition.java
```diff
@@ -43,10 +43,20 @@
...
+   * if that class cannot be found, return null.
    *
    * @return java class for this rest.li resource.
    */
-  Class<?> getResourceClass();
+  default Class<?> getResourceClass() {
+
+    String clazzName = getNamespace() + "." + getName();
+    try {
+      return Class.forName(clazzName);
```
karthikbalasub, line 55: This might be incorrect class, no? If the namespace is different from resource class's package. 
I think we can just throw unsupported exception instead

karthikbalasub (commented):
thanks for updating the ResourceModel.

restli-server/src/main/java/com/linkedin/restli/server/ResourceDefinition.java
mchen07: ok, that is even simpler.

karthikbalasub (approved):

PR #622, opened 2021-05-17
Avoid modifying the input schema when translating pegasus schemas to avro.

karthikbalasub:
Modifying the schema can result in situation when the schema used by other threads at runtime can be incorrect. This change ensures the original schema is unmodified and any changes are stored only as overrides.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.18.7] - 2021-05-17
+- Avoid modifying pegasus schemas when translating them to AVRO.
+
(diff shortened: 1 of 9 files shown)


karthikrg:
I tried this approach yesterday and didn't end up picking it because of 2 reasons:

- It is ~5x slower. Could be the frequent map lookups. Didn't check why.
- The code is a lot more invasive and convoluted. A large part of the convolution/fragility is because schema to avro JSON inherits from schema to JSON encoder. So the potential of a bug creeping in because override was not respected in the parent class is high.

Given all this, and also given that this is mostly invoked at build time I elected to use a copy approach.

karthikbalasub:
Closing this in favor of https://github.com/linkedin/rest.li/pull/621

PR #623, opened 2021-05-21
Remove primitive class constructors

evanw555:
Constructors the classes `Boolean`, `Integer`, `Long`, `Float`, and `Double` that accept a primitive input are deprecated in Java 11.

Suggested replacement (as per the [Oracle docs](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Double.html#%3Cinit%3E(double))) is to use `Integer.valueOf(1)` (rather than `new Integer(1)`).

Also, `Boolean.TRUE` and `Boolean.FALSE` are suggested for the `Boolean` class.


BrianPin (approved):

xuli-in (approved):

nickibi (approved):
Thanks for making this change.

PR #624, opened 2021-05-21
Introduce a readonly attribute on the action annotation

karthikrg:
For example, we can use this hint to decide which actions to pull into queries and which into mutations when building GQL schemas off rest.li resources.

### DIFF
--- a/restli-server/src/main/java/com/linkedin/restli/server/annotations/Action.java
+++ b/restli-server/src/main/java/com/linkedin/restli/server/annotations/Action.java
@@ -48,6 +48,15 @@ public @interface Action
   ResourceLevel resourceLevel() default ResourceLevel.ANY;
 
   /**
+   * Optional placeholder attribute used to indicate whether the action is read-only
+   * or not. Default is false which indicates that this action can result in writes. This
+   * parameter is a placeholder to provide a hint to metadata introspection systems. The
+   * framework itself does not and cannot actually enforce read-only behavior in actions
+   * annotated as such.
+   */
+  boolean readOnly() default false;
+
+  /**
    * Optional attribute used to indicate the desired typeref to use for primitive types.
    */
   Class<? extends TyperefInfo> returnTyperef() default RestAnnotations.NULL_TYPEREF_INFO.class;
--- a/restli-tools/src/test/java/com/linkedin/restli/tools/snapshot/circular/CircularResource.java
+++ b/restli-tools/src/test/java/com/linkedin/restli/tools/snapshot/circular/CircularResource.java
@@ -31,7 +31,7 @@ import com.linkedin.restli.tools.snapshot.circular.CircularTestDataModels.*;
 @RestLiActions(name="circular")
 public class CircularResource
 {
-  @Action(name="test")
+  @Action(name="test", readOnly = true)
   public void test(@ActionParam("a") A a,
                    @ActionParam("b") B b,
                    @ActionParam("c") C c,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.18.10] - 2021-05-24
+- Introduce a readonly attribute on the action annotation
+
 ## [29.18.9] - 2021-05-24
 - Initial support for the modern `ivy-publish` plugin when producing data-template artifacts
   - Use of `ivy-publish` plugin requires Gradle 6.1+.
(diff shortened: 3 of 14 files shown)


maximelamure:
LGTM, but I would like someone from Rest.li team to review it, at least for awareness purposes.

restli-tools/src/test/java/com/linkedin/restli/tools/snapshot/circular/CircularResource.java
```diff
@@ -31,7 +31,7 @@
 @RestLiActions(name="circular")
 public class CircularResource
 {
-  @Action(name="test")
+  @Action(name="test", readOnly = true)
```
nickibi, line 34: I am curious about how this hint will be consumed? By reading the resourceClass Java class?

maximelamure:
Actually, I don't think the Rest.li client will be impacted, but should we update the GraphQL backward compatibility checker to prevent GraphQL clients to be broken if we remove this annotation?

junchuanwang (approved):
To read that I think you still need to update `RestLiAnnotationReader` or add fields to `ActionSchema`.

But this PR can be a start to let writers to begin annotating. (reader's don't block writers)

restli-tools/src/test/java/com/linkedin/restli/tools/snapshot/circular/CircularResource.java
aman1309, line 34: have you considered post finder?

restli-server/src/main/java/com/linkedin/restli/server/annotations/Action.java
```diff
@@ -47,6 +47,15 @@
...
   ResourceLevel resourceLevel() default ResourceLevel.ANY;
 
+  /**
+   * Optional placeholder attribute used to indicate whether the action is read-only
+   * or not. Default is false which indicates that this action can result in writes. This
+   * parameter is a placeholder to provide a hint to metadata introspection systems. The
+   * framework itself does not and cannot actually enforce read-only behavior in actions
+   * annotated as such.
+   */
+  boolean readOnly() default false;
```
mchen07, line 57: Have you verified that it will appear in IDL? I suspect that this is not enough, and we may need to make change in ActionSchema used in ResourceModel to export to IDL so that `graphql-infra` can read from to get `ResourceSchema`.

mchen07 (commented):
Also, we need to agree on the compatibility checker impact of this annotation.

restli-server/src/main/java/com/linkedin/restli/server/annotations/Action.java
nickibi: Adding it here only will not exposed it in the IDL. That's the reason why I asked what's the way to consume this hint? If it will consumed from IDL, then ActionSchema, RestliAnnotationReader etc need to be updated to make it is exposed in the IDL.

mchen07: yeah, GraphQL infra needs to consume it from ResourceSchema to decide whether to generate this under Mutation or under Query. Of course, it can also be done through reading custom annotation (works but not very friendly). We can add it into ActionSchema to make it first-class citizen property.

evanw555:
I am skeptical about whether this needs to be defined as a first-class property on the `Action` annotation. In fact, I don't think it's appropriate because the native Rest.li framework doesn't act upon this property. Why are custom Rest.li annotations not sufficient in this case? That would be much less intrusive.

restli-server/src/main/java/com/linkedin/restli/server/annotations/Action.java
mchen07: > I am skeptical about whether this needs to be defined as a first-class property on the `Action` annotation. In fact, I don't think it's appropriate because the native Rest.li framework doesn't act upon this property. Why are custom Rest.li annotations not sufficient in this case? That would be much less intrusive.

rest.li compatibility checker needs to report warning if the value is changed, does this count as "rest.li framework acts upon this property"?

evanw555: @mchen07 Correct. If there will be behavior natively supported by the framework corresponding to this attribute (e.g. compatibility checker logic), then supporting the attribute as a first-class feature is warranted. Otherwise, implementing this as a decoupled extension (e.g. custom rest annotation) is more reasonable.

I think if we implement corresponding compatibility checker logic, then including `readOnly` as a first-class feature is okay.

restli-tools/src/test/java/com/linkedin/restli/tools/snapshot/circular/CircularResource.java
karthikrg: Yes I updated the PR to add this to the IDL and Snapshot.

karthikrg: We already talked offline with @mchen07 and @maximelamure as to why finder would not work here.

restli-server/src/main/java/com/linkedin/restli/server/annotations/Action.java
karthikrg: Ok added it to IDL, and added backward compat check for it. Making something readwrite as readonly, or vice-versa is a backward incompatible change.

The reason is that if something is read-write, it will appear as a GQL mutation, and if read-only as a query.

mchen07 (approved):

mchen07:
Check travis build to see if it is flaky

PR #625, opened 2021-05-21
Fix a bug in ZKDeterministicSubsettingMetadataProvider to make host set distinct

rickzx:
Fix a bug in ZKDeterministicSubsettingMetadataProvider to make host set distinct.
Also, make the info/debug level log more verbose.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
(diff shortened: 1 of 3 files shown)


rachelhanhan (approved):
Where does the duplicate come from? Is it due to multi announcement to both http and https?

rickzx:
> Where does the duplicate come from? Is it due to multi announcement to both http and https?

Yes, http and https will appear as two entries previously.

PR #626, opened 2021-05-24
Add support for returning location of schema elements from the PDL schema parser.

karthikbalasub:
This feature would be useful for tools that provide additional functionality like linting, code search etc on the PDL schemas.

### DIFF
--- a/data/src/main/java/com/linkedin/data/schema/grammar/PdlSchemaParser.java
+++ b/data/src/main/java/com/linkedin/data/schema/grammar/PdlSchemaParser.java
@@ -71,8 +71,10 @@ import java.io.InputStreamReader;
 import java.io.Reader;
 import java.io.StringReader;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.IdentityHashMap;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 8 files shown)


data/src/main/java/com/linkedin/data/schema/grammar/PdlSchemaParser.java
```diff
@@ -489,9 +583,24 @@ else if (namedType.enumDeclaration() != null)
...
 
+  private void recordLocation(Object schemaNode, ParserRuleContext context)
+  {
+    if (_isLocationNeeded)
+    {
+      // getCharPosition returns beginning of the last token. Add the token's length to get actual end position.
+      int endPosition = context.getStop().getCharPositionInLine() +
+          (context.getStop().getStopIndex() - context.getStop().getStartIndex());
+      // Parser columns are indexed at 0, so add 1 to get it 1 indexed.
+      _parseLocations.put(schemaNode,
```
BrianPin, line 598: So what is a schemaNode? 
Is it the whole PDL file record or the smallest unit on the schema (like the field)
If it is the whole file then there seems to be only one location allowed to be recorded?

karthikbalasub: schema node (I'll rename it as element) is any "referencable" element in the schema. It includes fields, union members, inline schemas, enum symbols etc. For the full list, look at the java doc of getParseLocations method.

BrianPin (approved):
LGTM

nickibi (approved):
LGTM

PR #629, opened 2021-05-26
Use daemon thread to unregister TimingKeys.

tjni:
This is safe because unregistering does no I/O; it only removes from a
concurrent map.

Without using daemon threads, the threads created by the ExecutorService
will never be shut down unless the Java process is killed forceably with
System.exit or by an external system.

We found that this is impacting many internal products that initialize their Espresso databases locally for integration testing. The tool that does that is kept alive indefinitely by the static thread pool in TimingKey. While this has been worked around in new versions of the tool by explicitly calling System.exit, it shouldn't be necessary; and there are compatibility tests that fail because rest.li is updated while the tool is not.

### DIFF
--- a/r2-core/src/main/java/com/linkedin/r2/message/timing/TimingKey.java
+++ b/r2-core/src/main/java/com/linkedin/r2/message/timing/TimingKey.java
@@ -35,7 +35,8 @@ import java.util.concurrent.Executors;
 public class TimingKey
 {
   private static final Map<String, TimingKey> _pool = new ConcurrentHashMap<>();
-  private static final ExecutorService _unregisterExecutor = Executors.newFixedThreadPool(1);
+  private static final ExecutorService _unregisterExecutor = Executors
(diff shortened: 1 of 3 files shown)


r2-core/src/main/java/com/linkedin/r2/message/timing/TimingKey.java
```diff
@@ -156,4 +157,9 @@ public static int getCount() {
     return _pool.size();
   }
 
+  private static final Thread createDaemonThread(Runnable runnable) {
+    Thread thread = Executors.defaultThreadFactory().newThread(runnable);
+    thread.setDaemon(true);
```
DPUkyle, line 162: This is a great change; thanks for doing this.

tjni:
@zizhong @nizarm 

Would you mind taking a look?

zizhong (approved):
LGTM!
Thanks for taking care of this!

PR #630, opened 2021-05-26
Fix race conditions in D2 cluster subsetting

rickzx:
1. Fix a race condition where two threads with same clusterGenerationId attempts to update subset simultaneously, one thread still gets the old cached subset. The fix is to make the update a critical section.
2. Fix a bug when client cluster changes, the clusterGenerationId passed into loadBalancerStrategy is not changed, causing stale state.
3. Fix a bug when minClusterSubsetSize changes, the client is still pulling the cached subset.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.18.15] - 2021-06-02
+- Fix race conditions in D2 cluster subsetting. Refactor subsetting cache to SubsettingState.
+
 ## [29.18.14] - 2021-05-27
 - Use class.getClassLoader() instead of thread.getContextClassLoader() to get the class loader.
 
@@ -4966,7 +4969,8 @@ patch operations can re-use these classes for generating patch messages.
 
 ## [0.14.1]
 
-[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.18.14...master
+[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.18.15...master
+[29.18.15]: https://github.com/linkedin/rest.li/compare/v29.18.14...v29.18.15
 [29.18.14]: https://github.com/linkedin/rest.li/compare/v29.18.13...v29.18.14
 [29.18.13]: https://github.com/linkedin/rest.li/compare/v29.18.12...v29.18.13
 [29.18.12]: https://github.com/linkedin/rest.li/compare/v29.18.11...v29.18.12
--- a/d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerState.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerState.java
@@ -33,8 +33,7 @@ import com.linkedin.d2.balancer.strategies.LoadBalancerStrategyFactory;
 import com.linkedin.d2.balancer.strategies.degrader.DegraderLoadBalancerStrategyV3;
 import com.linkedin.d2.balancer.strategies.relative.RelativeLoadBalancerStrategy;
 import com.linkedin.d2.balancer.subsetting.DeterministicSubsettingMetadataProvider;
-import com.linkedin.d2.balancer.subsetting.SubsettingStrategy;
-import com.linkedin.d2.balancer.subsetting.SubsettingStrategyFactory;
+import com.linkedin.d2.balancer.subsetting.SubsettingState;
 import com.linkedin.d2.balancer.subsetting.SubsettingStrategyFactoryImpl;
 import com.linkedin.d2.balancer.util.ClientFactoryProvider;
 import com.linkedin.d2.balancer.util.LoadBalancerUtil;
@@ -59,7 +58,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -145,8 +143,8 @@ public class SimpleLoadBalancerState implements LoadBalancerState, ClientFactory
   private final boolean       _isSSLEnabled;
   private final SslSessionValidatorFactory _sslSessionValidatorFactory;
 
-  private final SubsettingStrategyFactory _subsettingStrategyFactory;
-  private final ConcurrentMap<String, ConcurrentMap<Integer, Map<URI, TrackerClient>>> _weightedSubsetsCache;
+  private final DeterministicSubsettingMetadataProvider _subsettingMetadataProvider;
+  private final SubsettingState _subsettingState;
 
   /*
    * Concurrency considerations:
@@ -314,13 +312,13 @@ public class SimpleLoadBalancerState implements LoadBalancerState, ClientFactory
     _clusterListeners = Collections.synchronizedList(new ArrayList<>());
     if (deterministicSubsettingMetadataProvider != null)
     {
-      _subsettingStrategyFactory = new SubsettingStrategyFactoryImpl(deterministicSubsettingMetadataProvider, this);
+      _subsettingState = new SubsettingState(new SubsettingStrategyFactoryImpl(), deterministicSubsettingMetadataProvider);
     }
     else
     {
-      _subsettingStrategyFactory = SubsettingStrategyFactory.NO_OP_SUBSETTING_STRATEGY_FACTORY;
+      _subsettingState = null;
     }
-    _weightedSubsetsCache = new ConcurrentHashMap<>();
+    _subsettingMetadataProvider = deterministicSubsettingMetadataProvider;
   }
 
   public void register(final SimpleLoadBalancerStateListener listener)
@@ -674,57 +672,30 @@ public class SimpleLoadBalancerState implements LoadBalancerState, ClientFactory
   }
 
   @Override
-  public Map<URI, TrackerClient> getClientsSubset(String serviceName,
+  public SubsettingState.SubsetItem getClientsSubset(String serviceName,
                                                   int minClusterSubsetSize,
                                                   int partitionId,
-                                                  Map<URI, TrackerClient> potentialClients)
+                                                  Map<URI, TrackerClient> potentialClients,
+                                                  long version)
   {
-    SubsettingStrategy<URI> subsettingStrategy = _subsettingStrategyFactory.get(serviceName, minClusterSubsetSize, partitionId);
-
-    if (subsettingStrategy == null)
+    if (_subsettingState == null)
     {
-      return potentialClients;
-    }
-
-    // If cluster version is not changed, return the cached subset if possible
-    if (!subsettingStrategy.isSubsetChanged(_version.get()) &&
-        _weightedSubsetsCache.containsKey(serviceName) &&
-        _weightedSubsetsCache.get(serviceName).containsKey(partitionId))
-    {
-      return _weightedSubsetsCache.get(serviceName).get(partitionId);
-    }
-
-    Map<URI, Double> weightMap = potentialClients.entrySet().stream()
-        .collect(Collectors.toMap(Map.Entry::getKey, e -> e.getValue().getPartitionWeight(partitionId)));
-    Map<URI, Double> subsetMap = subsettingStrategy.getWeightedSubset(weightMap, _version.get());
-
-    if (subsetMap == null)
-    {
-      return potentialClients;
+      return new SubsettingState.SubsetItem(false, potentialClients);
     }
     else
     {
-      Map<URI, TrackerClient> subsetClients = new HashMap<>();
-      for (Map.Entry<URI, Double> entry: subsetMap.entrySet())
-      {
-        URI uri = entry.getKey();
-        TrackerClient client = potentialClients.get(uri);
-        client.setSubsetWeight(partitionId, subsetMap.get(uri));
-        subsetClients.put(uri, client);
-      }
+      SubsettingState.SubsetItem subsetItem = _subsettingState
+          .getClientsSubset(serviceName, minClusterSubsetSize, partitionId, potentialClients, version, this);
 
-      _weightedSubsetsCache.computeIfAbsent(serviceName, k -> new ConcurrentHashMap<>());
-      _weightedSubsetsCache.get(serviceName).put(partitionId, subsetClients);
-
-      debug(_log, "cluster subset updated for service ", serviceName, ": [",
-          subsetClients.values().stream()
-            .limit(LOG_SUBSET_MAX_SIZE)
-            .map(client -> client.getUri() + ":" + client.getSubsetWeight(partitionId))
-            .collect(Collectors.joining(",")),
-          " (total ", subsetClients.size(), ")]"
+      debug(_log, "get cluster subset for service ", serviceName, ": [",
+          subsetItem.getWeightedSubset().values().stream()
+              .limit(LOG_SUBSET_MAX_SIZE)
+              .map(client -> client.getUri() + ":" + client.getSubsetWeight(partitionId))
+              .collect(Collectors.joining(",")),
+          " (total ", subsetItem.getWeightedSubset().size(), ")], shouldForceUpdate = ", subsetItem.shouldForceUpdate()
       );
 
-      return subsetClients;
+      return subsetItem;
     }
   }
 
--- a/d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerState.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerState.java
@@ -142,8 +142,6 @@ public class SimpleLoadBalancerState implements LoadBalancerState, ClientFactory
   private final SSLParameters _sslParameters;
   private final boolean       _isSSLEnabled;
   private final SslSessionValidatorFactory _sslSessionValidatorFactory;
-
-  private final DeterministicSubsettingMetadataProvider _subsettingMetadataProvider;
   private final SubsettingState _subsettingState;
 
   /*
@@ -318,7 +316,6 @@ public class SimpleLoadBalancerState implements LoadBalancerState, ClientFactory
     {
       _subsettingState = null;
     }
-    _subsettingMetadataProvider = deterministicSubsettingMetadataProvider;
   }
 
   public void register(final SimpleLoadBalancerStateListener listener)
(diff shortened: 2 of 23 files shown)


d2/src/main/java/com/linkedin/d2/balancer/subsetting/SubsettingState.java
```diff
@@ -0,0 +1,121 @@
...
+      return potentialClients;
+    }
+
+    long peerClusterVersion = subsettingStrategy.getPeerClusterVersion();
+
+    synchronized (_lock)
+    {
+      if (version == _version &&
+          peerClusterVersion == _peerClusterVersion &&
+          minClusterSubsetSize == _minClusterSubsetSize &&
```
tjni, line 86: If I am understanding this code correctly, could we put these three conditions behind a method called `isCacheValid()` or something like that?

d2/src/main/java/com/linkedin/d2/balancer/subsetting/SubsettingState.java
```diff
@@ -0,0 +1,121 @@
...
+        for (Map.Entry<URI, Double> entry: subsetMap.entrySet())
+        {
+          URI uri = entry.getKey();
+          TrackerClient client = potentialClients.get(uri);
+          client.setSubsetWeight(partitionId, subsetMap.get(uri));
+          subsetClients.put(uri, client);
+        }
+
+        _weightedSubsetsCache.computeIfAbsent(serviceName, k -> new HashMap<>());
+        _weightedSubsetsCache.get(serviceName).put(partitionId, subsetClients);
```
tjni, line 113: If you save the return value of `computeIfAbsent` into a local variable, you can avoid the call to `get` on the next line.

tjni (commented):
Only looked at a small subset of the code out of curiosity. Interesting feature!

d2/src/main/java/com/linkedin/d2/balancer/subsetting/SubsettingState.java
rickzx: Thanks for reviewing the code! Appreciate your suggestions!

Agreed isCacheValid() would be cleaner

d2/src/main/java/com/linkedin/d2/balancer/subsetting/SubsettingState.java
```diff
@@ -0,0 +1,121 @@
...
+
+    synchronized (_lock)
+    {
+      if (version == _version &&
+          peerClusterVersion == _peerClusterVersion &&
+          minClusterSubsetSize == _minClusterSubsetSize &&
+          _weightedSubsetsCache.containsKey(serviceName) &&
+          _weightedSubsetsCache.get(serviceName).containsKey(partitionId))
+      {
+        return _weightedSubsetsCache.get(serviceName).get(partitionId);
```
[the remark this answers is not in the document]
rickzx: Fixed by adding a local variable

d2/src/main/java/com/linkedin/d2/balancer/subsetting/SubsettingState.java
```diff
@@ -0,0 +1,121 @@
...
+      }
+
+      Map<URI, Double> weightMap = potentialClients.entrySet().stream()
+          .collect(Collectors.toMap(Map.Entry::getKey, e -> e.getValue().getPartitionWeight(partitionId)));
+      Map<URI, Double> subsetMap = subsettingStrategy.getWeightedSubset(weightMap);
+
+      if (subsetMap == null)
+      {
+        return potentialClients;
+      }
```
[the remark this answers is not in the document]
rickzx: If cluster subsetting is configured correctly, this code path should never be hit.

d2/src/main/java/com/linkedin/d2/balancer/subsetting/SubsettingState.java
```diff
@@ -0,0 +1,121 @@
...
+      else
+      {
+        Map<URI, TrackerClient> subsetClients = new HashMap<>();
+        for (Map.Entry<URI, Double> entry: subsetMap.entrySet())
+        {
+          URI uri = entry.getKey();
+          TrackerClient client = potentialClients.get(uri);
+          client.setSubsetWeight(partitionId, subsetMap.get(uri));
```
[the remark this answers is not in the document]
rickzx: Fixed

d2/src/main/java/com/linkedin/d2/balancer/subsetting/SubsettingState.java
rickzx: Fixed

d2/src/main/java/com/linkedin/d2/balancer/subsetting/SubsettingState.java
```diff
@@ -0,0 +1,133 @@
...
+  private final Map<String, Map<Integer, Map<URI, TrackerClient>>> _weightedSubsetsCache;
+
+  @GuardedBy("_lock")
+  private long _version;
+
+  @GuardedBy("_lock")
+  private long _peerClusterVersion;
+
+  @GuardedBy("_lock")
+  private long _minClusterSubsetSize;
```
rachelhanhan, line 45: Does one subsettingState maps to one service in the cluster or the whole peer cluster? Based on the implementation, it seems that there is only one subsettingState for the whole peer cluster, but technically the minClusterSubsetSize can be different in different d2Services?

d2/src/main/java/com/linkedin/d2/balancer/subsetting/DeterministicSubsettingStrategy.java
```diff
@@ -82,42 +72,31 @@ public DeterministicSubsettingStrategy(DeterministicSubsettingMetadataProvider m
   }
 
   @Override
-  public boolean isSubsetChanged(long version)
+  public long getPeerClusterVersion()
```
rachelhanhan, line 75: Related to the other question - I believe one strategy maps to one d2Service, but the peerClusterVersion is for the whole cluster, I'm not sure if it's the right place to duplicate the cluster version inside each strategy

d2/src/main/java/com/linkedin/d2/balancer/subsetting/SubsettingState.java
rickzx: Good point! Fixed by adding a service -> minClusterSubsetSize map

d2/src/main/java/com/linkedin/d2/balancer/subsetting/DeterministicSubsettingStrategy.java
rickzx: This is fine. We would like to listen to the cluster changes that the service belongs to. When the cluster version is updated, the URIs that serve this service are also expected to be updated

rachelhanhan: Another idea I have is to have ZKDeterministicSubsettingMetadataProvider to directly update the subsetState in the SimpleLoadBalancerState. This way, each strategy doesn't have to interact with the peerVersion directly. Instead of pulling from ZKDeterministicSubsettingMetadataProvider every time from different service based strategy, we can push the update to  the state whenever we receive a new change. This is similar to what the uri listener does today.

d2/src/main/java/com/linkedin/d2/balancer/subsetting/SubsettingState.java
```diff
@@ -0,0 +1,164 @@
...
+   * Map from serviceName => Partition Id => all potential clients
+   */
+  @GuardedBy("_lock")
+  private final Map<String, Map<Integer, Set<URI>>> _potentialClientsMap;
+
+  @GuardedBy("_lock")
+  private final Map<String, Integer> _minClusterSubsetSizeMap;
+
+  @GuardedBy("_lock")
+  private long _version;
```
rachelhanhan, line 56: I didn't quite understand why do we need this version, can you explain? thx

d2/src/main/java/com/linkedin/d2/balancer/subsetting/SubsettingState.java
```diff
@@ -0,0 +1,164 @@
...
+    SubsettingStrategy<URI> subsettingStrategy = _subsettingStrategyFactory.get(serviceName, minClusterSubsetSize, partitionId);
+
+    if (subsettingStrategy == null)
+    {
+      return potentialClients;
+    }
+
+    long peerClusterVersion = subsettingStrategy.getPeerClusterVersion();
+
+    synchronized (_lock)
```
rachelhanhan, line 100: Do we need to share the same lock for all the services, or can we do one lock per service? When we implemented the lock in the strategy, it's one lock per service. My concern is the performance, using one lock may make multiple calls to different services wait on the same lock.

rachelhanhan (commented):
Is there any unit test that tests against the cases:
1. The host is not a new host and slow start is not performed
2. When multiple threads are trying to access at the same time.

d2/src/main/java/com/linkedin/d2/balancer/subsetting/DeterministicSubsettingStrategy.java
rickzx: The current implementation is built using the same idea. Even though there's one strategy for each D2 service, there's only a single shared ZKDeterministicSubsettingMetadataProvider instance. We use the URIProperties from LoadBalancerState as the source of truth for the latest peer cluster version. The SubsettingState is used for storing the cache of service -> subset map. The _peerClusterVersion in SubsettingState is not necessarily the latest one, but the last cached one.

d2/src/main/java/com/linkedin/d2/balancer/subsetting/SubsettingState.java
```diff
@@ -0,0 +1,164 @@
...
+    {
+      if (isCacheValid(version, peerClusterVersion, minClusterSubsetSize, serviceName))
+      {
+        Map<Integer, Map<URI, TrackerClient>> serviceCache = _weightedSubsetsCache.get(serviceName);
+        if (serviceCache != null && serviceCache.containsKey(partitionId))
+        {
+          return serviceCache.get(partitionId);
+        }
+      }
```
rachelhanhan, line 109: If we have multiple downstream services, now we have a new request calling service A. When we get the subset hosts for service A, we found that the peerClusterVersion is changed, so we updated the cache for service A.

Now there is a new request to service B, when we process the request to service B, minClusterSubsetSize remains the same, and the cached peerClusterVersion is up to date since we updated during service A request. In that case, will we not update the subset for service B?

rachelhanhan: The question is inspired by your reply to the other question. I was thinking about if the cache update should be done by request thread or uri listener thread. When I said why don't we push the update, I was thinking about if we should update as long as we receive an update from zk, but I found it's not that easy.

d2/src/main/java/com/linkedin/d2/balancer/subsetting/SubsettingStrategyFactoryImpl.java
```diff
@@ -48,7 +50,7 @@ public SubsettingStrategyFactoryImpl(DeterministicSubsettingMetadataProvider det
     if (_subsettingStrategyMap.containsKey(serviceName))
     {
       Map<Integer, SubsettingStrategy<URI>> strategyMap = _subsettingStrategyMap.get(serviceName);
-      if (strategyMap.containsKey(partitionId))
+      if (minClusterSubsetSize == _minClusterSubsetSizeMap.get(serviceName) && strategyMap.containsKey(partitionId))
```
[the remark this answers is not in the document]
rickzx: Good point!

d2/src/main/java/com/linkedin/d2/balancer/subsetting/SubsettingState.java
rickzx: I refactored the logic in the SubsettingState in the latest commit. We now have a serviceName -> subsetCache map which I believe is a lot clearer. Please help take another look

d2/src/main/java/com/linkedin/d2/balancer/subsetting/SubsettingState.java
rickzx: Good point. Change to service-based lock in the latest commit

d2/src/main/java/com/linkedin/d2/balancer/subsetting/SubsettingState.java
rickzx: Fixed in the latest commit

d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerState.java
```diff
@@ -145,8 +143,8 @@
   private final boolean       _isSSLEnabled;
   private final SslSessionValidatorFactory _sslSessionValidatorFactory;
 
-  private final SubsettingStrategyFactory _subsettingStrategyFactory;
-  private final ConcurrentMap<String, ConcurrentMap<Integer, Map<URI, TrackerClient>>> _weightedSubsetsCache;
+  private final DeterministicSubsettingMetadataProvider _subsettingMetadataProvider;
```
rachelhanhan, line 146: Do we need this?

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.18.12] - 2021-05-26
```
rachelhanhan: With new changes, do we need a new minor version?

rachelhanhan (commented):
Overall looks good to me!!! Clean and clear!
Do we need add more testing to cover the corner cases that this PR addresses?

d2/src/main/java/com/linkedin/d2/balancer/strategies/degrader/DegraderLoadBalancerStrategyV3.java
```diff
@@ -931,7 +931,8 @@ private static void enrollNewClientInRecoveryMap(Map<DegraderTrackerClient, Doub
         && !state.getTrackerClients().contains(client)                        // client is new
         && config.getRingRampFactor() > FAST_RECOVERY_THRESHOLD               // Fast recovery is enabled
         && degraderControl.getInitialDropRate() > SLOW_START_THRESHOLD        // Slow start is enabled
-        && !degraderControl.isHigh())                                         // current client is not degrading or QPS is too low
+        && !degraderControl.isHigh()                                          // current client is not degrading or QPS is too low
+        && !client.doNotSlowStart())                                          // doNotSlowStart is set to false
```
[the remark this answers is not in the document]
rachelhanhan: Any comment on this?

d2/src/main/java/com/linkedin/d2/balancer/subsetting/SubsettingState.java
```diff
@@ -0,0 +1,239 @@
...
+    DeterministicSubsettingMetadata metadata = _subsettingMetadataProvider.getSubsettingMetadata(state);
+
+    if (metadata == null)
+    {
+      return new SubsetItem(false, potentialClients);
+    }
+
+    synchronized (_lockMap.computeIfAbsent(serviceName, name -> new Object()))
+    {
+      SubsetCache subsetCacahe = _subsetCache.get(serviceName);
```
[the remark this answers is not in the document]
rickzx: Thanks for catching this, fixed.


> Do we need add more testing to cover the corner cases that this PR addresses?

The new SubsettingStateTest will cover single/multi-thread case and doNotSlowStart feature. Also added a unit test for the shouldForceUpdate feature.

d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerState.java
rickzx: Good point. Fixed

CHANGELOG.md
rickzx: I don't think we need a minor version bump. All changes in this PR are backward compatible

d2/src/main/java/com/linkedin/d2/balancer/strategies/degrader/DegraderLoadBalancerStrategyV3.java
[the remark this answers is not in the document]
rickzx: I think the current implementation will not enroll new tracker client into fast recovery if doNotSlowStart is enabled. 
https://github.com/linkedin/rest.li/blob/7372bc76f7ce78007ecb24ca80f17399e5246e13/d2/src/main/java/com/linkedin/d2/balancer/clients/DegraderTrackerClientImpl.java#L80
Here, if doNotSlowStart is enabled, initialDropRate will be set to 0, which fails the check on line 933 `degraderControl.getInitialDropRate() > SLOW_START_THRESHOLD`. My change ensures that if we override doNotSlowStart by setDoNotSlowStart(), the same behavior is expected.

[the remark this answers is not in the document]
rachelhanhan: It works before because when trackerClient set doNotSlowStart to be true, the initialDropRate will always be 0, and now it's not guaranteed? Is that why we need to add this? Or is this a bug before?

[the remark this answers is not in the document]
rickzx: setDoNotSlowStart() is a new method I exposed with this change. Previously, doNotSlowStart is only determined when creating the TrackerClient with `uriSpecificProperties`. Now, for subsetting, it introduces a new case where a TrackerClient can be created, but add to the load balancer later. So before adding the TrackerClient to the load balancer, we need to manually set doNotSlowStart and make it not enroll in fast recovery program.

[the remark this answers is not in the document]
rachelhanhan: Do we need to change the same logic in RLB QuarantineManager.java line 380?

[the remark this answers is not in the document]
rickzx: There are two cases for clients to be put into RecoveryMap:
1. It's a newly added client and slow start is enabled.
2. The client is just evicted from the quarantine.

The case I added above in degrader is for case 1, which is already implemented by RLB: https://github.com/linkedin/rest.li/blob/7372bc76f7ce78007ecb24ca80f17399e5246e13/d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/StateUpdater.java#L317.

For the quarantine case, the doNotSlowStart flag should not have an effect here. In degrader, it is handled by https://github.com/linkedin/rest.li/blob/7372bc76f7ce78007ecb24ca80f17399e5246e13/d2/src/main/java/com/linkedin/d2/balancer/strategies/degrader/DegraderLoadBalancerStrategyV3.java#L665.

rachelhanhan (approved):

PR #632, opened 2021-05-28
Use class.getClassLoader() instead thread.getContextClassLoader()

nickibi:
Use class.getClassLoader() instead thread.getContextClassLoader() to get the class loader due to the java 8 -> java 11 backward incompatible behavior.

Issue:
In Java 8 a new thread would “inherit” the classloader of the thread that forked it, but in Java 11 the new thread will use the default class loader. Because of this backward incompatible problem between Java 8 and Java 11, thread.getContextClassLoader() is not getting class loader in CustomTypeUtil.

Solution:
We are going to use class.getClassLoader() to get the class loader, if class.getClassLoader() is failing to get the class loader, we fall back to use thread.getContextClassLoader() to get the class loader.

### DIFF
--- a/data/src/main/java/com/linkedin/util/CustomTypeUtil.java
+++ b/data/src/main/java/com/linkedin/util/CustomTypeUtil.java
@@ -56,11 +56,14 @@ public class CustomTypeUtil
     {
       try
       {
-        bindingClass = Class.forName(javaCoercerClassName, false, Thread.currentThread().getContextClassLoader());
+        return Class.forName(javaCoercerClassName, false, CustomTypeUtil.class.getClassLoader());
       }
-      catch (ClassNotFoundException e)
+      catch (SecurityException | ClassNotFoundException e)
       {
-        throw new IllegalArgumentException("Unable to find java coercer class of " + javaCoercerClassName + " for schema " + schema.getUnionMemberKey());
+        // If CustomTypeUtil.class.getClassLoader() throws exception
+        // or CustomTypeUtil.class.getClassLoader() could not load class,
+        // fall back to use thread context class loader
+        return getBindingClass(javaCoercerClassName, Thread.currentThread().getContextClassLoader(), schema);
       }
     }
     else
@@ -90,5 +93,17 @@ public class CustomTypeUtil
 
     return (String) o2;
 
+  }
+
+  private static Class<?> getBindingClass(String javaCoercerClassName, ClassLoader classLoader, DataSchema schema)
+  {
+    try
+    {
+      return Class.forName(javaCoercerClassName, false, classLoader);
+    }
+    catch (ClassNotFoundException e)
+    {
+      throw new IllegalArgumentException("Unable to find java coercer class of " + javaCoercerClassName + " for schema " + schema.getUnionMemberKey());
+    }
   }
 }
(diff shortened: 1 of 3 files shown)


data/src/main/java/com/linkedin/util/CustomTypeUtil.java
```diff
@@ -49,18 +49,33 @@ public static String getJavaCustomTypeClassNameFromSchema(final DataSchema schem
...
+      ClassLoader cl = null;
       try
       {
-        bindingClass = Class.forName(javaCoercerClassName, false, Thread.currentThread().getContextClassLoader());
+        cl = CustomTypeUtil.class.getClassLoader();
+      }
+      catch (SecurityException e)
+      {
+        // If CustomTypeUtil.class.getClassLoader() throws exception, fall back to use thread context class loader
+        bindingClass = getBindingClass(javaCoercerClassName, Thread.currentThread().getContextClassLoader(), schema);
```
karthikbalasub, line 65: just do 
`return getBindingClass(..)` 
here

data/src/main/java/com/linkedin/util/CustomTypeUtil.java
```diff
@@ -49,18 +49,33 @@ public static String getJavaCustomTypeClassNameFromSchema(final DataSchema schem
...
+    Class<?> bindingClass = null;
     final String javaCoercerClassName = getJavaCustomTypeClassNameFromSchema(schema);
 
     if(javaCoercerClassName != null)
     {
+      ClassLoader cl = null;
       try
       {
-        bindingClass = Class.forName(javaCoercerClassName, false, Thread.currentThread().getContextClassLoader());
+        cl = CustomTypeUtil.class.getClassLoader();
```
karthikbalasub, line 60: Just do:
`
return Class.forName(javaCoercerClassName, false, CustomTypeUtil.class.getClassLoader());
`

And then catch ClassNotFoundException below to fallback to thread context classloader.

nickibi: Updated

data/src/main/java/com/linkedin/util/CustomTypeUtil.java
nickibi: Updated

data/src/main/java/com/linkedin/util/CustomTypeUtil.java
```diff
@@ -91,4 +94,18 @@ public static String getJavaCoercerClassFromSchema(final DataSchema schema)
     return (String) o2;
 
   }
+
+  private static Class<?> getBindingClass(String javaCoercerClassName, ClassLoader classLoader, DataSchema schema)
+  {
+    final Class<?> bindingClass;
+    try
+    {
+      bindingClass = Class.forName(javaCoercerClassName, false, classLoader);
```
karthikbalasub, line 103: You can simplify this by returning here.

karthikbalasub (approved):

data/src/main/java/com/linkedin/util/CustomTypeUtil.java
nickibi: Updated

PR #633, opened 2021-05-30
Replace custom Clock with java.time.Clock.

tjni:
This is a backwards incompatible code cleanup change. Since Java 8, the new java.time classes can be used instead of homespun versions.


evanw555:
Looks like there are some interfaces that accept a Rest.li Clock, so this may be troublesome for some developers when we finally migrate to the Java Clock. I'd suggest the following things:
1. Start by only replacing it in the underlying implementation (keep the Rest.li Clock in all public interfaces).
2. Deprecate the interfaces that accept Rest.li Clock, and add new interfaces that accept the Java Clock (leave a Javadoc note explaining that users should use the new interfaces instead).
3. Deprecate the Rest.li Clock classes and leave a Javadoc note explaining to use the Java Clock instead.
4. Add a method in the Rest.li Clock interface that can easily convert to a Java Clock (e.g. `#toClock` or something like that).

Then, we can just remove all the deprecated interfaces.

PR #636, opened 2021-06-05
Use ResourceContext's `isReturnEntityRequested()` in @ReturnEntity example

mohitk2088:
## Why
Doc uses deprecated `shouldReturnEntity()` which is misleading.

### DIFF
--- a/src/reference_return_entity.md
+++ b/src/reference_return_entity.md
@@ -140,7 +140,7 @@ public CreateKVResponse<Long, Greeting> create(Greeting entity)
 {
     
     final ResourceContext resourceContext = getContext();
-    if (resourceContext.shouldReturnEntity())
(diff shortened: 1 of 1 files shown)


src/reference_return_entity.md
```diff
@@ -165,4 +165,4 @@ optimization comes from the fact that the application developer can access this
 query parameter from the resource method, allowing them to conditionally avoid
 upstream service calls that would cause unnecessary slowdown.
 
-<a id="BATCH_PARTIAL_UPDATE"></a>
\ No newline at end of file
+<a id="BATCH_PARTIAL_UPDATE"></a>
```
mohitk2088: Did not edit this. GH auto-adds a whitespace change (tried it twice). I'm using edit file directly on GH.
Verified from rich text preview that there is no unintended changes so this should be safe.

PR #639, opened 2021-06-14
Add hooks for customizing documentation (OPTIONS) response.

karthikbalasub:
The changes in this PR would allow a rest.li service to use custom renderers for HTML or JSON documentation handler.

### DIFF
--- a/restli-docgen/src/main/java/com/linkedin/restli/docgen/DefaultDocumentationRequestHandler.java
+++ b/restli-docgen/src/main/java/com/linkedin/restli/docgen/DefaultDocumentationRequestHandler.java
@@ -19,7 +19,6 @@ package com.linkedin.restli.docgen;
 
 import com.linkedin.common.callback.Callback;
 import com.linkedin.data.schema.DataSchemaResolver;
-import com.linkedin.data.schema.SchemaParserFactory;
 import com.linkedin.data.schema.resolver.ClasspathResourceDataSchemaResolver;
 import com.linkedin.jersey.api.uri.UriBuilder;
 import com.linkedin.jersey.api.uri.UriComponent;
@@ -99,11 +98,23 @@ public class DefaultDocumentationRequestHandler implements RestLiDocumentationRe
     final ResourceSchemaCollection resourceSchemas = ResourceSchemaCollection.loadOrCreateResourceSchema(_rootResources);
     final RestLiResourceRelationship relationships = new RestLiResourceRelationship(resourceSchemas, schemaResolver);
 
-    _htmlRenderer = new RestLiHTMLDocumentationRenderer(_config.getServerNodeUri(),
+    _htmlRenderer = getHtmlDocumentationRenderer(schemaResolver, relationships);
+    _jsonRenderer = getJsonDocumentationRenderer(schemaResolver, relationships);
+  }
+
+  protected RestLiDocumentationRenderer getHtmlDocumentationRenderer(DataSchemaResolver schemaResolver,
+      RestLiResourceRelationship relationships)
+  {
+    return new RestLiHTMLDocumentationRenderer(_config.getServerNodeUri(),
         relationships,
         new VelocityTemplatingEngine(),
         schemaResolver);
-    _jsonRenderer = new RestLiJSONDocumentationRenderer(relationships);
+  }
+
+  protected RestLiDocumentationRenderer getJsonDocumentationRenderer(DataSchemaResolver schemaResolver,
+      RestLiResourceRelationship relationships)
+  {
+    return new RestLiJSONDocumentationRenderer(relationships);
   }
 
   @SuppressWarnings("fallthrough")
@@ -129,15 +140,16 @@ public class DefaultDocumentationRequestHandler implements RestLiDocumentationRe
         prefixSegment = pathSegments.get(1).getPath();
     }
 
-    assert(prefixSegment.equals(DOC_PREFIX) || (HttpMethod.valueOf(request.getMethod()) == HttpMethod.OPTIONS));
+    assert(DOC_PREFIX.equals(prefixSegment) || (HttpMethod.valueOf(request.getMethod()) == HttpMethod.OPTIONS));
 
     final ByteArrayOutputStream out = new ByteArrayOutputStream(BAOS_BUFFER_SIZE);
+    final RenderContext renderContext = new RenderContext(out, request.getHeaders());
     final RestLiDocumentationRenderer renderer;
 
     if (HttpMethod.valueOf(request.getMethod()) == HttpMethod.OPTIONS)
     {
       renderer = _jsonRenderer;
-      renderer.renderResource(prefixSegment, out);
+      renderer.renderResource(prefixSegment, renderContext);
     }
     else if (HttpMethod.valueOf(request.getMethod()) == HttpMethod.GET)
     {
@@ -166,16 +178,22 @@ public class DefaultDocumentationRequestHandler implements RestLiDocumentationRe
 
       if (renderer == _htmlRenderer)
       {
-        _htmlRenderer.setJsonFormatUri(UriBuilder.fromUri(request.getURI())
-                                                 .queryParam("format", DOC_JSON_FORMAT)
-                                                 .build());
+        _htmlRenderer.setFormatUriProvider(docFormat -> {
+          if (RestLiDocumentationRenderer.DocumentationFormat.JSON.equals(docFormat))
+          {
+            return UriBuilder.fromUri(_config.getServerNodeUri())
+                .path(request.getURI().getPath())
+                .queryParam("format", DOC_JSON_FORMAT).build();
+          }
+          return null;
+        });
       }
 
       try
       {
         if (typeSegment == null || typeSegment.isEmpty())
         {
-          renderer.renderHome(out);
+          renderer.renderHome(renderContext);
         }
         else
         {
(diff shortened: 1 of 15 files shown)


ktsimpso (commented):
If I'm reading this correctly, this will allow us to add extra schemas to the `relatedSchemas` section of the options call. (Which is definitely one thing we want). But don't see a way to inject information/meta-data into the resource itself. Which is something else we need.

restli-int-test/src/test/java/com/linkedin/restli/examples/documentation/TestCustomDocumentationHandler.java
```diff
@@ -0,0 +1,113 @@
...
+        return new RestLiJSONDocumentationRenderer(relationships)
+        {
+          @Override
+          public void renderDataModel(NamedDataSchema schema, DataMap output) throws IOException
+          {
+            super.renderDataModel(schema, output);
+            DataMap schemaData =  _codec.stringToMap(schema.toString());
+            String customName = schema.getFullName() + CUSTOM_SUFFIX;
+            schemaData.put("name", customName);
+            output.put(customName, schemaData);
```
bailan, line 72: How does it ensure the right schema is used for the versioned model. I only see the name of the schema has been extended with the suffix.

In addition, does this append the name of every model with the suffix or just the versioned one. I guess if it is not doing that, we can still customize the code to do that.

restli-int-test/src/test/java/com/linkedin/restli/examples/documentation/TestCustomDocumentationHandler.java
```diff
@@ -0,0 +1,113 @@
...
+public class TestCustomDocumentationHandler extends RestLiIntegrationTest
+{
+
+  public static final String CUSTOM_SUFFIX = ".Custom";
+
+  @BeforeClass
+  public void initClass() throws Exception
+  {
+    RestLiConfig config = new RestLiConfig();
+    config.setDocumentationRequestHandler(new DefaultDocumentationRequestHandler()
```
bailan, line 57: I did not know we can override the DocumentationRequestHandler. It looks like I can even rewrite a new DefaultDocumentationRequestHandler without the change you made to the RestLiDocumentationRenderer interface, though it will be a lot of duplicated code.

bailan (commented):
Thanks for making the change. I am not sure if this would help to get the schema for the versioned data models, because I am not sure how would the DocumentationHandler find the data model for a specific version (`Foo_$1` for example) to start with.

restli-int-test/src/test/java/com/linkedin/restli/examples/documentation/TestCustomDocumentationHandler.java
karthikbalasub: This is just a test to show how you can customize the response.
For versioned models, I'd expect the logic here to get the versioned schemas for the provided named data schema and then add them to the output.

restli-int-test/src/test/java/com/linkedin/restli/examples/documentation/TestCustomDocumentationHandler.java
karthikbalasub: You could also extend the DefaultDocumentationrequestHandler and just override how you construct the HTML/JSON renderers. 
Basically the hooks are provided at multiple levels:
 - Use a custom documentation request handler.
 - Use the default handler, but change the HTML/JSON renderers
 - Use the provided renderers, but change how you handle data schemas.

restli-docgen/src/main/java/com/linkedin/restli/docgen/RestLiHTMLDocumentationRenderer.java
```diff
@@ -285,9 +285,13 @@ public String getMIMEType()
     return "text/html";
   }
 
-  public void setJsonFormatUri(URI jsonFormatUri)
```
evanw555, line 288: Probably not high-risk, but just for good practice: I'd suggest deprecating this and having it redirect to the new method.

If you're worried about this getting lost and never being addressed, perhaps you can lead the way and start using [this pattern](https://cr.openjdk.java.net/~iris/se/12/latestSpec/api/java.base/java/lang/Deprecated.html#forRemoval()):
```
@Deprecated(forRemoval=true, since="29.19.6")
```
This could help us to keep track of deprecated interfaces.

restli-docgen/src/main/java/com/linkedin/restli/docgen/DefaultDocumentationRequestHandler.java
```diff
@@ -236,11 +248,11 @@ private static RoutingException createRoutingError(String path)
   private static final String DOC_VIEW_DOCS_ACTION = "docs";
   private static final String DOC_RESOURCE_TYPE = "rest";
   private static final String DOC_DATA_TYPE = "data";
-  private static final String DOC_JSON_FORMAT = "json";
+  public static final String DOC_JSON_FORMAT = "json";
```
evanw555, line 251: Instead of exposing this field, it could be cleaner to lower redundancy by instead using `RestLiDocumentationRenderer.DocumentationFormat.JSON.toString.toLowerCase()`.

restli-docgen/src/main/java/com/linkedin/restli/docgen/DefaultDocumentationRequestHandler.java
```diff
@@ -99,11 +98,23 @@ private void initializeRenderers()
...
+      RestLiResourceRelationship relationships)
+  {
+    return new RestLiHTMLDocumentationRenderer(_config.getServerNodeUri(),
         relationships,
         new VelocityTemplatingEngine(),
         schemaResolver);
-    _jsonRenderer = new RestLiJSONDocumentationRenderer(relationships);
+  }
+
+  protected RestLiDocumentationRenderer getJsonDocumentationRenderer(DataSchemaResolver schemaResolver,
```
evanw555, line 114: If I understand correctly, this (and `#renderDataModel`) is the "hook", right? Users will extend this class and override this method, using this subclass as the documentation handler instead of the default one. Please correct me if I'm wrong.

restli-docgen/src/main/java/com/linkedin/restli/docgen/RestLiHTMLDocumentationRenderer.java
```diff
@@ -285,9 +285,13 @@ public String getMIMEType()
...
+  @Override
+  public void setFormatUri(DocumentationFormat format, URI jsonFormatUri)
```
evanw555, line 289: nit: URI argument should just be called `uri`

restli-docgen/src/main/java/com/linkedin/restli/docgen/RestLiHTMLDocumentationRenderer.java
```diff
@@ -285,9 +285,13 @@ public String getMIMEType()
...
   {
-    _jsonFormatUri = URI.create(_serverNodeUri.toString() + jsonFormatUri.toString());
+    if (format == DocumentationFormat.JSON)
```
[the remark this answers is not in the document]
karthikbalasub: This API was odd/out of place and hard to fit. I've changed it to be pull so the renderer has better control on how it uses it.

restli-docgen/src/main/java/com/linkedin/restli/docgen/RestLiHTMLDocumentationRenderer.java
karthikbalasub: Added the deprecations, the additional attributes are available only in java 9, so can't use them now.

restli-docgen/src/main/java/com/linkedin/restli/docgen/DefaultDocumentationRequestHandler.java
karthikbalasub: you are correct. Users can also customize other render* methods.

karthikbalasub:
@bailan @ktsimpso I've updated the PR with changes to provide headers and ResourceModels in the renderer. PTAL.

junchuanwang:
looks good.

junchuanwang (approved):

bailan:
Thanks for adding the request headers to the renderer. Where can I find the ResourceModels in the renderer?

karthikbalasub:
> Thanks for adding the request headers to the renderer. Where can I find the ResourceModels in the renderer?

You can use ResourceSchemaCollection::getResourceModel.

For example, in RestLiHTMLDocumentationRenderer, you can do
` _resourceSchemas.getResourceModel(resourceName)`

bailan:
> > Thanks for adding the request headers to the renderer. Where can I find the ResourceModels in the renderer?
> 
> You can use ResourceSchemaCollection::getResourceModel.
> 
> For example, in RestLiHTMLDocumentationRenderer, you can do
> ` _resourceSchemas.getResourceModel(resourceName)`

How about RestLiJSONDocumentationRenderer? It seems like OPTIONS call uses JSON renderer.

karthikbalasub:
> > > Thanks for adding the request headers to the renderer. Where can I find the ResourceModels in the renderer?
> > 
> > 
> > You can use ResourceSchemaCollection::getResourceModel.
> > For example, in RestLiHTMLDocumentationRenderer, you can do
> > ` _resourceSchemas.getResourceModel(resourceName)`
> 
> How about RestLiJSONDocumentationRenderer? It seems like OPTIONS call uses JSON renderer.

It would be similar there too. You'd use 
`_relationships.getResourceSchemaCollection()` to get the resource schemas and then use `getResourceModel` to get resource models.

bailan:
> > > > Thanks for adding the request headers to the renderer. Where can I find the ResourceModels in the renderer?
> > > 
> > > 
> > > You can use ResourceSchemaCollection::getResourceModel.
> > > For example, in RestLiHTMLDocumentationRenderer, you can do
> > > ` _resourceSchemas.getResourceModel(resourceName)`
> > 
> > 
> > How about RestLiJSONDocumentationRenderer? It seems like OPTIONS call uses JSON renderer.
> 
> It would be similar there too. You'd use
> `_relationships.getResourceSchemaCollection()` to get the resource schemas and then use `getResourceModel` to get resource models.
Works out well, thank you!

PR #640, opened 2021-06-17
Do not apply Idea and Eclipse plugins.

tjni:
PegasusPlugin should not force its users to apply either the Idea or
the Eclipse plugins. In some setups, neither plugin is needed, such
as when using IntelliJ's native Gradle import.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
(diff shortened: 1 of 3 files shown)


aman1309 (commented):
Looks good, can you update PR with master and changelog also

tjni:
> Looks good, can you update PR with master and changelog also

Done!

karthikbalasub (approved):

PR #641, opened 2021-06-17
Remove three unused and deprecated configurations.

tjni:
The "dataTemplateGenerator", "restTools", and "avroSchemaGenerator"
configurations have been deprecated and are now unused.

This is backwards incompatible, so we can wait until the next major version
release, although it is in my opinion low risk to roll out even now.


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
+- Remove unused configurations "dataTemplateGenerator", "restTools", and "avroSchemaGenerator".
```
evanw555, line 16: Kinda a nit:
- Since this is backward-compatible and users may (theoretically) need to look at the changelog is something breaks/changes, I'd suggest being more specific here; call these "Gradle configurations" (not just "configurations") added to projects by the pegasus plugin. Remember that the Gradle plugin is just one small fraction of this codebase, so the changelog can be very ambiguous if you're not specific.
- Format this in markdown; wrap the configuration names (and all other code references, for that matter) in backticks.

evanw555 (commented):
Looks good except for the one nit comment. Although this is likely to not break anyone, I'd still suggest bundling this with a major version.

evanw555 (approved):

PR #642, opened 2021-06-17
Allow client-side RetriableRequestException to be retried after ClientRetryFilter

rickzx:


r2-core/src/test/java/test/r2/filter/TestClientRetryFilter.java
```diff
@@ -77,22 +78,20 @@ public void onRestError(Throwable ex, RequestContext requestContext, Map<String,
...
       {
-        Assert.assertEquals(exception, ex);
-        Assert.assertTrue(((RetriableRequestException) exception.getCause()).getDoNotRetryOverride());
+        Assert.assertTrue(ex instanceof RetriableRequestException);
+        Assert.assertFalse(((RetriableRequestException) ex).getDoNotRetryOverride());
       }
     };
     FilterChain filterChain = FilterChains.createRestChain(captureFilter, clientRetryFilter);
-    FilterUtil.fireRestError(filterChain, exception, new HashMap<>());
+    FilterUtil.fireRestError(filterChain, new WaiterTimeoutException("exception"), new HashMap<>());
```
zhang-chris, line 95: use new RetriableRequestException() instead of specific waitar time out

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.19.2] - 2021-06-17
+- Allow client-side RetiableRequestException to be retried after ClientRetryFilter
```
zhang-chris, line 18: Retriable*

r2-core/src/main/java/com/linkedin/r2/filter/transport/ClientRetryFilter.java
```diff
@@ -87,15 +87,7 @@ else if (ex instanceof StreamException)
...
     }
-    else {
-      Throwable[] throwables = ExceptionUtils.getThrowables(ex);
-      for (Throwable throwable : throwables) {
-        if (throwable instanceof RetriableRequestException) {
-          ((RetriableRequestException) throwable).setDoNotRetryOverride(true);
-          break;
-        }
-      }
-    }
```
zhang-chris, line 98: Why was this implemented in the first place? what scenario are we trying to avoid'

zhang-chris (commented):
just one question, see below

r2-core/src/main/java/com/linkedin/r2/filter/transport/ClientRetryFilter.java
rickzx: I originally thought client-side RetriableRequestException won't go through ClientRetryFilter. So as a safe-guard, I set do not retry override when the wire attribute is not present. But this is not necessary: 1. If we want to retry server-side exception, the wire attribute will be present and the exception will be wrapped into a retriable exception. 2. If we want to retry client-side exception, we should just pass the RetriableException to the next filter.

zhang-chris: makes sense, thanks for explaining

zhang-chris (approved):
thanks for the quick fix

PR #647, opened 2021-06-26
Fix validation of non-numeric float values

evanw555:
Support has already been added for non-numeric float/double values (`NaN`, `Infinity`, `-Infinity`) in Rest.li, but support wasn't added in the validation logic. This was causing validation failures for serialized (i.e. not in-memory) non-numeric float/double values, which only occured in upstream resources (services receiving these values in a response).

Updates the validation testing to more thoroughly test non-numeric float values.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.19.6] - 2021-06-28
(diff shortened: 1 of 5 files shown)


nickibi (approved):
Thanks for fixing it.

karthikbalasub (approved):

hezhao:
Thanks for the fix, Evan!

PR #651, opened 2021-07-12
Update RestliRouter to allow "action", "q" as parameter name

junchuanwang:
In previous RestLiRouter logic, we check:

If query parameter contains
"q" -> we treat that URL contains a finder name -> we thought that URL contains query parameters
"action" -> we treat URL contains  a action name -> we thought that URL contains action parameters

Therefore, a url contains "q" and "action" will be thought as a query which contains both QueryParameters and ActionParameters.

This is essentially werid logic (for example, deciding whether the URL contains query parameters by just checking single query name "q"), but that is what it was.

This bug fix however defined a new rule:

If GET and "bq" in parameters, treat this as a batchFinder, so "q" and "action" can be both query parameter name
If GET and "q" in parameters,  treat this as a Finder, so "action" can be both query parameter name
If Post and "action" in parameters,  treat this as a action, so "q" and "bq" can be both action parameter name

This could allow "action" "q" to be parameter name, while still be backward compatible and no intrution to current other existing logic.

### DIFF
--- a/restli-server/src/main/java/com/linkedin/restli/internal/server/RestLiRouter.java
+++ b/restli-server/src/main/java/com/linkedin/restli/internal/server/RestLiRouter.java
@@ -343,19 +343,19 @@ public class RestLiRouter
     boolean contextHasBatchFinderName = context.getRequestBatchFinderName() != null;
 
     // Explicitly setting other flags to false if condition meets:
-    if (context.getRequestMethod().equalsIgnoreCase("GET") && contextHasBatchFinderName )
+    if (context.getRequestMethod().equalsIgnoreCase("GET") && contextHasFinderName )
     {
-      // If "bq" in query params for "GET" methods, then this param considered as BatchFinder name
-      // and "q" and "action" can be used as query param name
-      contextHasActionName = false;
-      contextHasFinderName = false;
-    }
-    else if (context.getRequestMethod().equalsIgnoreCase("GET") && contextHasFinderName)
-    {
-      // If "bq" not in query params for "GET" methods but "q" seen, then this param considered as finder name
-      // and "action" can be used as query param name
+      // If "q" in query params for "GET" methods, then this param considered as Finder name
+      // and "bq" and "action" can be used as query param name
       contextHasActionName = false;
       contextHasBatchFinderName = false;
+    }
+    else if (context.getRequestMethod().equalsIgnoreCase("GET") && contextHasBatchFinderName)
+    {
+      // If "q" not in query params for "GET" methods but "bq" seen, then this param considered as batchFinder name
+      // and "action" can be used as query param name
+      contextHasActionName = false;
+      contextHasFinderName = false;
     }
     else if (context.getRequestMethod().equalsIgnoreCase("POST") && contextHasActionName)
     {
--- a/restli-server/src/main/java/com/linkedin/restli/internal/server/RestLiRouter.java
+++ b/restli-server/src/main/java/com/linkedin/restli/internal/server/RestLiRouter.java
@@ -338,12 +338,39 @@ public class RestLiRouter
   private ResourceMethod mapResourceMethod(final ServerResourceContext context,
                                            final ResourceLevel resourceLevel)
   {
+    boolean contextHasActionName = context.getRequestActionName() != null;
+    boolean contextHasFinderName = context.getRequestFinderName() != null;
+    boolean contextHasBatchFinderName = context.getRequestBatchFinderName() != null;
+
+    // Explicitly setting other flags to false if condition meets:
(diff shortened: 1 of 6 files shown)


restli-server/src/main/java/com/linkedin/restli/internal/server/RestLiRouter.java
```diff
@@ -338,12 +338,39 @@ private ResourceMethodDescriptor findMethodDescriptor(final ResourceModel resour
   private ResourceMethod mapResourceMethod(final ServerResourceContext context,
                                            final ResourceLevel resourceLevel)
   {
+    boolean contextHasActionName = context.getRequestActionName() != null;
+    boolean contextHasFinderName = context.getRequestFinderName() != null;
+    boolean contextHasBatchFinderName = context.getRequestBatchFinderName() != null;
+
+    // Explicitly setting other flags to false if condition meets:
+    if (context.getRequestMethod().equalsIgnoreCase("GET") && contextHasBatchFinderName )
```
nickibi, line 346: Can we add a positive test case for this branch?

restli-server/src/main/java/com/linkedin/restli/internal/server/RestLiRouter.java
```diff
@@ -338,12 +338,39 @@ private ResourceMethodDescriptor findMethodDescriptor(final ResourceModel resour
...
+      contextHasFinderName = false;
+    }
+    else if (context.getRequestMethod().equalsIgnoreCase("GET") && contextHasFinderName)
+    {
+      // If "bq" not in query params for "GET" methods but "q" seen, then this param considered as finder name
+      // and "action" can be used as query param name
+      contextHasActionName = false;
+      contextHasBatchFinderName = false;
+    }
+    else if (context.getRequestMethod().equalsIgnoreCase("POST") && contextHasActionName)
```
nickibi, line 360: Can we also add a positive test case for this branch?

restli-server/src/test/java/com/linkedin/restli/server/test/TestRestLiRouting.java
```diff
@@ -2676,13 +2684,14 @@ public void testNKeyAssociationRoutingBasicBatch(String uri,
   {
     return new Object[][]
       {
-        { AllProtocolVersions.RESTLI_PROTOCOL_1_0_0.getProtocolVersion(), "/accounts?action=register" },
-        { AllProtocolVersions.RESTLI_PROTOCOL_2_0_0.getProtocolVersion(), "/accounts?action=register" }
+        { AllProtocolVersions.RESTLI_PROTOCOL_1_0_0.getProtocolVersion(), "/accounts?action=register", "register" },
+        { AllProtocolVersions.RESTLI_PROTOCOL_2_0_0.getProtocolVersion(), "/accounts?action=register", "register"},
+          { AllProtocolVersions.RESTLI_PROTOCOL_2_0_0.getProtocolVersion(), "/accounts?action=noOps&q=wrong", "noOps" }
```
nickibi, line 2689: nit: indent inconsistent with previous 2 lines.

restli-server/src/main/java/com/linkedin/restli/internal/server/RestLiRouter.java
```diff
@@ -338,12 +338,39 @@ private ResourceMethodDescriptor findMethodDescriptor(final ResourceModel resour
...
+
+    // Explicitly setting other flags to false if condition meets:
+    if (context.getRequestMethod().equalsIgnoreCase("GET") && contextHasBatchFinderName )
+    {
+      // If "bq" in query params for "GET" methods, then this param considered as BatchFinder name
+      // and "q" and "action" can be used as query param name
+      contextHasActionName = false;
+      contextHasFinderName = false;
+    }
+    else if (context.getRequestMethod().equalsIgnoreCase("GET") && contextHasFinderName)
```
karthikbalasub, line 353: Since FINDER came first and BATCH_FINDER was added later, I prefer picking the FINDER param as the first choice.

junchuanwang: To be honest I think there is no enough reasoning to choose one against another.

Maybe you are referring to the fact that "FINDER" came first so it is more widely used.

But I am not sure what user are truely thinking when they are actually use "bq" and "q" simultanously and what we should put into the doc.

"bq" has a precedence over "q" is my natural thinking since this is the hierarchy in my mind..

junchuanwang: I decide to follow your comment.

nickibi (approved):
LGTM

karthikbalasub (approved):

PR #658, opened 2021-07-22
Add a predicate based bulk remove method for checkedMap

karthikrg:
This is useful because entrySet() by default returns an unmodifiableMap making bulk remove operations painful and inefficient (store all keys to be removed into another data structure and then do another pass to remove).


saponniah:
LGTM

mchen07 (approved):
LGTM

PR #660, opened 2021-07-26
Add support for validating aliased union members.

karthikbalasub:
Union members originally didn't support custom properties and thus validation on the members was not supported. With aliased unions, members now support custom properties and thus can specify custom validation.

This PR updates the validation logic so these custom validation are applied correctly.

### DIFF
--- a/data/src/main/java/com/linkedin/data/schema/UnionDataSchema.java
+++ b/data/src/main/java/com/linkedin/data/schema/UnionDataSchema.java
@@ -395,6 +395,18 @@ public final class UnionDataSchema extends ComplexDataSchema
   }
 
   /**
+   * Returns the member identified by its member key.
+   *
+   * @param memberKey Union member key of the member.
+   * @return the {@link Member} if key matches a member of the union, else return null.
+   */
+  public Member getMemberByKey(String memberKey)
+  {
+    Integer index = _memberKeyToIndexMap.get(memberKey);
+    return (index != null ? _members.get(index) : null);
+  }
+
+  /**
    * Checks if the union members have aliases specified. Since either all or none of the members can be aliased
    * in a union, a return value of true from this method means all the members (excluding null member, if present)
    * have been aliased and none otherwise.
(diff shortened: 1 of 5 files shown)


data/src/main/java/com/linkedin/data/schema/UnionDataSchema.java
```diff
@@ -394,6 +394,18 @@ public DataSchema getTypeByMemberKey(String memberKey)
     return (index != null ? _members.get(index).getType() : null);
   }
 
+  /**
+   * Returns the member identified by its member key.
+   *
+   * @param memberKey Union member key of the member.
+   * @return the {@link Member} if key matches a member of the union, else return null.
+   */
+  public Member getMemberByKey(String memberKey)
```
junchuanwang, line 403: Personal preference: getMemberByMemberKey (so it is consistent with others)

data/src/main/java/com/linkedin/data/schema/validator/DataSchemaAnnotationValidator.java
```diff
@@ -658,6 +683,21 @@ public void validate(ValidatorContext context)
           }
         }
       }
+      // check if the value belongs to a member in an union and if the member has
+      // validators.
```
junchuanwang, line 687: can you also mention this is for aliased union only?

data/src/main/java/com/linkedin/data/schema/validator/DataSchemaAnnotationValidator.java
```diff
@@ -658,6 +683,21 @@ public void validate(ValidatorContext context)
...
+      if (parentSchema != null && parentSchema.getType() == DataSchema.Type.UNION)
+      {
+        UnionDataSchema unionDataSchema = (UnionDataSchema) parentSchema;
+        Object name = element.getName();
+        if (unionDataSchema.areMembersAliased() && unionDataSchema.contains((String) name))
+        {
+          UnionDataSchema.Member member = unionDataSchema.getMemberByKey((String) name);
+          if (member != null)
```
junchuanwang, line 695: Do you know why this could be null ? (since we checked the schema contains the name?)

junchuanwang (approved):
Ship it!

nickibi (approved):

data/src/main/java/com/linkedin/data/schema/validator/DataSchemaAnnotationValidator.java
karthikbalasub: removed the null check.

evanw555 (approved):

PR #661, opened 2021-07-28
Fix lb strategy race condition

rachelhanhan:
This is to fix the race condition when user change from relative to degrader strategy.

How the race condition is triggered:
When there is a serviceProperty change in d2, there are 2 steps: 1) Update service strategies, 2) update tracker clients. In relative strategy and degrader strategy, they use different tracker clients, and the TrackerClient that's used in relative strategy isn't compatible with degrader strategy.
When a request comes between the 2 steps where the strategy is already changed to degrader, but the tracker clients are still the relative one, now there will be no tracker client added to the hash ring. Therefore the hashring will stay empty forever.

The fix in this PR does not directly address the race condition, instead, it triggers an update of the hash ring when it enters the bad state. What this PR does is to update the version of the uriItem after step 2 tracker clients being updated. In that way, a hash ring update will be triggered in the new strategy.

Other potential solutions
1. Let the user request wait if there is a serviceProperty update. This will eliminate the race condition, but it blocks requests and brings more complicated implementation. Since the race condition does not happen in all machines, the cost of not fixing the race condition is just to have that single request fail on ServiceUnavailableException, but it does not let the request wait, and the subsequent requests will succeed.
2. Do not update the version in the uriItem, instead compare the version of _simpleLoadBalancerState with the cached version. The problem with this is that the version in the _simpleLoadBalancerState is always going to be different compared with the cached version if we do not update uriItem's version, an update will be triggered for each request.

Testing
Tested using aldrin by adding Thread.sleep between the 2 steps mentioned above. Verified that without the change we can reproduce the ServiceUnavailableException, and after applying the change the problem is gone.

Version
The minor version is bumped so that it's easier for us to bump the container version internally at LinkedIn. Otherwise user will not be enforced to use the right pegasus version.


CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.20.0] - 2021-07-27
+- Fixed race condition when switching d2 load balancer strategies.
```
rickzx, line 18: So in the worst case, there will only be one request failure to trigger ring update, and all the subsequent requests  should succeed after the ring is updated. 

I was thinking to further eliminate the one-off failure, if `castToDegraderTrackerClients` fails, can we simply return any TrackerClient passed to the degrader strategy's `getTrackerClient`? It's similar to how ClientSelector works:

```
      // Pick one from the tracker clients passed from the request if the ring is completely out of date
      trackerClient = trackerClients.values().stream()
          .filter(latestTrackerClient -> !excludedUris.contains(latestTrackerClient.getUri()))
          .findAny().orElse(null);
```

rickzx (changes requested):

CHANGELOG.md
rickzx: Basically i think we're missing this logic in Degrader strategy https://github.com/linkedin/rest.li/blob/7372bc76f7ce78007ecb24ca80f17399e5246e13/d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/ClientSelector.java#L137

rachelhanhan: I think this is a very good idea to at least return something, but there is some logic that requires DegraderTrackerClient.

After a trackerClient is picked, there is a logic that checks whether the whole cluster is in the drop state, they requires the degraderTrackerClient instance to finish that check. It means if we pick a TrackerClientImpl instance, it will totally just skip that check, which will change the DegraderLoadBalancerV3 logic.

Now the trade-off becomes - should we try best effort to return a TrackerClientImpl instance without taking care of checkDrop() logic, or should we just let the request fail. What do you think?

rickzx: IMO, it would be better to at least return something other than fail a request. Given it's a one-off scenario, skipping checkDrop() seems okay to me. What are your thoughts?

karthikrg:
Why not do an atomic update of the map with something like computeIfAbsent?

d2/src/test/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerStateTest.java
```diff
@@ -1469,7 +1469,7 @@ public void testSSLDisabledWithHttpsInstances() throws URISyntaxException
     assertNull(client, "shouldn't pick an https uri");
 
     _state.refreshClients(serviceProperties);
-
+    assertEquals(_state.getUriProperties("cluster-1").getVersion(), _state.getVersion());
```
zizhong, line 1472: Since it's a race condition, is it better to run the test multiple times?

rachelhanhan:
> Why not do an atomic update of the map with something like computeIfAbsent?

The problem is not within the same map. The problem is that when D2 executor threads updates the trackerClients, if a request thread comes in the middle, it will uses the latest strategy with the old trackerClients. An atomic update of the map will not solve the problem.

d2/src/test/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerStateTest.java
rachelhanhan: It's very hard to reproduce the issue within unit tests. But I manually reproduced the issue by adding some time delay and verified the fix.

CHANGELOG.md
rachelhanhan: I'm not too worried about failing very few requests, but adding the check will make the casting logic more complicated. If it's not DegraderTrackerClient then we're using TrackerClientImpl again, so it uses different instances here and there, which makes the code a bit hard to follow. I don't have a strong opinion on this though, mainly because it will solve a very intermittent error.

rickzx (approved):

rachelhanhan:
Update the PR to remove the fix by updating the cluster item version, because the fix of inconsistency tracker clients should cover both GCNs.

For the 1st GCN, the symptom is that there is 0 tracker clients added into the point map for the 1st call after migration, later on when we have all the tracker clients we wouldn't be able to update the state so it remains that weird state. For the 2nd GCN we had partial tracker clients added in the point map, but we still ended up having the inconsistency. The fix checks the discrepancy between points map size and tracker clients size, and update the state when there is inconsistency.

Tests: unit tests and manual test.
For manual tests, I ran into the similar log we observed from 2nd GCN:
2021/08/10 15:46:49.970 WARN [DegraderLoadBalancerStrategyV3] [restliServlet.dispatcher.engine.taskExecutor-18-23] [aldrin-war] [QO/rYvUfutj2tRe4FXTprA==] Client passed to DegraderV3 not an instance of DegraderTrackerClient, will not load balance to it.TrackerClientImpl [_uri=https://lor1-0000961.int.linkedin.com:14559/aldrin-backend/resources, _partitionData={0=[ weight =1.0 ]}]
2021/08/10 15:46:49.973 WARN [LoggingFilter] [restliServlet.dispatcher.engine.taskExecutor-18-23] [aldrin-war] [QO/rYvUfutj2tRe4FXTprA==] onStreamError: ex: com.linkedin.r2.message.stream.StreamException (com.linkedin.restli.server.RestLiServiceException [HTTP Status:400]: Aldrin task clientRetry fails) wire: {} rpc-trace: (rhan-mn2.linkedin.biz,curli,2020/01/01 01:00:00.000)[rhan-mn2.linkedin.biz,aldrin-war,/aldrin/2 GET]
com.linkedin.r2.message.stream.StreamException: com.linkedin.restli.server.RestLiServiceException [HTTP Status:400]: Aldrin task clientRetry fails
	2021/08/10 15:46:59.081 WARN [DistributionNonDiscreteRing] [restliServlet.dispatcher.engine.taskExecutor-18-12] [aldrin-war] [WrdobVHbz6XUZRvPJlyx1A==] Calling get on an empty ring, null value will be returned
2021/08/10 15:46:59.081 WARN [DegraderLoadBalancerStrategyV3] [restliServlet.dispatcher.engine.taskExecutor-18-12] [aldrin-war] [WrdobVHbz6XUZRvPJlyx1A==] No client found. Degrader load balancer state is inconsistent with cluster manager
2021/08/10 15:46:59.081 WARN [DegraderLoadBalancerStrategyV3] [restliServlet.dispatcher.engine.taskExecutor-18-8] [aldrin-war] [Gb6whqRWtMYT2kGBZch+rA==] Client with URI https://lor1-0000961.int.linkedin.com:14559/aldrin-backend/resources is absent in point map, pointMap={}, quarantineMap={}
2021/08/10 15:46:59.081 WARN [DegraderLoadBalancerStrategyV3] [restliServlet.dispatcher.engine.taskExecutor-18-8] [aldrin-war] [Gb6whqRWtMYT2kGBZch+rA==] Client with URI https://lor1-0000522.int.linkedin.com:14559/aldrin-backend/resources is absent in point map, pointMap={}, quarantineMap={}
2021/08/10 15:46:59.081 WARN [DegraderLoadBalancerStrategyV3] [restliServlet.dispatcher.engine.taskExecutor-18-8] [aldrin-war] [Gb6whqRWtMYT2kGBZch+rA==] Client with URI https://lor1-0001031.int.linkedin.com:14559/aldrin-backend/resources is absent in point map, pointMap={}, quarantineMap={}
2021/08/10 15:46:59.081 WARN [DegraderLoadBalancerStrategyV3] [restliServlet.dispatcher.engine.taskExecutor-18-8] [aldrin-war] [Gb6whqRWtMYT2kGBZch+rA==] Client with URI https://lor1-0001548.int.linkedin.com:14559/aldrin-backend/resources is absent in point map, pointMap={}, quarantineMap={}
2021/08/10 15:46:59.081 WARN [DegraderLoadBalancerStrategyV3] [restliServlet.dispatcher.engine.taskExecutor-18-8] [aldrin-war] [Gb6whqRWtMYT2kGBZch+rA==] Client with URI https://lor1-0003060.int.linkedin.com:14559/aldrin-backend/resources is absent in point map, pointMap={}, quarantineMap={}
2021/08/10 15:46:59.081 WARN [DegraderLoadBalancerStrategyV3] [restliServlet.dispatcher.engine.taskExecutor-18-8] [aldrin-war] [Gb6whqRWtMYT2kGBZch+rA==] Client with URI https://lor1-0000346.int.linkedin.com:14559/aldrin-backend/resources is absent in point map, pointMap={}, quarantineMap={}
2021/08/10 15:46:59.081 WARN [DegraderLoadBalancerStrategyV3] [restliServlet.dispatcher.engine.taskExecutor-18-8] [aldrin-war] [Gb6whqRWtMYT2kGBZch+rA==] Client with URI https://lor1-0000772.int.linkedin.com:14559/aldrin-backend/resources is absent in point map, pointMap={}, quarantineMap={}
2021/08/10 15:46:59.081 WARN [DegraderLoadBalancerStrategyV3] [restliServlet.dispatcher.engine.taskExecutor-18-8] [aldrin-war] [Gb6whqRWtMYT2kGBZch+rA==] Client with URI https://lor1-0000791.int.linkedin.com:14559/aldrin-backend/resources is absent in point map, pointMap={}, quarantineMap={}
2021/08/10 15:46:59.081 WARN [DegraderLoadBalancerStrategyV3] [restliServlet.dispatcher.engine.taskExecutor-18-8] [aldrin-war] [Gb6whqRWtMYT2kGBZch+rA==] Client with URI https://lor1-0000907.int.linkedin.com:14559/aldrin-backend/resources is absent in point map, pointMap={}, quarantineMap={}
2021/08/10 15:46:59.081 WARN [DegraderLoadBalancerStrategyV3] [restliServlet.dispatcher.engine.taskExecutor-18-8] [aldrin-war] [Gb6whqRWtMYT2kGBZch+rA==] Client with URI https://lor1-0001465.int.linkedin.com:14559/aldrin-backend/resources is absent in point map, pointMap={}, quarantineMap={}
2021/08/10 15:46:59.081 WARN [DegraderLoadBalancerStrategyV3] [restliServlet.dispatcher.engine.taskExecutor-18-8] [aldrin-war] [Gb6whqRWtMYT2kGBZch+rA==] Client with URI https://lor1-0001226.int.linkedin.com:14559/aldrin-backend/resources is absent in point map, pointMap={}, quarantineMap={}
2021/08/10 15:46:59.081 WARN [DynamicClient] [restliServlet.dispatcher.engine.taskExecutor-18-12] [aldrin-war] [WrdobVHbz6XUZRvPJlyx1A==] unable to find service for: com.linkedin.r2.message.stream.StreamRequestImpl: [Service: clientRetry, Method: GET]
2021/08/10 15:46:59.081 WARN [DegraderLoadBalancerStrategyV3] [restliServlet.dispatcher.engine.taskExecutor-18-8] [aldrin-war] [Gb6whqRWtMYT2kGBZch+rA==] Client with URI https://lor1-0001587.int.linkedin.com:14559/aldrin-backend/resources is absent in point map, pointMap={}, quarantineMap={}
2021/08/10 15:46:59.081 WARN [DegraderLoadBalancerStrategyV3] [restliServlet.dispatcher.engine.taskExecutor-18-8] [aldrin-war] [Gb6whqRWtMYT2kGBZch+rA==] Client with URI https://lor1-0003006.int.linkedin.com:14559/aldrin-backend/resources is absent in point map, pointMap={}, quarantineMap={}
2021/08/10 15:46:59.081 WARN [DegraderLoadBalancerStrategyV3] [restliServlet.dispatcher.engine.taskExecutor-18-8] [aldrin-war] [Gb6whqRWtMYT2kGBZch+rA==] Client with URI https://lor1-0001561.int.linkedin.com:14559/aldrin-backend/resources is absent in point map, pointMap={}, quarantineMap={}
2021/08/10 15:46:59.082 WARN [DegraderLoadBalancerStrategyV3] [restliServlet.dispatcher.engine.taskExecutor-18-8] [aldrin-war] [Gb6whqRWtMYT2kGBZch+rA==] Client with URI https://lor1-0001631.int.linkedin.com:14559/aldrin-backend/resources is absent in point map, pointMap={}, quarantineMap={}
2021/08/10 15:46:59.082 WARN [DegraderLoadBalancerStrategyV3] [restliServlet.dispatcher.engine.taskExecutor-18-8] [aldrin-war] [Gb6whqRWtMYT2kGBZch+rA==] Client with URI https://lor1-app52506.int.linkedin.com:14559/aldrin-backend/resources is absent in point map, pointMap={}, quarantineMap={}
2021/08/10 15:46:59.082 WARN [DegraderLoadBalancerStrategyV3] [restliServlet.dispatcher.engine.taskExecutor-18-8] [aldrin-war] [Gb6whqRWtMYT2kGBZch+rA==] Client with URI https://lor1-0001027.int.linkedin.com:14559/aldrin-backend/resources is absent in point map, pointMap={}, quarantineMap={}
2021/08/10 15:46:59.082 WARN [DegraderLoadBalancerStrategyV3] [restliServlet.dispatcher.engine.taskExecutor-18-8] [aldrin-war] [Gb6whqRWtMYT2kGBZch+rA==] Client with URI https://lor1-0000834.int.linkedin.com:14559/aldrin-backend/resources is absent in point map, pointMap={}, quarantineMap={}
2021/08/10 15:46:59.082 WARN [DegraderLoadBalancerStrategyV3] [restliServlet.dispatcher.engine.taskExecutor-18-8] [aldrin-war] [Gb6whqRWtMYT2kGBZch+rA==] Client with URI https://lor1-0001283.int.linkedin.com:14559/aldrin-backend/resources is absent in point map, pointMap={}, quarantineMap={}
2021/08/10 15:46:59.082 WARN [DegraderLoadBalancerStrategyV3] [restliServlet.dispatcher.engine.taskExecutor-18-8] [aldrin-war] [Gb6whqRWtMYT2kGBZch+rA==] Client with URI https://lor1-0000975.int.linkedin.com:14559/aldrin-backend/resources is absent in point map, pointMap={}, quarantineMap={}

And it all get recovered in the next request.

PR #662, opened 2021-07-30
Bump netty version to use ALPN support needed for JDK8u282.

vivdesh:
With JDK8u282, the ALPN APIs are merged with JDK itself and the ALPN jar is not required. The support for this is added in the netty. Without this newer netty, we get "Unable to wrap SSLEngine of type sun.security.ssl.SSLEngineImpl" runtime error. So this PR updates the netty to avoid this error.

### DIFF
--- a/r2-netty/src/test/java/com/linkedin/r2/transport/http/client/TestHttpNettyClient.java
+++ b/r2-netty/src/test/java/com/linkedin/r2/transport/http/client/TestHttpNettyClient.java
@@ -261,7 +261,7 @@ public class TestHttpNettyClient
   {
     testHeaderSize(TEST_MAX_HEADER_SIZE - 1, RESPONSE_OK);
 
-    testHeaderSize(TEST_MAX_HEADER_SIZE, TOO_LARGE);
+    testHeaderSize(TEST_MAX_HEADER_SIZE, RESPONSE_OK);
 
     testHeaderSize(TEST_MAX_HEADER_SIZE + 1, TOO_LARGE);
   }
--- a/r2-netty/src/test/java/com/linkedin/r2/transport/http/client/TestHttpNettyClient.java
+++ b/r2-netty/src/test/java/com/linkedin/r2/transport/http/client/TestHttpNettyClient.java
@@ -261,7 +261,7 @@ public class TestHttpNettyClient
   {
     testHeaderSize(TEST_MAX_HEADER_SIZE - 1, RESPONSE_OK);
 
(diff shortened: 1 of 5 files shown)


vivdesh:
@nizarm @zizhong Could you please review this?

r2-netty/src/test/java/com/linkedin/r2/transport/http/client/TestHttpNettyClient.java
```diff
@@ -261,7 +261,7 @@ public void testMaxHeaderSize() throws InterruptedException, IOException, Timeou
   {
     testHeaderSize(TEST_MAX_HEADER_SIZE - 1, RESPONSE_OK);
 
-    testHeaderSize(TEST_MAX_HEADER_SIZE, RESPONSE_OK);
+    testHeaderSize(TEST_MAX_HEADER_SIZE, TOO_LARGE);
```
zizhong, line 264: @vivdesh Hi, could you explain why we need to change this test?

zizhong:
@vivdesh Is there any test done with this change? If possible, you can create a hotfix version of the container and test it with some services.

vivdesh:
> @vivdesh Is there any test done with this change? If possible, you can create a hotfix version of the container and test it with some services.

@zizhong We have some MPs with this netty version in exeternal dependency and using JDK8u282.

r2-netty/src/test/java/com/linkedin/r2/transport/http/client/TestHttpNettyClient.java
vivdesh: Mostly strict parsing of the http header causing this test to fail. https://github.com/netty/netty/pull/10058.
Please let me know where we can change the code according to this?

zizhong: @vivdesh Thanks for the additional information. If the original test failed because our header generator doesn't follow strictly the HTTP RFC, we probably better fix it.

zizhong: Actually, there is already a fix from https://github.com/linkedin/rest.li/pull/595/files#diff-bd070627be0acf661a65543c18644309364d2aaf26967596cfdce517c07a6a12R232 by @tjni 
Looking more, this change appears to be a subset of https://github.com/linkedin/rest.li/pull/595 which also introducing a newer version of Netty. It makes sense to me to review and merge that commit instead.

zizhong (approved):

PR #664, opened 2021-08-04
Squash a flea. ConstantQps burst value should never eval to 0

lesterhaynes:
Bug is preventing us from ramping the feature into production. Appears that maybe the clusterInfoProvider returns 0 before it gets a chance to discover what's actually there?


rickzx (changes requested):
Looks good. Please update CHANGELOG and bump the patch version in gradle.properties.

rickzx (approved):

PR #665, opened 2021-08-05
Add support for resolving from multiple schema source directories.

karthikbalasub:
This change also introduces the concept of "source" and "resolver"
directories when creating a `DataSchemaParser`. "Source" directories are
used to parse/load the input schemas, while the "resolver" directories
will only be used for resolving referenced schemas.

### DIFF
--- a/data/src/main/java/com/linkedin/data/schema/resolver/SchemaDirectory.java
+++ b/data/src/main/java/com/linkedin/data/schema/resolver/SchemaDirectory.java
@@ -1,5 +1,5 @@
 /*
-   Copyright (c) 2020 LinkedIn Corp.
+   Copyright (c) 2021 LinkedIn Corp.
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
--- a/data/src/main/java/com/linkedin/data/schema/resolver/SchemaDirectory.java
+++ b/data/src/main/java/com/linkedin/data/schema/resolver/SchemaDirectory.java
@@ -17,31 +17,23 @@
 package com.linkedin.data.schema.resolver;
 
 /**
- * Directories within resource jar file that holds different types of pegasus schemas. Ex. Data or Extensions
+ * Represents a schema directory relative to the resolver path (directory or Jar file). This is used to customize
+ * schema resolvers to limit which directories are used for resolving schema references.
  *
- * @author Aman Gupta
+ * @author Karthik Balasubramanian
  */
-public enum SchemaDirectoryName implements SchemaDirectory
+public interface SchemaDirectory
 {
   /**
-   * Directory holds the pegasus schemas. Pegasus parsers and resolvers look for pegasus
-   * files(*.pdl, *.pdsc) only within this directory.
+   * Return the schema directory name.
    */
-  PEGASUS("pegasus"),
+  String getName();
+
   /**
-   * Directory holds the Entity Relationship pegasus schemas.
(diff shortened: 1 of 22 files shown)


data/src/main/java/com/linkedin/data/schema/resolver/SchemaDirectory.java
```diff
@@ -0,0 +1,39 @@
+/*
+   Copyright (c) 2020 LinkedIn Corp.
```
junchuanwang, line 2: 2021

data/src/main/java/com/linkedin/data/schema/resolver/SchemaDirectory.java
```diff
@@ -0,0 +1,39 @@
...
+
+package com.linkedin.data.schema.resolver;
+
+/**
+ * Represents a schema directory relative to the resolver path (directory or Jar file). This is used to customize
+ * schema resolvers to limit which directories are used for resolving schema references.
+ *
+ * @author Karthik Balasubramanian
+ */
+public interface SchemaDirectory
```
junchuanwang, line 25: Suggest to rename to something like "SchemaDirectoryRepresentation" to indicate this could be an interface

generator/src/main/java/com/linkedin/pegasus/generator/DataSchemaParser.java
```diff
@@ -277,4 +298,50 @@ public ParseResult addMessage(String message)
       return this;
     }
   }
+
+  public static class Builder
```
junchuanwang, line 302: Usually the builder pattern create a new builder each time call a static method.

For example
DataSchemaParser.Builder.<>.<>.build().

Do you think we can do that?

generator/src/main/java/com/linkedin/pegasus/generator/DataSchemaParser.java
```diff
@@ -54,38 +56,57 @@
 
   /**
    * @param resolverPath provides the search paths separated by the system file separator, or null for no search paths.
+   * @deprecated Use {@link Builder} to construct the parser.
```
junchuanwang, line 59: Top Level comment (line 51 ish) is probably needed to guide users to use Builder pattern since all public constructors are deprecated

data/src/main/java/com/linkedin/data/schema/resolver/SchemaDirectory.java
karthikbalasub: I prefer to keep this name. It actually captures what this interface represents. 
Naming for interfaces and classes should be similar (i.e, usually nouns). See https://google.github.io/styleguide/javaguide.html#s5.2.2-class-names

Also, IDEs do a good job of indicating when a type is an interface.

generator/src/main/java/com/linkedin/pegasus/generator/DataSchemaParser.java
karthikbalasub: There is no convention that builder should have static methods. Infact, the accepted pattern is to use the constructor for the Builder which accepts all required paramters.

Added a static "newBuilder" method for convenience.

data/src/main/java/com/linkedin/data/schema/resolver/SchemaDirectory.java
junchuanwang: I just think hard to image I will name an interface as "SchemaDirectory" (which is ideally a class, a noun)

junchuanwang (approved):

PR #666, opened 2021-08-09
Optimize in-process request execution.

karthikrg:
Provide the ability to set cookies and projection params in request context's local attributes to avoid serializing/deserializing them for requests that are executed in-process.

### DIFF
--- a/restli-server/src/main/java/com/linkedin/restli/internal/server/ResourceContextImpl.java
+++ b/restli-server/src/main/java/com/linkedin/restli/internal/server/ResourceContextImpl.java
@@ -45,6 +45,7 @@ import com.linkedin.restli.internal.common.URIParamUtils;
 import com.linkedin.restli.internal.server.util.ArgumentUtils;
 import com.linkedin.restli.internal.server.util.MIMEParse;
 import com.linkedin.restli.internal.server.util.RestLiSyntaxException;
+import com.linkedin.restli.server.LocalRequestProjectionMask;
 import com.linkedin.restli.server.ProjectionMode;
 import com.linkedin.restli.server.RestLiResponseAttachments;
 import com.linkedin.restli.server.RestLiServiceException;
@@ -139,6 +140,7 @@ public class ResourceContextImpl implements ServerResourceContext
    * @throws RestLiSyntaxException if the syntax of query parameters in the request is
    *           incorrect
    */
+  @SuppressWarnings("unchecked")
   public ResourceContextImpl(final MutablePathKeys pathKeys,
                              final Request request,
                              final RequestContext requestContext) throws RestLiSyntaxException
@@ -148,7 +150,17 @@ public class ResourceContextImpl implements ServerResourceContext
     _requestHeaders = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
     _requestHeaders.putAll(request.getHeaders());
     _responseHeaders = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
-    _requestCookies = new ArrayList<>(CookieUtil.decodeCookies(_request.getCookies()));
+
+    List<HttpCookie> contextRequestCookies = (List<HttpCookie>) requestContext.getLocalAttr(CONTEXT_COOKIES_KEY);
+    if (contextRequestCookies != null)
+    {
+      _requestCookies = contextRequestCookies;
+    }
+    else
+    {
+      _requestCookies = CookieUtil.decodeCookies(_request.getCookies());
+    }
+
     _responseCookies = new ArrayList<>();
     _requestContext = requestContext;
     _responseAttachmentsAllowed = isResponseAttachmentsAllowed(request);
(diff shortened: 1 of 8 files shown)


restli-server/src/main/java/com/linkedin/restli/internal/server/ResourceContextImpl.java
```diff
@@ -184,31 +196,42 @@ public ResourceContextImpl(final MutablePathKeys pathKeys,
...
 
-    if (_parameters.containsKey(RestConstants.METADATA_FIELDS_PARAM))
-    {
-      _metadataProjectionMask = ArgumentUtils.parseProjectionParameter(getParameter(RestConstants.METADATA_FIELDS_PARAM));
-    }
-    else
-    {
-      _metadataProjectionMask = null;
-    }
+      if (_parameters.containsKey(RestConstants.METADATA_FIELDS_PARAM))
```
junchuanwang, line 218: Is there inconsistent tab/space uses this line? (github diff seems to have highlighted this)

restli-server/src/main/java/com/linkedin/restli/server/LocalProjectionMaskAttr.java
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * Encapsulates projection {@link MaskTree} that can be set in local attributes of
+ * {@link com.linkedin.r2.message.RequestContext} with the
+ * {@link com.linkedin.restli.internal.server.ServerResourceContext#CONTEXT_PROJECTION_MASKS_KEY}.
+ *
+ * <p>This enables a nifty performance optimization that avoids serializing and deserializing the projection
+ * masks for in-process request execution.</p>
+ */
+public class LocalProjectionMaskAttr
```
junchuanwang, line 31: I would suggest rename to relate it with more context,  e.g. LocalRequestProjectionMask, what do you think.

karthikbalasub, line 31: This can be package private, we don't expect this class to be used outside.

restli-server/src/main/java/com/linkedin/restli/server/LocalProjectionMaskAttr.java
```diff
@@ -0,0 +1,93 @@
...
+  private final MaskTree _pagingProjectionMask;
+
+  /**
+   * Constructor
+   *
+   * @param projectionMask           Projection mask for root entity.
+   * @param metadataProjectionMask   Projection mask for metadata.
+   * @param pagingProjectionMask     Projection mask for paging.
+   */
+  public LocalProjectionMaskAttr(@Nullable MaskTree projectionMask,
```
karthikbalasub, line 58: package private

restli-server/src/main/java/com/linkedin/restli/internal/server/ServerResourceContext.java
```diff
@@ -46,6 +46,17 @@
  */
 public interface ServerResourceContext extends ResourceContext
 {
+  /**
+   * Local attribute key for request cookies. Value should be a {@link List} of {@link HttpCookie} objects.
+   */
+  public static final String CONTEXT_COOKIES_KEY = "$Cookies$";
```
junchuanwang, line 52: Not sure why these constants are put in `ServerResrouceContext` interface? As I see it, they are more related to request?

karthikbalasub, line 52: It looks like there are two existing patterns for these attribute names:
R2 attributes are "ALL_CAPS"
Only attribute in rest.li uses the convention "classname.attributename"

As the attribute names are shared across the framework and also the application code, I like the second approach.

So these params would be:
`ServerResourceContext.class.getCannonicalName() + ".cookie".`

restli-server/src/main/java/com/linkedin/restli/internal/server/ResourceContextImpl.java
karthikrg: It looks ok in the IDE and in local git diff... Not sure why.

restli-server/src/main/java/com/linkedin/restli/server/LocalProjectionMaskAttr.java
karthikrg: Great idea. renamed.

restli-server/src/main/java/com/linkedin/restli/internal/server/ServerResourceContext.java
karthikrg: Because they are used in the Server resource context's local attributes.

restli-server/src/main/java/com/linkedin/restli/server/LocalProjectionMaskAttr.java
karthikrg: Cannot be package private. This will be set by client code when making short-circuited in process requests.

restli-server/src/main/java/com/linkedin/restli/internal/server/ServerResourceContext.java
karthikrg: Fixed

restli-server/src/main/java/com/linkedin/restli/server/LocalProjectionMaskAttr.java
karthikrg: We do expect it to be used outside. Client app code will create instances of this class to set projection masks in request context when creating short-circuit inprocess requests.

karthikbalasub (approved):

junchuanwang (approved):

PR #668, opened 2021-08-12
Minimize computations for requests resolved via in-process Rest.li servers

karthikrg:


restli-server/src/main/java/com/linkedin/restli/internal/server/ServerResourceContext.java
```diff
@@ -58,6 +58,17 @@
    */
   String CONTEXT_PROJECTION_MASKS_KEY = ServerResourceContext.class.getName() + ".projectionMasks";
 
+  /**
+   * Local attribute key to indicate that this request will be served by an in-process Rest.li server. Value must
+   * be a {@link Boolean}.
+   */
+  String CONTEXT_IN_PROCESS_RESOLUTION_KEY = ServerResourceContext.class.getName() + ".inProcessResolution";
```
mchen07, line 65: Where did you set these two newly added context key?

restli-server/src/main/java/com/linkedin/restli/server/BaseRestLiServer.java
```diff
@@ -117,9 +118,14 @@ private boolean isSupportedProtocolVersion(ProtocolVersion clientProtocolVersion
...
    */
-  private void ensureRequestUsesValidRestliProtocol(final Request request) throws RestLiServiceException
+  private void ensureRequestUsesValidRestliProtocol(final Request request, final RequestContext requestContext)
+      throws RestLiServiceException
   {
-    ProtocolVersion clientProtocolVersion = ProtocolVersionUtil.extractProtocolVersion(request.getHeaders());
+    ProtocolVersion clientProtocolVersion =
+        Optional.ofNullable(requestContext.getLocalAttr(ServerResourceContext.CONTEXT_PROTOCOL_VERSION_KEY))
+        .map(ProtocolVersion.class::cast)
+        .orElseGet(() -> ProtocolVersionUtil.extractProtocolVersion(request.getHeaders()));
```
mchen07, line 127: Why do you feel that this is more performant than previous just map lookup?

restli-server/src/main/java/com/linkedin/restli/internal/server/ServerResourceContext.java
karthikrg: I will set it in the clients when making the request. For example in RemoteAndLocalRestClient in our internal GQL repo.

mchen07 (approved):

PR #673, opened 2021-08-18
Catch the error from the executor task so that it does not fail the schedule silently

rachelhanhan:
Recently we found a bug that D2 executor in StateUpdater cancelled the update schedule silently. In order to fix this, there are 2 places to fix:

1. Error handling for tasks. Whenever this is an error, we just catch the error instead of let it fail the whole schedule.
2. After some investigation, the quarantine logic can throw a runtime exception when the quarantine latency is longer than the check interval. To reach the parity with degrader strategy, this PR makes sure that the quarantineLatency will not exceed the check interval.

There are other design flaws of quarantine logic in relative load balancer, especially around the dynamic latency calculation, but those are not as urgent as fixing the bug that potentially breaks D2.


d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/QuarantineManager.java
```diff
@@ -99,7 +100,11 @@ public void updateQuarantineState(PartitionState newPartitionState,
       PartitionState oldPartitionState, long clusterAvgLatency)
   {
     long quarantineLatency = Math.max((long) (clusterAvgLatency * _relativeLatencyLowThresholdFactor),
-        MIN_QUANRANTINE_LATENCY_MS);
+        MIN_QUARANTINE_LATENCY_MS);
+    if (quarantineLatency > MAX_QUARANTINE_LATENCY_MS)
```
rickzx, line 104: nit: `quarantineLatency = min(MAX_QUARANTINE_LATENCY_MS, quarantineLatency)`?

d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/StateUpdater.java
```diff
@@ -167,12 +167,18 @@ int getFirstValidPartitionId()
...
+      for (Integer partitionId : _partitionLoadBalancerStateMap.keySet())
+      {
+        PartitionState partitionState = _partitionLoadBalancerStateMap.get(partitionId);
+        updateStateForPartition(partitionState.getTrackerClients(), partitionId, partitionState, partitionState.getClusterGenerationId());
+      }
+    } catch (Exception ex)
     {
-      PartitionState partitionState = _partitionLoadBalancerStateMap.get(partitionId);
-      updateStateForPartition(partitionState.getTrackerClients(), partitionId, partitionState, partitionState.getClusterGenerationId());
+      LOG.error("Failed update the event.", ex);
```
rickzx, line 179: Maybe `"Failed to update the state for service: " + _serviceName`?

rickzx (changes requested):
LGTM

d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/QuarantineManager.java
rachelhanhan: Fixed

d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/StateUpdater.java
rachelhanhan: Fixed

rickzx (approved):
Thanks for the quick fix!

PR #674, opened 2021-08-19
Added 2 simulation tests to explicitly reproduce the issue when the quarantine latency is bigger than 1000ms and verify the fix.

rachelhanhan:
This PR adds 2 simulation tests that simulates the following scenario:
A host is unhealthy and needs to be quarantined, when it's quarantined, the quarantine latency is greater than 1000ms (with 4.0 as low relative factor, avgClusterLatency > 250ms)

I explicitly verified when the last fix is not in place, we will receive the error stating the quarantine latency is greater than 1000ms. After the fix, the quarantine logic works well. These 2 simulation tests gives us confidence about our last fix in 29.21.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.21.1] - 2021-08-18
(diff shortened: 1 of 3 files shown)


ChrisCarini:

PR #675, opened 2021-08-19
Enable skipping request and response validation via the use of request headers

karthikrg:
This is useful if clients are doing validation or don't care about it.

### DIFF
--- a/restli-server/src/main/java/com/linkedin/restli/server/validation/RestLiValidationFilter.java
+++ b/restli-server/src/main/java/com/linkedin/restli/server/validation/RestLiValidationFilter.java
@@ -29,6 +29,7 @@ import com.linkedin.restli.common.HttpStatus;
 import com.linkedin.restli.common.PatchRequest;
 import com.linkedin.restli.common.ProtocolVersion;
 import com.linkedin.restli.common.ResourceMethod;
+import com.linkedin.restli.common.RestConstants;
 import com.linkedin.restli.common.UpdateEntityStatus;
 import com.linkedin.restli.common.util.ProjectionMaskApplier;
 import com.linkedin.restli.common.util.ProjectionMaskApplier.InvalidProjectionException;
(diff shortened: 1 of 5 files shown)


restli-server/src/main/java/com/linkedin/restli/server/validation/RestLiValidationFilter.java
```diff
@@ -143,6 +144,11 @@ public RestLiValidationFilter(Collection<String> nonSchemaFieldsToAllowInProject
       }
     }
 
+    if (!shouldValidateOnRequest(requestContext))
```
mchen07, line 147: prefer to move this to the first line in this call, then you can remove another `if`.

karthikrg: It is unsafe to move this earlier, because we first check shouldValidateOnResponse and if true we construct validating schema and put it in the scratchpad.

If we move this earlier, we will miss doing that and cases where request validation is skipped but response validation is enabled will bomb.

mchen07: oh, I misread the first if, it is `shouldValidateOnResponse`.

mchen07 (approved):

restli-server/src/main/java/com/linkedin/restli/server/validation/RestLiValidationFilter.java
```diff
@@ -449,8 +455,22 @@ private void validateCreateCollectionResponse(RestLiDataValidator validator,
     }
   }
 
-  private boolean shouldValidateOnResponse(FilterRequestContext requestContext)
+  protected boolean shouldValidateOnRequest(FilterRequestContext requestContext)
+  {
+    // Skip request validation if the header to skip request validation is set.
+    return !(Boolean.TRUE.toString().equals(
+        requestContext.getRequestHeaders().get(RestConstants.HEADER_SKIP_REQUEST_VALIDATION)));
```
karthikbalasub, line 462: This is dangerous. A client can simply add this header and skip server side validation. We should not roll this out to all services.

We should add a hook to this filter that a single service can customize and use a header specific to that service to skip validation.

syzboy: Arun mentioned we have a similar case in deco and we worked with ATS to block that header so it cannot be set from requests from outside of our network. We probably should do the similar thing here.

karthikrg: Discussed offline.

Raised https://github.com/linkedin/rest.li/pull/676 to address this issue.

PR #680, opened 2021-08-27
Enable customizing MethodAdapterRegistry

bailan:
Make MethodAdapterRegistry customizable, so we can plugin custom argument builders and response builders.

This pull request adds MethodAdapterRegistry to RestLiConfig and updates BaseRestLiServer to use custom MethodAdapterRegistry when provided. MethodAdapterRegistry is also extracted to an interface and existing one is renamed to DefaultMethodAdapterRegistry.

### DIFF
--- a/restli-int-test/src/test/java/com/linkedin/restli/examples/method/TestCustomMethodBuildersRegistry.java
+++ b/restli-int-test/src/test/java/com/linkedin/restli/examples/method/TestCustomMethodBuildersRegistry.java
@@ -16,16 +16,27 @@
 
 package com.linkedin.restli.examples.method;
 
+import com.linkedin.data.DataMap;
 import com.linkedin.r2.RemoteInvocationException;
 import com.linkedin.restli.client.CreateIdRequest;
 import com.linkedin.restli.client.GetRequest;
 import com.linkedin.restli.client.Response;
 import com.linkedin.restli.common.IdResponse;
+import com.linkedin.restli.common.ResourceMethod;
 import com.linkedin.restli.examples.RestLiIntegrationTest;
 import com.linkedin.restli.examples.greetings.api.Greeting;
 import com.linkedin.restli.examples.greetings.api.Tone;
 import com.linkedin.restli.examples.greetings.client.GreetingsRequestBuilders;
-import com.linkedin.restli.internal.server.methods.DefaultMethodAdapterRegistry;
+import com.linkedin.restli.internal.server.RoutingResult;
+import com.linkedin.restli.internal.server.methods.DefaultMethodBuildersRegistry;
+import com.linkedin.restli.internal.server.methods.MethodBuildersRegistry;
+import com.linkedin.restli.internal.server.methods.arguments.CreateArgumentBuilder;
+import com.linkedin.restli.internal.server.methods.arguments.RestLiArgumentBuilder;
+import com.linkedin.restli.internal.server.response.ErrorResponseBuilder;
+import com.linkedin.restli.server.ErrorResponseFormat;
+import com.linkedin.restli.server.RestLiConfig;
+import com.linkedin.restli.server.RestLiRequestData;
+import com.linkedin.restli.server.RestLiRequestDataImpl;
 import org.testng.Assert;
 import org.testng.annotations.AfterClass;
 import org.testng.annotations.BeforeClass;
@@ -33,14 +44,34 @@ import org.testng.annotations.Test;
 
 
 /**
- * Integration tests for {@link DefaultMethodAdapterRegistry}.
+ * Integration tests for a custom {@link MethodBuildersRegistry}.
  */
-public class TestDefaultMethodAdapterRegistry extends RestLiIntegrationTest
+public class TestCustomMethodBuildersRegistry extends RestLiIntegrationTest
 {
+  private static final Greeting FRIENDLY = new Greeting().setMessage("Friendly").setTone(Tone.FRIENDLY);
+
   @BeforeClass
   public void initClass() throws Exception
   {
-    super.init();
+    RestLiConfig config = new RestLiConfig();
+    config.setMethodBuildersRegistry(new DefaultMethodBuildersRegistry(new ErrorResponseBuilder(ErrorResponseFormat.MESSAGE_AND_SERVICECODE)) {
+      @Override
+      public RestLiArgumentBuilder getArgumentBuilder(ResourceMethod resourceMethod) {
+        // Override the behavior of the CREATE argument builder
+        if (resourceMethod == ResourceMethod.CREATE) {
+          return new CreateArgumentBuilder() {
+            @Override
+            public RestLiRequestData extractRequestData(RoutingResult routingResult, DataMap dataMap) {
+                // Always use the FRIENDLY record regardless of the actual data
+                return new RestLiRequestDataImpl.Builder().entity(FRIENDLY).build();
+            }
+          };
+        } else {
+          return super.getArgumentBuilder(resourceMethod);
+        }
+      }
+    });
+    super.init(false, config);
   }
 
   @AfterClass
@@ -50,7 +81,7 @@ public class TestDefaultMethodAdapterRegistry extends RestLiIntegrationTest
   }
 
   @Test
-  public void testCreateAndGetTheSameRecord() throws RemoteInvocationException
+  public void testCreateAndGetOverriddenRecord() throws RemoteInvocationException
   {
     Greeting insulting = new Greeting().setMessage("Insulting").setTone(Tone.INSULTING);
     CreateIdRequest<Long, Greeting> createRequest = new GreetingsRequestBuilders().create().input(insulting).build();
@@ -64,7 +95,7 @@ public class TestDefaultMethodAdapterRegistry extends RestLiIntegrationTest
 
     Assert.assertFalse(getResponse.hasError());
     Greeting actualEntity = getResponse.getEntity();
-    Assert.assertEquals(actualEntity.getMessage(), insulting.getMessage());
-    Assert.assertEquals(actualEntity.getTone(), insulting.getTone());
+    Assert.assertEquals(actualEntity.getMessage(), FRIENDLY.getMessage());
+    Assert.assertEquals(actualEntity.getTone(), FRIENDLY.getTone());
   }
 }
--- a/restli-server/src/main/java/com/linkedin/restli/internal/server/methods/MethodAdapterRegistry.java
+++ b/restli-server/src/main/java/com/linkedin/restli/internal/server/methods/MethodAdapterRegistry.java
@@ -1,130 +1,23 @@
-/*
-   Copyright (c) 2012 LinkedIn Corp.
-
-   Licensed under the Apache License, Version 2.0 (the "License");
-   you may not use this file except in compliance with the License.
-   You may obtain a copy of the License at
-
-       http://www.apache.org/licenses/LICENSE-2.0
-
-   Unless required by applicable law or agreed to in writing, software
-   distributed under the License is distributed on an "AS IS" BASIS,
-   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-   See the License for the specific language governing permissions and
-   limitations under the License.
-*/
-
 package com.linkedin.restli.internal.server.methods;
 
 import com.linkedin.restli.common.ResourceMethod;
-import com.linkedin.restli.internal.server.methods.arguments.ActionArgumentBuilder;
-import com.linkedin.restli.internal.server.methods.arguments.BatchCreateArgumentBuilder;
-import com.linkedin.restli.internal.server.methods.arguments.BatchDeleteArgumentBuilder;
-import com.linkedin.restli.internal.server.methods.arguments.BatchGetArgumentBuilder;
-import com.linkedin.restli.internal.server.methods.arguments.BatchPatchArgumentBuilder;
-import com.linkedin.restli.internal.server.methods.arguments.BatchUpdateArgumentBuilder;
-import com.linkedin.restli.internal.server.methods.arguments.CollectionArgumentBuilder;
-import com.linkedin.restli.internal.server.methods.arguments.CreateArgumentBuilder;
-import com.linkedin.restli.internal.server.methods.arguments.GetArgumentBuilder;
-import com.linkedin.restli.internal.server.methods.arguments.PatchArgumentBuilder;
 import com.linkedin.restli.internal.server.methods.arguments.RestLiArgumentBuilder;
-import com.linkedin.restli.internal.server.methods.arguments.UpdateArgumentBuilder;
-import com.linkedin.restli.internal.server.response.ActionResponseBuilder;
-import com.linkedin.restli.internal.server.response.BatchCreateResponseBuilder;
-import com.linkedin.restli.internal.server.response.BatchDeleteResponseBuilder;
-import com.linkedin.restli.internal.server.response.BatchFinderResponseBuilder;
-import com.linkedin.restli.internal.server.response.BatchGetResponseBuilder;
-import com.linkedin.restli.internal.server.response.BatchPartialUpdateResponseBuilder;
-import com.linkedin.restli.internal.server.response.BatchUpdateResponseBuilder;
-import com.linkedin.restli.internal.server.response.CreateResponseBuilder;
-import com.linkedin.restli.internal.server.response.DeleteResponseBuilder;
-import com.linkedin.restli.internal.server.response.ErrorResponseBuilder;
-import com.linkedin.restli.internal.server.response.FinderResponseBuilder;
-import com.linkedin.restli.internal.server.response.GetAllResponseBuilder;
-import com.linkedin.restli.internal.server.response.GetResponseBuilder;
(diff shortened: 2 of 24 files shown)


restli-server/src/main/java/com/linkedin/restli/server/BaseRestLiServer.java
```diff
@@ -87,7 +89,10 @@
     _methodInvoker = new RestLiMethodInvoker(resourceFactory, engine, config.getInternalErrorMessage());
 
     _errorResponseBuilder = errorResponseBuilder;
-    _responseHandler = new RestLiResponseHandler(_errorResponseBuilder);
+    _methodAdapterRegistry = Optional
+            .ofNullable(config.getMethodAdapterRegistry())
+            .orElse(new DefaultMethodAdapterRegistry(_errorResponseBuilder));
```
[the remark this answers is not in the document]
ktsimpso: It looks like `ErrorResponseBuilder` is required for the `batch` methods.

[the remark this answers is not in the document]
bailan: I'd argue that it does not have to be implemented with `ErrorResponseBuilder`. Anyway, if you think and we can decide to have this dependency, I can add it to the interface with an initialization method like https://github.com/linkedin/rest.li/blob/7ecc85455b8a00e6d71986f97a3c02242749a9b4/restli-server/src/main/java/com/linkedin/restli/server/RestLiDocumentationRequestHandler.java#L30

restli-server/src/main/java/com/linkedin/restli/internal/server/response/RestLiResponseHandler.java
```diff
@@ -70,7 +71,7 @@ public RestLiResponseHandler()
 
   public RestLiResponseHandler(ErrorResponseBuilder errorResponseBuilder)
   {
-    this(new MethodAdapterRegistry(errorResponseBuilder), errorResponseBuilder);
+    this(new DefaultMethodAdapterRegistry(errorResponseBuilder), errorResponseBuilder);
   }
```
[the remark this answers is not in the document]
bailan: Deleted the constructor in 947ed58

restli-server/src/main/java/com/linkedin/restli/internal/server/response/RestLiResponseHandler.java
```diff
@@ -108,7 +109,7 @@ public RestLiResponseHandler build()
       }
       if (_methodAdapterRegistry == null)
       {
-        _methodAdapterRegistry = new MethodAdapterRegistry(_errorResponseBuilder);
+        _methodAdapterRegistry = new DefaultMethodAdapterRegistry(_errorResponseBuilder);
       }
       return new RestLiResponseHandler(_methodAdapterRegistry, _errorResponseBuilder);
```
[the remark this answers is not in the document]
bailan: Deleted the Builder in 947ed58

restli-server/src/main/java/com/linkedin/restli/server/BaseRestLiServer.java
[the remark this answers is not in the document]
bailan: Updated to create `DefaultMethodAdapterRegistry` using global `ErrorResponseBuilder` in `RestLiConfnig` in 947ed58

restli-server/src/main/java/com/linkedin/restli/server/RestLiConfig.java
```diff
@@ -102,6 +104,7 @@
 
   /** configuration for whether to attach stacktrace for {@link com.linkedin.r2.message.rest.RestException} */
   private boolean _writableStackTrace = true;
+  private MethodAdapterRegistry _methodAdapterRegistry = null;
```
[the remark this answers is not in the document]
bailan: This still has to be null for the get method.
```
  public MethodAdapterRegistry getMethodAdapterRegistry() {
    return Optional.ofNullable(_methodAdapterRegistry)
            .orElse(new DefaultMethodAdapterRegistry(new ErrorResponseBuilder(_errorResponseFormat)));
  }
  ```

File: restli-server/src/main/java/com/linkedin/restli/server/RestLiConfig.java [hunk 5]
```diff
@@ -593,4 +597,23 @@ public void setFillInDefaultValues(boolean fillInDefaultValues)
...
+  }
+
+  /**
+   * @return Return the custom {@link MethodAdapterRegistry} in the config. Return null if no custom
+   * {@link MethodAdapterRegistry} is provided, then the {@link DefaultMethodAdapterRegistry} will be used for setting
+   * up rest.li server.
+   */
+  public MethodAdapterRegistry getMethodAdapterRegistry() {
+    return Optional.ofNullable(_methodAdapterRegistry)
+            .orElse(new DefaultMethodAdapterRegistry(new ErrorResponseBuilder(_errorResponseFormat)));
```
ktsimpso, line 617: I assume we need to default this for backwards compatibility?

bailan: Not only for backwards compatibility. If users do not specify `MethodAdapterRegistry` in `RestLiConfig`, this guarantees the use of the default one.

bailan:
> Do we need a major version bump for these changes? (Can we try snapshotting this and testing against some repos?)

We probably need a major version bump.

I did a test by overriding `GetResponseBuilder` and verified the response was built by the overridden builder.

restli-server/src/main/java/com/linkedin/restli/internal/server/methods/MethodAdapterRegistry.java
```diff
@@ -1,130 +1,23 @@
...
-    result.put(ResourceMethod.BATCH_PARTIAL_UPDATE, new BatchPartialUpdateResponseBuilder(errorResponseBuilder));
-    result.put(ResourceMethod.BATCH_CREATE, new BatchCreateResponseBuilder(errorResponseBuilder));
-    result.put(ResourceMethod.BATCH_DELETE, new BatchDeleteResponseBuilder(errorResponseBuilder));
-    result.put(ResourceMethod.BATCH_FINDER, new BatchFinderResponseBuilder(errorResponseBuilder));
-    result.put(ResourceMethod.GET_ALL, new GetAllResponseBuilder());
-
-    return Collections.unmodifiableMap(result);
-  }
 
+public interface MethodAdapterRegistry {
```
evanw555, line 7: Simply changing this class to an interface is backward-incompatible. There may be places where this is constructed explicitly. I suggest keeping `MethodAdapterRegistry`, deprecating it (with a javadoc saying to use `DefaultMethodAdapterRegistry`), and having its implementation defer to the `DefaultMethodAdapterRegistry` (perhaps by extending it). This means we will need to introduce the interface as a new file with a unique name.

Since I don't want something like `MethodAdapterRegistryInterface`, I suggest we consider renaming this altogether. I propose using the naming convention `MethodBuildersRegistry`, `DefaultMethodBuildersRegistry`, and the now-deprecated class `MethodAdapterRegistry` will extend `DefaultMethodBuildersRegistry` (so that it relies on its functionality).

The exact naming can be adjusted, but my core idea is to use new naming.

restli-server/src/main/java/com/linkedin/restli/internal/server/response/RestLiResponseHandler.java
```diff
@@ -63,57 +64,12 @@
   private final MethodAdapterRegistry _methodAdapterRegistry;
   private final ErrorResponseBuilder _errorResponseBuilder;
 
-  public RestLiResponseHandler()
```
evanw555, line 66: This is backward-incompatible. Please deprecate the existing constructors, and have them defer to the new classes while also maintaining existing behavior. Please leave a javadoc explaining the suggested alternative for the user.

restli-server/src/main/java/com/linkedin/restli/internal/server/response/RestLiResponseHandler.java
```diff
@@ -63,57 +64,12 @@
...
-  }
-
   public RestLiResponseHandler(MethodAdapterRegistry methodAdapterRegistry, ErrorResponseBuilder errorResponseBuilder)
   {
     _methodAdapterRegistry = methodAdapterRegistry;
     _errorResponseBuilder = errorResponseBuilder;
   }
 
-  /**
-   * @deprecated Use appropriate constructors.
```
evanw555, line 83: Don't remove this just yet. Leave it deprecated. We can remove all of the deprecated stuff together in a follow-up PR.

restli-server/src/main/java/com/linkedin/restli/server/RestLiConfig.java
```diff
@@ -593,4 +597,23 @@ public void setFillInDefaultValues(boolean fillInDefaultValues)
   {
     _fillInDefaultValues = fillInDefaultValues;
   }
+
+  /**
+   * Set a custom {@link MethodAdapterRegistry} in the config.
+   *
+   * @param methodAdapterRegistry a custom to be set in the config.
+   */
+  public void setMethodAdapterRegistry(MethodAdapterRegistry methodAdapterRegistry) {
```
evanw555, line 606: If we follow my advice about altering the naming, the naming here should be consistent. The exact naming is up for debate.

restli-server/src/main/java/com/linkedin/restli/server/RestLiConfig.java
evanw555: I'm skeptical about including the default-constructed object in this override getter. I'm not sure if this will cause any real issues, but we're constructing a new `ErrorResponseBuilder` here rather than using the one defined in `BaseRestLiServer`. It may be better to return null (or an `Optional`) here, then construct the default in `BaseRestLiServer` using the existing `ErrorResponseBuilder`.

evanw555:
@bailan Do not bump the major version. Please follow my compatibility recommendations and bump the minor version (since we're introducing new interfaces and methods). We can clean up the deprecated stuff in a follow-up PR that will be bundled with the next major version.

restli-server/src/main/java/com/linkedin/restli/server/RestLiConfig.java
bailan: I think there are two folds of you comments: (1) using existing `ErrorResponseBuilder` and (2) having default in override getter (RestLiConfig).

For (1), the reason to construct the default in the config is to make sure it uses the `ErrorResponseFormat` in the config rather than the local one embedded in `ErrorResponseBuilder`. This should be the intention based on the following analysis. So, I think this would be a good opportunity to make this idea explicit.

Only three dependency chains I can find for `BaseRestLiServer`. They are end at `RestLiServer`.
```
BaseRestLiServer <- RestRestLiServer <- RestLiServer
BaseRestLiServer <- RestRestLiServer <- StreamRestLiServer <- AttachmentHandlingRestLiServer <- RestLiServer
BaseRestLiServer <- StreamRestLiServer <- AttachmentHandlingRestLiServer <- RestLiServer
```
In `RestLiServer`, the `ErrorResponseBuilder`, `RestRestLiServer` and `AttachmentHandlingRestLiServer` are initialized as follows.
```
    ErrorResponseBuilder errorResponseBuilder = new ErrorResponseBuilder(config.getErrorResponseFormat());

    _restRestLiServer = new RestRestLiServer(config,
        resourceFactory,
        engine,
        rootResources,
        errorResponseBuilder);
    _streamRestLiServer = new AttachmentHandlingRestLiServer(config,
        resourceFactory,
        engine,
        rootResources,
        errorResponseBuilder);
```
https://github.com/linkedin/rest.li/blob/710acd54c1ce4ec8cc6615ccf1d95250e8a09ed2/restli-server/src/main/java/com/linkedin/restli/server/RestLiServer.java#L112

Given that this particular builder is immutable (builder might not be the best name, but this is a different issue), this suggests that constructing it in the config or using the local one is the same. Therefore, I think the intention is to have the `DefaultMethodAdapterRegistry` using the `ErrorResponseFormat` in the config.


For (2), I could move this out of the config, but in the `BaseRestLiServer` the initialization will not use the local `ErrorResponseBuilder`.
```
_methodBuildersRegistry = Optional.ofNullable(config.getMethodAdapterRegistry())
            .orElse(new DefaultMethodAdapterRegistry(new ErrorResponseBuilder(_errorResponseFormat)));
```

restli-server/src/main/java/com/linkedin/restli/internal/server/methods/MethodBuildersRegistry.java
```diff
@@ -0,0 +1,23 @@
+package com.linkedin.restli.internal.server.methods;
+
+import com.linkedin.restli.common.ResourceMethod;
+import com.linkedin.restli.internal.server.methods.arguments.RestLiArgumentBuilder;
+import com.linkedin.restli.internal.server.response.RestLiResponseBuilder;
+
```
junchuanwang, line 6: java doc for public class please

restli-int-test/src/test/java/com/linkedin/restli/examples/method/TestCustomMethodBuildersRegistry.java
```diff
@@ -0,0 +1,101 @@
...
+
+  @BeforeClass
+  public void initClass() throws Exception
+  {
+    RestLiConfig config = new RestLiConfig();
+    config.setMethodBuildersRegistry(new DefaultMethodBuildersRegistry(new ErrorResponseBuilder(ErrorResponseFormat.MESSAGE_AND_SERVICECODE)) {
+      @Override
+      public RestLiArgumentBuilder getArgumentBuilder(ResourceMethod resourceMethod) {
+        // Override the behavior of the CREATE argument builder
+        if (resourceMethod == ResourceMethod.CREATE) {
```
junchuanwang, line 61: newline for curly brace please

restli-int-test/src/test/java/com/linkedin/restli/examples/method/TestCustomMethodBuildersRegistry.java
```diff
@@ -0,0 +1,101 @@
...
+          return new CreateArgumentBuilder() {
+            @Override
+            public RestLiRequestData extractRequestData(RoutingResult routingResult, DataMap dataMap) {
+                // Always use the FRIENDLY record regardless of the actual data
+                return new RestLiRequestDataImpl.Builder().entity(FRIENDLY).build();
+            }
+          };
+        } else {
```
junchuanwang, line 69: else needs to be in separate line

junchuanwang (commented):
please follow Evan's suggestion for backward incompatible changes

restli-server/src/main/java/com/linkedin/restli/server/RestLiConfig.java
evanw555: You bring up a good point about "making this idea explicit". The issue is that if we don't use the error response builder in the constructor, then the API is misleading. If the constructor takes in an error response builder, then it implies that this is the one that will be used.

In order to preserve consistency, we should go with one of two routes:
1. Remove the `ErrorResponseBuilder` argument from the constructor and the constructor of all subclasses. The config argument is present, so response builders should just be constructed using the error response format config. Since this is backward-compatible (users may define base classes extending BaseRestLiServer and pass up their own response builder). This is backward-incompatible, so instead please deprecate the existing constructors (with javadoc notes) and create new package-private constructors without the error response builder argument.
2. Follow my original suggestion and use the error response builder provided in the constructor.

I suggest going with option 1, since option 2 would result in a config method that returns an `Optional`, and it would be incompatible to go back.

restli-server/src/main/java/com/linkedin/restli/internal/server/methods/MethodBuildersRegistry.java
```diff
@@ -0,0 +1,23 @@
...
+public interface MethodBuildersRegistry {
```
evanw555, line 7: After some offline discussion, I've decided that the new preferred naming for this interface is `MethodAdapterProvider`. Please update all subclasses/tests/methods accordingly.

Sorry for the overhead, just trying to get the naming right 😃

bailan: Sounds good. Updated in dce626d

restli-int-test/src/test/java/com/linkedin/restli/examples/method/TestCustomMethodBuildersRegistry.java
bailan: Updated

restli-int-test/src/test/java/com/linkedin/restli/examples/method/TestCustomMethodBuildersRegistry.java
bailan: Updated

restli-server/src/main/java/com/linkedin/restli/server/RestLiConfig.java
bailan: Thank you for the suggestions. I updated code with option 1 dce626d

evanw555:
LGTM, I'm going to run internal wc-tests today and I'll approve/merge once those pass

evanw555:
Tests pass. Going to merge.

evanw555 (approved):

PR #681, opened 2021-08-30
DarkClusterConfig's dispatcherOutboundTargetRate: int -> float

lesterhaynes:
No reason this shouldn't be a float. This will support usecases with low qps.


d2-schemas/src/main/pegasus/com/linkedin/d2/DarkClusterConfig.pdl
```diff
@@ -13,7 +13,7 @@ record DarkClusterConfig {
   /**
    * Desired query rate to be maintained to the dark cluster per dark cluster host by the CONSTANT_QPS strategy. Measured in qps.
    */
-  dispatcherOutboundTargetRate: int = 0
+  dispatcherOutboundTargetRate: float = 0
```
rickzx, line 16: Can we add some unit tests to make sure that fractional QPS works?

rickzx (changes requested):
Please also update the changelog and gradle.properties.

PR #682, opened 2021-08-31
Expose an API to build a URI without query params. Expose a local attr for passing query params for in-process calls.

karthikrg:
This helps avoid query param ser/deser for in-process rest.li calls.


mchen07 (approved):
Looks good to me.

PR #684, opened 2021-09-01
ConstantQpsDarkClusterStrategy post-prod fixes

lesterhaynes:
- DarkClusterConfig's dispatcherOutboundTargetRate: int -> float
This change enables low qps use cases to more finely configure their per-host inbound dark qps.

- ConstantQpsRateLimiter - Introduce randomness while maintaining constant per period rate
This addresses a thundering herd problem resulting from a central signaling of many replicas simultaneously running ConstantQpsRateLimiter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,9 @@ and what APIs have changed, if applicable.
 ## [Unreleased]
 
 ## [29.21.5] - 2021-08-31
-- DarkClusterConfig's dispatcherOutboundTargetRate: int -> float
+- ConstantQpsDarkClusterStrategy post-prod fixes
+  - DarkClusterConfig's dispatcherOutboundTargetRate: int -> float
+  - ConstantQpsRateLimiter - Introduce randomness while maintaining constant per period rate
 
 ## [29.21.4] - 2021-08-30
 - Expose an API to build a URI without query params. Expose a local attr for passing query params for in-process calls. 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ and what APIs have changed, if applicable.
 
 ## [29.21.5] - 2021-08-31
 - ConstantQpsDarkClusterStrategy post-prod fixes
-  - DarkClusterConfig's dispatcherOutboundTargetRate: int -> float
+  - Change the type of dispatcherOutboundTargetRate in DarkClusterConfig.pdl from int to float.
   - ConstantQpsRateLimiter - Introduce randomness while maintaining constant per period rate
 
 ## [29.21.4] - 2021-08-30
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.21.5] - 2021-08-31
+- DarkClusterConfig's dispatcherOutboundTargetRate: int -> float
+
 ## [29.21.4] - 2021-08-30
 - Expose an API to build a URI without query params. Expose a local attr for passing query params for in-process calls. 
 
@@ -5070,7 +5073,8 @@ patch operations can re-use these classes for generating patch messages.
 
 ## [0.14.1]
 
-[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.21.4...master
+[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.21.5...master
+[29.21.5]: https://github.com/linkedin/rest.li/compare/v29.21.4...v29.21.5
 [29.21.4]: https://github.com/linkedin/rest.li/compare/v29.21.3...v29.21.4
 [29.21.3]: https://github.com/linkedin/rest.li/compare/v29.21.2...v29.21.3
 [29.21.2]: https://github.com/linkedin/rest.li/compare/v29.21.1...v29.21.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.21.6] - 2021-09-09
+- ConstantQpsDarkClusterStrategy post-prod fixes
+   - Change the type of dispatcherOutboundTargetRate in DarkClusterConfig.pdl from int to float.
+   - ConstantQpsRateLimiter - Introduce randomness while maintaining constant per period rate
+
 ## [29.21.5] - 2021-09-09
 - Fix a bug in DataTranslator where accessing non-existent fields under avro 1.10+ throws
 
@@ -5073,7 +5078,8 @@ patch operations can re-use these classes for generating patch messages.
 
 ## [0.14.1]
 
-[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.21.5...master
+[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.21.6...master
+[29.21.6]: https://github.com/linkedin/rest.li/compare/v29.21.5...v29.21.6
 [29.21.5]: https://github.com/linkedin/rest.li/compare/v29.21.4...v29.21.5
 [29.21.4]: https://github.com/linkedin/rest.li/compare/v29.21.3...v29.21.4
 [29.21.3]: https://github.com/linkedin/rest.li/compare/v29.21.2...v29.21.3
--- a/r2-core/src/main/java/com/linkedin/r2/transport/http/client/ConstantQpsRateLimiter.java
+++ b/r2-core/src/main/java/com/linkedin/r2/transport/http/client/ConstantQpsRateLimiter.java
@@ -16,9 +16,11 @@
 
 package com.linkedin.r2.transport.http.client;
 
+import com.linkedin.r2.transport.http.client.ratelimiter.Rate;
 import com.linkedin.r2.transport.http.client.ratelimiter.RateLimiterExecutionTracker;
 import com.linkedin.util.clock.Clock;
 import java.time.temporal.ChronoUnit;
+import java.util.Random;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -67,6 +69,7 @@ public class ConstantQpsRateLimiter extends SmoothRateLimiter
   private static class UnboundedRateLimiterExecutionTracker implements RateLimiterExecutionTracker
   {
     private final AtomicBoolean _paused = new AtomicBoolean(true);
+    private final Random _random = new Random();
 
     public int getPending()
     {
@@ -96,6 +99,11 @@ public class ConstantQpsRateLimiter extends SmoothRateLimiter
     public int getMaxBuffered()
     {
       return Integer.MAX_VALUE;
+    }
+
+    public int getNextExecutionDelay(Rate rate)
+    {
+      return _random.nextInt(Math.max(1, (int) (rate.getPeriodRaw() / rate.getEventsRaw())));
     }
   }
 }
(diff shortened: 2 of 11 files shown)


r2-core/src/main/java/com/linkedin/r2/transport/http/client/ConstantQpsRateLimiter.java
```diff
@@ -97,5 +100,10 @@ public int getMaxBuffered()
     {
       return Integer.MAX_VALUE;
     }
+
+    public int getNextExecutionDelay(Rate rate)
```
rickzx, line 104: This seems to me only adds a delay before the first execution. The naming seems confusing here.  Maybe `getInitialDelay()`?

r2-core/src/main/java/com/linkedin/r2/transport/http/client/SmoothRateLimiter.java
```diff
@@ -225,6 +224,7 @@ public int getPendingTasksCount(){
     private int _permitAvailableCount;
     private int _permitsInTimeFrame;
     private long _nextScheduled;
+    private boolean _wasDelayed;
```
rickzx, line 227: Shall we initialize this to `false` in the constructor?

r2-core/src/main/java/com/linkedin/r2/transport/http/client/SmoothRateLimiter.java
```diff
@@ -272,6 +272,13 @@ public void loop()
 
       if (_permitAvailableCount > 0)
       {
+        if (!_wasDelayed)
+        {
+          _wasDelayed = true;
+          _scheduler.schedule(this::loop, _executionTracker.getNextExecutionDelay(_rate), TimeUnit.MILLISECONDS);
+          return;
+        }
+        _wasDelayed = false;
```
rickzx, line 281: And we don't need to set it here right?

r2-core/src/main/java/com/linkedin/r2/transport/http/client/ratelimiter/RateLimiterExecutionTracker.java
```diff
@@ -53,4 +53,9 @@
      * @return maximum number of callbacks that can be stored in the RateLimiter
      */
     int getMaxBuffered();
+
+    /**
+     * @return amount of delay to be incurred before executing the next callback, based on provided rate
+     */
+    int getNextExecutionDelay(Rate rate);
```
rickzx, line 60: We can implement `return 0` as the default method here to make the interface backward compatible. And also, we don't need to override in the `SmoothRateLimiter` class

CHANGELOG.md
```diff
@@ -14,6 +14,11 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.21.5] - 2021-08-31
+- ConstantQpsDarkClusterStrategy post-prod fixes
+  - DarkClusterConfig's dispatcherOutboundTargetRate: int -> float
```
rickzx, line 19: Let's make the changelog more obvious. e.g. Change the type of dispatcherOutboundTargetRate in DarkClusterConfig.pdl from int to float.

r2-core/src/test/java/com/linkedin/r2/transport/http/client/ratelimiter/TestConstantQpsRateLimiter.java
```diff
@@ -88,21 +90,57 @@ public void eventLoopStopsWhenTtlExpiresAllRequests()
...
+    ClockedExecutor executor = new ClockedExecutor();
+    ConstantQpsRateLimiter rateLimiter =
+        new ConstantQpsRateLimiter(executor, executor, executor, TestEvictingCircularBuffer.getBuffer(executor));
+    rateLimiter.setRate(200d, ONE_SECOND, 1);
+    rateLimiter.setBufferCapacity(1);
+    TattlingCallback<None> tattler = new TattlingCallback<>(executor);
+    rateLimiter.submit(tattler);
+    executor.runFor(ONE_SECOND * TEST_NUM_CYCLES);
+    long prevTime = 0;
+    ArrayList<Long> timeDeltas = new ArrayList<>();
```
rickzx, line 105: ArrayList -> List for the type

r2-core/src/test/java/com/linkedin/r2/transport/http/client/ratelimiter/TestConstantQpsRateLimiter.java
```diff
@@ -88,21 +90,57 @@ public void eventLoopStopsWhenTtlExpiresAllRequests()
...
+    for (Long stamp : tattler.getOccurrences())
+    {
+      timeDeltas.add(stamp - prevTime);
+      prevTime = stamp;
+    }
+    // Ensure variance up to 10 possible time deltas given a rate of 200 requests per second
+    HashSet<Long> uniqueTimeDeltas = new HashSet<>(timeDeltas);
```
rickzx, line 112: Ditto. HashSet -> Set

r2-core/src/test/java/com/linkedin/r2/transport/http/client/ratelimiter/TestConstantQpsRateLimiter.java
```diff
@@ -88,21 +90,57 @@ public void eventLoopStopsWhenTtlExpiresAllRequests()
...
+    assert(uniqueTimeDeltas.size() > 5 && uniqueTimeDeltas.size() < 11);
+  }
+
   private static class TattlingCallback<T> implements Callback<T>
   {
-    private  AtomicInteger _interactCount = new AtomicInteger();
+    private AtomicInteger _interactCount = new AtomicInteger();
+    private ArrayList<Long> _occurrences = new ArrayList<>();
```
rickzx, line 119: Ditto

r2-core/src/test/java/com/linkedin/r2/transport/http/client/ratelimiter/TestConstantQpsRateLimiter.java
```diff
@@ -88,21 +90,57 @@ public void eventLoopStopsWhenTtlExpiresAllRequests()
...
       _interactCount.incrementAndGet();
+      _occurrences.add(_clock.currentTimeMillis());
     }
 
     public int getInteractCount()
     {
       return _interactCount.intValue();
     }
+
+    public ArrayList<Long> getOccurrences()
```
rickzx, line 141: Ditto

rickzx (changes requested):

r2-core/src/main/java/com/linkedin/r2/transport/http/client/SmoothRateLimiter.java
lesterhaynes: Same same. I can explicitly set to false in the constructor if that's the preferred style.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/SmoothRateLimiter.java
lesterhaynes: There might be a more readable way of doing this, totally open to suggestions. The whole purpose of lines 275 - 281 is to stop prematurely every other iteration of the loop, taking care to schedule the next loop iteration after the delay. This works because the delay is always set such that there's time enough in the rest of the period to honor all the permits.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/ratelimiter/RateLimiterExecutionTracker.java
lesterhaynes: Good idea

CHANGELOG.md
lesterhaynes: Changelog updated verbatim

r2-core/src/test/java/com/linkedin/r2/transport/http/client/ratelimiter/TestConstantQpsRateLimiter.java
lesterhaynes: Alrighty

r2-core/src/test/java/com/linkedin/r2/transport/http/client/ratelimiter/TestConstantQpsRateLimiter.java
lesterhaynes: 🙌

r2-core/src/test/java/com/linkedin/r2/transport/http/client/ratelimiter/TestConstantQpsRateLimiter.java
lesterhaynes: 👍

r2-core/src/test/java/com/linkedin/r2/transport/http/client/ratelimiter/TestConstantQpsRateLimiter.java
lesterhaynes: 👌

r2-core/src/main/java/com/linkedin/r2/transport/http/client/ConstantQpsRateLimiter.java
lesterhaynes: The intent of this method is indeed to provide a delay in ms to be incurred before every execution.

I'd like to get your opinion here. Given that this method requires intimate knowledge of a Rate, it may make more sense to make it a static method in Rate called "getRandomAcceptableDelay" or something like that. Basically something that clearly describes this delay as a random delay no greater than the amount of free time in a given period while still satisfying the rate.

A degree of randomness is ideal in any traffic synthetically generated, and it's possible that there may be synthetic use cases down the line that aren't handled by ConstantQpsRateLimiter.

If we did go this route, maybe we could then convert RateLimiterExecutionTracker.getNextExecutionDelay into a boolean getter that is used to decide whether or not we call the new static method in Rate.

r2-core/src/test/java/com/linkedin/r2/transport/http/client/ratelimiter/TestConstantQpsRateLimiter.java
```diff
@@ -88,21 +92,57 @@ public void eventLoopStopsWhenTtlExpiresAllRequests()
     Assert.assertSame(executor.getExecutedTaskCount(), prevTaskCount);
   }
 
+  @Test
+  public void ensureRandomButConstantRate()
+  {
+    ClockedExecutor executor = new ClockedExecutor();
+    ConstantQpsRateLimiter rateLimiter =
+        new ConstantQpsRateLimiter(executor, executor, executor, TestEvictingCircularBuffer.getBuffer(executor));
+    rateLimiter.setRate(200d, ONE_SECOND, 1);
```
rickzx, line 101: If i change this to 0.05d per second, and have executor run for 200s, the test was stopped after just 1 execution. I think we should use a different executor for circular buffer as in your previous test cases: `TestEvictingCircularBuffer.getBuffer(circularBufferExecutor));`

rickzx (changes requested):

r2-core/src/main/java/com/linkedin/r2/transport/http/client/ConstantQpsRateLimiter.java
rickzx: I would prefer to keep `Rate` as is. It's more like a configuration class to me

r2-core/src/test/java/com/linkedin/r2/transport/http/client/ratelimiter/TestConstantQpsRateLimiter.java
lesterhaynes: Done

rickzx (approved):

PR #686, opened 2021-09-09
dont access avro record fields that may not exist in DataTranslator

radai-rosenblatt:
avro 1.10+ is more strict and throws an exception when trying to access a field not in the record schema.


mchen07 (approved):
LGTM

mchen07:
Please update CHANGELOG and version to get a version released.

mchen07 (approved):

PR #687, opened 2021-09-09
Allow disabling of load balancing for a specific host.

zhang-chris:
Allow disabling of load balancing for a specific host. Verified by JMX with local deploy of client, server, and zk that it's working as expected. 

Will update changelog and gradle.properties at the end when ready to merge.

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/LoadBalancerServer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/LoadBalancerServer.java
@@ -49,6 +49,23 @@ public interface LoadBalancerServer
                     boolean doNotSlowStart,
                     Callback<None> callback);
 
+  /**
+   * 1. Gets existing {@link UriProperties} for given cluster and add doNotSlowStart property
+   * for given uri.
+   * 2. Mark down existing node.
+   * 3. Mark up new node for uri with modified UriProperties and given partitionDataMap.
+   *
+   * @param uriSpecificPropertiesName Name of uri specific property to add.
+   * @param uriSpecificPropertiesValue Value of uri specific property to add.
+   */
+  void addUriSpecificProperty(String clusterName,
+                              String operationName,
+                              URI uri,
+                              Map<Integer, PartitionData> partitionDataMap,
+                              String uriSpecificPropertiesName,
+                              Object uriSpecificPropertiesValue,
+                              Callback<None> callback);
+
   void start(Callback<None> callback);
 
   void shutdown(Callback<None> callback);
--- a/d2/src/main/java/com/linkedin/d2/balancer/LoadBalancerServer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/LoadBalancerServer.java
@@ -50,7 +50,7 @@ public interface LoadBalancerServer
                     Callback<None> callback);
 
   /**
-   * 1. Gets existing {@link UriProperties} for given cluster and add doNotSlowStart property
+   * 1. Gets existing {@link UriProperties} for given cluster and add property
    * for given uri.
    * 2. Mark down existing node.
    * 3. Mark up new node for uri with modified UriProperties and given partitionDataMap.
--- a/d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
@@ -31,6 +31,7 @@ import java.util.concurrent.CancellationException;
 import com.linkedin.common.callback.Callback;
 import com.linkedin.common.util.None;
 import com.linkedin.d2.balancer.properties.PartitionData;
+import com.linkedin.d2.balancer.properties.PropertyKeys;
 import com.linkedin.d2.balancer.properties.UriProperties;
 import com.linkedin.d2.balancer.util.partitions.DefaultPartitionAccessor;
 import com.linkedin.d2.discovery.stores.zk.ZooKeeperEphemeralStore;
@@ -343,6 +344,27 @@ public class ZooKeeperAnnouncer
       }
     });
     _log.info("changeWeight called for uri = {}.", _uri);
+  }
+
+  public synchronized void doNotLoadBalance(final Callback<None> callback, boolean doNotLoadBalance)
+  {
+    _server.addUriSpecificProperty(_cluster, "doNotLoadBalance", _uri, _partitionDataMap, PropertyKeys.DO_NOT_LOAD_BALANCE, doNotLoadBalance, new Callback<None>()
+    {
+      @Override
+      public void onError(Throwable e)
+      {
+        _log.warn("doNotLoadBalance for uri = {} failed.", _uri);
+        callback.onError(e);
+      }
+
+      @Override
+      public void onSuccess(None result)
+      {
+        _log.info("doNotLoadBalance for uri = {} succeeded.", _uri);
+        callback.onSuccess(result);
+      }
+    });
+    _log.info("doNotLoadBalance called for uri = {}.", _uri);
   }
 
   public String getCluster()
--- a/d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
@@ -346,7 +346,7 @@ public class ZooKeeperAnnouncer
     _log.info("changeWeight called for uri = {}.", _uri);
   }
 
-  public synchronized void doNotLoadBalance(final Callback<None> callback, boolean doNotLoadBalance)
+  public synchronized void setDoNotLoadBalance(final Callback<None> callback, boolean doNotLoadBalance)
   {
     _server.addUriSpecificProperty(_cluster, "doNotLoadBalance", _uri, _partitionDataMap, PropertyKeys.DO_NOT_LOAD_BALANCE, doNotLoadBalance, new Callback<None>()
     {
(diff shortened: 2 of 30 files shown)


d2/src/main/java/com/linkedin/d2/balancer/simple/UriLoadBalancerSubscriber.java
```diff
@@ -75,7 +76,16 @@ protected void handlePut(final String cluster, final UriProperties uriProperties
...
+
+            Optional<Map<String, Object>> newUriSpecificProperties = Optional.ofNullable(uriProperties.getUriSpecificProperties())
+              .map(uriSpecificProperties -> uriSpecificProperties.get(uri));
+
+            Optional<Map<String, Object>> oldUriSpecificProperties = Optional.ofNullable(_simpleLoadBalancerState.getUriProperties(clusterName))
+              .map(LoadBalancerStateItem::getProperty)
+              .map(UriProperties::getUriSpecificProperties)
+              .map(uriSpecificProperties -> uriSpecificProperties.get(uri));
+
+            if (client == null || !client.getPartitionDataMap().equals(partitionDataMap) || !newUriSpecificProperties.equals(oldUriSpecificProperties))
```
rickzx, line 88: Do we need to add this additional check `!newUriSpecificProperties.equals(oldUriSpecificProperties)`? I think adding uri specific properties will involve marking down and marking up nodes anyways.

d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/StateUpdater.java
```diff
@@ -318,13 +318,18 @@ else if (trackerClientState.getHealthScore() < MAX_HEALTH_SCORE
         {
           trackerClientState.setHealthState(TrackerClientState.HealthState.NEUTRAL);
         }
+
+        if (trackerClient.doNotLoadBalance())
```
rickzx, line 322: Can we skip the above health score logic (line 267-320) if the client has set doNotLoadBalancer? We can just keep HealthState.HEALTHY right?

d2/src/main/java/com/linkedin/d2/jmx/ZooKeeperAnnouncerJmxMXBean.java
```diff
@@ -47,6 +47,13 @@
    */
   void changeWeight(boolean doNotSlowStart) throws PropertyStoreException;
 
+  /**
+   * Set {@link com.linkedin.d2.balancer.properties.PropertyKeys#DO_NOT_LOAD_BALANCE} for a given uri.
+   *
+   * @param doNotLoadBalance Flag to let clients know if load balancing should be disabled for a host.
+   */
+  void doNotLoadBalance(boolean doNotLoadBalance) throws PropertyStoreException;
```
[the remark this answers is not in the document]
rickzx: Or setDoNotLoadBalance

d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/StateUpdater.java
zhang-chris: good catch, added

d2/src/main/java/com/linkedin/d2/jmx/ZooKeeperAnnouncerJmxMXBean.java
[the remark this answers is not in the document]
zhang-chris: sure will rename to setDoNotLoadBalance. I'd rather keep the explicit "do not" to make it clear this is not default behavior. 

The method at line 48 is different because it's changing the weight, not only disabling/enabling slow start.

d2/src/main/java/com/linkedin/d2/balancer/simple/UriLoadBalancerSubscriber.java
zhang-chris: I'm not sure why but when testing locally the client did not receive the markDown update, only markUp. I'd need to ask some one more familiar with the zookeeper internals to understand why this is happening. But I can reproduce it and confirm the node is actually marked down/up in zk by looking at cversion. So looks like the check is needed.

d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/StateUpdater.java
rickzx: I mean, we don't need to execute through line 278-230 if doNotLoadBalance is set, do we? Can we check the flag first at line 278
```
if (trackerClient.doNotLoadBalancer()) {
    // set health state and health score
} else {
    // normal health score adjustment logic
}
```

rickzx: Also have we agreed on not load balancing on error rate as well? Looks like we're turning off load balancing completely.

zhang-chris: We agreed to not load balance on error, no? 

I see your point, we can skip that logic.

d2/src/main/java/com/linkedin/d2/balancer/LoadBalancerServer.java
```diff
@@ -49,6 +49,23 @@ void changeWeight(String clusterName,
                     boolean doNotSlowStart,
                     Callback<None> callback);
 
+  /**
+   * 1. Gets existing {@link UriProperties} for given cluster and add doNotSlowStart property
```
bohhyang, line 53: this comment should not have "add doNotSlowStart property"

d2/src/test/java/com/linkedin/d2/balancer/strategies/relative/StateUpdaterTest.java
```diff
@@ -163,8 +164,44 @@ public void testInitializePartitionWithSlowStartInitialHealthScore(boolean doNot
...
+
+    assertEquals(_stateUpdater.getPointsMap(DEFAULT_PARTITION_ID).get(trackerClient0.getUri()).intValue(),
+                 (int) (initialHealthScore * RelativeLoadBalancerStrategyFactory.DEFAULT_POINTS_PER_WEIGHT));
+    assertFalse(_stateUpdater.getPartitionState(DEFAULT_PARTITION_ID).getTrackerClientStateMap().get(trackerClient0).isUnhealthy());
+
+    if (!doNotLoadBalance)
+    {
+      assertEquals(_stateUpdater.getPointsMap(DEFAULT_PARTITION_ID).get(trackerClient1.getUri()).intValue(),
+                   (int) (initialHealthScore * RelativeLoadBalancerStrategyFactory.DEFAULT_POINTS_PER_WEIGHT));
+      assertFalse(_stateUpdater.getPartitionState(DEFAULT_PARTITION_ID).getTrackerClientStateMap().get(trackerClient0).isUnhealthy());
```
bohhyang, line 193: should this be "get(tracker**Client1**)" (instead of tracker**Client0**)?

d2/src/test/java/com/linkedin/d2/balancer/strategies/relative/StateUpdaterTest.java
```diff
@@ -163,8 +164,44 @@ public void testInitializePartitionWithSlowStartInitialHealthScore(boolean doNot
...
+    assertTrue(_stateUpdater.getPointsMap(DEFAULT_PARTITION_ID).isEmpty(), "There should be no state before initialization");
+
+    _stateUpdater.updateState(new HashSet<>(trackerClients), DEFAULT_PARTITION_ID, DEFAULT_CLUSTER_GENERATION_ID, false);
+
+    final TrackerClient trackerClient0 = trackerClients.get(0);
+    final TrackerClient trackerClient1 = trackerClients.get(1);
+
+    assertEquals(_stateUpdater.getPointsMap(DEFAULT_PARTITION_ID).get(trackerClient0.getUri()).intValue(),
+                 (int) (initialHealthScore * RelativeLoadBalancerStrategyFactory.DEFAULT_POINTS_PER_WEIGHT));
+    assertFalse(_stateUpdater.getPartitionState(DEFAULT_PARTITION_ID).getTrackerClientStateMap().get(trackerClient0).isUnhealthy());
```
bohhyang, line 187: nitpick: this looks like fit into a simple helper method to DRY for line 191-193 and 197-198 to avoid errors.

d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/StateUpdater.java
bohhyang: [our discussion ](https://docs.google.com/document/d/1C2IVFf74zjFO1H3Mgk0JGpmJumGjN7Rg2K23_gGhFOc/edit?disco=AAAAOAnK3uM) over the design doc was to load balance on error rate as the bottom line guard. But it seems you and Dyno team had some new updates.

zhang-chris: Added some info, btw I don't think we should include links to internal docs since this is a public repo.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -345,6 +346,27 @@ public void onSuccess(None result)
     _log.info("changeWeight called for uri = {}.", _uri);
   }
 
+  public synchronized void setDoNotLoadBalance(final Callback<None> callback, boolean doNotLoadBalance)
+  {
+    _server.addUriSpecificProperty(_cluster, "doNotLoadBalance", _uri, _partitionDataMap, PropertyKeys.DO_NOT_LOAD_BALANCE, doNotLoadBalance, new Callback<None>()
```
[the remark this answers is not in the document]
zhang-chris: good suggestion

d2/src/test/java/com/linkedin/d2/balancer/strategies/relative/StateUpdaterTest.java
zhang-chris: good catch

bohhyang:
LGTM. thanks.

bohhyang (approved):

rickzx (approved):

PR #689, opened 2021-09-13
Add a pegasus plugin config to use case sensitive path in datetemplate generation and rest client generation

junchuanwang:
In previous PR

https://github.com/linkedin/rest.li/commit/e720a5bf4ff110aa0981c5e5f1ab0f7b01367688#diff-881c39f18a279b2c4590938b0e7eb00edaf336ee6ad1291488f80b7f589d13c9R102

The contributor forgot to add the config into PegasusPlugin so user can opt in or opt out.

This PR is to add that back.


nickibi (approved):
LGTM
This config is used to opt-out this feature, right?

junchuanwang:
> LGTM
> This config is used to opt-out this feature, right?

Correct

nickibi:
Will you raise another PR for CHANGELOG?

junchuanwang:
> Will you raise another PR for CHANGELOG?

Yeah I might need to include it.

Initially I think we might not needed, since this is a bug fix, the original changelog has included it. But I checked the original changelog, it does not mention pegasus plugin config thing.

PR #690, opened 2021-09-14
Make extensions directory as the resource root

nickibi:
This change is making /extensions directory as the resource root. It was a miss when we introduced /extensions directory.

Tested the snapshot in a MP.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,11 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
-- Add a pegasus plugin config to use case sensitive path in dataTemplate generation and rest client generation
 
+
+## [29.22.1] - 2021-09-13
+- Mark the `extensions` directory as a resource root in the Gradle plugin.
+- Add a pegasus plugin config to use case sensitive path in dataTemplate generation and rest client generation
 
 ## [29.22.0] - 2021-09-09
 - Allow customizing `MethodAdapterRegistry` (now called `MethodAdapterProvider`) via `RestLiConfig`.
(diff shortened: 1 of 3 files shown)


evanw555:
You need to update the exclusion logic (down a few lines from your impl) to make sure the extensions don't get packaged into the default JAR as resources.

nickibi:
> You need to update the exclusion logic (down a few lines from your impl) to make sure the extensions don't get packaged into the default JAR as resources.

Good catch, updated accordingly.

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.22.1] - 2021-09-13
+- Add extensions directory to the resource root.
```
evanw555, line 18: This isn't quite accurate. It's not "adding the directory to the resource root", it's marking the directory _as a_ resource root. A given source set has multiple resource roots (directories that are treated as the root directory containing resources to be copied over in the build/publish process). By default, only `resources` it treated as a resource root for a given source set, but here we're also adding `pegasus` and `extensions` (or whatever the user configures) as extra resource roots.

I would also clarify that this is done by the Gradle plugin (I doubt that it will affect anyone, but it's best to keep things accurate and specific).

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -2296,11 +2296,18 @@ private static void configureDataSchemaResourcesRoot(Project project, SourceSet
...
+      final File extensionsSchemaRoot = project.file(extensionsSchemaPath);
+      sourceDirectorySet.srcDir(extensionsSchemaPath);
+      project.getLogger().info("Adding resource root '{}'", extensionsSchemaPath);
+
+      // Exclude the data schema and extensions schema directory from being copied into the default Jar task
       sourceDirectorySet.getFilter().exclude(fileTreeElement -> {
         final File file = fileTreeElement.getFile();
         // Traversal starts with the children of a resource root, so checking the direct parent is sufficient
-        final boolean exclude = dataSchemaRoot.equals(file.getParentFile());
+        final boolean excludeDataSchema = dataSchemaRoot.equals(file.getParentFile());
```
evanw555, line 2308: nit: these variable names are misleading. It seems to indicate whether data schema/extension dirs will be excluded, but obviously we always want to do that. I would change to something like `underDataSchemaRoot` and `underExtensionsRoot`, which indicates whether this file is the child of the data schema/extension dirs.

CHANGELOG.md
nickibi: Updated

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
nickibi: Updated.

evanw555 (approved):

PR #692, opened 2021-09-15
Change JSR330Adapter to use getDeclaredConstructors

jasongilanfarr:
In section 4.1 of JSR 330, there is explicitly a package protected
constructor: https://jcp.org/en/jsr/detail?id=330#orig

This is also consistent with guice and others as well as errorprone:
https://errorprone.info/bugpattern/UnnecessarilyVisible

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
+- Improve support for JSR330 by allowing package protected constructors annotated with @Inject
 
 ## [29.22.2] - 2021-09-20
 - Add server config support to define supported accept types
@@ -37,7 +38,7 @@ and what APIs have changed, if applicable.
 - Fix a bug in `DataTranslator` where accessing non-existent fields under avro 1.10+ throws an exception.
 
 ## [29.21.4] - 2021-08-30
-- Expose an API to build a URI without query params. Expose a local attr for passing query params for in-process calls. 
+- Expose an API to build a URI without query params. Expose a local attr for passing query params for in-process calls.
 
 ## [29.21.3] - 2021-08-25
 - Fix a bug in `SmoothRateLimiter` where `getEvents` will always return `0`.
@@ -62,7 +63,7 @@ and what APIs have changed, if applicable.
 
 ## [29.19.16] - 2021-08-09
 - Add support for resolving from multiple schema source directories.
-  - This change also introduces the concept of "source" and "resolver" directories when 
+  - This change also introduces the concept of "source" and "resolver" directories when
     creating a `DataSchemaParser`. "Source" directories are used to parse/load the input
     schemas, while the "resolver" directories will only be used for resolving referenced
     schemas.
@@ -86,16 +87,16 @@ serializing/deserializing them for requests that are executed in-process.
 
 ## [29.19.11] - 2021-07-20
 - Add compatibility level config for extension schema compatibility check.
-   - "pegasusPlugin.extensionSchema.compatibility" is the compatibility level config for extension schema compatibility check. 
+   - "pegasusPlugin.extensionSchema.compatibility" is the compatibility level config for extension schema compatibility check.
       It supports following 4 levels:
      - "off": the extension schema compatibility check will not be run.
      - "ignore": the extension schema compatibility check will run, but it allows backward incompatible changes.
      - "backwards": Changes that are considered backwards compatible will pass the check, otherwise changes will fail the check.
      - "equivalent": No changes to extension schemas will pass.
    - If this config is not provided by users, by default the extension schema compatibility check is using "backwards".
-   - How to use it: users could add 'pegasusPlugin.extensionSchema.compatibility=<compatibility level>' in the gradle.properties file 
+   - How to use it: users could add 'pegasusPlugin.extensionSchema.compatibility=<compatibility level>' in the gradle.properties file
      or directly add this property '-PpegasusPlugin.extensionSchema.compatibility=<compatibility level>' to the gradle build.
-   
+
 - Revert "Relax extension schema check to make '@extension' annotation is optional for 1-to-1 injections."
 
 ## [29.19.10] - 2021-07-16
@@ -255,7 +256,7 @@ serializing/deserializing them for requests that are executed in-process.
 - Add `UnionTemplate.memberKeyName()` to directly return the key name for a union member.
 
 ## [29.15.1] - 2021-02-18
-- Cleanup compression code to reduce duplication and minimize memcopies 
+- Cleanup compression code to reduce duplication and minimize memcopies
 
 ## [29.15.0] - 2021-02-17
 - Always enable client compression filter so that responses can be decompressed. If the request already has an accept encoding header set do not overwrite it.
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -13,7 +13,7 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
-
+- Improve support for JSR330 by allowing package protected constructors.
```
evanw555, line 16: So how exactly does this change behavior from the user's perspective? Does it currently only support public constructors, but now it will support public and package-private (or protected?) constructors? Please help me to understand.

evanw555 (commented):
Since we currently have PR builds disabled. Please build and test this locally to ensure we aren't going to break master. The full build-and-test should take about 30-40 minutes.

CHANGELOG.md
evanw555: Normally for something like this, we would update tests in the PR to demonstrate (as well as test...) the new behavior. If there are any tests for this case, please update them accordingly to demonstrate and test the new behavior.

jasongilanfarr: public, protected and package protected constructors are now Injectable instead of only public.

There were no existing tests for this behavior. Happy to add one.

Also, is HEAD passing?


Failure message for com.linkedin.pegasus.generator.TestDataTemplateGeneratorCmdLineApp > testGeneratorWithCustomResolverDirectory:
    Incorrect @Generated annotation, expected: Generated from /var/folders/3w/tzz49g9n0yj2lpcy1938897w000t07/T/TestDataTemplateGeneratorCmdLineApp1631738171784638724531855579560/testWithResolverDirectory.jar:custom/CustomResolverFoo.pdl expected [true] but found [false]

Gradle suite > Gradle test > com.linkedin.pegasus.generator.TestDataTemplateGeneratorCmdLineApp > testGeneratorWithCustomResolverDirectory FAILED
    java.lang.AssertionError at TestDataTemplateGeneratorCmdLineApp.java:176

Failure message for com.linkedin.pegasus.generator.TestDataTemplateGeneratorCmdLineApp > testGeneratorWithMultipleCustomResolverDirectories:
    Incorrect @Generated annotation, expected: Generated from /var/folders/3w/tzz49g9n0yj2lpcy1938897w000t07/T/TestDataTemplateGeneratorCmdLineApp16317381720085544734219936146515/testWithResolverDirectories.jar:custom1/CustomResolverFoo.pdl expected [true] but found [false]

Gradle suite > Gradle test > com.linkedin.pegasus.generator.TestDataTemplateGeneratorCmdLineApp > testGeneratorWithMultipleCustomResolverDirectories FAILED
    java.lang.AssertionError at TestDataTemplateGeneratorCmdLineApp.java:176

restli-server/src/test/java/com/linkedin/restli/server/resources/TestInjectResourceFactory.java
```diff
@@ -217,4 +218,36 @@ public void testInjectConstructorArgs()
     assertNotNull(r1.getDependency2());
     assertNull(r1.getNonInjectedDependency());
   }
+
+  @Test
+  public void testInjectDefaultConstructorArgs()
```
evanw555, line 223: Seems like this case is exactly the same as the one above. Could you please refactor into one method that accepts the resource class as a parameter + add a data provider to provide `[ConstructorArgResource.class, DefaultConstructorArgResource.class]` to the test method?

restli-server/src/test/java/com/linkedin/restli/server/resources/fixtures/DefaultConstructorArgResource.java
```diff
@@ -0,0 +1,57 @@
+/*
+   Copyright (c) 2012 LinkedIn Corp.
```
evanw555, line 2: nit: `2021` for new source files

restli-server/src/test/java/com/linkedin/restli/server/resources/fixtures/DefaultConstructorArgResource.java
```diff
@@ -0,0 +1,57 @@
...
+
+package com.linkedin.restli.server.resources.fixtures;
+
+import com.linkedin.restli.server.TestRecord;
+import com.linkedin.restli.server.annotations.RestLiCollection;
+import com.linkedin.restli.server.resources.CollectionResourceTemplate;
+import javax.inject.Inject;
+import javax.inject.Named;
+
+@RestLiCollection(name="defaultConstructorArgResource",
```
evanw555, line 25: Just for clarity, leave a javadoc saying that the purpose of this class is to test injecting resources via package-private constructors (but worded more better)

restli-server/src/main/java/com/linkedin/restli/server/resources/Jsr330Adapter.java
```diff
@@ -118,7 +118,7 @@ private void scan(final Collection<Class<?>> managedBeans)
   private void scanInjectableConstructors(Class<?> beanClazz)
   {
     int annotatedConstructors = 0;
-    for (Constructor<?> constructor : beanClazz.getConstructors())
+    for (Constructor<?> constructor : beanClazz.getDeclaredConstructors())
```
evanw555, line 121: Could you reasonably foresee this loosened behavior breaking any existing use cases? Could someone have written a resource with package-private or private constructors expecting them to not be used?

restli-server/src/test/java/com/linkedin/restli/server/resources/TestInjectResourceFactory.java
evanw555: Could be worthwhile to also test private and protected constructors as well.

CHANGELOG.md
evanw555: Discussed offline. Head is passing.

restli-server/src/main/java/com/linkedin/restli/server/resources/Jsr330Adapter.java
jasongilanfarr: Its highly unlikely that that would be the case, especially if the consumers are using any other JSR330 IOC framework.

restli-server/src/test/java/com/linkedin/restli/server/resources/fixtures/DefaultConstructorArgResource.java
jasongilanfarr: Added.

restli-server/src/test/java/com/linkedin/restli/server/resources/fixtures/DefaultConstructorArgResource.java
jasongilanfarr: doh

restli-server/src/test/java/com/linkedin/restli/server/resources/TestInjectResourceFactory.java
jasongilanfarr: I suspect that is very uncommonly used and the change itself is covered with the common case of using default protection. Open to both though.

jasongilanfarr: There is some argument to be made that protected and private really shouldn't be encouraged, so not having them explicitly tested could discourage it.

jasongilanfarr:
Local Build passes.

junchuanwang (approved):
Ship it!

very reasonable change. Thank you for checking into spec details and also fo your contribution.

evanw555 (commented):
LGTM, following up offline about additional testing

PR #695, opened 2021-09-16
Add server config support to define supported accept types

aman1309:
Add server config support to define supported accept types
1) if this is set server will support only the accept types defined in list and if non of accept type requested by client is supported it will default to empty type ie json
2) if not set server will support all supported content types ie. custom or defined within restli.

### DIFF
--- a/restli-server/src/main/java/com/linkedin/restli/server/BaseRestLiServer.java
+++ b/restli-server/src/main/java/com/linkedin/restli/server/BaseRestLiServer.java
@@ -69,6 +69,7 @@ abstract class BaseRestLiServer
   private final ErrorResponseBuilder _errorResponseBuilder;
   private final List<Filter> _filters;
   private final Set<String> _customContentTypes;
+  private final List<String> _supportedAcceptTypes;
   private final ResourceMethodConfigProvider _methodConfigProvider;
   private final boolean _fillInDefaultValueConfigured;
   private final MethodAdapterProvider _methodAdapterProvider;
@@ -81,6 +82,7 @@ abstract class BaseRestLiServer
     _customContentTypes = config.getCustomContentTypes().stream()
         .map(ContentType::getHeaderKey)
         .collect(Collectors.toSet());
+    _supportedAcceptTypes = config.getSupportedAcceptTypes();
 
     _router = new RestLiRouter(rootResources, config);
     resourceFactory.setRootResources(rootResources);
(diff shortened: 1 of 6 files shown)


restli-server/src/main/java/com/linkedin/restli/server/BaseRestLiServer.java
```diff
@@ -172,8 +175,8 @@ protected RoutingResult getRoutingResult(Request request, RequestContext request
     try
     {
       ServerResourceContext context = new ResourceContextImpl(new PathKeysImpl(), request, requestContext);
-      RestUtils.validateRequestHeadersAndUpdateResourceContext(
-          request.getHeaders(), _customContentTypes, context, requestContext);
+      RestUtils.validateRequestHeadersAndUpdateResourceContext(request.getHeaders(), _supportedAcceptTypes,
+              _customContentTypes, context, requestContext);
```
junchuanwang, line 179: can you fix the tab/spce issue? The diff file shows you might be using inconsistent tab/spaces here.

junchuanwang (commented):
can you add description regarding *Why* the change of this PR is needed?

aman1309:
> can you add description regarding _Why_ the change of this PR is needed?

why this is needed is internal to linkedin. lets discuss offline.

junchuanwang:
> > can you add description regarding _Why_ the change of this PR is needed?
> 
> why this is needed is internal to linkedin. lets discuss offline.

Two things
(1) A lot of libraries solves problem to its creators too, yet once they still try to summarize that as generic use case and open source to end users. 
(2) I might be ok with no description if the title and change is self-explanary. In this case I am afraid not. No config should be added without detailed doc explanining its impact, usage pattern and side-effect.

aman1309:
> > > can you add description regarding _Why_ the change of this PR is needed?
> > 
> > 
> > why this is needed is internal to linkedin. lets discuss offline.
> 
> Two things
> (1) A lot of libraries solves problem to its creators' internal companies too, yet they still try to summarize that as generic use case and open source to end users.
> (2) I might be ok with no description if the title and change is self-explanary. In this case I am afraid not. No config should be added without detailed doc explanining its impact, usage pattern and side-effect.

Although I think the commit msg is enough still for the sake of brevity I have answered your point 2 in desc.

junchuanwang (approved):

junchuanwang:
if you want to release the chagne in this version, please update changelog and version number.

PR #697, opened 2021-09-23
Fix bug in SmoothRateLimiter where executionDelay is not honored

lesterhaynes:
Previous code tends to not actually introduce randomness before executing callbacks. This is caused by event loop iterations being rapidly scheduled back to back (such as setting the rate), which flip the boolean that tracks whether or not a delay should be incurred. This patch replaces the hasDelayed boolean with a more durable timestamp to compare against.

Most important change is a robust unit test that should accurately simulate production. Let me know if I'm missing something with this test setup. The new test fails spectacularly without the patch to SmoothRateLimiter applied.

### DIFF
--- a/r2-core/src/main/java/com/linkedin/r2/transport/http/client/SmoothRateLimiter.java
+++ b/r2-core/src/main/java/com/linkedin/r2/transport/http/client/SmoothRateLimiter.java
@@ -224,7 +224,7 @@ public class SmoothRateLimiter implements AsyncRateLimiter
     private int _permitAvailableCount;
     private int _permitsInTimeFrame;
     private long _nextScheduled;
-    private boolean _wasDelayed = false;
+    private long _delayUntil;
 
     EventLoop(Clock clock)
     {
@@ -243,8 +243,14 @@ public class SmoothRateLimiter implements AsyncRateLimiter
       // before entering the next period
       _permitAvailableCount = Math.max(rate.getEvents() - (_permitsInTimeFrame - _permitAvailableCount), 0);
       _permitsInTimeFrame = rate.getEvents();
+      _delayUntil = _clock.currentTimeMillis() + _executionTracker.getNextExecutionDelay(_rate);
 
       loop();
+    }
+
+    private int getRemainingTimeInPeriod()
+    {
+      return (int) (_rate.getPeriod() - (_clock.currentTimeMillis() - _permitTime));
     }
 
     public void loop()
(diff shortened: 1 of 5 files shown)


r2-core/src/main/java/com/linkedin/r2/transport/http/client/SmoothRateLimiter.java
```diff
@@ -243,10 +243,16 @@ private void updateWithNewRate()
       // before entering the next period
       _permitAvailableCount = Math.max(rate.getEvents() - (_permitsInTimeFrame - _permitAvailableCount), 0);
       _permitsInTimeFrame = rate.getEvents();
+      _delayUntil = _clock.currentTimeMillis() + _executionTracker.getNextExecutionDelay(_rate);
 
       loop();
     }
 
+    private int getRemainingTimeInPeriod()
```
rickzx, line 251: Is this function used anywhere?

r2-core/src/main/java/com/linkedin/r2/transport/http/client/SmoothRateLimiter.java
```diff
@@ -317,12 +318,12 @@ public void loop()
         try
         {
           // avoids executing too many duplicate tasks
-          long nextRunRelativeTime = Math.max(0, _permitTime + rate.getPeriod() - _clock.currentTimeMillis());
-          long nextRunAbsolute = _clock.currentTimeMillis() + nextRunRelativeTime;
-          if (_nextScheduled > nextRunAbsolute || _nextScheduled <= _clock.currentTimeMillis())
+          // reschedule next iteration of the event loop to the next delay, or the beginning of the next period
+          long nextRunRelativeTime = _permitAvailableCount > 0 ? _delayUntil - now : Math.max(0, _permitTime + rate.getPeriod() - now);
+          long nextRunAbsolute = now + nextRunRelativeTime;
```
rickzx, line 323: This looks a bit confusing to me. How about keep nextRunRelativeTime as is and change
`long nextRunAbsolute = _permitAvailableCount > 0 ? _delayUntil : now + nextRunRelativeTime`

lesterhaynes: The suggested change results in EventLoop failing to honor the delay, with all unit tests failing in TestConstantQpsRateLimiter, though all tests in TestSmoothRateLimiter continue to pass. We need to augment both nextRunRelative and nextRunAbsolute with the delay in order for the next iteration of the loop to be scheduled with the delay in mind.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/SmoothRateLimiter.java
lesterhaynes: Thanks, should remove it.

rickzx (approved):

PR #707, opened 2021-10-15
SmoothRateLimiter - Make setRate idempotent

lesterhaynes:
ConstantQpsDarkClusterStrategy calls setRate on ConstantQpsRateLimiter(SmoothRateLimiter) with every inbound request using the latest state from the ClusterInfoProvider. This follows the pattern of existing DarkClusterStrategy implementations but currently causes SmoothRateLimiter to exhibit lower throughput caused by repeatedly rescheduling the event loop. Since rescheduling the event loop seems to incur a cost, I recommend only doing so when the rate actually changes.


r2-core/src/main/java/com/linkedin/r2/transport/http/client/ratelimiter/Rate.java
```diff
@@ -16,6 +16,9 @@
 
 package com.linkedin.r2.transport.http.client.ratelimiter;
 
+import java.util.Objects;
```
rickzx, line 19: Unused import?

r2-core/src/main/java/com/linkedin/r2/transport/http/client/SmoothRateLimiter.java
```diff
@@ -179,9 +179,14 @@ public void setRate(double permitsPerPeriod, long periodMilliseconds, int burst)
     ArgumentUtil.checkArgument(permitsPerPeriod >= 0, "permitsPerPeriod");
     ArgumentUtil.checkArgument(periodMilliseconds > 0, "periodMilliseconds");
     ArgumentUtil.checkArgument(burst > 0, "burst");
-
-    _rate = new Rate(permitsPerPeriod, periodMilliseconds, burst);
-    _scheduler.execute(_eventLoop::updateWithNewRate);
+    Rate newRate = new Rate(permitsPerPeriod, periodMilliseconds, burst);
```
rickzx, line 182: Can we not creating a new Rate object when permitsPerPeriod and periodMilliseconds are equal to the same rate? (So instead of creating a new Rate object and using Object.eqauls, we just directly compare permitsPerPeriod and periodMilliseconds)

r2-core/src/main/java/com/linkedin/r2/transport/http/client/ratelimiter/Rate.java
lesterhaynes: Between draft iterations, this one's missing the equals method that uses Objects. 😄

lesterhaynes:
Merged master and updated changelog. Ready to merge at your discretion. Thanks!

PR #715, opened 2021-10-26
Make SmoothRateLimiter setRate idempotent

lesterhaynes:
Further testing shows benefit in randomness of query distribution over a given period when setRate is idempotent. Current implementation of setRate tends to favor maintaining throughput over maintaining a consistently random distribution over a period.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.22.13] - 2021-11-05
+- Make SmoothRateLimiter setRate idempotent
+
 ## [29.22.12] - 2021-10-28
 - add canaries to service and cluster properties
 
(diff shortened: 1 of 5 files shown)


rickzx (commented):
The change looks good to me. But could you help me understand why

> Current implementation of setRate tends to favor maintaining throughput over maintaining a consistently random distribution over a period.

And how changing setRate to be idempotent helps with this? Is it because we're calling updateWithNewRate too often?

lesterhaynes:
> The change looks good to me. But could you help me understand why
> 
> > Current implementation of setRate tends to favor maintaining throughput over maintaining a consistently random distribution over a period.
> 
> And how changing setRate to be idempotent helps with this? Is it because we're calling updateWithNewRate too often?

That's basically it. In cases where setRate is called very frequently (in the case of ConstantQpsDarkClusterStrategy, setRate is called with every incoming query), the delay until the next dispatched query is re-evaluated. Since we're already waiting, calling setRate repeatedly gives a higher chance of encountering a lower delay, especially after discounting the amount of time already waited.

rickzx (approved):
LGTM

PR #718, opened 2021-10-27
Reduce garbage generated when searching for MIME attachment delimiters

awin775:
Change ByteString.indexOfBytes() to re-use the ByteIterator object when
backtracking after a failed match, rather than creating a new one. This
will reduce the amount of memory used when MultiPartMIMEReader searches
through a large binary attachment for the MIME delimiter.


ChrisCarini:

PR #720, opened 2021-11-03
Backfill unit tests for RestLiRouter, RestLiAnnotationReader.

dlessa-li:
Adds unit tests for RestLiRouter and RestLiAnnotationReader.


restli-server/src/test/java/com/linkedin/restli/internal/server/model/TestRestLiAnnotationReader.java
```diff
@@ -0,0 +1,966 @@
...
+import java.util.stream.Collectors;
+import org.junit.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+
+public class TestRestLiAnnotationReader
+{
+
+  @Test(description = "resource: action; type: root; annotations: {name}; model: none; key: none]")
```
mchen07, line 72: this description is a bit cryptic

restli-server/src/test/java/com/linkedin/restli/internal/server/model/TestRestLiAnnotationReader.java
```diff
@@ -0,0 +1,966 @@
...
+  {
+
+    final String expectedNamespace = "com.linkedin.model.actions";
+    final String expectedName = "actionsOnly";
+
+    @RestLiActions(
+        name = "actionsOnly",
+        namespace = "com.linkedin.model.actions"
+    )
+    class ActionsOnlyResource {
```
mchen07, line 83: why not using `SampleResources` there?

restli-server/src/test/java/com/linkedin/restli/internal/server/TestRestLiRouter.java
```diff
@@ -0,0 +1,237 @@
...
+      doReturn(null).when(_context).getMethodName(eq(ResourceMethod.GET));
+    }
+
+    private void mockContextForOperationNotFound(String path) throws URISyntaxException
+    {
+      mockCommon();
+      doReturn(null).when(_pathKeys).getBatchIds();
+      doReturn(new URI(path)).when(_context).getRequestURI();
+      doReturn("POST").when(_context).getRequestMethod(); // http method
+      doReturn("CREATE").when(_context).getRestLiRequestMethod(); // header value
```
mchen07, line 219: what does this //header value mean?

restli-server/src/test/java/com/linkedin/restli/internal/server/TestRestLiRouter.java
```diff
@@ -0,0 +1,237 @@
...
+      _context = mock(ServerResourceContext.class);
+      _pathKeys = mock(MutablePathKeys.class);
+      _rootModel = RestLiAnnotationReader.processResource(RootResource.class);
+      _pathToModelMap = new HashMap<>();
+      _config = new RestLiConfig();
+      _router = new RestLiRouter(_pathToModelMap, _config);
+      _requestContext = new RequestContext();
+    }
+
+    private void mockCommon() {
```
junchuanwang, line 195: Can this be setup using "beforeMethod"?

restli-server/src/test/java/com/linkedin/restli/internal/server/model/TestRestLiAnnotationReader.java
junchuanwang: Is this my display issue? Looks like there is not maching left bracket "[", same for every description below?

`"resource: action; type: root; annotations: {name}; model: none; key: none]"`

junchuanwang (approved):
overall looks good to me

evanw555:
Thanks for backfilling a lot of these tests. My only comment is that you should check out this class: https://github.com/linkedin/rest.li/blob/master/restli-server/src/test/java/com/linkedin/restli/internal/server/model/SampleResources.java

I refactored a lot of the dummy test resources into this file to clean the tests up and make them reusable. You may want to look through this file to see if there are any resources you could reuse for your own testing. You may want to also refactor your test resources into this file as well.

Also, I think the reason we didn't have direct tests for `RestLiAnnotationReader` is that we tested its behavior via `TestRestLiApiBuilder`. Whenever I made changes to `RestLiAnnotationReader`, I added tests to `TestRestLiApiBuilder`, since `RestLiApiBuilder` is basically a wrapper around `RestLiAnnotationReader`

restli-server/src/test/java/com/linkedin/restli/internal/server/model/TestRestLiAnnotationReader.java
dlessa-li: Is the suggestion to move the local test resource classes in the `SampleResources` file? I can definitely do it. I just thought these use cases are very specific to this test class due to the annotations I have on these classes.

restli-server/src/test/java/com/linkedin/restli/internal/server/TestRestLiRouter.java
dlessa-li: It means that the string returned would come from the "X-RestLi-Method" request header.

restli-server/src/test/java/com/linkedin/restli/internal/server/TestRestLiRouter.java
dlessa-li: Yes it can, along with the test setup creation. Will make the change.

dlessa-li:
> Thanks for backfilling a lot of these tests. My only comment is that you should check out this class: https://github.com/linkedin/rest.li/blob/master/restli-server/src/test/java/com/linkedin/restli/internal/server/model/SampleResources.java
> 
> I refactored a lot of the dummy test resources into this file to clean the tests up and make them reusable. You may want to look through this file to see if there are any resources you could reuse for your own testing. You may want to also refactor your test resources into this file as well.
> 
> Also, I think the reason we didn't have direct tests for `RestLiAnnotationReader` is that we tested its behavior via `TestRestLiApiBuilder`. Whenever I made changes to `RestLiAnnotationReader`, I added tests to `TestRestLiApiBuilder`, since `RestLiApiBuilder` is basically a wrapper around `RestLiAnnotationReader`. I fear that many of your tests may be duplicates of the test cases in that file. If that's the case, we should try and reuse the testing/resources as much as possible.

I used some resources from `SampleResources` in some of the happy path tests. In the other happy path tests, there were aspects that were not covered in `SampleResources` classes; I can double check and either reuse an existing resource or move the resource to that class. For the negative cases, all of them are exercising one particular failed validation most of which I did not see covered before I had the tests in place.

I'll update to follow the model you explained above, trying to reuse more than I did.

As for test duplication, I personally think it's better to test `RestLiAnnotationReader` and `RestLiApiBuilder` separately. The rationale here is that when we update `RestLiAnnotationReader` or find a bug in it, we know exactly the test class that needs to be updated.  

I understand that having utility classes with static methods forces us into coupling the unit tests. But I believe we could still focus on just the `RestLiApiBuilder` logic in the `TestRestLiApiBuilder` respective test class, in the spirit of single responsibility. That said, I'm open to discuss this further.

restli-server/src/test/java/com/linkedin/restli/internal/server/model/TestRestLiAnnotationReader.java
dlessa-li: Updated descriptions to plain english.

restli-server/src/test/java/com/linkedin/restli/internal/server/model/TestRestLiAnnotationReader.java
dlessa-li: Moved some resources to `SampleResources` since no existing resource in that file supported some of the new test cases added here. Other tests in this file had already been using resources from `SampleResources` where it fit the test case.

restli-server/src/test/java/com/linkedin/restli/internal/server/TestRestLiRouter.java
dlessa-li: Updated the inline comment to make it clear.

restli-server/src/test/java/com/linkedin/restli/internal/server/TestRestLiRouter.java
dlessa-li: Actually, I take it back: `TestRestLiRouter#failsOnRootResourceNotFound` requires no mocking.

restli-server/src/test/java/com/linkedin/restli/internal/server/model/SampleResources.java
```diff
@@ -171,7 +178,7 @@ public String code()
     namespace = "com.linkedin.restli.internal.server.model",
     parent = ParentResource.class
   )
-  class TestResource extends CollectionResourceTemplate<String, EmptyRecord>
+  static class TestResource extends CollectionResourceTemplate<String, EmptyRecord>
```
mchen07, line 181: You make this change because you put these two new test classes into different packages, why
cannot we put them together in the same package as existing `TestRestLiApiBuilder` for easier discovery later?

dlessa-li: They are in the same package! The problem was that I was importing `static` in the new test file. Fixed that and reverted the static modifiers in `SampleResources`.

evanw555 (approved):

PR #724, opened 2021-11-06
Update get_started.md

haroldl:
Adding details about the versions of the JDK and Gradle that are needed to build successfully.

### DIFF
--- a/src/get_started.md
+++ b/src/get_started.md
@@ -22,8 +22,6 @@ Follow the steps below to try Rest.li quickly and get a basic idea of how it wor
 
 JDK: You will need a Java 1.8 JDK to build the code. Using Java 1.9 or later will fail because Rest.li uses some APIs that were deprecated in 1.9 and the build treats warnings as error in order to ensure high code quality. On Linux systems openjdk-8-jdk is known to work.
 
-Gradle: You will need to install [Gradle](http://www.gradle.org/) before you can build Rest.li. Gradle 5.2.1 is known to work.
-
 ## Build
 
 Rest.li uses [Gradle](http://www.gradle.org/) as the build system. The following points summarize some basic tasks that can be run:
--- a/src/get_started.md
+++ b/src/get_started.md
@@ -12,10 +12,17 @@ index: 1
 
 Follow the steps below to try Rest.li quickly and get a basic idea of how it works:
 
-* [Build](#Build)
+* [Setup](#setup)
+* [Build](#build)
 * [Run the Examples](#run-the-examples)
 * [API](#api)
 * [Code Generator](#code-generator)
+
+## Setup
+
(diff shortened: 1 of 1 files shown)


src/get_started.md
```diff
@@ -12,11 +12,18 @@ index: 1
...
+* [Build](#build)
 * [Run the Examples](#run-the-examples)
 * [API](#api)
 * [Code Generator](#code-generator)
 
+## Setup
+
+JDK: You will need a Java 1.8 JDK to build the code. Using Java 1.9 or later will fail because Rest.li uses some APIs that were deprecated in 1.9 and the build treats warnings as error in order to ensure high code quality. On Linux systems openjdk-8-jdk is known to work.
+
+Gradle: You will need to install [Gradle](http://www.gradle.org/) before you can build Rest.li. Gradle 5.2.1 is known to work.
```
evanw555, line 25: Suggest using the `gradlew` Gradle wrapper. This is much easier than installing Gradle for OSS development.

haroldl: I removed the Gradle portion. If the JDK part is useful then we can push that.

src/get_started.md
```diff
@@ -12,11 +12,16 @@ index: 1
...
-* [Build](#Build)
+* [Setup](#setup)
+* [Build](#build)
 * [Run the Examples](#run-the-examples)
 * [API](#api)
 * [Code Generator](#code-generator)
 
+## Setup
+
+JDK: You will need a Java 1.8 JDK to build the code. Using Java 1.9 or later will fail because Rest.li uses some APIs that were deprecated in 1.9 and the build treats warnings as error in order to ensure high code quality. On Linux systems openjdk-8-jdk is known to work.
```
junchuanwang, line 23: Unfortunately this is not entirely true. Certain version of openJDK 11  on linux can work as well. We mostly heard build failures only in MAC.

Maybe this should be one of the FAQ sections to suggest if build failed,try reverting back to java 1.8 :-)

haroldl: I think it would be useful to warn people up front that they likely need to use an older version of Java. My first try was with openjdk 14 on Linux and it failed. Is there a recommended version to use? Is it expected to build and work with newer versions like 14 (default on Ubuntu) or 17 (default when downloading from Oracle)?

junchuanwang: No problem, we can keep it here in this section for now. My point is just to say Java1.9's later version might still build.

Personally this is my java version, seems to work.

```
openjdk version "11.0.11" 2021-04-20 LTS
OpenJDK Runtime Environment 18.9 (build 11.0.11+9-LTS)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.11+9-LTS, mixed mode, sharing
```

ChrisCarini:

PR #725, opened 2021-11-13
Add support for a rate limiter supplier to enable multiple dark clusters with the CONSTANT_QPS strategy

lesterhaynes:
Current implementation of ConstantQpsDarkClusterStrategy takes a ConstantQpsRateLimiter as a parameter. This works fine when a single dark cluster is defined that uses CONSTANT_QPS strategy, but results in unexpected behavior when more than one is specified (both share the same rate limiter). This change adds the support for passing a supplier instead.


r2-core/src/main/java/com/linkedin/r2/transport/http/client/CallbackWrappingConstantQpsRateLimiterFactory.java
```diff
@@ -0,0 +1,34 @@
+package com.linkedin.r2.transport.http.client;
+
+import com.linkedin.r2.transport.http.client.ratelimiter.CallbackWrapper;
+import com.linkedin.util.clock.Clock;
+import java.time.temporal.ChronoUnit;
+import java.util.concurrent.Executor;
+import java.util.concurrent.ScheduledExecutorService;
+
+
+public class CallbackWrappingConstantQpsRateLimiterFactory implements ConstantQpsRateLimiterFactory {
```
rickzx, line 10: What is this RateLimiterFactory used for? Only for test purpose? Then can we make this an inner class of the test file

lesterhaynes: Glad you asked! Before this change, we'd pass in a ConstantQpsRateLimiter that we could extend with whatever additional functionality we desired through an inheritance programming model. The move to supply a ConstantQpsRateLimiterFactory changes this to a composition model, where we're now instead passing the functionality we need in the form of a wrapper.

Name dropping class names from LI internal libraries: We'll no longer be able to use ICAwareCallbackCircularBuffer in the package that pulls in rest.li, and will instead be passing a ICAwareCallbackWrapper.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/ratelimiter/CallbackWrapper.java
```diff
@@ -0,0 +1,9 @@
+package com.linkedin.r2.transport.http.client.ratelimiter;
+
+import com.linkedin.common.callback.Callback;
+import com.linkedin.common.util.None;
+
+
+public interface CallbackWrapper {
```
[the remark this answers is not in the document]
lesterhaynes: Glad you asked! Before this change, we'd pass in a ConstantQpsRateLimiter that we could extend with whatever additional functionality we desired through an inheritance programming model. The move to supply a ConstantQpsRateLimiterFactory changes this to a composition model, where we're now instead passing the functionality we need in the form of a wrapper.

Name dropping class names from LI internal libraries: We'll no longer be able to use ICAwareCallbackCircularBuffer in the package that pulls in rest.li, and will instead be passing a ICAwareCallbackWrapper.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/CallbackWrappingConstantQpsRateLimiterFactory.java
lesterhaynes: RateLimiterFactory is the essential change in this PR. By supplying a RateLimiterFactory, it grants rest.li the ability to create its own RateLimiters, and it's able to create a separate RateLimiter for every downstream dark cluster. Without this change, rest.li only has the one rate limiter, so it has all downstream dark clusters sharing that one rate limiter. This gets pretty gnarly when each ConstantQpsDarkClusterStrategy independently calls setRate (for example) on that rate limiter.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/ratelimiter/CallbackWrapper.java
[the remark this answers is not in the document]
cx-super: 1. Instead of passing in a callback wrapper, have you considered passing a FactoryFactory so that you can use your own factory to generate ratelimiter you want?
2. Please do not define new interface. Use CallbackAdaptor instead.

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
```diff
@@ -65,15 +67,15 @@
   private final Random _random;
   private final LoadBalancerClusterListener _clusterListener;
   private final DarkClusterVerifierManager _verifierManager;
-  private final ConstantQpsRateLimiter _rateLimiter;
+  private final ConstantQpsRateLimiterFactory _rateLimiterFactory;
 
   public DarkClusterStrategyFactoryImpl(@Nonnull Facilities facilities,
```
rickzx, line 72: Let's make this constructor the last constructor in the class

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
```diff
@@ -82,18 +84,30 @@ public DarkClusterStrategyFactoryImpl(@Nonnull Facilities facilities,
     _random = random;
     _darkClusterDispatcher = darkClusterDispatcher;
     _verifierManager = verifierManager;
-    _rateLimiter = rateLimiter;
+    _rateLimiterFactory = rateLimiterFactory;
     _clusterListener = new DarkClusterListener();
   }
 
+  public DarkClusterStrategyFactoryImpl(@Nonnull Facilities facilities,
```
rickzx, line 91: Can we remove this constructor?

r2-core/src/main/java/com/linkedin/r2/transport/http/client/ratelimiter/CallbackWrapper.java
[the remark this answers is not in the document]
rickzx: +1 for using com.linkedin.common.callback.CallbackAdapter.

Also as Chao mentioned, using FactoryFactory pattern has the advantage that you will be able to define an `ICAwareRateLimiterFactory` in LI internal libraries to generate IC aware rate limiter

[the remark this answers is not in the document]
lesterhaynes: Thanks for pushing back on this. The latest diff uses a Supplier instead of a static factory, which I think takes care of the concerns raised here. Let me know what you think.

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
lesterhaynes: Moved this constructor to the end

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
lesterhaynes: Added a Deprecated decorator. Removing the constructor would break backward compatibility.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/CallbackWrappingConstantQpsRateLimiterFactory.java
lesterhaynes: ConstantQpsRateLimiterFactory has been removed in the latest iteration of the PR. Instead, the API takes a Supplier of the same rate limiter to accomplish the same goal while requiring no new classes.

rickzx (approved):
LGTM. Let's also update the CHANGELOG

lesterhaynes:
@rickzx Thanks for taking a look. Updated change log

PR #726, opened 2021-11-16
Add support for dark warm-up

rpuranda:
- Modifying doMarkUp method in ZooKeeperAnnouncer to optionally run warm up first by announcing to a D2 dark cluster


d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -211,8 +219,102 @@ public void onSuccess(None result)
...
+      public void onError(Throwable e)
+      {
+        // this would ideally never happen
+        callback.onError(e);
+      }
+
+      @Override
+      public void onSuccess(None result)
+      {
+        _log.info("doing mark up of main cluster " + _cluster);
```
mariabornski, line 235: "main" cluster is potentially confusing here.  How about "primary" or "live" or "regular" something similar?

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -211,8 +219,102 @@ public void onSuccess(None result)
...
+      public void onError(Throwable e) {
+        _log.info("mark down warm up cluster " + _warmupClusterName + " has failed");
+        // what to do here ? - go ahead continue to announce main cluster
+        // doAnnounceCallback.onSuccess(None.none());
+      }
+
+      @Override
+      public void onSuccess(None result) {
+        _log.info("mark down warm up cluster " + _warmupClusterName +
+            " has completed, now mark up of main cluster " + _cluster);
```
mariabornski, line 251: "main" cluster is potentially confusing here. How about "primary" or "live" or "regular" something similar?

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -211,8 +219,102 @@ public void onSuccess(None result)
...
+        _server.markUp(_warmupClusterName, _uri, _partitionDataMap, _uriSpecificProperties, doWarmupWaitCallback);
+      }
+    };
+
     _log.info("overrideMarkUp is called for uri = " + _uri);
+    if (_isEnabledD2WarmupCluster) {
+      _log.info("Calling mark up for warm up cluster " + _warmupClusterName);
+      doWarmupMarkUpCallback.onSuccess(None.none());
+    } else {
+      _log.info("Calling mark up of main cluster " + _cluster);
```
mariabornski, line 314: "main" cluster is potentially confusing here. How about "primary" or "live" or "regular" something similar?

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -162,11 +172,10 @@ public synchronized void markUp(final Callback<None> callback)
 
   private synchronized void doMarkUp(Callback<None> callback)
   {
-    _server.markUp(_cluster, _uri, _partitionDataMap, _uriSpecificProperties, new Callback<None>()
-    {
+
+    final Callback<None> markUpCallback = new Callback<None>() {
```
mariabornski, line 176: Looks like the style in this file is to have the opening curly brace on a separate line.  It would be good to match the existing style & move your curly brace down onto the next line.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -162,11 +172,10 @@ public synchronized void markUp(final Callback<None> callback)
...
       @Override
-      public void onError(Throwable e)
-      {
+      public void onError(Throwable e) {
```
mariabornski, line 178: Looks like the style in this file is to have the opening curly brace on a separate line.  It would be good to match the existing style & move your curly brace down onto the next line.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -185,8 +194,7 @@ public void onError(Throwable e)
       }
 
       @Override
-      public void onSuccess(None result)
-      {
+      public void onSuccess(None result) {
```
mariabornski, line 197: Looks like the style in this file is to have the opening curly brace on a separate line.  It would be good to match the existing style & move your curly brace down onto the next line.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -211,8 +219,102 @@ public void onSuccess(None result)
         }
         runNextMarkUpOrMarkDown();
       }
-    });
+    };
+
+    final Callback<None> doMarkUpCallback = new Callback<None>() {
```
mariabornski, line 224: Looks like the style in this file is to have the opening curly brace on a separate line.  It would be good to match the existing style & move your curly brace down onto the next line (unless the curly brace has to stay on the same line to compile correctly)

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -211,8 +219,102 @@ public void onSuccess(None result)
...
+
+      @Override
+      public void onSuccess(None result)
+      {
+        _log.info("doing mark up of main cluster " + _cluster);
+        _server.markUp(_cluster, _uri, _partitionDataMap, _uriSpecificProperties, markUpCallback);
+      }
+    };
+
+    final Callback<None> warmupMarkDownCallback = new Callback<None>() {
```
mariabornski, line 240: Looks like the style in this file is to have the opening curly brace on a separate line.  It would be good to match the existing style & move your curly brace down onto the next line (unless the curly brace has to stay on the same line to compile correctly)

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -211,8 +219,102 @@ public void onSuccess(None result)
...
+      @Override
+      public void onError(Throwable e) {
```
mariabornski, line 242: Looks like the style in this file is to have the opening curly brace on a separate line.  It would be good to match the existing style & move your curly brace down onto the next line (unless the curly brace has to stay on the same line to compile correctly)

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -211,8 +219,102 @@ public void onSuccess(None result)
...
+        _log.info("mark down warm up cluster " + _warmupClusterName + " has failed");
+        // what to do here ? - go ahead continue to announce main cluster
+        // doAnnounceCallback.onSuccess(None.none());
+      }
+
+      @Override
+      public void onSuccess(None result) {
```
mariabornski, line 249: Looks like the style in this file is to have the opening curly brace on a separate line.  It would be good to match the existing style & move your curly brace down onto the next line (unless the curly brace has to stay on the same line to compile correctly)

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -211,8 +219,102 @@ public void onSuccess(None result)
...
+        _log.info("mark down warm up cluster " + _warmupClusterName +
+            " has completed, now mark up of main cluster " + _cluster);
+        doMarkUpCallback.onSuccess(None.none());
+      }
+    };
+
+    final Callback<None> doWarmupMarkDownCallback = new Callback<None>() {
```
mariabornski, line 256: Looks like the style in this file is to have the opening curly brace on a separate line.  It would be good to match the existing style & move your curly brace down onto the next line (unless the curly brace has to stay on the same line to compile correctly)

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -211,8 +219,102 @@ public void onSuccess(None result)
...
+      @Override
+      public void onError(Throwable e) {
```
mariabornski, line 258: Looks like the style in this file is to have the opening curly brace on a separate line.  It would be good to match the existing style & move your curly brace down onto the next line (unless the curly brace has to stay on the same line to compile correctly)

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -211,8 +219,102 @@ public void onSuccess(None result)
...
+        _log.info("failure in scheduling warm up mark down");
+        // what to do here ? - go ahead continue to announce main cluster
+        // doAnnounceCallback.onSuccess(None.none());
+      }
+
+      @Override
+      public void onSuccess(None result) {
```
mariabornski, line 265: Looks like the style in this file is to have the opening curly brace on a separate line.  It would be good to match the existing style & move your curly brace down onto the next line (unless the curly brace has to stay on the same line to compile correctly)

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -211,8 +219,102 @@ public void onSuccess(None result)
...
+        _log.info("warm up for cluster " + _warmupClusterName + " completed, now calling mark down on the warm up cluster");
+        _server.markDown(_warmupClusterName, _uri, warmupMarkDownCallback);
+      }
+    };
+
+    final Callback<None> doWarmupWaitCallback = new Callback<None>() {
```
mariabornski, line 271: Looks like the style in this file is to have the opening curly brace on a separate line.  It would be good to match the existing style & move your curly brace down onto the next line (unless the curly brace has to stay on the same line to compile correctly)

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -211,8 +219,102 @@ public void onSuccess(None result)
...
+      public void onSuccess(None result)
+      {
+        _log.info("mark up for warm up cluster " + _warmupClusterName + " succeeded");
+        ScheduledExecutorService warmup = Executors.newSingleThreadScheduledExecutor();
+        _log.info("scheduled executor service to run after warmup");
+        warmup.schedule(() -> doWarmupMarkDownCallback.onSuccess(None.none()), _warmupDuration, TimeUnit.MILLISECONDS);
+      }
+    };
+
+    final Callback<None> doWarmupMarkUpCallback = new Callback<None>() {
```
mariabornski, line 293: Looks like the style in this file is to have the opening curly brace on a separate line.  It would be good to match the existing style & move your curly brace down onto the next line (unless the curly brace has to stay on the same line to compile correctly)

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -211,8 +219,102 @@ public void onSuccess(None result)
...
+      @Override
+      public void onSuccess(None result)
+      {
+        _log.info("Calling ZooKeeperServer's mark up for warm up cluster " + _warmupClusterName);
+        _server.markUp(_warmupClusterName, _uri, _partitionDataMap, _uriSpecificProperties, doWarmupWaitCallback);
+      }
+    };
+
     _log.info("overrideMarkUp is called for uri = " + _uri);
+    if (_isEnabledD2WarmupCluster) {
```
mariabornski, line 310: Looks like the style in this file is to have the opening curly brace on a separate line.  It would be good to match the existing style & move your curly brace down onto the next line (unless the curly brace has to stay on the same line to compile correctly)

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -211,8 +219,102 @@ public void onSuccess(None result)
...
+      _log.info("Calling mark up for warm up cluster " + _warmupClusterName);
+      doWarmupMarkUpCallback.onSuccess(None.none());
+    } else {
```
mariabornski, line 313: Looks like the style in this file is to have the curly braces surrounding "else" on separate lines.  It would be good to match the existing style & move your curly braces down onto the next lines

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -211,8 +219,102 @@ public void onSuccess(None result)
...
+            " has completed, now mark up of main cluster " + _cluster);
+        doMarkUpCallback.onSuccess(None.none());
+      }
+    };
+
+    final Callback<None> doWarmupMarkDownCallback = new Callback<None>() {
+      @Override
+      public void onError(Throwable e) {
+        _log.info("failure in scheduling warm up mark down");
+        // what to do here ? - go ahead continue to announce main cluster
```
rickzx, line 260: imo, there are a couple of things we can do:
1. Notice that in this class we also have a `ZKPersistentConnection.EventListener` that subscribes to zk events. I would consider recording a failed markdown, and retry the operation whenever zk connection is back.
2. As you mentioned, we should go ahead and continue to announce the main cluster

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -211,8 +219,102 @@ public void onSuccess(None result)
...
+        _log.info("warm up for cluster " + _warmupClusterName + " completed, now calling mark down on the warm up cluster");
+        _server.markDown(_warmupClusterName, _uri, warmupMarkDownCallback);
+      }
+    };
+
+    final Callback<None> doWarmupWaitCallback = new Callback<None>() {
+      @Override
+      public void onError(Throwable e)
+      {
+        _log.info("failure in mark up of warm up cluster " + _warmupClusterName);
```
rickzx, line 275: This deserves a WARN level log imo.
And i prefer
```
_log.warn("failed to mark up warm-up cluster {} due to {}.", _warmupClusterName, e.getClass().getSimpleName());
```

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -211,8 +219,102 @@ public void onSuccess(None result)
...
+        // OR 2. mark announcement fail, move on to next pending mark up / mark down
+        // announceCallback.onError(e);
+      }
+
+      @Override
+      public void onSuccess(None result)
+      {
+        _log.info("mark up for warm up cluster " + _warmupClusterName + " succeeded");
+        ScheduledExecutorService warmup = Executors.newSingleThreadScheduledExecutor();
+        _log.info("scheduled executor service to run after warmup");
```
rickzx, line 288: This doesn't need to be in INFO log. Debug log should suffice

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rickzx, line 293: Ditto. We don't need the callback pattern here

rickzx (changes requested):
Thanks for working on this! Great start

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: Thank you for pointing this out. Correcting in the next version of the PR.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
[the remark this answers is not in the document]
rpuranda: Thank you for suggested the correct way. Correcting in the next version.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: Thank you for suggestions and explaining that failed markdown for warm up should be retried. About announcing to main cluster, in the markUp method of ZooKeeperServer, in the getCallback the check for uri already existing will not be able to catch that the uri exists but for a different cluster name (warm up cluster's name) and will add association of the main cluster name with the uri in the store, in this case should we still announce to the main cluster or keep retrying mark down for the warm up cluster?https://github.com/linkedin/rest.li/blob/7372bc76f7ce78007ecb24ca80f17399e5246e13/d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperServer.java#L145

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: thanks for pointing this. I have deleted the log statement and only log if mark up succeeds or fails and using the name of the cluster instead of "main", to avoid confusion.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: Thanks for pointing this. I have changed it to
 _log.info("markDown for uri {} on warm-up cluster {} has completed, now marking up regular cluster {}", _uri, _warmupClusterName, _cluster);

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: thanks for pointing this. I have deleted the log statement and only log if mark up succeeds or fails and using the name of the cluster instead of "main", to avoid confusion.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: Thanks! Done.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: Thanks! Done

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: Thanks! Done.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: Thanks! Done.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: Thanks! Done

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: Thanks! Done

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: Thanks! Done

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: Thanks! Done

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: Thanks! Done

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: Thanks! Actually this callback is deleted in the updated commit of the PR. Have fixed these styling issues.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: Thanks! Actually this callback is deleted in the updated commit of the PR. Have fixed these styling issues.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: Thanks! Fixed.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: Thanks! Fixed.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: Thanks! Fixed it in the new commit.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: Thanks! Fixed and changed to 
_log.debug("warm-up will run for {}ms",_warmupDuration);

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: Thank you! Fixed and changed to
_log.warn("failure to markUp uri {} of warm-up cluster {} due to {}.", _uri, _warmupClusterName, e.getClass().getSimpleName());

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -72,6 +78,10 @@
   private boolean _isRunningMarkUpOrMarkDown;
   private volatile boolean _shuttingDown;
 
+  private boolean _isEnabledD2WarmupCluster = true;  //temporary - for testing flow
```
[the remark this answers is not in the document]
rpuranda: thanks! I have added an additional class constructor . I have added these constants
  public static final boolean DEFAULT_D2_WARMUP_CLUSTER_ENABLED = false;
  public static final int DEFAULT_D2_WARMUP_DURATION = 0;
  
  but not used them anywhere yet. @rickzx  @lesterhaynes Can you please suggest how I should use these values?

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -211,8 +219,102 @@ public void onSuccess(None result)
...
+      {
+        _log.info("doing mark up of main cluster " + _cluster);
+        _server.markUp(_cluster, _uri, _partitionDataMap, _uriSpecificProperties, markUpCallback);
+      }
+    };
+
+    final Callback<None> warmupMarkDownCallback = new Callback<None>() {
+      @Override
+      public void onError(Throwable e) {
+        _log.info("mark down warm up cluster " + _warmupClusterName + " has failed");
```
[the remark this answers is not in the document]
rpuranda: thank you! fixed to
_log.warn("failed to markDown uri {} on warm-up cluster {} due to {}.", _uri, _warmupClusterName, e.getClass().getSimpleName());

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -211,8 +219,102 @@ public void onSuccess(None result)
...
+        // what to do here ? - go ahead continue to announce main cluster
+        // doAnnounceCallback.onSuccess(None.none());
+      }
+
+      @Override
+      public void onSuccess(None result) {
+        _log.info("mark down warm up cluster " + _warmupClusterName +
+            " has completed, now mark up of main cluster " + _cluster);
+        doMarkUpCallback.onSuccess(None.none());
```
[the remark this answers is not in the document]
rpuranda: Thank you! Fixed.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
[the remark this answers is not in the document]
rpuranda: Thank you! Fixed.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -211,8 +219,102 @@ public void onSuccess(None result)
...
+        // doAnnounceCallback.onSuccess(None.none());
+        // OR 2. mark announcement fail, move on to next pending mark up / mark down
+        // announceCallback.onError(e);
+      }
+
+      @Override
+      public void onSuccess(None result)
+      {
+        _log.info("mark up for warm up cluster " + _warmupClusterName + " succeeded");
+        ScheduledExecutorService warmup = Executors.newSingleThreadScheduledExecutor();
```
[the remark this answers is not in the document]
rpuranda: In the last commit, I have added a new constructor which supplies ScheduledExecutorService.
Also added a setter method setExecutorServiceAndWarmupParameters to set _executorService and warm up parameters for the announcer. Using any of the 2 ways - the new constructor or the setter method, will need a change in container code, and the services that directly call methods on ZooKeeperAnnouncer if they have use to warmup. @rickzx @lesterhaynes please review and suggest.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -452,4 +576,17 @@ private int getNumberOfPartitions()
...
+   * A way to force users to set both executor service and warm up parameters, if they want to use d2 warmup
+   */
+  public void setExecutorServiceAndWarmupParameters(ScheduledExecutorService executorService, boolean isEnabledD2WarmupCluster,
+      String warmupClusterName, int warmupDuration)
+  {
+    _executorService = executorService;
+    _isEnabledD2WarmupCluster = isEnabledD2WarmupCluster;
+    _warmupClusterName = warmupClusterName;
+    _warmupDuration = warmupDuration;
+  }
```
lesterhaynes, line 590: What are you seeing in the container code that makes this necessary (rather than just using the new constructor)?

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -130,6 +167,12 @@ synchronized void retry(Callback<None> callback)
       {
         markDown(callback);
       }
+
+      // If markDown for warm-up cluster is pending, complete it
+      if(!_pendingWarmUpMarkDown.isEmpty() && _isWarmUp)
+      {
+        _server.markDown(_warmupClusterName, _uri, callback);
+      }
```
lesterhaynes, line 175: May want to move this to the top, above the check for `if (_isUp)`. Having it at the end seems like it could short-circuit the warmup that was initialized from the retried markUp.

lesterhaynes (commented):
Looks good, few minor things. Seems like it will do a good job in continuing to honor the markUp contract (making warmUp best effort). Seems like it should do well in cleaning up warmup announcements as well. Test cases to validate zookeeper connection failure scenarios would help shore up confidence in that.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -65,13 +70,26 @@
    * it will try to bring up the server again on ZK if the connection goes down, or a new store is set
    */
   private boolean _isUp;
+
+  // Field to indicate if warm up was started. If it is true, it will try to end the warm up
+  // by marking down on ZK if the connection goes down
+  private boolean _isWarmUp;
```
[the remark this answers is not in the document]
rpuranda: Thanks for pointing this out. Updated to _isWarmingUp.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -82,8 +100,27 @@ public ZooKeeperAnnouncer(ZooKeeperServer server, boolean initialIsUp)
     _server = server;
     // initialIsUp is used for delay mark up. If it's false, there won't be markup when the announcer is started.
     _isUp = initialIsUp;
+    _isWarmUp = false;
+    _pendingMarkDown = new ArrayDeque<>();
+    _pendingMarkUp = new ArrayDeque<>();
+    _pendingWarmUpMarkDown = new ArrayDeque<>();
+
```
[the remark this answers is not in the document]
rpuranda: thanks for pointing this out. Corrected.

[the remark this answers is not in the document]
rpuranda: Though I haven't given any default for DEFAULT_D2_WARMUP_CLUSTER_NAME. Setting warmClusterName as null for now, as by default we do not enable warm up and we require the cluster name mentioned for warm up to be actually configured.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: Thanks for noticing this! In the container code, currently ZooKeeperAnnouncer object is by using the second constructor
ZooKeeperAnnouncer(ZooKeeperServer server, boolean initialIsUp)

and setter methods are used for other class parameters like _cluster, _uri, _partitionDataMap and _uriSpecificProperties.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: thanks! Fixed it.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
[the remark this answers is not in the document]
rpuranda: I have used these defaults in the second constructor. Please review.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -211,8 +248,88 @@ public void onSuccess(None result)
...
+      @Override
+      public void onSuccess(None result)
+      {
+        // Mark _isWarmingUp to true to indicate warm up is completed
+        _isWarmingUp = false;
+        // Note that the pending callbacks we see at this point are from the requests that are filed before us because
+        // zookeeper guarantees the ordering of callback being invoked.
+        synchronized (ZooKeeperAnnouncer.this)
+        {
+          // drain _pendingMarkDown with successful result.
```
mariabornski, line 288: Should this comment reference _pendingWarmUpMarkDown, not _pendingMarkDown?  The code below it is draining _pendingWarmUpMarkDown.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -211,8 +248,88 @@ public void onSuccess(None result)
...
+      public void onSuccess(None result)
+      {
+        _log.info("markUp for uri {} on warm-up cluster {} succeeded", _uri, _warmupClusterName);
+        // Mark _isWarmingUp to true to indicate warm up in progress
+        _isWarmingUp = true;
+        // Add mark down as pending, so that in case of ZK connection loss, on retry there is a mark down attempted
+        // for warm-up cluster
+        _pendingWarmUpMarkDown.add(callback);
+        // Run warm-up for _warmupDuration ms and then schedule a mark down for the warm-up cluster
+        _log.debug("warm-up will run for {}ms",_warmupDuration);
```
mariabornski, line 316: minor: please add a space between  the "," and "_warmupDuration"

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -211,8 +248,88 @@ public void onSuccess(None result)
...
+        _executorService.schedule(() -> _server.markDown(_warmupClusterName, _uri, warmUpMarkDownCallback), _warmupDuration,
+              TimeUnit.MILLISECONDS);
+      }
+    };
+
     _log.info("overrideMarkUp is called for uri = " + _uri);
+    if (_isEnabledD2WarmupCluster)
```
mariabornski, line 323: Should we also check that "_warmupDuration > 0" before announcing to the warmup cluster?

If someone has warmup enabled but has set the duration to 0, it seems like a lot of churn to announce to the warmup cluster then immediately de-announce.  Additionally, I'm not sure what "_executorService.schedule" would do with a negative value (though hopefully that information is documented somewhere)

mariabornski (commented):
This is definitely moving along!  See inline for my comments & questions.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -53,6 +55,9 @@
 
 public class ZooKeeperAnnouncer
```
rickzx, line 56: 1. Can we add some unit tests in `ZookeeperConnectionManagerTest` for warm-up scenarios?
2. Have you tested the flow locally? You can publish a SNAPSHOT version of pegasus and try with local ZooKeeper. Please let me know if you need any help onthis

rickzx (changes requested):
Great work! Left some comments

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: Should I pasd just the warm up parameters through the new constructor? or others  like _cluster, _uri, _partitionDataMap and _uriSpecificProperties. too?

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: Thank you for reviewing.
1. Added unit tests in ZookeeperConnectionManagerTest for warm-up scenarios. 
2. Have tested the flow locally using a SNAPSHOT version of pegasus, with a SNAPSHOT version of container for testing dark warm-up feature with local ZooKeeper for a midtier service.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -118,6 +148,12 @@ synchronized void retry(Callback<None> callback)
   {
     // If we have pending operations failed because of a connection loss,
     // retry the last one.
```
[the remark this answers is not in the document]
rpuranda: As per the current proposed design, we are not keeping track here of _pendingWarmUpMarkUp and not retrying it. This is because dark warm-up is a best effort. If it fails due to ZooKeeper Connection loss or Session Expiry, then the retry method will try markUp again during which dark warm-up will also be retied in the flow. If dark warm-up fails due to any other exception then it is not retried and flow will proceed to markUp the regular cluster. Please suggest if you think this design should be changed.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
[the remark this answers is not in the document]
rpuranda: As per discussion with @rickzx  on slack, modified to only use the new constructor for ZooKeeperAnnouncer, because we don’t expect warm-up parameters to change at run-time.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: Updated the design to retry markDown for warm-up cluster in case of ZooKeeper connection loss or session expiry. In case of other exceptions will continue to announce to main cluster.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -82,8 +100,27 @@ public ZooKeeperAnnouncer(ZooKeeperServer server, boolean initialIsUp)
...
     _isUp = initialIsUp;
+    _isWarmUp = false;
+    _pendingMarkDown = new ArrayDeque<>();
+    _pendingMarkUp = new ArrayDeque<>();
+    _pendingWarmUpMarkDown = new ArrayDeque<>();
+
+  }
+
+  public ZooKeeperAnnouncer(ZooKeeperServer server, boolean initialIsUp, ScheduledExecutorService executorService,
+      boolean isEnabledD2WarmupCluster, String warmupClusterName, int warmupDuration)
```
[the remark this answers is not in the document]
rpuranda: Fixed.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: Deleted this method as for warm-up parameters only constructor should be used, as these should not change at runtime.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: Thank you for spotting this! Fixed.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
rpuranda: Fixed. Thank you for pointing this out. The condition will now check if warm up is enabled, _warmupDuration > 0 , _warmClusterName is not null and _executorService is not null.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -53,6 +55,9 @@
 
 public class ZooKeeperAnnouncer
 {
+  public static final boolean DEFAULT_D2_WARMUP_CLUSTER_ENABLED = false;
```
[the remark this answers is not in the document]
rpuranda: Fixed.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -162,14 +198,15 @@ public synchronized void markUp(final Callback<None> callback)
...
+
+    final Callback<None> markUpCallback = new Callback<None>()
     {
       @Override
       public void onError(Throwable e)
       {
         if (e instanceof KeeperException.ConnectionLossException || e instanceof KeeperException.SessionExpiredException)
         {
-          _log.debug("failed to mark up uri {} due to {}.", _uri, e.getClass().getSimpleName());
+          _log.info("failed to mark up uri {} for cluster {} due to {}.", _uri, _cluster, e.getClass().getSimpleName());
```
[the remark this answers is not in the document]
rpuranda: Fixed.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -211,8 +248,88 @@ public void onSuccess(None result)
...
+        _log.debug("warm-up will run for {}ms",_warmupDuration);
+        _executorService.schedule(() -> _server.markDown(_warmupClusterName, _uri, warmUpMarkDownCallback), _warmupDuration,
+              TimeUnit.MILLISECONDS);
+      }
+    };
+
     _log.info("overrideMarkUp is called for uri = " + _uri);
+    if (_isEnabledD2WarmupCluster)
+    {
+      _log.info("Warm-up is enabled with warm-up cluster {}", _warmupClusterName);
```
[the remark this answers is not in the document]
rpuranda: Fixed.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -65,25 +70,50 @@
...
+
+  // Field to indicate if warm up was started. If it is true, it will try to end the warm up
+  // by marking down on ZK if the connection goes down
+  private boolean _isWarmingUp;
+
   private final Deque<Callback<None>> _pendingMarkDown;
   private final Deque<Callback<None>> _pendingMarkUp;
 
+  // Queue to store pending mark down for warm-up cluster
+  private final Deque<Callback<None>> _pendingWarmUpMarkDown;
```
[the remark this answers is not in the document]
rpuranda: Fixed.

d2/src/test/java/com/linkedin/d2/balancer/servers/ZookeeperConnectionManagerTest.java
```diff
@@ -471,6 +472,311 @@ public void testMarkUpAndDownMultipleTimesFinalUp()
...
+    int warmupDuration = 5; //run warm up for 5 sec
+
+    ZooKeeperAnnouncer announcer = getZooKeeperWarmupAnnouncer(_cluster, _uri, WEIGHT, isDarkWarmupEnabled, warmupClusterName, warmupDuration, warmupExecutorService);
+    ZooKeeperConnectionManager manager = createManagerForWarmupTests(false, warmupDuration, announcer);
+    ZooKeeperEphemeralStore<UriProperties> store = createAndStartUriStore();
+    FutureCallback<None> managerStartCallback = new FutureCallback<>();
+    manager.start(managerStartCallback);
+
+    //ugly but to ensure that manager has started and has announcer has announced the regular cluster
+    Thread.sleep(1000);
```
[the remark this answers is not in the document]
rickzx: Heavy +1 on this. Please avoid using Thread.sleep if possible, so that running all the tests won't be cumbersome. Please see the example provided by Lester on using a ClockedExecutor

d2/src/test/java/com/linkedin/d2/balancer/servers/ZookeeperConnectionManagerTest.java
```diff
@@ -471,6 +472,311 @@ public void testMarkUpAndDownMultipleTimesFinalUp()
     executorService.shutdown();
   }
 
+  @Test
+  public void testNoWarmupWhenDisabled() throws Exception
```
rickzx, line 476: Thanks for writing all the unit tests. These are very comprehensive and convincing

rickzx (changes requested):
LGTM, one small comment about the issue raised by Lester

d2/src/test/java/com/linkedin/d2/balancer/servers/ZookeeperConnectionManagerTest.java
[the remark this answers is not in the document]
rpuranda: I couldn't get the ClockedExecutor to work. The schedule method of the Clocked Executor  (https://github.com/rpuranda/rest.li/blob/master/test-util/src/main/java/com/linkedin/test/util/ClockedExecutor.java#L94) wasn't working to schedule the markDown for the warm-up cluster (https://github.com/linkedin/rest.li/pull/726/files#diff-ce9b6eecd634167a616ca98580fd1b888067137f94c92c2e92026dc182f946ebR361). I have changed the test cases in a way to get the callback of ZooKeeperConnectionManager after a timeout to ensure warm-up has begun, to avoid using Thread.sleep. There is only 1 test case now that needs using a Thread.sleep after zkServer restart, to check that zkServer has started. Please review if this is acceptable.

d2/src/test/java/com/linkedin/d2/balancer/servers/ZookeeperConnectionManagerTest.java
```diff
@@ -471,6 +472,297 @@ public void testMarkUpAndDownMultipleTimesFinalUp()
...
+
+    // If warm up has happened, mark down for the warm up cluster should be successful
+    properties = store.get(warmupClusterName);
+    assertNotNull(properties);
+    assertEquals(properties.Uris().size(), 0);
+
+    shutdownManager(manager);
+  }
+
+  @Test (invocationCount = 10, timeOut = 15000)
```
rickzx, line 593: Each invocation takes 5s+, 10 invocations would add 1min runtime to the test suite. Can we do a single invoke and reduce warmup duration to something like 2s?

rickzx (changes requested):
Looks good! One minor comment.
Let's also change the title of the PR to take it out of draft status. And also update the changelog and bump minor version. You can refer to https://iwww.corp.linkedin.com/wiki/cf/pages/viewpage.action?spaceKey=ENGS&title=Pegasus+Git+Workflow#PegasusGitWorkflow-Changelog
Thanks!

d2/src/test/java/com/linkedin/d2/balancer/servers/ZookeeperConnectionManagerTest.java
rpuranda: Resolved.

rpuranda:
> Looks good! One minor comment. Let's also change the title of the PR to take it out of draft status. And also update the changelog and bump minor version. You can refer to https://iwww.corp.linkedin.com/wiki/cf/pages/viewpage.action?spaceKey=ENGS&title=Pegasus+Git+Workflow#PegasusGitWorkflow-Changelog Thanks!

Updated. Please review.

rickzx (approved):
LGTM. Thanks for the great work!

PR #727, opened 2021-11-16
Fixed issues with potential duplicate TimingKeys being registered.

zhang-chris:
Change to log warning instead of throwing an error when a duplicate timing key name is registered.
Guarantee uniqueness of filter timing key names.

### DIFF
--- a/r2-core/src/main/java/com/linkedin/r2/message/timing/TimingKey.java
+++ b/r2-core/src/main/java/com/linkedin/r2/message/timing/TimingKey.java
@@ -23,7 +23,10 @@ import java.util.concurrent.ConcurrentHashMap;
 import com.linkedin.r2.message.RequestContext;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicInteger;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * A timing key uniquely identifies a timing record, which will be saved in {@link RequestContext}.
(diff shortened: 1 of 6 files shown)


r2-core/src/main/java/com/linkedin/r2/message/timing/TimingKey.java
```diff
@@ -79,7 +84,7 @@ private static TimingKey registerNewKey(TimingKey timingKey)
   {
     if (_pool.putIfAbsent(timingKey.getName(), timingKey) != null)
     {
-      throw new IllegalStateException("Timing key " + timingKey.getName() + " has already been registered!");
+      LOG.warn("Timing key " + timingKey.getName() + " has already been registered!");
```
cx-super, line 87: Given that now we have unique key, this exception should be kept, right?

cx-super (changes requested):
I'm curious why we're getting duplicated keys. Was it because this filter is instantiated twice? That sounds an error to me.

r2-core/src/main/java/com/linkedin/r2/message/timing/TimingKey.java
zhang-chris: github doesn't let us reply to the top level comment.

"I'm curious why we're getting duplicated keys. Was it because this filter is instantiated twice? That sounds an error to me."
If you have multiple clients it's possible a collision happens. This is more likely in offline jobs where they spawn many clients, but it can also happen for online apps. 

I'd still rather log a warning than fail deployment for future potential edge cases like this one.

cx-super: What edge case we're talking about? We have logic to make the key unique. If it is not, it's a bug. We should fix it instead of let it go with a warning. What's wrong if we fail the deployment like this?

zhang-chris: Imo it's not a severe enough error to justify failing the deployment. The app can still operate completely ok - just the timings may be inaccurate.

cx-super: OK

cx-super (approved):

PR #733, opened 2021-11-29
Add a Spotbug plugin that detect and prevent use of avro1.4 and avro1.6 API

junchuanwang:
This plugin can prevent user from using of old API before we release the new major version that completely get rid of legacy Avro versions (1.4 & 1.6)

After all migration is finished, we might or might not need to remove it. (It should be OK to remove it)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.22.15] - 2021-11-29
+- Add SpotBugs plugin that detect use of Avro1.4 and Avro1.6 API uses during build time
+
 ## [29.22.14] - 2021-11-24
 - Fix bug where content type was not set for stream requests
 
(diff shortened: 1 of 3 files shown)


nickibi:
How this plugin works？If there is any avro 1.4 or 1.6 APIs, it will trigger build failure?

junchuanwang:
> How this plugin works？If there is any avro 1.4 or 1.6 APIs, it will trigger build failure?

Correct

nickibi:
> 

There are some "deprecated" classes, for example: [AvroAdapter](https://github.com/linkedin/rest.li/blob/7372bc76f7ce78007ecb24ca80f17399e5246e13/data-avro/src/main/java/com/linkedin/data/avro/AvroAdapter_1_4.java#L36), are still using avro 1.4 or 1.6 APIs, whether those will trigger build failure?

junchuanwang:
> > 
> 
> There are some "deprecated" classes, for example: [AvroAdapter](https://github.com/linkedin/rest.li/blob/7372bc76f7ce78007ecb24ca80f17399e5246e13/data-avro/src/main/java/com/linkedin/data/avro/AvroAdapter_1_4.java#L36), are still using avro 1.4 or 1.6 APIs, whether those will trigger build failure?

Those files will not trigger build failure. But if user tries to use them, then it will.

These files will be cleaned up next major release for backward incompatibility.

nickibi:
> 

"User" here you mean other libraries which use Rest.li? Then that is a behavior change for them, we cannot release such change in patch version.

junchuanwang:
> > 
> 
> "User" here you mean other libraries which use Rest.li? Then that is a behavior change for them, we cannot release such change in patch version.

My bad. I mean any use of these files inside this **Rest.Li library**. Corrected wording

nickibi:
> 

If it will only affect Rest.li itself, why we need to release such change?

junchuanwang:
> > 
> 
> If it will only affect Rest.li itself, why we need to release such change?

Make sense. Not releasing this for new revision

nickibi (approved):

PR #736, opened 2021-12-01
Add file existence check before performing compatibility report check

junchuanwang:
Previously we didn't write any compatibility report if no files found. 

That case typically happens for the "test" sourceSet. Since normally Gradle process "main" sourceset first and then "test" sourceset, therefore this bug is not causing trouble since "test" sourceset just reuse "main" sourceset result.

But it will cause IOException in some corner cases, such as user directly try to process test sourceset.

This change add an additional check. As a result, if no files found, no compatibility report will be return and no compatibility report will be checked

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,10 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
+
+## [29.22.15] - 2021-11-30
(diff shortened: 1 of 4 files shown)


nickibi (commented):
Whether same thing applies to CheckRestModelTask as well?

evanw555:
Does it make sense to publish empty compatibility report files? Could that be misleading?

Have we considered the alternative: update the reader logic to gracefully handle cases where there's no compatibility report file?

junchuanwang:
@nickibi @evanw555 
Talked with Evan offline and later adopted his suggestion.

nickibi (approved):

PR #740, opened 2021-12-06
Introduce d2ServiceName annotation on rest.li resources

karthikrg:
This is an optional param that is meant to be populated by resources for whom the resource name is not the same as d2 service name (if not populated we default to null). This is useful for situations like generating routing hints for graphQL.

This value is also emitted into the IDL. Any changes to this value are considered backward incompatible.

### DIFF
--- a/restli-server/src/main/java/com/linkedin/restli/internal/server/model/ResourceModel.java
+++ b/restli-server/src/main/java/com/linkedin/restli/internal/server/model/ResourceModel.java
@@ -57,6 +57,7 @@ public class ResourceModel implements ResourceDefinition
 
   private final String                          _name;
   private final String                          _namespace;
+  private final String                          _d2ServiceName;
 
   private final Class<?>                        _resourceClass;
   private final ResourceType                    _resourceType;
@@ -114,6 +115,39 @@ public class ResourceModel implements ResourceDefinition
                        final ResourceType resourceType,
                        final String namespace)
   {
+    this(primaryKey, keyKeyClass, keyParamsClass, keys, valueClass, resourceClass, parentResourceClass, name,
+        resourceType, namespace, null);
+  }
+
+  /**
+   * Constructor.
+   *
+   * @param primaryKey the primary {@link Key} of this resource
+   * @param keyKeyClass class of the key part of a {@link ComplexResourceKey} if this is a
+   *          {@link ComplexKeyResource}
+   * @param keyParamsClass class of the param part of a {@link ComplexResourceKey} if this
+   *          is a {@link ComplexKeyResource}
+   * @param keys set of resource keys
+   * @param valueClass resource value class
+   * @param resourceClass resource class
+   * @param parentResourceClass parent resource class
+   * @param name resource name
+   * @param resourceType {@link ResourceType}
+   * @param namespace namespace
+   * @param d2ServiceName The d2 service name for the resource
+   */
+  public ResourceModel(final Key primaryKey,
+      final Class<? extends RecordTemplate> keyKeyClass,
+      final Class<? extends RecordTemplate> keyParamsClass,
+      final Set<Key> keys,
+      final Class<? extends RecordTemplate> valueClass,
+      final Class<?> resourceClass,
+      final Class<?> parentResourceClass,
+      final String name,
+      final ResourceType resourceType,
+      final String namespace,
+      final String d2ServiceName)
+  {
     _keyKeyClass = keyKeyClass;
     _keyParamsClass = keyParamsClass;
     _keys = keys;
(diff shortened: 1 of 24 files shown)


restli-server/src/main/java/com/linkedin/restli/internal/server/model/ResourceModel.java
```diff
@@ -164,6 +199,38 @@ public ResourceModel(final Class<? extends RecordTemplate> valueClass,
          namespace);
   }
 
+  /**
+   * Constructor.
```
mchen07, line 203: Is this only introduced for simple resource? If so, maybe add this in comment.

mchen07 (commented):
Where is compatibility change?

mchen07:
Also, we need to expose this d2ServiceName into IDL so that graphql-infra can get this to generate into GraphQL schema.

restli-tools/src/main/java/com/linkedin/restli/tools/compatibility/ResourceCompatibilityChecker.java
```diff
@@ -224,6 +225,20 @@ private boolean checkEqualSingleValue(RecordDataSchema.Field field, Object prevD
     return true;
   }
 
+  private boolean checkD2ServiceName(String prevData, String currData) {
```
mchen07, line 228: what if `prevData` is null, but `currData` is set to the same as `resource name`?

restli-server/src/main/java/com/linkedin/restli/internal/server/model/ResourceModel.java
```diff
@@ -164,6 +198,38 @@ public ResourceModel(final Class<? extends RecordTemplate> valueClass,
...
+   *
+   * @param valueClass resource value class
+   * @param resourceClass resource class
+   * @param parentResourceClass parent resource class
+   * @param name resource name
+   * @param resourceType {@link ResourceType}
+   * @param namespace namespace
+   * @param d2ServiceName The d2 service name for the resource
+   */
+  public ResourceModel(final Class<? extends RecordTemplate> valueClass,
```
mchen07, line 212: Is this new constructor ever used? No need to introduce this variant if not used.

restli-tools/src/main/java/com/linkedin/restli/tools/compatibility/ResourceCompatibilityChecker.java
karthikrg: So I don't want to enforce that it should be the same as the resource name. That is a convention that people may not always choose to follow. For the purposes of GraphQL routing alone, I will add a check that if null, use the resource name, since we are confident about that convention there.

restli-server/src/main/java/com/linkedin/restli/internal/server/model/ResourceModel.java
karthikrg: This new one is used by the annotation reader.

restli-tools/src/test/resources/idls/curr-greeting-simple-fail.restspec.json
```diff
@@ -1,5 +1,6 @@
 {
   "name" : "greeting",
+  "d2ServiceName": "greetingSimpleD2",
```
evanw555, line 3: There are a couple concerns I have with this change:
1. Should the d2 service name be surfaced in the resource class? This seems like a tight coupling that assumes Rest.li resources will be used with d2. That may not always be the case.
2. If we decide that it should be surfaced in the resource class, should it be a first-class citizen on the resource annotation? This seems to imply that this annotation will affect native framework behavior. As far as I can tell, there are no corresponding changes in d2. Perhaps then it would be more appropriate as a custom annotation defined for the D2-specific use case, since it's just inert arbitrary data added onto the resource.

mchen07: custom annotation is a good point

restli-tools/src/main/java/com/linkedin/restli/tools/compatibility/ResourceCompatibilityChecker.java
karthikrg: Resolved offline with @mchen07

restli-tools/src/test/resources/idls/curr-greeting-simple-fail.restspec.json
karthikrg: 1. No other easy way to handle routing in a clean way.
2. Custom annotations are weird because we do want to use this to check backward compatibility in the framework

Discussed offline to update the javadocs to indicate the same.

mchen07 (approved):

PR #741, opened 2021-12-09
update DefaultSymbolTableProvider to take default headers as input

aman1309:
This is required to add any static headers required for the application downstream calls


gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=29.23.2
+version=29.23.3
```
nickibi, line 1: Adding new public APIs requires a minor version bump

PR #746, opened 2022-01-11
Fail documentation requests while renderers are being lazily initialized

evanw555:
In `DefaultDocumentationRequestHandler`, the documentation renderers are lazily initialized. Request threads will block on the completion of this initialization, which could cause problems if it takes too long. This changes that, as request threads will simply fail if it's not complete.


mchen07 (approved):
LGTM

nickibi (approved):

PR #747, opened 2022-01-13
Fix for SI-24257 (NPS from PropertyEventBusImpl)

jpstewart:
Fix for SI-24257, where NPS is thrown when registering listener before publisher exists on an event bus.

Some event buses are created without a publisher, as a publisher can be set later via setPublisher. However the register listener function assumes that a publisher exists when a first listener is added for a property name, and attempts to tell it to start publishing. Fix is to simply check for a null publisher. Same issue potentially exists in unregister. setPublisher already correctly handles starting publishing if listeners are already registered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.26.2] - 2022-01-13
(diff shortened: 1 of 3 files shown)


li-bsoetarman (approved):
LGTM

PR #748, opened 2022-01-19
Generate compile time constants for union members

karthikrg:
This is useful if if people want to write build time switch case to check for things based on member


mchen07 (approved):
LGTM

PR #750, opened 2022-01-20
Modified the service-cluster mapping for regular service Znodes to be…

JieyingXu:
Modified the service-cluster mapping for regular service Znodes to be consistent with the change in d2-config-service.

Before the change, regular service Znode will be mapped to clusterA-localColo, e.g. clusterA-WestCoast.
After the change, it will be mapped to clusterA. The local colo suffix will not present.
This is to remove the requirements on operation orders when enabling/disabling cross colo routing to prevent errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and what APIs have changed, if applicable.
 ## [Unreleased]
 
 ## [29.26.4] - 2022-01-24
-- Modified the service-cluster mapping for regular service Znodes to be consistent with the change in d2-config-service.
+- Map local variant of service ZNodes to cluster without colo suffix
 
 ## [29.26.3] - 2022-01-18
 - Generate compile time constants for union members if people want to write build time switch case to check for things based on member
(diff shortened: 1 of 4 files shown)


rickzx (changes requested):
Please make sure to update gradle.properties and run ./script/update_changelog

bohhyang (commented):
echo Rick's comment.

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.26.4] - 2022-01-24
+- Modified the service-cluster mapping for regular service Znodes to be consistent with the change in d2-config-service.
```
rickzx, line 18: Let's not mention d2-config-service as this is a LI-internal MP.
I would prefer `Map local variant of service ZNodes to cluster without colo suffix`

rickzx (changes requested):

rickzx (approved):

bohhyang (approved):
lgtm. thx

PR #751, opened 2022-01-24
Update avro to 1.9.2

junchuanwang:
Upgrade declared avro version to 1.9.2 only.

The deprecated avro util classes will be delete in next major version release


data/src/main/java/com/linkedin/data/codec/AbstractJacksonDataCodec.java
```diff
@@ -57,10 +58,12 @@
    *
    * <a href="https://github.com/FasterXML/jackson-docs/wiki/Presentation:-Jackson-Performance">Jackson Performance</a>
    */
-  public static final JsonFactory JSON_FACTORY = new JsonFactory();
+  public static final JsonFactoryBuilder JSON_FACTORY_BUILDER = new JsonFactoryBuilder();
```
aman1309, line 61: Builder can be moved to static block

gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=29.26.3
+version=29.26.4
```
nickibi, line 1: You introduce new APIs in AvroUtil, it should bump minor version rather than patch version.

data/src/main/java/com/linkedin/data/codec/AbstractJacksonDataCodec.java
junchuanwang: this is just perference, right?

gradle.properties
junchuanwang: adopted

gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=29.26.3
+version=29.27.1
```
evanw555, line 1: Wy not `29.27.0`?

junchuanwang: did

nickibi (approved):

data/src/main/java/com/linkedin/data/codec/AbstractJacksonDataCodec.java
aman1309: this would save space also. we don't need to keep reference to it in codec.

PR #752, opened 2022-01-28
Add RestLiTraceInfo to the request context for both client and server requests

dmessink:
I'd like to expose the RestRequest template to both client and server requests.

I'm trying to work around several issues...

- The R2 client and server don't have access to the resource template (including subresources and params).
- We have an rest client interface - but it's not used everywhere. Using the RestClient class directly is much more common - so we can't decorate user's clients. This means I need to do this work in the RestClient class.
- It seems the restli server doesn't use Request objects that know about the template.
- The template doesn't seem to be available in the ResourceModel... but we may be able to reconstruct it.
- We have a restli filter chain on the server, but not the client. So I could implement this as a filter for the server-side, but not the client.

I did the best I could -- but didn't finish it. Thinking we should discuss all this before committing more time to it.


sudheerv:
Hey @dmessink - How is the new context info from the Client Request passed along to the callee? ie, are we able to have access to the context within the client request from the server request at the callee? Do we plan to use IC for that or OTEL context headers (tracecontext)?

dmessink:
It's not actually passed. The PR generates the new target/operation values for both the client and the server and adds them to the request context.

Our filters will pull that data and add it to the span/sce.

So nothing needs to be passed.
  Dave
________________________________

PR #753, opened 2022-01-31
fix request service -> symbol table cache to invalidate stale entry

aman1309:


karthikrg (changes requested):
This is a dangerous change and may cause thrashing if we have multiple versions of a service concurrently deployed.

PR #754, opened 2022-02-04
Fix ConcurrentModificationException thrown in R2 RequestContext

rickzx:
A constructor in RequestConext is inherently not thread-safe.

As per javadoc: It is imperative that the user manually synchronize on the returned map when iterating over any of its collection views


bbarkley (approved):
Please add tests to cover this case in a followup PR.

PR #755, opened 2022-02-05
support picking d2 config canary

bohhyang:
**Background**
As defined for D2 config phased deployment, D2 client needs to be able to pick between canary vs stable configs when canary config is present in D2 config.

**Changes**
1. Added a canary distribution provider interface (which will be implemented in container) which will decide the canary distribution result. 
2. Passing in the provider from D2 client builder to simple load balancer state. 
3. Added logic in cluster and service config subscriber to leverage the provider and use picked config (Will create a follow-up PR to add metrics for tracking).

**Tests Done**
Included snapshot in container and run wc-test.


d2/src/main/java/com/linkedin/d2/balancer/util/CanaryDistributionProvider.java
```diff
@@ -0,0 +1,27 @@
+package com.linkedin.d2.balancer.util;
+
+import com.linkedin.d2.D2CanaryDistributionStrategy;
+
+
+/**
+ * Canary Distribution Provider is supposed to provide information about canary distributions.
+ */
+public interface CanaryDistributionProvider {
```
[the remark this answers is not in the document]
rickzx: And if we agreed to implement this in pegasus, i feel like it'd be sufficient to just have the class, rather than class + interface. The polymorphism is already presented in D2CanaryDistributionStrategy

rickzx (changes requested):
Thanks for making the change. Left some comments

d2/src/main/java/com/linkedin/d2/balancer/util/CanaryDistributionProvider.java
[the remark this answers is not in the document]
bohhyang: we discussed this with the team in one of the standups last year, and the team concluded with putting the implementation class in container for the reasons:
1. the distribution logic may change in the future to involve more LI-specific info. We don't want to wait till then to move it with backward-compatibility problems. The suggestion is to abstract the operation and not the data. This will help us use the same abstraction layer to implement different canary strategies depending on future data that is not anticipated today and not readily available in pegasus. If we have to do the canary picking logic in pegasus - we have to build the abstraction around how to pass the data into pegasus, and that abstraction is easily broken if we need additional data for future canary use case that is not covered today.
2. the canary distribution logic could be re-used by other features in both pegasus and container in future too.

The reason to have inter + class is because D2CanaryDistributionStrategy only defines different canary strategies (ramp, target host, target apps) but there could be different implementations in the distribution logic. My current implementation uses XXHash, while future implementations may add different hash logic or depend on completely different mechanisms.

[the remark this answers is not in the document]
bohhyang: I thought about this too. But "Distributor" kind of limits what the class can do to only "distribute the canary". What if we add another function to the interface in the future that has nothing to do with distributing (like combining an old and a new strategy?). "Provider" sounds more general as an info provider to me.

d2/src/main/java/com/linkedin/d2/balancer/simple/ClusterLoadBalancerSubscriber.java
```diff
@@ -43,13 +46,24 @@ public ClusterLoadBalancerSubscriber(SimpleLoadBalancerState simpleLoadBalancerS
...
   protected void handlePut(final String listenTo, final ClusterProperties discoveryProperties)
   {
-    // TODO: pick stable or canary configs if canary exists
     if (discoveryProperties != null)
     {
+      ClusterProperties pickedProperties = discoveryProperties;
+      if (discoveryProperties instanceof ClusterPropertiesWithCanary) {
+        // has canary config, distribute to use either stable config or canary config
+        ClusterPropertiesWithCanary propertiesWithCanary = (ClusterPropertiesWithCanary) discoveryProperties;
+        CanaryDistributionProvider.Distribution distribution = _simpleLoadBalancerState.getCanaryDistributionProvider()
```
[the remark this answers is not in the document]
bohhyang: They are guaranteed non-null in the ClusterPropertiesJsonSerializer(with default values), same as the service properties serializer. I will add a comment to clarify. thanks.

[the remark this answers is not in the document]
bohhyang: the distribution provider could be null though. I will add a null check for that.

d2/src/main/java/com/linkedin/d2/balancer/util/CanaryDistributionProvider.java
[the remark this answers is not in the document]
rickzx: IMO, even if we may need LI-specific info in the future, we should still strive to provide at least a basic implementation in pegasus, and have a wrapper in container to extend the basic implementation. For example, `BaseDarkClusterDispatcherImpl` is the concrete class of `DarkClusterDispatcher`, which is extended by container to include LI-specific info like IC. Otherwise, this canary config feature is missing a core component in the open-source version.

d2/src/main/java/com/linkedin/d2/balancer/simple/ClusterLoadBalancerSubscriber.java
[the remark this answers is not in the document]
bohhyang: in fact, the canary data may be null once failover data is added (which extends ClusterPropertiesWithCanary). I will do null checks, thanks.

d2/src/main/java/com/linkedin/d2/balancer/util/CanaryDistributionProvider.java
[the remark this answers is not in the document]
bohhyang: make sense. I will add a very basic implementation in pegasus and let the container one extends it.

d2/src/main/java/com/linkedin/d2/balancer/simple/ClusterLoadBalancerSubscriber.java
```diff
@@ -43,13 +46,24 @@ public ClusterLoadBalancerSubscriber(SimpleLoadBalancerState simpleLoadBalancerS
   @Override
   protected void handlePut(final String listenTo, final ClusterProperties discoveryProperties)
   {
-    // TODO: pick stable or canary configs if canary exists
     if (discoveryProperties != null)
     {
+      ClusterProperties pickedProperties = discoveryProperties;
+      if (discoveryProperties instanceof ClusterPropertiesWithCanary) {
```
bbarkley, line 52: Sorry I missed the original review where the canary classes were introduced, I should have brought this up there. What was the motivation for creating subclasses (`ClusterPropertiesWithCanary` and `ServicePropertiesWithCanary`) instead of adding the new data to the base classes and having it be null by default? The issue with using subclassing is it locks you in to a strict structure and doesn't allow for composition. As you mention below with the current model the failover data would need to extend the canary classes, but these features aren't necessarily related in that way. 
Is this necessary to preserve binary compatibility on the base classes? If not I think modifying them and dropping the subclasses would be a cleaner approach.

bbarkley (commented):
What test cases can be written for this? I think we've found that test coverage overall for Rest.li isn't sufficient and has led to some recent GCNs.

d2/src/main/java/com/linkedin/d2/balancer/util/CanaryDistributionProvider.java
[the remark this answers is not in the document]
bohhyang: added. thanks.

gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=29.27.0
+version=29.27.1
```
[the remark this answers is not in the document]
bohhyang: bumped, thanks.

d2/src/main/java/com/linkedin/d2/balancer/simple/ClusterLoadBalancerSubscriber.java
bohhyang: added composite classes and tests for picking between canary and stable configs.

d2/src/main/java/com/linkedin/d2/balancer/simple/ClusterLoadBalancerSubscriber.java
```diff
@@ -43,13 +46,28 @@ public ClusterLoadBalancerSubscriber(SimpleLoadBalancerState simpleLoadBalancerS
...
+        if (clusterStoreProperties.hasCanary() && canaryDistributionProvider != null)
+        {
+          // Canary config and canary distribution provider exist, distribute to use either stable config or canary config.
+          distribution = canaryDistributionProvider
+              .distribute(CanaryDistributionStrategyConverter.toConfig(clusterStoreProperties.getCanaryDistributionStrategy()));
+        }
+        pickedProperties = clusterStoreProperties.getDistributedClusterProperties(distribution);
+      }
+      // TODO: set canary/stable config metric
+
```
rickzx, line 66: nit: Maybe extract this to a separate helper function?

d2/src/main/java/com/linkedin/d2/balancer/simple/ServiceLoadBalancerSubscriber.java
```diff
@@ -47,25 +50,39 @@ public ServiceLoadBalancerSubscriber(PropertyEventBus<ServiceProperties> eventBu
...
+      CanaryDistributionProvider canaryDistributionProvider = _simpleLoadBalancerState.getCanaryDistributionProvider();
+      if (serviceStoreProperties.hasCanary() && canaryDistributionProvider != null) {
+        // Canary config and canary distribution provider exist, distribute to use either stable config or canary config.
+        distribution = canaryDistributionProvider
+            .distribute(CanaryDistributionStrategyConverter.toConfig(serviceStoreProperties.getCanaryDistributionStrategy()));
+      }
+      pickedProperties = serviceStoreProperties.getDistributedServiceProperties(distribution);
+    }
+    // TODO: set canary/stable config metric
+
```
rickzx, line 70: ditto

rickzx (commented):
LGTM, just some minor suggestions

d2/build.gradle
```diff
@@ -24,6 +24,7 @@ dependencies {
   testCompile externalDependency.commonsIo
   testCompile externalDependency.easymock
   testCompile externalDependency.mockito
+  testCompile externalDependency.guava
```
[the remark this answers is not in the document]
bohhyang: nope, this is needed in the test for ImmutableMap.

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.30.0] - 2022-02-12
+support d2 config canary
```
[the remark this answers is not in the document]
bohhyang: thanks for catching. fixed.

d2/src/main/java/com/linkedin/d2/balancer/simple/ClusterLoadBalancerSubscriber.java
bohhyang: sure, cleaner then. done.

d2/src/main/java/com/linkedin/d2/balancer/simple/ServiceLoadBalancerSubscriber.java
bohhyang: done.

d2/src/main/java/com/linkedin/d2/balancer/properties/ClusterPropertiesWithCanary.java
```diff
@@ -1,13 +1,37 @@
...
+
 /**
- * Cluster properties with canary configs and a distribution strategy, serialized and stored on service registry.
+ * ClusterPropertiesWithCanary includes {@link ClusterProperties} and a canary version of the cluster properties, and a distribution strategy.
+ * It is serialized and stored on service registry (like Zookeeper).
+ * Canary cluster properties are to be used when a D2 client falls into the canary group, as opposed to the original cluster properties.
+ *
+ * NOTE: {@link ClusterStoreProperties} includes ALL properties on a cluster store on service registry (zookeeper).
+ *
+ * DEPRECATED: use {@link ClusterStoreProperties} instead.
```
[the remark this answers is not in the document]
bohhyang: I was concerned that removing public classes (even unused) will cause failures in container wc-test for backward compatibility. Just tried removing them and running a container wc-test, and it succeeded.

d2/src/main/java/com/linkedin/d2/balancer/properties/ClusterStoreProperties.java
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  @Override
+  public boolean equals(Object obj)
+  {
+    if (!super.equals(obj))
+    {
+      return false;
+    }
+    ClusterStoreProperties other = (ClusterStoreProperties) obj;
```
bbarkley, line 178: This cast isn't safe. Even if the superclass is doing a type check you're only guaranteed to have a ClusterProperties object. Unfortunately you have the general problem of non-symmetric equals since you have a subclass situation and the it may return true when comparing two objects of different types, but false when the comparison order is reversed. See item 10 from Effective Java for more info. I don't see a good way around this given other constraints, but the cast should be type checked first at least.

d2/src/main/java/com/linkedin/d2/balancer/properties/ServiceStoreProperties.java
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  @Override
+  public boolean equals(Object obj)
+  {
+    if (!super.equals(obj))
+    {
+      return false;
+    }
+    ServiceStoreProperties other = (ServiceStoreProperties) obj;
```
bbarkley, line 198: Cast issue

d2/src/main/java/com/linkedin/d2/balancer/simple/ClusterLoadBalancerSubscriber.java
```diff
@@ -68,4 +72,29 @@ protected void handleRemove(final String listenTo)
...
+  /**
+   * Pick the active properties (stable or canary configs) based on canary distribution strategy.
+   * @param discoveryProperties a composite properties containing all data on the cluster store (stable configs, canary configs, etc.).
+   * @return the picked active properties
+   */
+  private ClusterProperties pickActiveProperties(final ClusterProperties discoveryProperties)
+  {
+    ClusterProperties pickedProperties = discoveryProperties;
+    CanaryDistributionProvider.Distribution distribution = CanaryDistributionProvider.Distribution.STABLE;
+    if (discoveryProperties instanceof ClusterStoreProperties) // this should always be true since the serializer returns the composite class
```
bbarkley, line 85: Should we be doing something if the assumption isn't true? Logging at least?

d2/src/main/java/com/linkedin/d2/balancer/util/canary/BasicCanaryDistributionProviderImpl.java
```diff
@@ -0,0 +1,127 @@
...
+    if (percentageProperties == null) {
+      _log.warn("Empty percentage properties in distribution strategy type.");
+      return Distribution.STABLE;
+    }
+    return isCanaryByRampScope(percentageProperties.getScope()) ? Distribution.CANARY : Distribution.STABLE;
+  }
+
+  protected String getServiceName()
+  {
+    return _serviceName == null ? "" : _serviceName;
```
bbarkley, line 100: Why not do this null prevention logic in the constructor?

d2/src/test/java/com/linkedin/d2/balancer/properties/ClusterPropertiesSerializerTest.java
```diff
@@ -70,30 +70,34 @@ public void testClusterPropertiesSerializer() throws PropertySerializationExcept
...
-    assertEquals(jsonSerializer.fromBytes(jsonSerializer.toBytes(property)), property);
+    ClusterStoreProperties storeProperties = new ClusterStoreProperties(property, null, null);
+    // cluster properties will be serialized then deserialized as cluster store properties
+    assertEquals(jsonSerializer.fromBytes(jsonSerializer.toBytes(property)), storeProperties);
+    // cluster store properties will be serialized then deserialized as cluster store properties
+    assertEquals(jsonSerializer.fromBytes(jsonSerializer.toBytes(storeProperties)), storeProperties);
 
     property = new ClusterProperties("test", schemes);
-    assertEquals(jsonSerializer.fromBytes(jsonSerializer.toBytes(property)), property);
+    assertEquals(jsonSerializer.fromBytes(jsonSerializer.toBytes(property)), new ClusterStoreProperties(property, null, null));
```
bbarkley, line 80: For backward compatibility the old test of comparing it to the property should still pass too right?

d2/src/test/java/com/linkedin/d2/balancer/util/canary/TestBasicCanaryDistributionProviderImpl.java
```diff
@@ -0,0 +1,210 @@
...
+    // verify hash results are almost uniformly distributed
+    int bucketCount = 100;
+    int attemptCount = 1000;
+    int[] hits = new int[bucketCount]; // to count hits in each bucket 0 ~ 99.
+    String serviceName = "test-service";
+    String hostName = "lor1-app";
+    int bucketIdx;
+    int hashResult;
+    for (int i = 0; i < attemptCount; i++) {
+      hashResult = (serviceName + hostName + i).hashCode();
```
bbarkley, line 198: This should be using the actual implementation in BasicCanaryDistributionProviderImpl, otherwise that might change and this test will still pass.

d2/src/test/java/com/linkedin/d2/balancer/util/canary/TestBasicCanaryDistributionProviderImpl.java
```diff
@@ -0,0 +1,210 @@
...
+
+  @Test
+  public void testHashDistribution()
+  {
+    // verify hash results are almost uniformly distributed
+    int bucketCount = 100;
+    int attemptCount = 1000;
+    int[] hits = new int[bucketCount]; // to count hits in each bucket 0 ~ 99.
+    String serviceName = "test-service";
+    String hostName = "lor1-app";
```
bbarkley, line 194: I believe the host names that the service gets are fully qualified - this should probably test with that, or using both.

d2/src/test/java/com/linkedin/d2/balancer/util/canary/TestBasicCanaryDistributionProviderImpl.java
```diff
@@ -0,0 +1,210 @@
...
+    int bucketIdx;
+    int hashResult;
+    for (int i = 0; i < attemptCount; i++) {
+      hashResult = (serviceName + hostName + i).hashCode();
+      bucketIdx = abs(hashResult) % 100; // hash "test-service"+"lor1-appXXX"
+      hits[bucketIdx]++;
+    }
+
+    // compare each bucket hits with the average hit, verify offsets are < 4
```
bbarkley, line 203: Using 4 seems like a large range in distribution (almost 50% right?). Does if fail if you decrease that? Ideally I'd think we'd want it to be within 10% which is the limit of the test as it's set up.

bbarkley (commented):
Great progress here. Thanks for all of the additional tests. Mostly minor things, though the hash distribution could be an issue still.

d2/src/test/java/com/linkedin/d2/balancer/util/canary/TestBasicCanaryDistributionProviderImpl.java
bohhyang: could you specify what do you mean here? Do you mean to use the real host and service name that the test gets? We still need to append the string with the different attempt number, so it becomes xxx0, xxx1, xxx2, xxx3, ... xxx1000. It will be the same as using dummy service and host names, won't it?

d2/src/main/java/com/linkedin/d2/balancer/properties/ClusterStoreProperties.java
bohhyang: the super class's type check "getClass() != obj.getClass()" actually will fail if we are comparing a ClusterStoreProperties with a ClusterProperties (checked by stepping thru the code). It will see "ClusterStoreProperties" != "ClusterProperties". 
So only when the class names are exactly the same, (comparing a ClusterProperties with a ClusterProperties, or a ClusterStoreProperties with a ClusterStoreProperties), the super class type check will pass. This means at line 178 in this class, the class type is guaranteed ClusterStoreProperties.

d2/src/test/java/com/linkedin/d2/balancer/properties/ClusterPropertiesSerializerTest.java
bohhyang: nope. As explained in my other comment, The super class's equals method will fail since it sees the class names are not the same (ClusterProperties != ClusterStoreProperties).

bohhyang: Backward compatibility is still ensured since the serializer can serialize a ClusterProperties into bytes, and when deserializing (taking bytes back from ZK), it will create a ClusterStoreProperties, which will then be broken down to ClusterProperties again to be saved in LoadBalancerState.

d2/src/main/java/com/linkedin/d2/balancer/util/canary/BasicCanaryDistributionProviderImpl.java
bohhyang: just wanted to be tolerant with this basic impl, since the hash logic will technically still work will null values (will use empty strings). The container extension uses serviceInfo, so it won't have null values.

d2/src/test/java/com/linkedin/d2/balancer/properties/ClusterPropertiesSerializerTest.java
bohhyang: We essentially changed the underlining data used in ZK sync (without changing the interface leveraging the superclass-subclass trick), so we bumped the minor version of pegasus to reflect this change.

d2/src/test/java/com/linkedin/d2/balancer/util/canary/TestBasicCanaryDistributionProviderImpl.java
bohhyang: makes sense. fixed.

d2/src/test/java/com/linkedin/d2/balancer/util/canary/TestBasicCanaryDistributionProviderImpl.java
bohhyang: The total skewed hits are actually < 7% (about 62 / 1000). Please see my updated test.

d2/src/main/java/com/linkedin/d2/balancer/simple/ClusterLoadBalancerSubscriber.java
bohhyang: we have changed the serialiers to always return the composite class. If it doesn't, the unit tests will fail.

rickzx (commented):
LGTM, thanks for addressing the comments! i'm ready to give approval if @bbarkley is ready

d2/src/main/java/com/linkedin/d2/balancer/properties/ClusterStoreProperties.java
bbarkley: Okay, thanks for confirming. In general it's not safe to assume the behavior of a super class will remain unchanged - doing the type check here for safety and clarity is still a good idea.

d2/src/main/java/com/linkedin/d2/balancer/simple/ClusterLoadBalancerSubscriber.java
bbarkley: Are they unit tests for this class? If not it'd be possible to make that change, see the tests failing, and then fix the tests. But this would still break.

d2/src/test/java/com/linkedin/d2/balancer/properties/ClusterPropertiesSerializerTest.java
bbarkley: Thanks for the explanation.

d2/src/test/java/com/linkedin/d2/balancer/util/canary/TestBasicCanaryDistributionProviderImpl.java
bbarkley: Yes, I mean using the full host name including the other domains. The distribution of the hashing algorithm could be different between something like lor1-app001 and lor1-app001.linkedin.prod. I think we should be validating with data as similar to production as possible.

d2/src/test/java/com/linkedin/d2/balancer/util/canary/TestBasicCanaryDistributionProviderImpl.java
```diff
@@ -183,28 +183,44 @@ public void testEdgeCases(D2CanaryDistributionStrategy strategy)
...
     int[] hits = new int[bucketCount]; // to count hits in each bucket 0 ~ 99.
     String serviceName = "test-service";
     String hostName = "lor1-app";
     int bucketIdx;
-    int hashResult;
+    BasicCanaryDistributionProviderImpl impl = spy(new BasicCanaryDistributionProviderImpl(serviceName, hostName));
     for (int i = 0; i < attemptCount; i++) {
-      hashResult = (serviceName + hostName + i).hashCode();
-      bucketIdx = abs(hashResult) % 100; // hash "test-service"+"lor1-appXXX"
+      when(impl.getHashKey()).thenReturn(serviceName + hostName + i);
```
bbarkley, line 203: Is there a reason to not use the actual implementation of getHashKey here? Otherwise if it changes the test won't break.

d2/src/test/java/com/linkedin/d2/balancer/util/canary/TestBasicCanaryDistributionProviderImpl.java
```diff
@@ -183,28 +183,44 @@ public void testEdgeCases(D2CanaryDistributionStrategy strategy)
...
+  // in Map with String keys. But we explicitly test it to verify its behavior.
   @Test
   public void testHashDistribution()
   {
     // verify hash results are almost uniformly distributed
-    int bucketCount = 100;
+    int bucketCount = 100; // since ramp is 100% based, we test with 100 buckets
     int attemptCount = 1000;
+    // count how many attempts are skewed
+    // (e.g: one bucket gets 13 will cause another bucket gets 7, the attempts skewed in this pair of buckets is ((13 - 10) + (10 - 7)) / 2 = 3
```
bbarkley, line 195: I don't know if we should be dividing by two here. One bucket will receive 30% more, and one will receive 30% less, but the way you're doing the calculation would be against 20 attempts, and it would come out as 15%.

d2/src/test/java/com/linkedin/d2/balancer/util/canary/TestBasicCanaryDistributionProviderImpl.java
bbarkley: Thanks, left some comments on the new test code - I'm not sure I follow/agree with the skew calculation.

d2/src/main/java/com/linkedin/d2/balancer/simple/ClusterLoadBalancerSubscriber.java
bohhyang: there is a TestClusterLoadBalancerSubscriber (also one for the service load balancer subscriber). If this breaks, that test will fail.

d2/src/test/java/com/linkedin/d2/balancer/util/canary/TestBasicCanaryDistributionProviderImpl.java
bohhyang: The percentage calculated here is not how skewed a single bucket is, but how skewed the overall attempts are. If we want to distribute 1000 upstream client instances into 100 buckets, what we care about most is how many instances are mis-distributed (for example, expecting 400 instances fall into canary, but actually got 420, then the skewed instances are 20 / 400 = 0.05 = 5%, which means 5% of the clients will be mis-distributed).

So for this test, it first calculates how many attempts are mis-distributed (for example, expecting 10 attempts in each bucket, actually one bucket got 13, another got 7, then the mis-distributed attempts are 3, not 6), then divides it by the total attempts (1000) to get the percentage of mis-distributed attempts.

d2/src/test/java/com/linkedin/d2/balancer/util/canary/TestBasicCanaryDistributionProviderImpl.java
bohhyang: fixed. thanks.

d2/src/test/java/com/linkedin/d2/balancer/util/canary/TestBasicCanaryDistributionProviderImpl.java
bohhyang: modified the test to check the exact mis-distributed canary instances and added more comments to clarify what's tested. Moved the test to container since it uses linkedin-style host names for the test.

d2/src/main/java/com/linkedin/d2/balancer/properties/ClusterStoreProperties.java
bohhyang: makes sense. fixed. thanks.

bbarkley (approved):
Thanks for your patience with all of the change requests and the refactoring. Looks great!

rickzx (approved):

PR #756, opened 2022-02-09
Revert the avro upgrade introduced in 29.27.0

rickzx:


gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=29.28.0
+version=29.28.1
```
aman1309, line 1: so you will be bumping the minor version later? otherwise, new version won't be resolved transitively.

rickzx: Fixed, bumping minor version

aman1309 (approved):

junchuanwang:
please add description

PR #757, opened 2022-02-10
Make DarkGateKeeper configurable given a dark cluster name

rickzx:
Previously, DarkGateKeeper can only apply to all dark clusters. Exposing an API to make it configurable for each dark cluster

### DIFF
--- a/darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterManagerImpl.java
+++ b/darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterManagerImpl.java
@@ -20,16 +20,13 @@ import com.linkedin.darkcluster.api.DarkGateKeeper;
 import com.linkedin.darkcluster.api.DarkRequestHeaderGenerator;
 import com.linkedin.r2.message.rest.RestRequestBuilder;
 import java.net.URI;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.regex.Pattern;
 
-import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
 
 import com.linkedin.common.util.Notifier;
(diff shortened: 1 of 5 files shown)


darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterManagerImpl.java
```diff
@@ -127,10 +127,13 @@ public boolean handleDarkRequest(RestRequest originalRequest, RequestContext ori
         DarkClusterConfigMap configMap = _facilities.getClusterInfoProvider().getDarkClusterConfigMap(_sourceClusterName);
         for (String darkClusterName : configMap.keySet())
         {
-          RestRequest newD2Request = rewriteRequest(reqCopy, darkClusterName);
-          // now find the strategy appropriate for each dark cluster
-          DarkClusterStrategy strategy = _darkClusterStrategyFactory.get(darkClusterName);
-          darkRequestSent = strategy.handleRequest(reqCopy, newD2Request, newRequestContext);
+          if (_darkGateKeeper.shouldDispatchToDark(originalRequest, originalRequestContext, darkClusterName))
```
JieyingXu, line 130: Do we still need the (_darkGateKeeper.shouldDispatchToDark(originalRequest, originalRequestContext) outside of the for loop. Since we need to check shouldDispatchToDark with the dark cluster name anyway.
Also should we add unit tests for this new logic?

rickzx: I leave it there for maintaining backward compatibility. If the service owner is on a older version of pegasus, they don't need to change their DarkGateKeeperImpl to migrate to this new interface.

I'll add a unit test later. Thanks for reviewing

rickzx: You're right, we don't need the check outside the for loop. I can actually deprecate the old API, as the use case of it can be fully covered by the new API.

Also added unit tests

darkcluster/src/test/java/com/linkedin/darkcluster/TestDarkClusterManager.java
```diff
@@ -16,8 +16,7 @@
 import com.linkedin.r2.message.rest.RestRequestBuilder;
 
 import static com.linkedin.darkcluster.DarkClusterTestUtil.createRelativeTrafficMultiplierConfig;
-import static com.linkedin.darkcluster.TestDarkClusterStrategyFactory.DARK_CLUSTER_NAME;
-import static com.linkedin.darkcluster.TestDarkClusterStrategyFactory.SOURCE_CLUSTER_NAME;
+import static com.linkedin.darkcluster.TestDarkClusterStrategyFactory.*;
```
bohhyang, line 19: nit: I think we want to avoid importing with *, right?

bohhyang (approved):
lgtm. thx.

darkcluster/src/test/java/com/linkedin/darkcluster/TestDarkClusterManager.java
rickzx: Fixed. Thanks!

PR #764, opened 2022-03-01
Support null in extract properties

li-ukumar:
Stopped raising error if null value was found in Datamap which started after Avro schema parser started allowing null values (commit : "AVRO-1157. Java: Extend schema and protocol property support from string-only to full JSON.")


nickibi (commented):
I think you need to rebase master and put your commit on top of master. This PR shows diffs from previous commits, it's hard to review your change.

PR #770, opened 2022-03-16
Fix a bug during documentation rendering where it does not account for api-name in the path when searching for restepec file

junchuanwang:
Our documentationRender would respond to Options calls by checking the restspec json files from class path using a direct resource name as the file name. 

But some cusomters added a prefix as api-name to pegasus plugin's idl options when generating the idl files, therefore the idl file in class path will have that prefix and documentationRender will not find it for option calls.

This change, is to fix this case by searching for all restspec.json files in the classpath, so that the restspec.json which prefixed by api-names are also taken into account.

Tested by local optionss call, e.g.
`curl "localhost:9000/<service-name>/<resource-name>" -X OPTIONS`

### DIFF
--- a/restli-server/src/main/java/com/linkedin/restli/internal/server/model/ResourceModelEncoder.java
+++ b/restli-server/src/main/java/com/linkedin/restli/internal/server/model/ResourceModelEncoder.java
@@ -76,17 +76,24 @@ import com.linkedin.restli.server.ResourceLevel;
 import com.linkedin.restli.server.annotations.BatchFinder;
 import com.linkedin.restli.server.errors.ServiceError;
 import com.linkedin.restli.server.errors.ParametersServiceError;
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.reflect.Method;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.net.URLClassLoader;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.jar.JarEntry;
+import java.util.jar.JarInputStream;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import org.apache.commons.io.IOUtils;
@@ -108,6 +115,10 @@ public class ResourceModelEncoder
   private static final Pattern UNNECESSARY_WHITESPACE_PATTERN = Pattern.compile("[ \\t]+");
 
   private final DataCodec codec = new JacksonDataCodec();
+
+  // Used to cache the mapping between restspec file and
+  // the class loader that can get the resource
+  private Map<String, ClassLoader> _restSpecToClassLoaderMap = null;
 
   /**
    * Provides documentation strings from a JVM language to be incorporated into ResourceModels.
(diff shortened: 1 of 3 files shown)


restli-server/src/main/java/com/linkedin/restli/internal/server/model/ResourceModelEncoder.java
```diff
@@ -314,6 +338,53 @@ public ResourceSchema loadOrBuildResourceSchema(final ResourceModel resourceMode
     }
   }
 
+  private InputStream getResourceStreamBySearchingRestSpec(String resourceName, ClassLoader... classLoaders)
+  {
+    final String REST_SPEC_JSON_SUFFIX = "restspec.json";
```
zhihaodev, line 343: Move to the constant of the class?

restli-server/src/main/java/com/linkedin/restli/internal/server/model/ResourceModelEncoder.java
```diff
@@ -314,6 +338,53 @@ public ResourceSchema loadOrBuildResourceSchema(final ResourceModel resourceMode
...
+    if (!resourceName.endsWith(REST_SPEC_JSON_SUFFIX))
+    {
+      return null;
+    }
+    if (_restSpecToClassLoaderMap == null)
+    {
+      _restSpecToClassLoaderMap = new HashMap<>();
+      for(ClassLoader classLoader: classLoaders)
+      {
+        URL[] urls = ((URLClassLoader) classLoader).getURLs();
```
zhihaodev, line 353: Are all the classLoaders always `URLClassLoader` in this case?

restli-server/src/main/java/com/linkedin/restli/internal/server/model/ResourceModelEncoder.java
```diff
@@ -109,6 +116,10 @@
 
   private final DataCodec codec = new JacksonDataCodec();
 
+  // Used to cache the mapping between restspec file and
+  // the class loader that can get the resource
+  private Map<String, ClassLoader> _restSpecToClassLoaderMap = null;
```
zhihaodev, line 121: Would there be any concurrency concerns here? I'm not sure whether the method `loadOrBuildResourceSchema` can be run concurrently or not.

restli-server/src/main/java/com/linkedin/restli/internal/server/model/ResourceModelEncoder.java
junchuanwang: Yes. In fact it should be either AppClassLoader or ExtClassLoader which all extends URLClassLoader

restli-server/src/main/java/com/linkedin/restli/internal/server/model/ResourceModelEncoder.java
junchuanwang: There will be no concurreny issue as far as I am aware of. Also I did the testing by single steps and I found the execution is through the loop

restli-server/src/main/java/com/linkedin/restli/internal/server/model/ResourceModelEncoder.java
junchuanwang: moved

zhihaodev (approved):

nickibi:
I am trying to understanding this change. It used to use resource java classes to generate option call data which shows the resource specs. Now we rely on both java classes and respect.json to generate the spec response？

junchuanwang:
> I am trying to understanding this change. It used to use resource java classes to generate option call data which shows the resource specs. Now we rely on both java classes and respect.json to generate the spec response？

we use restspec.json for options calls. And if not found, we will fall back to use java classes. But using java classes is not desired (see customer complains in detail)

So we need to make sure restspec.json are found in this corner case.

nickibi:
> 

Thanks for the explanation.

PR #771, opened 2022-03-17
Failout delegator client

jpstewart:
Add the client for failout. This largely defers its functionality to the service architecture-specific implementation of the various failout interfaces, so the core complexity is mostly outside of pegasus.


d2/src/main/java/com/linkedin/d2/balancer/clients/FailoutClient.java
```diff
@@ -0,0 +1,124 @@
...
+
+  private final FailoutRedirectStrategy _rewriteStrategy;
+  private final LoadBalancerWithFacilities _balancer;
+
+  public FailoutClient(D2Client d2Client, LoadBalancerWithFacilities balancer,
+      FailoutRedirectStrategy rewriteStrategy)
+  {
+    super(d2Client);
+    _balancer = balancer;
+    _rewriteStrategy = rewriteStrategy;
```
shangchengying, line 49: nit: rewrite -> redirect to make it consistent

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
```diff
@@ -181,7 +183,8 @@ public D2Client build()
                   _config.deterministicSubsettingMetadataProvider,
                   _config.canaryDistributionProvider,
                   _config.enableClusterFailout,
-                  _config.failoutConfigProviderFactory);
+                  _config.failoutConfigProviderFactory,
+                  _config._failoutRedirectStrategy);
```
shangchengying, line 187: nit: remove the leading `_`

d2/src/main/java/com/linkedin/d2/balancer/clients/FailoutRedirectStrategy.java
```diff
@@ -0,0 +1,32 @@
...
+ * A provider for rewriting of D2 URIs when necessary to reroute away from/around failed out clusters.
+ */
+public interface FailoutRedirectStrategy {
+  /**
+   * Rewrites a D2 URI to avoid a failed out cluster.
+   *
+   * @param peerClusters a set of peer clusters which are candidates for redirection targets.
+   * @param uri the D2 URI to rewrite.
+   *
+   * @return A new URI to another cluster.
```
shangchengying, line 29: it could still be local right?

d2/src/main/java/com/linkedin/d2/balancer/clients/FailoutClient.java
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  @Override
+  public void streamRequest(StreamRequest request, final RequestContext requestContext,
+      final Callback<StreamResponse> callback)
+  {
+    final FailoutConfig config = getClusterFailoutConfig(request);
+    if (config != null && config.isFailedOut()) {
+      request = request.builder().setURI(
+          _rewriteStrategy.redirect(config.getPeerClusters(), request.getURI())).build(
```
shangchengying, line 98: i think we need the entire config here? We will match the failed out buckets against member's assigned bucket and then use redirect configs to figure out the peer cluster.

jpstewart: I've made this change as it is a more generic interface and allows implementations to handle more cases. Is your expectation that we will do the bucket matching in the redirect, and if we don't see a the service's assigned bucket is in the failed out range the URI returned is the same as the original? In my current implementation, I am doing that logic in the `isFailedOut()` call, so the redirect call is only made if we have already determined a redirect is needed.

shangchengying: I see. I'm thinking of `isFailedOut` as a short-circuiting check to see if the cluster has any active fail out vs. individual member. One example scenario where it would return false is when all the buckets are online but config has not been removed yet.

shangchengying (approved):

PR #773, opened 2022-03-23
Adds ByteString[] as Query param type

dg-builder:
According to [Restli documentation](https://linkedin.github.io/rest.li/user_guide/restli_server), we support Arrays of `ByteString`s as a query parameter type. This PR adds support for `ByteString[]`.

Screen Shot 2022-03-23 at 1 22 43 PM

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-version=29.32.5
+version=29.32.6
 group=com.linkedin.pegasus
 org.gradle.configureondemand=true
 org.gradle.parallel=true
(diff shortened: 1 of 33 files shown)


nickibi:
> Hi @nickibi it looks like the build is failing when running `:restli-server-testutils:test` but it doesn't seem to fail locally.
> 
> https://github.com/linkedin/rest.li/runs/5721565282?check_suite_focus=true

Looks like a flakey R2/D2 test, you may want to rebuild and check.

gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=29.32.5
+version=29.32.6
```
nickibi, line 1: This PR is introducing a new behavior - support ByteStringArray as finder param, I think it should be a minor version bump.

nickibi (approved):

PR #779, opened 2022-04-11
Add connection warmup handler

shangchengying:
Add a handler for handling warm up connections to peer clusters after watchers are established.

Since establishing connections can be expensive and all the clients need to connect to hosts in the downstream's peer clusters, by pre-establishing the connections, we can avoid the thundering herd effect of all clients trying to connect at the same time.

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/clusterfailout/FailedoutClusterConnectionWarmUpHandler.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/clusterfailout/FailedoutClusterConnectionWarmUpHandler.java
@@ -1,0 +1,20 @@
+package com.linkedin.d2.balancer.clusterfailout;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/**
+ * A handler for handling connection warm up to peer clusters.
+ */
+public interface FailedoutClusterConnectionWarmUpHandler
+{
+  /**
+   * Warms up connections to the given cluster with provided failout config.
+   */
+  void warmUpConnections(@Nonnull String clusterName, @Nullable FailoutConfig config);
+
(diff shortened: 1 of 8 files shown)


d2/src/main/java/com/linkedin/d2/balancer/clusterfailout/FailedoutClusterConnectionWarmUpHandler.java
```diff
@@ -0,0 +1,20 @@
+package com.linkedin.d2.balancer.clusterfailout;
```
jpstewart, line 1: License header

jpstewart (approved):

d2/src/main/java/com/linkedin/d2/balancer/clusterfailout/ZKFailoutConfigProvider.java
```diff
@@ -100,4 +101,10 @@ public void onClusterRemoved(String clusterName)
       manager.updateFailoutConfig(null);
     }
   }
+
+  @Nullable
+  public FailedoutClusterConnectionWarmUpHandler createConnectionWarmUpHandler()
+  {
+    return null;
```
bbarkley, line 108: Might want to indicate this is meant to be overridden by subclasses

bbarkley (approved):
Suggestion for clearing up some code that isn't obvious, but otherwise looks good.

d2/src/main/java/com/linkedin/d2/balancer/clusterfailout/ZKFailoutConfigProvider.java
shangchengying: Forgot to add docstring. Fixed

PR #780, opened 2022-04-13
Shade Netty into r2.

jpstewart:
Shades Netty into a restli-netty package, with relocates the classes under com.linkedin.pegasus, to avoid conflicts with end users' own Netty version. Based on previous PR, with some modifications and fixes for some outstanding issues inthe shadow plugin and IntelliJ handling for shadowing. (https://github.com/linkedin/rest.li/pull/595).


build_script/publications.gradle
```diff
@@ -61,13 +61,19 @@ def manipulatePomXml = {
 publishing {
   publications {
     // Define the "release" publication
-    release(MavenPublication) {
-      from components.java
+    release(MavenPublication) { publication ->
+      if (project.name == 'restli-netty') {
+        apply plugin: 'com.github.johnrengelman.shadow'
```
evanw555, line 66: Why are we applying a plugin in the Maven publication closure? Seems pretty weird... Also, isn't this plugin already applied in this project's `build.gradle` file?

build_script/publications.gradle
```diff
@@ -61,13 +61,19 @@ def manipulatePomXml = {
...
+      } else {
+        from components.java
+      }
 
       afterEvaluate {
         // Add all extra archives (sources, javadoc, any custom archives e.g. all)
         project.configurations.archives.allArtifacts.forEach {
-          if (it.classifier) {
+          // Skip 'slim' identifier, otherwise this logic attemps to publish the disabled slim jar for restli-netty.
+          if (it.classifier && it.classifier != 'slim') {
```
evanw555, line 76: What exact JARs will this new module publish and with what exact classifiers? This is not clear looking at the code.

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [30.0.0] - 2022-03-24
```
evanw555, line 17: Delete these changes and redo with minor version bump.

build_script/publications.gradle
jpstewart: The restli-netty jar is published with classifier shadow (the standard classifier for the shaded jar with this plugin). The "slim" seen here is the default jar task in restli-netty.

The details on why it is here and the classifier for the default jar task was set to "slim" can be found in build.gradle for restli-netty. There is a longstanding bug in IntelliJ when using the shadow plugin; IntelliJ will still think it should use the classpaths as it would determine from analyzing the default jar task. This means after shading Netty, while the builds all work, IntelliJ will give everyone spurious errors in their code that uses shaded Netty classes. The workaround is to assign a classifier to the default jar task. This issue is being tracked as IDEA-163411 on their Youtrack.

CHANGELOG.md
jpstewart: Done

build_script/publications.gradle
jpstewart: This does seem to be required here for running the `local-release` script or the attempt to access the `shadow` property fails. If only for a normal release/publish case, I did not see an issue with omitting the plugin apply here.

r2-netty/src/test/java/com/linkedin/r2/transport/http/client/TestHttpNettyStreamClient.java
```diff
@@ -1111,6 +1111,7 @@ public void onError(Throwable e)
                 latch.countDown();
               }
             }));
+        Thread.sleep(5);
```
FreeMem17, line 1114: What is this for?

r2-perf-test/build.gradle
```diff
@@ -6,6 +6,7 @@ dependencies {
   compile project (':r2-sample')
   compile project (':pegasus-common')
   compile project (':test-util')
+  implementation project(path: ':restli-netty', configuration: 'shadow')
```
FreeMem17, line 9: Why does `r2-perf-test` add a dependency here? Did it depend on Netty before?

r2-netty/src/test/java/com/linkedin/r2/transport/http/client/TestHttpNettyStreamClient.java
jpstewart: This addresses a non-deterministic failure in the test, which happens even without the netty shading. It seems the Netty server gets overloaded by the requests, and we are relying on the retry to get past the failures, but sometimes the retry is not enough.

FreeMem17: Thanks for sharing the context. How about adding the information to the comment?

build_script/publications.gradle
evanw555: Thanks for the explanation. Could you leave a TODO in the code here so that in the future, the context isn't totally lost on us?

jpstewart: Done.

r2-perf-test/build.gradle
jpstewart: It depends on Netty. Previously this came through r2-netty, but with the shadow plugin this needs to be explicitly placed here now.

restli-netty/build.gradle
```diff
@@ -0,0 +1,33 @@
...
+  dependencies {
+    exclude dependency(externalDependency.slf4jApi)
+  }
+
+  exclude 'META-INF/io.netty.versions.properties'
+  exclude 'META-INF/maven/'
+  exclude 'META-INF/native-image/'
+
+  relocate 'io.netty', 'com.linkedin.pegasus.io.netty'
+  relocate 'META-INF/native/libnetty', 'META-INF/native/libcom_linkedin_pegasus_netty'
```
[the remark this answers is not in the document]
jpstewart: It doesn't seem to be, from looking at other projects that have done this. However I will cover one additional case I see used for GRPC.

r2-netty/src/test/java/com/linkedin/r2/transport/http/client/TestHttpNettyStreamClient.java
jpstewart: Done

FreeMem17 (approved):
Looks good to me.

li-bsoetarman (approved):
What about the sources jar and the javadoc jar for the shadow?  The debugger would not the sources jar.

ChrisCarini:

PR #798, opened 2022-05-31
Fix passing in canary distribution provider from ZKFS load balancer factory

bohhyang:
Previously, canary distribution provider was passed in only from last-seen load balancer factory, which causes the canary distribution provider being null for ZKFS toggling load balancer. This PR is to fix that by passing in the provider from ZKFS load balancer factory as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.34.1] - 2022-05-28
+- fix passing in canary distribution provider from ZKFS load balancer factory.
+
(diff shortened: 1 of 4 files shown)


rickzx (approved):

gu0keno0:
LGTM, let's consider adding some test coverage in the future? The UT coverage for these classes looks very sparse.

bohhyang:
That's true. Lots of the pegasus classes have very few UT. I will create a work item to improve test coverage for our planning tomorrow. Thanks.

> LGTM, let's consider adding some test coverage in the future? The UT coverage for these classes looks very sparse.

gu0keno0:
Great to hear! Thank you.

PR #799, opened 2022-06-03
Provide a way to link a finder with a functionally equivalent batch finder declaratively

karthikrg:
This provides a way for client applications to batch multiple finder requests together into a single batch finder request to optimize downstream fanout.

### DIFF
--- a/restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
+++ b/restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
@@ -1969,8 +1969,8 @@ public final class RestLiAnnotationReader
       }
 
       addActionResourceMethod(model, method);
-      addFinderResourceMethod(model, method);
       addBatchFinderResourceMethod(model, method);
+      addFinderResourceMethod(model, method);
       addTemplateResourceMethod(resourceClass, model, method);
       addCrudResourceMethod(resourceClass, model, method);
     }
@@ -2111,6 +2111,8 @@ public final class RestLiAnnotationReader
             method.getName()));
       }
 
+      String linkedBatchFinderName =
+          RestAnnotations.DEFAULT.equals(finderAnno.linkedBatchFinderName()) ? null : finderAnno.linkedBatchFinderName();
       List<Parameter<?>> queryParameters = getParameters(model, method, ResourceMethod.FINDER);
 
       Class<? extends RecordTemplate> metadataType = getCustomCollectionMetadata(method, DEFAULT_METADATA_PARAMETER_INDEX);
@@ -2123,9 +2125,10 @@ public final class RestLiAnnotationReader
                                                                                                  queryType,
                                                                                                  metadataType,
                                                                                                  getInterfaceType(method),
-                                                                                                 annotationsMap);
+                                                                                                 annotationsMap,
+                                                                                                 linkedBatchFinderName);
 
-      validateFinderMethod(finderMethodDescriptor, model);
+      validateFinderMethod(finderMethodDescriptor, model, linkedBatchFinderName);
       addServiceErrors(finderMethodDescriptor, method);
       addSuccessStatuses(finderMethodDescriptor, method);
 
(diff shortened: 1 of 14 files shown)


restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
```diff
@@ -2694,9 +2697,142 @@ else if (CollectionResult.class.isAssignableFrom(returnType))
...
+          if (countField == null || countField.getType().getType() != DataSchema.Type.INT)
+          {
+            throw new ResourceConfigException("Linked @BatchFinder method '" + linkedBatchFinderName
+                + "' from @Finder Method named '" + finder.getFinderName()
+                + "' on class '" + resourceModel.getResourceClass().getName()
+                + "' has a @PagingContext or @Context param but no int '" + RestConstants.COUNT_PARAM
+                + "' field in the criteria object");
+          }
+
+          if (countField.getDefault() == null || !countField.getDefault().equals(defaultContext.getStart()))
```
mchen07, line 2809: should be `defaultContext.getCount()`

restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
```diff
@@ -2694,9 +2697,142 @@ else if (CollectionResult.class.isAssignableFrom(returnType))
...
+    Set<String> fieldNames = criteriaSchema.getFields()
+        .stream()
+        .map(RecordDataSchema.Field::getName)
+        .collect(Collectors.toCollection(HashSet::new));
+    finder.getParameters().forEach(parameter ->
+    {
+      switch (parameter.getParamType())
+      {
+        case QUERY:
+        case ASSOC_KEY_PARAM:
```
mchen07, line 2746: Since you include deprecated @Context here, you also should include deprecated @AssocKey.

restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
```diff
@@ -2694,9 +2697,226 @@ else if (CollectionResult.class.isAssignableFrom(returnType))
...
+    Set<String> fieldNames = criteriaSchema.getFields()
+        .stream()
+        .map(RecordDataSchema.Field::getName)
+        .collect(Collectors.toCollection(HashSet::new));
+    finder.getParameters().forEach(parameter ->
+    {
+      switch (parameter.getParamType())
+      {
+        case QUERY:
+        case KEY:
```
mchen07, line 2795: I think that should be ASSOC_KEY, not KEY

restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
```diff
@@ -2694,9 +2697,226 @@ else if (CollectionResult.class.isAssignableFrom(returnType))
...
+                + "' has a @PagingContext or @Context param but the '" + RestConstants.COUNT_PARAM
+                + "' field default values don't match");
+          }
+
+          fieldNames.remove(RestConstants.START_PARAM);
+          fieldNames.remove(RestConstants.COUNT_PARAM);
+          break;
+        case PROJECTION:
+        case PROJECTION_PARAM:
+          RecordDataSchema.Field fieldsField = criteriaSchema.getField(RestConstants.FIELDS_PARAM);
```
mchen07, line 2873: This is different from my understanding of batchFinder, paging should be specified as a separate parameter in BatchFinder instead of putting inside Criteria schema. See https://github.com/linkedin/rest.li/blob/master/restli-int-test-server/src/main/java/com/linkedin/restli/examples/greetings/server/BatchFinderResource.java#L54

restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
karthikrg: KEY (@AssocKey) is deprecated version of ASSOC_KEY (@AssocKeyParam)

restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
karthikrg: Ok fair point. If we do it this way then we will lose what the framework gives us. Simplified and changed. We will change batching logic accordingly.

mchen07 (approved):
LGTM

PR #800, opened 2022-06-06
Translate Data.null to Avro null if schema field is optional

aman1309:
This PR updates data schema to avro translation to handle Data.Null value in pegasus DataMap and translate it to Avro null


data-avro/src/test/java/com/linkedin/data/avro/TestDataTranslator.java
```diff
@@ -1133,7 +1134,7 @@ private void testDataTranslation(String schemaText, String[][] row) throws IOExc
       {
         // translate from Avro back to Pegasus
         DataMap dataMapResult = DataTranslator.genericRecordToDataMap(avroRecord, recordDataSchema, avroSchema);
-        ValidationResult vr = ValidateDataAgainstSchema.validate(dataMap,
+        ValidationResult vr = ValidateDataAgainstSchema.validate(dataMapResult,
```
evanw555, line 1137: Why did this need to be changed? Was this a bug in the original testing logic?

aman1309: yes looks like it as we would want to validate the datamap converted from avro not the datamap from test input.

evanw555 (approved):
LGTM

PR #802, opened 2022-06-08
Add Enum symbols order change as compatible message in checker

aman1309:
This is added to fail equivalent compatibility check and also avoid skipping publish snapshot/idl task which are only executed if snapshot/idl are not equivalent.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-version=29.35.0
+version=29.36.0
 group=com.linkedin.pegasus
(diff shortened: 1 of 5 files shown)


gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=29.34.3
+version=29.34.4
```
nickibi, line 1: Since we introduced a new enum symbol in a public enum, let's bump minor version here.

PR #803, opened 2022-06-15
Avoid using JsonFactoryBuilder to be more compatible with pre 2.10 jackson at runtime

cgtz:
Some users have runtimes that pin jackson-core to older pre 2.10
versions. This introduces compatibility issues with the newer
JsonFactoryBuilder APIs. For the remaining jackson-core 2.x releases we
can continue to use the deprecated JsonFactory.enable/disable methods.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-version=29.34.3
+version=29.34.4
 group=com.linkedin.pegasus
 org.gradle.configureondemand=true
 org.gradle.parallel=true
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-version=29.34.4
+version=29.35.0
 group=com.linkedin.pegasus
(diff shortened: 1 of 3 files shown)


junchuanwang (approved):

nickibi:
Please bump minor version in gradle.properties and update changelog.

gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=29.34.3
+version=29.34.4
```
nickibi, line 1: This change removed a public builder, please bump minor version.

cgtz: Thanks for catching this. I have bumped the minor version. FWIW, I didn't see anyone using the public builder in codesearch.

evanw555: @nickibi Just to be clear: removing a public interface is a major version change, not a minor version change (however, it seems like we're electing to take the risk here since no one is apparently using this)

PR #804, opened 2022-06-17
Fix FailoutClient invoking delegated client's restRequest twice

shangchengying:
There was an extra call to `super.restRequest(request, requestContext, callback);` in FailoutClient.
Removing the extra call and adding unit tests

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/clients/FailoutClient.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/clients/FailoutClient.java
@@ -90,8 +90,6 @@ public class FailoutClient extends D2ClientDelegator
         FailoutClient.super.restRequest(redirectedRequest, requestContext, callback);
       }
     });
-
-    super.restRequest(request, requestContext, callback);
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


jpstewart (approved):

d2/src/main/java/com/linkedin/d2/balancer/clients/FailoutClient.java
```diff
@@ -90,8 +90,6 @@ public void onSuccess(URI result)
         FailoutClient.super.restRequest(redirectedRequest, requestContext, callback);
```
rickzx, line 90: nit: Can we just do `super.restRequest`. Same for line 83

rickzx (approved):

d2/src/main/java/com/linkedin/d2/balancer/clients/FailoutClient.java
shangchengying: I think we cannot resolve `super` correctly in the callback.
```
Cannot resolve method 'restRequest' in 'Object'
```

PR #805, opened 2022-06-23
Translate restli-common PDL files to PDSC files to support legacy web use cases

aman1309:
This is required since PDSCs are still used in web products and Restli common models jar only packed PDL files. 
This PR manually adds translated PDSC files to the restli-common module as the auto-translation task can't be added due to the circular dependency of restli-tools on restli-common.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and what APIs have changed, if applicable.
 ## [Unreleased]
 
 ## [29.37.0] - 2022-06-23
-- Translate `:restli-common` PDL files to PDSC files to support legacy web use cases.
+- Package translated legacy PDSC models into `:restli-common` JAR
 
 ## [29.36.1] - 2022-06-22
 - Fix FailoutClient delegated client's restRequest invocation 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.37.0] - 2022-06-23
+- Translate `:restli-common` PDL files to PDSC files to support legacy web use cases.
+
 ## [29.36.1] - 2022-06-22
 - Fix FailoutClient delegated client's restRequest invocation 
 
@@ -5261,7 +5264,8 @@ patch operations can re-use these classes for generating patch messages.
 
 ## [0.14.1]
 
-[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.36.1...master
+[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.37.0...master
+[29.37.0]: https://github.com/linkedin/rest.li/compare/v29.36.1...v29.37.0
 [29.36.1]: https://github.com/linkedin/rest.li/compare/v29.36.0...v29.36.1
 [29.36.0]: https://github.com/linkedin/rest.li/compare/v29.35.0...v29.36.0
 [29.35.0]: https://github.com/linkedin/rest.li/compare/v29.34.3...v29.35.0
--- a/restli-common/src/main/resources/legacyPegasusSchemas/com/linkedin/restli/common/CollectionMetadata.pdsc
+++ b/restli-common/src/main/resources/legacyPegasusSchemas/com/linkedin/restli/common/CollectionMetadata.pdsc
@@ -1,0 +1,26 @@
+{
+  "type" : "record",
+  "name" : "CollectionMetadata",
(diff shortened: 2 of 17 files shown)


restli-common/src/test/java/com/linkedin/restli/common/TestLegacyPdscEquivalence.java
```diff
@@ -0,0 +1,74 @@
...
+public class TestLegacyPdscEquivalence
+{
+  private final String pdlSrcDir = System.getProperty("srcDir", "src/main") + "/pegasus";
+  private final File legacyPdscDir =
+      new File(System.getProperty("resourcesDir", "src/main/resources") + "/legacyPegasusSchemas");
+
+  @Test
+  public void testPdscEquivalence() throws IOException
+  {
+    final Map<String, File> pdscFiles = FileUtil.listFiles(legacyPdscDir,
```
evanw555, line 50: Add an assertion guaranteeing that this isn't empty, since we want to confirm that the test is actually running

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.37.0] - 2022-06-23
+- Translate `:restli-common` PDL files to PDSC files to support legacy web use cases.
```
evanw555, line 18: nit: remove LI-specific comment, just say "to be consistent with legacy PDSC JAR publishing behavior"

evanw555, line 18: This is wrong, it's not translating anything. I'd say "Package translate legacy PDSC models into :restli-common JAR"

Also remove LI-specific comment (i.e. "web")

restli-common/src/main/resources/legacyPegasusSchemas/com/linkedin/restli/common/CollectionMetadata.pdsc
```diff
@@ -0,0 +1,26 @@
+{
```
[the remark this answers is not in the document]
aman1309: yes, I saw the jar generated from build.

nickibi:
I don't think those PDSC files will be wrapped into the Jar. https://github.com/linkedin/rest.li/blob/master/build_script/dataTemplate.gradle#L23

restli-common/src/main/resources/legacyPegasusSchemas/com/linkedin/restli/common/CollectionMetadata.pdsc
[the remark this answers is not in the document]
aman1309: my bad I didn't check the dataTemplate jar. will update the dataTemplate script

aman1309:
> I don't think those PDSC files will be wrapped into the Jar. https://github.com/linkedin/rest.li/blob/master/build_script/dataTemplate.gradle#L23

resources dir is auto included in jar, if we want legacy schemas to be outside resources like pdl in pegasus dir we could update this script

nickibi:
> resources dir is auto included in jar, if we want legacy schemas to be outside resources like pdl in pegasus dir we could update this script
You may need to check how the web plugin is fetching the PDSC files, whether it looks for a specific directory? If you could generate a snapshot and verify the whole flow that will be good.
IIRC, in ligradle-pegasus we specifically check whether a MP contains PDSC files or not. Please make sure this change will not make other MPs build failing. 
Make sure run container wc tests before merging this change.

evanw555 (approved):

evanw555:
> > I don't think those PDSC files will be wrapped into the Jar. https://github.com/linkedin/rest.li/blob/master/build_script/dataTemplate.gradle#L23

> resources dir is auto included in jar, if we want legacy schemas to be outside resources like pdl in pegasus dir we could update this script

@nickibi @aman1309 Note that the native Gradle JAR task auto-packages everything under `main/resources/` at the root of the JAR. So if the legacy schemas are at `main/resources/legacyPegasusSchemas/com/linkedin/common/Foo.pdsc` then it'll be packaged in the JAR at `legacyPegasusSchemas/com/linkedin/common/Foo.pdsc` automatically. No need for any special configuration here.

nickibi:
> nickibi @aman1309 Note that the native Gradle JAR task auto-packages everything under `main/resources/` at the root of the JAR. So if the legacy schemas are at `main/resources/legacyPegasusSchemas/com/linkedin/common/Foo.pdsc` then it'll be packaged in the JAR at `legacyPegasusSchemas/com/linkedin/common/Foo.pdsc` automatically. No need for any special configuration here.

Yes, we discussed offline and verified it.

PR #806, opened 2022-06-29
Handle method order when validating methods to ensure consistent linked batch finder validation

karthikrg:
Finders can have linked batch finders associated with them. To ensure that the validation of these linked batch finders happens correctly, we need to process all the batch finder methods before the finder methods when processing annotations.

### DIFF
--- a/restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
+++ b/restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
@@ -126,6 +126,7 @@ import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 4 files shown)


restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
```diff
@@ -1978,6 +1980,21 @@ private static void addResourceMethods(final Class<?> resourceClass,
     validateResourceModel(model);
   }
 
+  static int getMethodIndex(Method method)
+  {
+    if (method.getAnnotation(BatchFinder.class) != null)
+    {
+      return 1;
```
zsmoore, line 1987: Any reason we do 1 instead of 0? I haven't written a comparator in forever

zsmoore (approved):

restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
```diff
@@ -1956,18 +1957,19 @@ private static boolean checkParameterType(final Class<?> type,
...
-      if (method.isSynthetic())
-      {
-        continue;
-      }
+    // this ignores methods declared in superclasses (e.g. template methods) and synthetic type-erased methods.
+    // We sort methods such that batch finders are always processed before finders. This ensures that any validation
+    // of linked batch finder methods from finders works as expected. We don't care about the order in which other
+    // methods are processed.
+    List<Method> methods = Arrays.stream(resourceClass.getDeclaredMethods())
+        .filter(method -> !method.isSynthetic())
```
mchen07, line 1967: Instead of sorting this, maybe you can do this cross-method validation in `validateResourceModel` method.

restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
karthikrg: Doesn't matter, just need them in increasing order. I just liked 1-2-3 better.

restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
karthikrg: I thought about that. It makes the code uglier.

restli-server/src/main/java/com/linkedin/restli/internal/server/model/RestLiAnnotationReader.java
```diff
@@ -1978,6 +1980,21 @@ private static void addResourceMethods(final Class<?> resourceClass,
     validateResourceModel(model);
   }
 
+  static int getMethodIndex(Method method)
```
evanw555, line 1983: nit: Maybe your comment about the motivation for sorting batchFinders before finders should be left as a javadoc here so this method itself is clear. other than this LGTM

mchen07 (approved):

PR #807, opened 2022-07-18
Validate tunneling HTTP method override content.

jpstewart:
This will filter out requests which use a non-HTTP verb for the content of X-HTTP-Method-Override.

### DIFF
--- a/r2-core/src/main/java/com/linkedin/r2/message/QueryTunnelUtil.java
+++ b/r2-core/src/main/java/com/linkedin/r2/message/QueryTunnelUtil.java
@@ -26,6 +26,8 @@ import com.linkedin.r2.message.rest.RestRequestBuilder;
 import com.linkedin.r2.message.stream.StreamRequest;
 import com.linkedin.r2.util.IOUtil;
 
+import java.util.HashSet;
+import java.util.Set;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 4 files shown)


r2-core/src/main/java/com/linkedin/r2/message/QueryTunnelUtil.java
```diff
@@ -408,6 +421,12 @@ else if(existingQuery.isEmpty())
     }
     requestBuilder.setEntity(entity);
     requestBuilder.setHeaders(h);
+
+    String overrideHeader = request.getHeader(HEADER_METHOD_OVERRIDE);
+    if (!VALID_HTTP_VERBS.contains(overrideHeader)) {
+      LOG.warn("Invalid HTTP method override header " + overrideHeader + " was received, rejecting request.");
```
li-bsoetarman, line 427: Since overrideHeader can contain anything, let's not log the actual string here, unless there's a way to sanitize it.

r2-core/src/main/java/com/linkedin/r2/message/QueryTunnelUtil.java
```diff
@@ -408,6 +421,12 @@ else if(existingQuery.isEmpty())
     }
     requestBuilder.setEntity(entity);
     requestBuilder.setHeaders(h);
+
+    String overrideHeader = request.getHeader(HEADER_METHOD_OVERRIDE);
```
li-bsoetarman, line 425: This header is first set on line 181; shall we move this check there?  This is to make sure bad method string is not saved into the headers at all.

r2-core/src/main/java/com/linkedin/r2/message/QueryTunnelUtil.java
jpstewart: Done

r2-core/src/main/java/com/linkedin/r2/message/QueryTunnelUtil.java
jpstewart: I don't want to only validate in encode since an alternate client can still send invalid content. I've refactored the validation into its own method so we can validate in both locations.

li-bsoetarman (approved):

PR #809, opened 2022-07-25
Add support for r2 Netty client DNS metrics, and clarify timeout errors for DNS timeouts.

jpstewart:
There can be a misleading situation where DNS resolution is timing out due to host system DNS issues, which present as request timeouts in the error messages. In these situations the request is actually never made, since the DNS resolution is done within the same timeout boundaries as the request. This will clarify the error messages and also makes it possible to use JMX metrics to track DNS latency.

### DIFF
--- a/r2-netty/src/main/java/com/linkedin/r2/netty/client/HttpNettyClient.java
+++ b/r2-netty/src/main/java/com/linkedin/r2/netty/client/HttpNettyClient.java
@@ -18,6 +18,9 @@ package com.linkedin.r2.netty.client;
 
 import com.linkedin.common.callback.Callback;
 import com.linkedin.common.callback.MultiCallback;
+import com.linkedin.common.stats.LongStats;
+import com.linkedin.common.stats.LongTracker;
+import com.linkedin.common.stats.LongTracking;
 import com.linkedin.common.util.None;
 import com.linkedin.r2.filter.R2Constants;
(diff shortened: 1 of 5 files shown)


r2-netty/src/main/java/com/linkedin/r2/netty/client/HttpNettyClient.java
```diff
@@ -294,18 +302,23 @@ private void sendRequest(Request request, RequestContext requestContext, Map<Str
...
+        new TimeoutException("Exceeded request timeout of " + resolvedRequestTimeout + "ms" +
+            (requestContext.getLocalAttr(R2Constants.REMOTE_SERVER_ADDR) == null ? " (timeout during DNS resolution)" : "")))));
 
     // resolve address
     final SocketAddress address;
     try
     {
       TimingContextUtil.markTiming(requestContext, TIMING_KEY);
+      _dnsResolutions.incrementAndGet();
+      long startTime = _clock.currentTimeMillis();
```
gu0keno0, line 314: Consider to use nanoTime() to be monotonic : https://stackoverflow.com/questions/2978598/will-system-currenttimemillis-always-return-a-value-previous-calls

gu0keno0 (approved):
Consider adding some UTs ? I'm unblocking this because the changes look safe and straightforward enough.

Issue #810, opened 2022-07-26
Update ZooKeeper client dependency to latest stable version 3.7.1


abhilash1in:
https://github.com/linkedin/rest.li/blob/6823381dd41a4e965b96442ce12f6e86889fcb80/build.gradle#L105

ZK v3.4 has been EOL'd from 1st June 2020 ([announcement](https://lists.apache.org/thread/xckr6nnsg9rxchkbvltkvt7hr2d0mhbo)) and ZK v3.5 has been EOL'd on 1st June 2022 (see https://zookeeper.apache.org/releases.html). 

ZK v3.4 and v3.5 versions:
- Are not guaranteed to run with new JDK versions
- Will not receive any security updates
- Will not receive any bug fixes

Please upgrade ZK client to the latest stable version 3.7.1.

PR #811, opened 2022-07-26
Bump zookeeper to latest stable version 3.7.1 in d2

abhilash1in:
ZK 3.4 is EOL as of 1st June, 2020 ([announcement](https://lists.apache.org/thread/xckr6nnsg9rxchkbvltkvt7hr2d0mhbo)), so this PR bumps ZK dependency to latest stable version 3.7.1. 

There should be no breaking changes between the two versions (see https://zookeeper.apache.org/releases.html).

Fixes linkedin/rest.li#810


d2/build.gradle
```diff
@@ -19,6 +19,9 @@ dependencies {
   compile externalDependency.jdkTools
   compile externalDependency.zero_allocation_hashing
   compile externalDependency.xchart
+  compile externalDependency.findbugs
```
li-bsoetarman, line 22: compileOnly?  or does it actually need compile

abhilash1in: `compileOnly` was the first thing I tried and is technically the right option to use as we don't need findbugs during runtime, but with `compileOnly`, `./gradlew compileTestJava` still fails with the error:

```
> Task :d2-test-api:compileJava FAILED
/Users/abkishor/.gradle/caches/modules-2/files-2.1/org.apache.zookeeper/zookeeper/3.7.1/b598708e153ff74dd2ebb800419df96df8a01c2d/zookeeper-3.7.1.jar(org/apache/zookeeper/server/ZooKeeperServer.class): warning: Cannot find annotation method 'value()' in type 'SuppressFBWarnings': class file for edu.umd.cs.findbugs.annotations.SuppressFBWarnings not found
/Users/abkishor/.gradle/caches/modules-2/files-2.1/org.apache.zookeeper/zookeeper/3.7.1/b598708e153ff74dd2ebb800419df96df8a01c2d/zookeeper-3.7.1.jar(org/apache/zookeeper/server/ZooKeeperServer.class): warning: Cannot find annotation method 'justification()' in type 'SuppressFBWarnings'
/Users/abkishor/.gradle/caches/modules-2/files-2.1/org.apache.zookeeper/zookeeper/3.7.1/b598708e153ff74dd2ebb800419df96df8a01c2d/zookeeper-3.7.1.jar(org/apache/zookeeper/server/ZooKeeperServer.class): warning: Cannot find annotation method 'value()' in type 'SuppressFBWarnings'
/Users/abkishor/.gradle/caches/modules-2/files-2.1/org.apache.zookeeper/zookeeper/3.7.1/b598708e153ff74dd2ebb800419df96df8a01c2d/zookeeper-3.7.1.jar(org/apache/zookeeper/server/ZooKeeperServer.class): warning: Cannot find annotation method 'justification()' in type 'SuppressFBWarnings'
error: warnings found and -Werror specified
1 error
4 warnings
```

The only way I could get rid of it was by using `compile`.

li-bsoetarman: ah maybe you need both compileOnly and testCompileOnly.

abhilash1in: That works, thanks! 

Updated the PR. Also had to add `compileOnly` / `testCompileOnly` to `:d2-int-test` and `:d2-test-api`.

abhilash1in: Checks are passing as well.

junchuanwang:
except I don't know d2 logic, this PR looks good to me

junchuanwang (approved):

jpstewart (approved):

PR #814, opened 2022-08-02
Ensure skip deprecation config is only applied to the direct input schema

imagiczhang:
Skip deprecation config will only remain in the direct input schema and not passed down recursively. This is to ensure there is no conflict behavior in common shared models between difference schema.

### DIFF
--- a/generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
+++ b/generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
@@ -140,7 +140,6 @@ public class TemplateSpecGenerator
    * Generate {@link ClassTemplateSpec} from the specified {@link DataSchema} and its location.
    *
    * @param skipDeprecatedField flag to indicate whether to skip spec generation for deprecated fields, their types and referenced types.
-   *                            This behavior affect the spec generation recursively.
    */
   public ClassTemplateSpec generate(DataSchema schema, DataSchemaLocation location, boolean skipDeprecatedField)
   {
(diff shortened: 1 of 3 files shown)


generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
```diff
@@ -137,15 +145,14 @@ public ClassTemplateSpec generate(DataSchema schema, DataSchemaLocation location
...
-   *                            This behavior affect the spec generation recursively.
+   * Do not use this deprecated method. If need to skip deprecated field when generating specs, please use setSkipDeprecatedPatterns
+   * and generate(DataSchema schema, DataSchemaLocation location) instead.
    */
+  @Deprecated
   public ClassTemplateSpec generate(DataSchema schema, DataSchemaLocation location, boolean skipDeprecatedField)
   {
     pushCurrentLocation(location);
-    final ClassTemplateSpec result = processSchema(schema, null, null, skipDeprecatedField);
+    final ClassTemplateSpec result = processSchema(schema, null, null);
```
evanw555, line 155: This deprecated method should call the non-deprecated method, not vice versa (core logic should be moved out of deprecated methods marked for deletion).

generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
```diff
@@ -811,6 +829,10 @@ else if (property instanceof String)
     }
   }
 
+  public boolean shouldSkipDeprecatedFields(NamedDataSchema dataSchema) {
```
evanw555, line 832: Internal utility should be private

generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
```diff
@@ -155,6 +162,16 @@ public Collection<ClassTemplateSpec> getGeneratedSpecs()
     return _classTemplateSpecs;
   }
 
+  /**
+   * Set the regex name patterns for schemas which needs to skip deprecated fields when generating the specs
+   *
+   * @param skipDeprecatedPatterns list of regex name patterns to be set. They will be used to match full name of schemas.
+   */
+  public void setSkipDeprecatedPatterns(@Nonnull List<Pattern> skipDeprecatedPatterns) {
```
evanw555, line 170: Rename to be more clear `setSkipDeprecatedFieldPatterns` and `_skipDeprecatedFieldPatterns`.

generator/src/main/java/com/linkedin/pegasus/generator/TemplateSpecGenerator.java
```diff
@@ -155,6 +162,16 @@ public Collection<ClassTemplateSpec> getGeneratedSpecs()
     return _classTemplateSpecs;
   }
 
+  /**
+   * Set the regex name patterns for schemas which needs to skip deprecated fields when generating the specs
```
evanw555, line 166: Include a more functional explanation: "For any record schema with a fully-qualified name that matches any of the provided patterns, all its fields marked as deprecated will be skipped during spec generation."

evanw555 (approved):

PR #815, opened 2022-08-02
Support removing cluster watches after failout is over

shangchengying:
When a failout is over or when a peer cluster is no longer accepting request, we should remove clusters/uris watches for that peer cluster to reduce the load on ZK:
* Cancels any pending connection warm-up requests
* Support cluster watches removal to be scheduled to sometime in the future to allow in-flight requests to finish.
* If there was no existing watches on the peer cluster, it's likely that cluster failout is the only consumer of the peer cluster info. Only remove the watches on the peer cluster if this is the case.
* Added a `stopListenToCluster` method to `LoadBalancerState`
* `SimpleLoadBalancerState` will invoke `ensureNotListening` on `cluster` and `uri` subscribers

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/LoadBalancerState.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/LoadBalancerState.java
@@ -76,12 +76,9 @@ public interface LoadBalancerState
   /**
    * Stops listening to a cluster
    * @param clusterName the cluster to stop listening to.
-   * @return true if listeners successfully stopped.
+   * @param callback callback to be invoked after stopped listening to the cluster.
    */
-  default boolean stopListenToCluster(String clusterName)
-  {
-    return false;
-  }
+  default void stopListenToCluster(String clusterName, LoadBalancerStateListenerCallback callback) {}
 
   void start(Callback<None> callback);
 
--- a/d2/src/main/java/com/linkedin/d2/balancer/LoadBalancerState.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/LoadBalancerState.java
@@ -73,6 +73,16 @@ public interface LoadBalancerState
 
   void listenToCluster(String clusterName, LoadBalancerStateListenerCallback callback);
 
+  /**
+   * Stops listening to a cluster
+   * @param clusterName the cluster to stop listening to.
+   * @return true if listeners successfully stopped.
+   */
+  default boolean stopListenToCluster(String clusterName)
+  {
+    return false;
+  }
+
   void start(Callback<None> callback);
 
   void shutdown(PropertyEventShutdownCallback shutdown);
--- a/d2/src/main/java/com/linkedin/d2/balancer/clusterfailout/FailedoutClusterManager.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/clusterfailout/FailedoutClusterManager.java
@@ -95,7 +95,8 @@ public class FailedoutClusterManager
 
   public void shutdown()
   {
-    if (_connectionWarmUpHandler != null) {
+    if (_connectionWarmUpHandler != null)
+    {
       _connectionWarmUpHandler.shutdown();
     }
   }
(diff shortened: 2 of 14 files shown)


d2/src/main/java/com/linkedin/d2/balancer/LoadBalancerState.java
```diff
@@ -73,6 +73,16 @@
 
   void listenToCluster(String clusterName, LoadBalancerStateListenerCallback callback);
 
+  /**
+   * Stops listening to a cluster
+   * @param clusterName the cluster to stop listening to.
+   * @return true if listeners successfully stopped.
+   */
+  default boolean stopListenToCluster(String clusterName)
```
rickzx, line 81: The signature of `listenToCluster` is taking a callback. I think the reason is that we can supply a `FutureCallback` and provide a timeout value in case this operation is blocking for too long: https://github.com/linkedin/rest.li/blob/ac4ad1f91b91c0f7f974d6af1f4b60b3bbcace2e/d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancer.java#L503.
Shall we do the same here by passing in a callback, and invoke `callback.done` when the subscriber is unregistered?

d2/src/main/java/com/linkedin/d2/balancer/clusterfailout/FailedoutClusterManager.java
```diff
@@ -155,16 +165,23 @@ private void addClusterWatches(@Nonnull Set<String> clustersToWatch, @Nonnull Fa
...
     for (final String cluster : clustersToWatch)
     {
-      _clusterListenerCallbacks.computeIfAbsent(cluster, clusterName -> {
-        LoadBalancerStateListenerCallback listenerCallback = (type, name) -> {
+      _peerWatches.computeIfAbsent(cluster, clusterName ->
+      {
+        boolean watchExistsBeforeFailout = _loadBalancerState.isListeningToCluster(clusterName);
+        PeerWatchState peerWatchState = new PeerWatchState(watchExistsBeforeFailout);
+
+        LoadBalancerStateListenerCallback listenerCallback = (type, name) ->
```
rickzx, line 173: Same here, do we want to use a FutureCallback, and provide a timeout value? We can probably use the same timeout value as `SimpleLoadBalancer._timeout`

d2/src/main/java/com/linkedin/d2/balancer/clusterfailout/FailedoutClusterManager.java
```diff
@@ -177,11 +194,51 @@ private void removeClusterWatches(@Nonnull Set<String> clustersToRemove)
...
+      if (peerWatchState == null)
+      {
+        continue;
+      }
+      if (_connectionWarmUpHandler != null)
+      {
+        _log.debug("Cancel pending requests to: {}", cluster);
+        _connectionWarmUpHandler.cancelPendingRequests(cluster);
+      }
+      if (peerWatchState.shouldUnregisterWatches())
```
rickzx, line 207: Maybe add a comment here: we only unregister the watches if they do not exist prior to fail out.

d2/src/main/java/com/linkedin/d2/balancer/LoadBalancerState.java
shangchengying: In this case, do we expect caller to verify if the listener is stopped successfully? Or should I create another callback that returns the success/failure as part of the `callback.done`?

d2/src/main/java/com/linkedin/d2/balancer/clusterfailout/FailedoutClusterManager.java
shangchengying: Looking at the logic, I think we don't need to block/wait for the call to finish. It will be a best effort clean-up.

d2/src/main/java/com/linkedin/d2/balancer/LoadBalancerState.java
rickzx: I think the current logic looks good. But just that this `ensureNotListening` is not really ensuring (e.g. in case the queue is not closed). Maybe change the name to `stopListening` would better describe what it does.

rickzx (approved):
LGTM. Replied to one comment. Thanks for making the changes!

d2/src/main/java/com/linkedin/d2/balancer/LoadBalancerState.java
shangchengying: Good point. Renamed to `tryStopListening` since it's not guaranteed.

rickzx (approved):
LGTM! Thanks

PR #782, opened 2022-04-15
Implement rest.li changes for adding D2 Canary metrics

gu0keno0:
Implement rest.li changes for adding D2 Canary metrics. This commit adds the following items:

1. Add canary distribution policy into ClusterInfo and SimpleLoadBalancerState to facilitate
   constructing corresponding JMX beans
2. New callbacks in SimpleLoadBalancerState for notifying listeners about ServiceProperties updates, the callbacks can then be used to register / deregister corresponding JMX and Sensor in the follow-up diff.
3. Simple JMX and MBeans for ClusterInfo and ServiceProperties, for now we only expose canary distribution policy. In the future, we can add more metrics about Service or Cluster properties by resuing the same framework.

PR #817, opened 2022-08-08
Add null-checks for cluster and service properties in D2ClientJmxManager

bohhyang:
This is a follow-up PR for a [previous change](https://github.com/linkedin/rest.li/pull/782/files#diff-9993e287812781a541ea17425748df60fe6966b7a77ea69eba72fcf8a5890102) to add null checks for cluster and service properties in D2ClientJmxManager which caused NPE in some rare edge cases (when the cluster or service properties received from PropertyEventBus is null, as seen in EXC-320587 and SI-27628).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.37.9] - 2022-08-07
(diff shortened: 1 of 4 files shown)


gu0keno0 (approved):
Thanks for the fix!

PR #818, opened 2022-08-08
Fix PrimitiveTemplateSpec not having className

junchuanwang:
Prior to the change, `<PrimitiveTemplateSpec>.getBindingName()` returns null, this is because the PrimitiveTemplateSpec does not have its class name properly set.


mchen07 (approved):
Looks good to me

PR #821, opened 2022-08-17
Exclude transitive Netty dependency for ZooKeeper client.

jpstewart:
The updated ZooKeeper client depends on Netty, but uses a newer version.
Prior to that version Netty switched its netty-all Maven coordinates from
being a super jar to being a pom that transitively depends on the other
Netty dependencies. This makes it possible for it to bring in a conflicting
version of Netty, resulting in two versions in the classpath. ZooKeeper
appears fine with our current Netty version, so excluding the transitive
dependencies seems the safest way to ensure no issues are caused by
upgrading Netty to match the ZooKeeper version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,9 +14,11 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.37.15] - 2022-08-23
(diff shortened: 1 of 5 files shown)


bohhyang (approved):
lgtm, thx

rickzx (approved):

PR #822, opened 2022-08-19
Avoid casting classloader to URLLoader in ResourceModelEncoder to be compatible with Java 11

junchuanwang:
Prior to Java 9, AppClassLoader is a subclass of URLClassLoader, but this is not the case after Java 9.

Therefore casting AppClassLoader to URLClassLoader will fail in Java 11.

This change removed related logic contains casting and instead use ClassGraph to search for restspec file instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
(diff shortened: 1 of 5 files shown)


evanw555 (approved):
LGTM, but have you tested this with restspecs from external JARs?

jessejie7:
nice change! much simpler

PR #824, opened 2022-08-24
Revert DefaultDocumentationRequestHandler to be blocking again

evanw555:
Back in January, we changed the blocking DDRH to be non-blocking (fail requests sent during initialization)
to solve an edge-case issue of request thread pool exhaustion. This started causing a lot more complaints,
as many more use cases were now seeing errors for `OPTIONS` calls that would otherwise be successful.

This change brings back the old blocking behavior as default, and adds `NonBlockingDocumentationRequestHandler`
as a subclass of the DDRH which users can configure if they see thread pool exhaustion in their use case.

### DIFF
--- a/restli-docgen/src/main/java/com/linkedin/restli/docgen/DefaultDocumentationRequestHandler.java
+++ b/restli-docgen/src/main/java/com/linkedin/restli/docgen/DefaultDocumentationRequestHandler.java
@@ -34,18 +34,20 @@ import com.linkedin.restli.common.RestConstants;
 import com.linkedin.restli.internal.server.model.ResourceModel;
 import com.linkedin.restli.server.RestLiConfig;
 import com.linkedin.restli.server.RestLiDocumentationRequestHandler;
-import com.linkedin.restli.server.RestLiServiceException;
 import com.linkedin.restli.server.RoutingException;
 
 import java.io.ByteArrayOutputStream;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 
 /**
  * Default {@link RestLiDocumentationRequestHandler} that serves both HTML and JSON documentation.
+ * This request handler initializes its renderers lazily (on the first request).
+ * The initializing is blocking, meaning that subsequent request threads will block until it's done.
+ *
+ * If a non-blocking request handler is needed, consider using {@link NonBlockingDocumentationRequestHandler}.
  *
  * @author Keren Jin
  */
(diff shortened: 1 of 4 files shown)


restli-docgen/src/main/java/com/linkedin/restli/docgen/DefaultDocumentationRequestHandler.java
```diff
@@ -260,5 +269,5 @@ private static RoutingException createRoutingError(String path)
   private RestLiDocumentationRenderer _jsonRenderer;
   private RestLiConfig _config;
   private Map<String, ResourceModel> _rootResources;
-  private final AtomicBoolean _shouldInitialize = new AtomicBoolean(true);
+  private boolean _initialized = false;
```
zhihaodev, line 272: Wondering if `volatile` keyword is needed for the variable?

zhihaodev (commented):
Is it possible have unit tests for the changes?

restli-docgen/src/main/java/com/linkedin/restli/docgen/DefaultDocumentationRequestHandler.java
cgtz: May not be needed since they use a double check pattern where _initialized is rechecked/manipulated within a synchronized block.

restli-docgen/src/main/java/com/linkedin/restli/docgen/NonBlockingDocumentationRequestHandler.java
```diff
@@ -0,0 +1,62 @@
...
+ * Non-blocking extension of the default {@link RestLiDocumentationRequestHandler} that is needed in special use cases.
+ * This implementation blocks on the request thread that lazily initializes the renderers, but refuses to block
+ * subsequent request threads, failing the requests instead.
+ *
+ * The advantage of this approach is that the request thread pool may avoid exhaustion in the case of lengthy initialization.
+ * The downside is that requests sent during initialization will fail, which may cause client problems and confusion.
+ *
+ * @author Evan Williams
+ */
+public class NonBlockingDocumentationRequestHandler extends DefaultDocumentationRequestHandler
```
cgtz, line 39: How does one configure their app to use this class instead of the default?

cgtz (approved):
approved after reviewer questions answered.

restli-docgen/src/main/java/com/linkedin/restli/docgen/NonBlockingDocumentationRequestHandler.java
evanw555: `RestLiConfig#setDocumentationRequestHandler` in the boot listener

zhihaodev (approved):

PR #827, opened 2022-08-31
Bump Netty version to latest (4.1.79) and remove ZooKeeper Netty exclusions.

jpstewart:
This is to resolve issues with netty-all and netty-* being resolved to different
versions by Gradle. Newer versions of netty-all are a POM dependency that transtively
depends on the others, allowing Gradle to pull up any conflicting versions properly.


CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.37.19] - 2022-08-31
```
rickzx, line 17: We probably need to bump minor version for this. 29.38.0. Because it introduces some backward incompatible changes, and we would like this to get user attention at gradle build time.

jpstewart: Done

rickzx (approved):
LGTM. Thanks!

PR #828, opened 2022-09-01
R2 Support for UDS

nizarm:


gu0keno0 (commented):
Thanks for filing this PR, Nizar!

Just to confirm, is this ready for review? I see a previous commit in it, and the socket creation is TBD.

PR #830, opened 2022-09-01
Add Support for UDS transport in R2

nizarm:


r2-netty/src/main/java/com/linkedin/r2/netty/client/http2/Http2ChannelPoolFactory.java
```diff
@@ -87,18 +94,33 @@ public Http2ChannelPoolFactory(
...
-        option(ChannelOption.CONNECT_TIMEOUT_MILLIS, connectTimeout).
-        handler(initializer);
     _ssl = sslContext != null && sslParameters != null;
   }
 
   @Override
   public AsyncPool<Channel> getPool(SocketAddress address)
   {
+    if (!StringUtils.isEmpty(_udsAddress)) {
+      address = new DomainSocketAddress(_udsAddress);
```
gu0keno0, line 121: Appreciate the simplification, yet I just feel that since getPool is asked to get a pool for SocketAddress, then we probably should not override it with UDS? 

I looked at the code further and maybe we should change the socket at a higher level, like in https://github.com/linkedin/rest.li/blob/6c51a48c8f1a68ed7a395dc7a3b4270771f69feb/r2-netty/src/main/java/com/linkedin/r2/netty/client/HttpNettyClient.java#L330 ?

PR #834, opened 2022-09-12
Java 11 - export jvm stat for java version >= 1_9

li-ukumar:
- Internal apis encapsulated starting 1.9, exporting jvmstat

- Updated constructor calls, deprecated in newer versions of Java

### DIFF
--- a/d2/build.gradle
+++ b/d2/build.gradle
@@ -1,5 +1,5 @@
-// required starting 1.11
-if(JavaVersion.current() >= JavaVersion.VERSION_11){
+// internal api inaccessible starting 1.9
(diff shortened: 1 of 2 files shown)


cgtz:
Could we change the CI config to also test with java 11 or are there more changes to make? https://github.com/linkedin/rest.li/blob/master/.github/workflows/build.yml#L14

evanw555:
@cgtz I think we need to resolve blockers such as this before making that change to ensure we don't block the CI. We could make the build jobs optional, but I still want the Java 8 build to be required

evanw555 (approved):

PR #838, opened 2022-09-16
Add a validation option to coerce base64 encoded fixed values

karthikrg:
Some clients cannot encode fixed values as avro encoded strings and instead use base64 encoding. Currently, these fail validation in the RestliValidationFilter. 

Provide an option to customize the validation options used in the filter, and add a validation option to fix base64 encoded fixed values.

### DIFF
--- a/data/src/main/java/com/linkedin/data/schema/validation/ValidationOptions.java
+++ b/data/src/main/java/com/linkedin/data/schema/validation/ValidationOptions.java
@@ -257,6 +257,36 @@ public final class ValidationOptions
     return _avroUnionMode;
   }
 
+  /**
+   * Set whether we should detect and fix base64 encoded fixed values.
+   *
+   * <p>Some clients cannot encode avro encoded strings, and send Base64 encoded fixed values instead. If this is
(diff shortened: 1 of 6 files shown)


data/src/main/java/com/linkedin/data/schema/validation/ValidationOptions.java
```diff
@@ -257,6 +257,36 @@ public boolean isAvroUnionMode()
...
   }
 
+  /**
+   * Set whether we should detect and fix base64 encoded fixed values.
+   *
+   * <p>Some clients cannot encode avro encoded strings, and send Base64 encoded fixed values instead. If this is
+   * set to true, then the validator attempts to detect this case and decode the base64 encoded string into fixed bytes
+   * to avoid breaking validation.</p>
+   *
+   * @param value True if base64 fixing mode for fixed values is enabled, false otherwise.
```
mchen07, line 267: `base64 fixing mode for fixed values` is reading awkward. Maybe "if flag to fix base64 encoding for fixed value is enabled"?

mchen07 (approved):

PR #839, opened 2022-09-16
Catch exceptions when connection state change event is received after ZK connection shutdown

bohhyang:
As titled, we used to catch the exception and log a debug message for async-shutdown only. But it was reported for regular shutdown as well. It's safe to catch the exception instead of throwing for all shutdown.


rickzx (approved):
LGTM. Please modify the CHANGELOG and version bump as well. Thanks!

PR #846, opened 2022-10-04
Add RestLiTraceInfo to the request context for both client and server…

dmessink:
The intent is to provide additional Rest.Li trace information to R2 by packaging this info into a RestLiTraceInfo bean and adding it to the request context for both incoming and outgoing requests. The main id is the resourceMethodIdentifier in the form: stripped(baseUriTemplate) + ':' + <operation name>. Example: album/photo:get or album:action:purgeAll. Note that keys are stripped from the baseUriTemplate when generating the identifier.

For outgoing requests, the identifier is available via Request.getResourceMethodIdentifier(). For incoming requests, the identifier is available via ResourceMethodDescriptor.getResourceMethodIdentifier(). (The baseUriTemplate is also available from the ResourceMethodDescriptor.)

### DIFF
--- a/restli-client/src/main/java/com/linkedin/restli/client/Request.java
+++ b/restli-client/src/main/java/com/linkedin/restli/client/Request.java
@@ -26,6 +26,7 @@ import com.linkedin.data.transform.filter.request.MaskTree;
 import com.linkedin.jersey.api.uri.UriTemplate;
 import com.linkedin.restli.common.HttpMethod;
 import com.linkedin.restli.common.ResourceMethod;
+import com.linkedin.restli.common.ResourceMethodIdentifierGenerator;
 import com.linkedin.restli.common.ResourceProperties;
 import com.linkedin.restli.common.ResourceSpec;
 import com.linkedin.restli.common.RestConstants;
@@ -71,6 +72,7 @@ public class Request<T>
   private final Map<String, Class<?>>       _queryParamClasses; // Used for coercing query params. In case of collection or iterable, contains the type parameter class.
   private final String                      _methodName; // needed to identify finders and actions. null for everything else
   private final String                      _baseUriTemplate;
+  private final String                      _resourceMethodIdentifier;
   private final Map<String, Object>         _pathKeys;
   private final List<Object>                _streamingAttachments; //Usually null since streaming is rare. Creating an empty List is wasteful.
   private RestliRequestOptions              _requestOptions;
@@ -113,6 +115,7 @@ public class Request<T>
     _queryParamClasses = queryParamClasses;
     _methodName = methodName;
     _baseUriTemplate = baseUriTemplate;
+    _resourceMethodIdentifier = ResourceMethodIdentifierGenerator.generate(baseUriTemplate, method, methodName);
     _pathKeys = pathKeys;
 
     if (_baseUriTemplate != null && _pathKeys != null)
@@ -215,6 +218,10 @@ public class Request<T>
   public String getBaseUriTemplate()
   {
     return _baseUriTemplate;
+  }
+
+  public String getResourceMethodIdentifier() {
+    return _resourceMethodIdentifier;
   }
 
   public Map<String, Object> getPathKeys()
--- a/restli-common/src/main/java/com/linkedin/restli/common/ResourceMethodIdentifierGenerator.java
+++ b/restli-common/src/main/java/com/linkedin/restli/common/ResourceMethodIdentifierGenerator.java
@@ -1,5 +1,5 @@
 /*
-   Copyright (c) 2013 LinkedIn Corp.
+   Copyright (c) 2022 LinkedIn Corp.
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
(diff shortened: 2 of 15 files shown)


restli-server/src/main/java/com/linkedin/restli/server/ResourceDefinition.java
```diff
@@ -52,6 +52,12 @@
    */
   String getD2ServiceName();
 
+  /**
+   * Gets the base uri template.
+   * @return the base uri template.
+   */
+  String getBaseUriTemplate();
```
evanw555, line 59: Should we add a default implementation here to prevent build incompatibilities?

restli-client/src/main/java/com/linkedin/restli/client/Request.java
```diff
@@ -26,6 +26,7 @@
 import com.linkedin.jersey.api.uri.UriTemplate;
 import com.linkedin.restli.common.HttpMethod;
 import com.linkedin.restli.common.ResourceMethod;
+import com.linkedin.restli.common.ResourceMethodIdentifierGenerator;
```
evanw555, line 29: Are you adding this class? I don't see it in the existing code or in this PR

restli-server/src/main/java/com/linkedin/restli/internal/server/model/ResourceMethodDescriptor.java
```diff
@@ -434,6 +434,24 @@ public String getMethodName()
     return _type.toString();
   }
 
+  /**
+   * Returns the resource method identifier if it can be determined, or null otherwise.
+   * This is identical to Request.getResourceMethodIdentifier().
+   * @return the resource method identifier if it can be determined, or null otherwise.
+   */
+  public String getResourceMethodIdentifier() {
```
evanw555, line 442: Theoretically, this can become stale if the resource model's URI template is updated after parents/subs are added. Would it be ok in just calling the `generate` method every time? Would the performance impact be that great?

restli-server/src/main/java/com/linkedin/restli/server/ResourceDefinition.java
dmessink: There is only one implementation available (ResourceModel). Do you think it may be extended externally? Or is this just being extra safe?

restli-client/src/main/java/com/linkedin/restli/client/Request.java
dmessink: Yes. Thanks, I had to reclone off my fork when I couldn't commit directly - and I missed adding these new files.

restli-server/src/main/java/com/linkedin/restli/internal/server/model/ResourceMethodDescriptor.java
dmessink: The scan for properties isn't free. Do you think either the ResourceModel or the ResourceModel.baseUriTemplate changes on the fly?

restli-server/src/main/java/com/linkedin/restli/server/ResourceDefinition.java
evanw555: I searched and didn't find any other internal usages. It's likely safe, however our policy is that we follow semver, so I'd suggest you just provide a default implementation that throws a `NotImplementedException` (or whatever it's called)

restli-common/src/main/java/com/linkedin/restli/common/ResourceMethodIdentifierGenerator.java
```diff
@@ -0,0 +1,70 @@
+/*
+   Copyright (c) 2013 LinkedIn Corp.
```
evanw555, line 2: nit: update year

restli-common/src/main/java/com/linkedin/restli/common/ResourceMethodIdentifierGenerator.java
```diff
@@ -0,0 +1,70 @@
...
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package com.linkedin.restli.common;
+
+/**
+ * @author dmessink
```
evanw555, line 20: This is a new concept, so I'd suggest we document a little more heavily. I suggest explaining that this generates a unique identifier for a given Rest.li method, fully qualified by its resource hierarchy

restli-common/src/main/java/com/linkedin/restli/common/RestLiTraceInfo.java
```diff
@@ -0,0 +1,48 @@
+package com.linkedin.restli.common;
```
evanw555, line 1: nit: we should add licenses to test files as well

restli-server/src/main/java/com/linkedin/restli/server/ResourceDefinition.java
dmessink: Done

evanw555:
You're adding new APIs in various places, so please bump the _minor_ version.

Also, update `CHANGELOG.md` accordingly. Follow the directions at the top of the file to ensure it's updated correctly.

evanw555 (approved):

PR #851, opened 2022-10-21
[rest.li] Update IndentedPdlBuilder to remove extra whitespace when writing a blank comment line

vshwnth2:
In some scenarios, we have a PDL where we need to write a comment that looks like 
`Flavors for an elementary particle and a fundamental constituent of matter\n\nThis is to test multi-line comment`
This is a 3 line comment with an empty 2nd line. However, since the 2nd line is empty, we do not want a trailing whitespace on the right (see image below). This PR helps address this.
Screen Shot 2022-10-21 at 12 22 04 AM

### DIFF
--- a/data/src/main/java/com/linkedin/data/schema/IndentedPdlBuilder.java
+++ b/data/src/main/java/com/linkedin/data/schema/IndentedPdlBuilder.java
@@ -115,9 +115,9 @@ class IndentedPdlBuilder extends PdlBuilder
       writeLine("/**");
       for (String line : doc.split("\n"))
       {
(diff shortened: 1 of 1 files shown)


evanw555:
This change seems appropriate. If you can get all the tests to pass, indicating that this doesn't break any of their assumptions, then I'll approve.

The only thing I wonder is if this trailing whitespace affects the parser. In other words, if you feed this comment back into the parser, will it parse the line with trailing whitespace as being empty or will it have that whitespace in it?

evanw555 (approved):

PR #852, opened 2022-10-24
Test Pegasus <--> Avro bijectivity

dg-builder:
This PR adds tests to check if converting between Avro and Pegasus schemas results in the same schema. For example, say we start with `DataSchema originalPegasusSchema`. We want to check if `originalPegasusSchema == avroToDataSchema(dataToAvroSchema(originalPegasusSchema))`. 

This PR shows that this conversion fails in certain cases and that `SchemaTranslator#dataToAvroSchema` and `SchemaTranslator#avroToDataSchema` are not inverse functions. Below are the failure cases:

## Issues with Avro --> Pegasus --> Avro Conversion
1. **`null` issues in unions**. There are 2 main issues: a) if a `null` type exists in an Avro union, it is reordered to be the first element of the union and b) a new default field is introduced to Avro schemas with `"default": null`. This occurs on Avro to Pegasus translation. The Avro `null` type in a union is converted to marking the union (if the Avro union contained 2+ non-null elements) or field (if the Avro union only contained 1 non-null element) as `optional`.

## Issues with Pegasus --> Avro --> Pegasus conversion
1. `typerefs` - Typerefs are completely stripped when converting to Avro. For example, a `tyepref`ed  `int` becomes a plain `int`
2. `includes` - fields that are added to a record via an `includes` statement are  kept but their status as `include`-ed is forgotten. For example, fields that are included in a schema become first-class fields upon conversion back to Pegasus.
3. `alias`es - A union of aliases is converted to a `record` where the aliases are field names.

# Future work
Conversion between Avro and Pegasus should result in the same schema. Any suggestions or gotchas to look out for when fixing these issues would be appreciated


data-avro/src/test/java/com/linkedin/data/avro/AvroSchemaEquals.java
```diff
@@ -0,0 +1,260 @@
+package com.linkedin.data.avro;
```
evanw555, line 1: nit: add license to the top of all new source files (with current year)

data-avro/src/test/java/com/linkedin/data/avro/AvroSchemaEquals.java
```diff
@@ -0,0 +1,260 @@
...
+import org.apache.avro.JsonProperties;
+import org.apache.avro.Schema;
+
+
+/**
+ * Compare 2 Avro schemas for equality.
+ *
+ * NOTE: This is temporary until avro-util supports an equality check between 2 avro schemas
+ */
+public class AvroSchemaEquals {
```
evanw555, line 19: nit: braces on new lines in Rest.li (I'm sorry)

evanw555 (commented):
Overall LGTM, just a couple nits

evanw555 (approved):

PR #853, opened 2022-10-25
Refactor DNS metrics for the `HttpNettyClient` to be easier to use.

jpstewart:
The current model requires injecting the `HttpNettyClient` into a JMX MBean. This doesn't work well with the current pattern of using the client, since the client type is abstracted and multiple clients could be created, resulting in a need to manage multiple MBeans even though the DNS resolution logic is shared globally.

This change lets the MBean itself hold the stats, and the Netty client will call back to it to increment the counters. This way when the `HttpClientFactory` is created we can provide it with a single MBean to track DNS across all clients, and the factory can handle injecting it properly into an applicable client instances it creates.


r2-netty/src/main/java/com/linkedin/r2/netty/client/JmxDnsMetricsCallback.java
```diff
@@ -0,0 +1,59 @@
...
+import com.linkedin.common.stats.LongStats;
+import com.linkedin.common.stats.LongTracker;
+import com.linkedin.common.stats.LongTracking;
+import java.util.concurrent.atomic.AtomicLong;
+
+
+public class JmxDnsMetricsCallback implements HttpNettyClientJmxMBean, DnsMetricsCallback {
+  private final AtomicLong _dnsResolutionErrors = new AtomicLong(0);
+  private final AtomicLong _dnsResolutions = new AtomicLong(0);
+  private final LongTracker _dnsResolutionLatency = new LongTracking();
```
gu0keno0, line 28: will be clearer to include time unit in the var name, e.g. _dnsResolutionLatencyMs , also good to keep it consistent

gu0keno0 (approved):

r2-netty/src/main/java/com/linkedin/r2/netty/client/JmxDnsMetricsCallback.java
jpstewart: Changed dnsResolutionLatency to dnsResolutionLatencyMs.

PR #854, opened 2022-10-25
Add channel writability to streaming timeout exception

TylerHorth:
This should help with determining the cause of the timeout. If the channel is not writable, then it is likely that the remote endpoint has stopped consuming date. Otherwise, the local endpoint is the likely culprit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.40.12] - 2022-11-30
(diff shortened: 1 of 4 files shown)


cgtz (approved):
LGTM. Rerunning tests since one of the flaky ones failed.

cgtz:
One thing you will have to do before we merge this is to update the changelog/version and run the update-changelog script: https://iwww.corp.linkedin.com/wiki/cf/display/ENGS/Pegasus+Git+Workflow#PegasusGitWorkflow-Implementation

PR #860, opened 2022-10-31
Reenable LastSeenZKStoreTest test

rickzx:
Reenable LastSeenZKStoreTest to make sure it is included in the CI/CD pipeline. Not creating a new version for this, so it will be bundled with the next release.


gu0keno0 (approved):
This reads like an important test, any idea for why this was disabled in the past?

rickzx:
> This reads like an important test, any idea for why this was disabled in the past?
@gu0keno0 I found this RB: https://rb.corp.linkedin.com/r/1336587/ that disables this test 4 years ago due to consistently failing. However, i tested locally and it passed without problem. I agree that it's an important test, so i'll wait and see if CI/CD is able to pass with this test reenabled.

PR #865, opened 2022-11-14
Fix the deprecated configuration name used in the PegasusPlugin

njuliyu:
The pegasus used the `Implementation` and `testImplementation` configuration in the code, this PR is to fix that, and also added a integration test to verify data-template classes in api moudle can be consumed by other modules

### DIFF
--- a/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
+++ b/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
@@ -1838,7 +1838,7 @@ public class PegasusPlugin implements Plugin<Project>
 
     // include additional dependencies into the appropriate configuration used to compile the input source set
     // must include the generated data template classes and their dependencies the configuration
-    String compileConfigName = isTestSourceSet(sourceSet) ? "testCompile" : "compile";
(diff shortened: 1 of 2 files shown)


gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1838,7 +1838,14 @@ protected GenerateDataTemplateTask configureDataTemplateGeneration(Project proje
 
     // include additional dependencies into the appropriate configuration used to compile the input source set
     // must include the generated data template classes and their dependencies the configuration
-    String compileConfigName = isTestSourceSet(sourceSet) ? "testCompile" : "compile";
+    String compileConfigName;
+    if (isAtLeastGradle7()) {
```
evanw555, line 1842: Leave a comment here explaining the context for this check, and maybe add a TODO stating that we can remove this once the min required version (this is defined as a constant near the top of this file) is 7+

njuliyu: add a comment

evanw555 (approved):

PR #867, opened 2022-11-16
[rest.li] Add util class to convert generic List/Map to DataList/DataMap or vice versa

vshwnth2:
Currently, there are some internal implementations of this, but logic is pretty standard conversion between java and restli objects.

### DIFF
--- a/data/src/main/java/com/linkedin/util/DataComplexUtil.java
+++ b/data/src/main/java/com/linkedin/util/DataComplexUtil.java
@@ -1,0 +1,185 @@
+package com.linkedin.util;
+
+import com.linkedin.data.Data;
+import com.linkedin.data.DataList;
+import com.linkedin.data.DataMap;
+import com.linkedin.data.DataMapBuilder;
+import com.linkedin.data.collections.CheckedUtil;
+import com.linkedin.data.template.DataTemplateUtil;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+
+/**
+ * Utils to convert generic List/Map to DataList/DataMap or vice versa
+ */
+public class DataComplexUtil {
+
+  private DataComplexUtil() { }
+
+  /**
+   * Converts the input {@link Map} to a {@link DataMap} recursively.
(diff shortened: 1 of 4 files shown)


data/src/main/java/com/linkedin/util/DataComplexUtil.java
```diff
@@ -0,0 +1,187 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+
+/**
+ * Utils to convert generic List/Map to DataList/DataMap or vice versa
+ */
+public class DataComplexUtil {
```
saponniah, line 18: Can you move the relevant unit test from here?

https://git.corp.linkedin.com:1367/a/plugins/gitiles/multiproducts/graphql-infra/+/master/util/src/test/java/com/linkedin/graphql/DataComplexUtilTest.java

data/src/main/java/com/linkedin/util/DataComplexUtil.java
```diff
@@ -0,0 +1,187 @@
...
+
+  private DataComplexUtil() { }
+
+  /**
+   * Converts the input {@link Map} to a {@link DataMap} recursively.
+   * Optional to stringfy the value of the leaf nodes in the datacomplex
+   */
+  public static DataMap convertMap(Map<String, ?> input, boolean stringify) {
```
aman1309, line 26: It's annoying but can follow the Restli repo indentation ie '{' in a new line

saponniah (approved):

junchuanwang:
please add changelog and upgrade pegasus version?

junchuanwang (changes requested):
please add changelog and upgrade pegasus version?

data/src/main/java/com/linkedin/util/DataComplexUtil.java
junchuanwang: yeah it is call AllMan's style. https://en.wikipedia.org/wiki/Indentation_style

Java code doesn't usually do that though

junchuanwang (approved):

PR #875, opened 2022-12-20
Expose enumeration of clusters to `ZKFailoutConfigProvider`.

jpstewart:
This assists higher level functions that need to gather data from all of the failout configs. It is left off the interface since the interface should support implementations which do not have a defined set of known clusters.

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/clusterfailout/ZKFailoutConfigProvider.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/clusterfailout/ZKFailoutConfigProvider.java
@@ -20,6 +20,8 @@ import com.linkedin.d2.balancer.LoadBalancerState;
 import com.linkedin.d2.balancer.LoadBalancerStateItem;
 import com.linkedin.d2.balancer.properties.FailoutProperties;
 
+import java.util.HashSet;
+import java.util.Set;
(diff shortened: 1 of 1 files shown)


d2/src/main/java/com/linkedin/d2/balancer/clusterfailout/ZKFailoutConfigProvider.java
```diff
@@ -90,6 +91,10 @@ public FailoutConfig getFailoutConfig(String clusterName)
     return failedoutClusterManager != null ? failedoutClusterManager.getFailoutConfig() : null;
   }
 
+  public Set<String> getClusters() {
+    return _failedoutClusterManagers.keySet();
```
shangchengying, line 95: Should we make a copy [given the doc](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#keySet--)? Wondering if it will give an exception when the underlying map is modified while we are iterating the set.

jpstewart: Done. The docs say that the iterator will not throw an exception but reflects at a state at the time of creation or in the future and are weakly consistent. For a metrics use case I think this is acceptable (we have that issue reading each config over time as well).

shangchengying (approved):

PR #883, opened 2023-02-09
Split getPotentialClients between subsetting and not-subsetting

idarmans:
~55% of CPU usage consumed during getPotentialClients is used for constructing weightedUris as shown in the following flamegraph with the Stream collecting.
Screenshot 2023-02-09 at 11 57 01 AM

To minimize this operation, this change is to avoid doing it for clusters that don't have subsetting enabled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
+
+## [29.41.7] - 2023-02-13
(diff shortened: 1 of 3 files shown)


rickzx (commented):
Looks good to me. Thanks for making the changes to make D2 more efficient!
Please follow https://iwww.corp.linkedin.com/wiki/cf/pages/viewpage.action?spaceKey=ENGS&title=Pegasus+Git+Workflow#PegasusGitWorkflow-Implementation to update CHANGELOG.md as well

rickzx (approved):

PR #884, opened 2023-02-09
Mark AlternativeKey annotation as deprecated

nickibi:
In the future, we will no longer support AlternativeKey. Therefore, mark the AlternativeKey annotation as deprecated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.41.7] - 2023-02-09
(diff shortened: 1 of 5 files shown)


mchen07 (approved):

nickibi:
Mark @AltertiveKey as deprecated will cause compile errors in the classes where annotation has been used, for example: https://github.com/linkedin/rest.li/actions/runs/4139749453. I am afraid we cannot roll out this change.

junchuanwang:
I think our build is too strict. I am surprised we fail on Error. We should remove -Werror for `restli-int-test` modules.  Otherwise we will never roll out any feature that mark things deprecated.

PR #886, opened 2023-02-15
fix translation of float/double's -Infinity/Infinity/NaN

dg-builder:
Bug fix -  The scenario is as follows:
* a data map contains a deserialized `Double.POSITIVE_INFINITY`, `Double.NEGATIVE_INFINITY`, or `Double.NaN` 
* a call to `DataTranslator#dataMapToGenericRecord` or `DataTranslator#dataMapToSpecificRecord` is made with the data map.

This flow throws an exception because those double values are serialized as strings (`Infinity`, `-Infinity`, `NaN`) and are then attempted to be deserialized directly into a Double (`ClassCastException: java.lang.String cannot be cast to java.lang.Number`).

### DIFF
--- a/data-avro/src/test/java/com/linkedin/data/avro/TestDataTranslator.java
+++ b/data-avro/src/test/java/com/linkedin/data/avro/TestDataTranslator.java
@@ -1075,6 +1075,40 @@ public class TestDataTranslator
     }
   }
 
+  @Test
+  public void testInfinityAndNan() throws IOException {
+    String schemaText =
+        "{\n" +
+          "  \"type\" : \"record\",\n" +
+          "  \"name\" : \"Foo\",\n" +
+          "  \"fields\" : [\n" +
+          "    { \"name\" : \"doubleRequired\", \"type\" : \"double\" }\n" +
+          "  ]\n" +
+          "}\n";
+    // First element is the input, second element is the expected output
+    Object[][] inputs = {
+        {
+            "{ \"doubleRequired\" : \"Infinity\"}",
+            Double.POSITIVE_INFINITY
+        }, {
+            "{ \"doubleRequired\" : \"NaN\"}",
+            Double.NaN
+        }, {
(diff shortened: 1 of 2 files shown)


data-avro/src/test/java/com/linkedin/data/avro/TestDataTranslator.java
```diff
@@ -1075,6 +1075,40 @@ public void testDataTranslator() throws IOException
     }
   }
 
+  @Test
+  public void testInfinityAndNan() throws IOException {
+    String schemaText =
+        "{\n" +
```
nickibi, line 1081: Shall we also add a test case for float?

junchuanwang (approved):
+1 to give complete test cases

data-avro/src/test/java/com/linkedin/data/avro/TestDataTranslator.java
```diff
@@ -1075,6 +1075,41 @@ public void testDataTranslator() throws IOException
...
+            "{ \"doubleRequired\" : \"-Infinity\", \"floatRequired\" : \"-Infinity\"}",
+            Double.NEGATIVE_INFINITY
+        }
+    };
+    RecordDataSchema recordDataSchema = (RecordDataSchema) TestUtil.dataSchemaFromString(schemaText);
+    for (Object[] input : inputs) {
+      DataMap dataMap = TestUtil.dataMapFromString((String) input[0]);
+      Schema avroSchema = SchemaTranslator.dataToAvroSchema(recordDataSchema);
+      GenericRecord avroRecord = DataTranslator.dataMapToGenericRecord(dataMap, recordDataSchema, avroSchema);
+      assertEquals(avroRecord.get("doubleRequired"), input[1]);
```
nickibi, line 1107: assert float as well.

dg-builder: shoot, sorry about that. I merged before I saw your comment. I'll put up another PR with the fix

PR #888, opened 2023-02-17
[rest.li] Update Data.TraverseCallback to have a callback for 'endKey'

vshwnth2:
This helps provide a way to traverse through the DataMap and strip out fields if needed. See https://rb.corp.linkedin.com/r/3855625/diff/1/#1 as a WIP example of a use case

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,11 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
-## [29.41.10] - 2023-02-16
+## [29.41.11] - 2023-03-07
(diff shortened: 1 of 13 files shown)


nickibi:
This is a backward incompatible change, as the public API has been changed. Anyone who uses those APIs will break. We should be very cautious to roll-out such change in Rest.li. It will require a major version bump.

vshwnth2:
>

vshwnth2:
> This is a backward incompatible change, as the public API has been changed. Anyone who uses those APIs will break. We should be very cautious to roll-out such change in Rest.li. It will require a major version bump.

I kept the existing key and added a new one to rename it something more descriptive. I can revert that part of the change if that's safer

PR #889, opened 2023-02-23
use proper UTF8 encoding in AvroUtil.jsonFromGenericRecord, also deprecate AvroUtil and bump avro-util dependency

radai-rosenblatt:
OutputStrean.toString() decodes bytes using the system default encoding, which might not be unicode, resulting in corrupt text

### DIFF
--- a/data-avro/src/main/java/com/linkedin/data/avro/util/AvroUtil.java
+++ b/data-avro/src/main/java/com/linkedin/data/avro/util/AvroUtil.java
@@ -21,6 +21,7 @@ import com.linkedin.avroutil1.compatibility.AvroVersion;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericDatumReader;
 import org.apache.avro.generic.GenericDatumWriter;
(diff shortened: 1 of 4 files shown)


data-avro/src/main/java/com/linkedin/data/avro/util/AvroUtil.java
```diff
@@ -61,7 +67,7 @@ private static String jsonFromGenericRecord(
     writer.setSchema(record.getSchema());
     writer.write(record, jsonEncoder);
     jsonEncoder.flush();
-    return outputStream.toString();
```
karthikrg, line 64: Why not use https://docs.oracle.com/javase/7/docs/api/java/io/ByteArrayOutputStream.html#toString(java.lang.String)?

radai-rosenblatt: the argument to that is a string, which causes a charset lookup, whereas what i used uses an already-looked-up Charset object.
will change

xiaoyu-yang-gh (approved):

nickibi:
Please update CHANGELOG.md and gradle.properties, ref: https://iwww.corp.linkedin.com/wiki/cf/pages/viewpage.action?spaceKey=ENGS&title=Pegasus+Git+Workflow

nickibi (approved):

mchen07 (approved):

PR #892, opened 2023-04-05
Let FieldDef lazy-calculate and cache hash code

zhaih:
Sometimes `_dataSchema` could be very complex and makes computing hash code for this class extremely slow. Since all data fields participated in hash code calculation is final, we should able to calculate it at construction time and avoid paying the cost repeatedly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.46.9] - 2023-11-02
+- Update FieldDef so that it will lazily cache the hashCode.
+
 ## [29.46.8] - 2023-10-11
 - add metrics about xds connection status and count
 
@@ -5554,7 +5557,8 @@ patch operations can re-use these classes for generating patch messages.
 
 ## [0.14.1]
 
-[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.46.8...master
+[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.46.9...master
+[29.46.9]: https://github.com/linkedin/rest.li/compare/v29.46.8...v29.46.9
 [29.46.8]: https://github.com/linkedin/rest.li/compare/v29.46.7...v29.46.8
 [29.46.7]: https://github.com/linkedin/rest.li/compare/v29.46.6...v29.46.7
 [29.46.6]: https://github.com/linkedin/rest.li/compare/v29.46.5...v29.46.6
--- a/data/src/main/java/com/linkedin/data/template/FieldDef.java
+++ b/data/src/main/java/com/linkedin/data/template/FieldDef.java
@@ -37,6 +37,7 @@ public class FieldDef<T>
   private final DataSchema _dataSchema;
   private final Class<?> _dataClass;
   private final RecordDataSchema.Field _field;
+  private final int _hashCode;
 
   public FieldDef(String name, Class<T> type)
   {
@@ -48,6 +49,7 @@ public class FieldDef<T>
     _name = name;
     _type = type;
     _dataSchema = dataSchema;
+    _hashCode = computeHashCode();
     /**
      * FieldDefs representing context, pagination, or things relating to synchronization will not
      * have schemas, so dataSchema and thus dataClass can be null.
@@ -126,6 +128,11 @@ public class FieldDef<T>
 
   @Override
   public int hashCode()
+  {
+      return _hashCode;
+  }
+
+  private int computeHashCode()
   {
     return 13*_name.hashCode() + 17*_type.hashCode() + 23*(_dataSchema == null? 1 :_dataSchema.hashCode());
   }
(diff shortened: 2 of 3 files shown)


zackthehuman:
Thanks for the contribution, Patrick! Based on the usage of `FieldDef` I think we should understand the implications and tradeoffs with how we compute the hashcode.

Here are a few things to consider:

- `FieldDef` is not `final`, and can be extended. If we precompute the hashcode by calling `hashCode()` in the constructor, then we'll do it for all subclasses too, even if that isn't the desired behavior. 
- Caching the hashcode will improve performance as long as the _same_ instance is reused -- do we know if that is the usage pattern? If we're creating new `FieldDef` or derived class instances frequently, then this caching and eager computation may actually _negatively_ impact performance.
- There are classes like `Parameter` which extend `FieldDef`, and the above points may apply here.

zhaih:
Hi Zack, thank you for the comment:
1. For child classes, I think usually people either use `super.hashCode()` as part of the calculation or just do it from scratch, I don't think either way will be a true blocker
2. For our use case (and I believe for maybe most use cases using RestLi), the [ActionRequestBuilder](https://github.com/linkedin/rest.li/blob/master/restli-client/src/main/java/com/linkedin/restli/client/ActionRequestBuilder.java#L58) is the only place we're using it, and the key is get from a `ResourceSpecImpl` class which stores those `FieldsDef` as long as the service is running. I can't say that's for every use cases, but I do think the use case should be fairly normal.

Regarding the potential cost to the constructor, it can be solved by lazily initialize the `_hashCode` field, but maybe need some synchronization to avoid multiple concurrent write (but maybe that's acceptable since the hash code should not change?)

Let me know what you think, thanks!

zackthehuman:
> Let me know what you think, thanks!

I chatted with some folks about this internally and our recommendation is to compute the value lazily so that it is computed the first time `hashCode()` is called. Your point about synchronization is valid, but I'm convinced that it's not a huge problem since it should compute to the same value (as you mentioned).

How does this sound?

zhaih:
Sure, I just changed it to compute lazily

data/src/main/java/com/linkedin/data/template/FieldDef.java
```diff
@@ -126,6 +127,16 @@ public boolean equals(Object object)
 
   @Override
   public int hashCode()
+  {
+      if (_hashCode == null) {
+          // If this method is called by multiple thread, there might be multiple concurrent write
+          // here, but since the hashCode should be the same it is tolerable
+          _hashCode = computeHashCode();
+      }
+      return _hashCode;
```
zackthehuman, line 136: Leaving this comment here for my future self and/or other maintainers:

We compute the hash code of `DataMap` lazily using `synchronize(this)` and a form of double-checked locking:

```
public int dataComplexHashCode()
{
  if (_dataComplexHashCode != 0)
  {
    return _dataComplexHashCode;
  }

  synchronized (this)
  {
    if (_dataComplexHashCode == 0)
    {
      _dataComplexHashCode = DataComplexHashCode.nextHashCode();
    }
  }

  return _dataComplexHashCode;
}
```

This approach avoids auto boxing/unboxing between `Integer` and `int` but makes the assumption that the computed hash code will never be 0. Doing something like this is also an option we can take in the future if we want to explore this more.

zhaih: A similar but more general idea maybe just use another boolean to track whether the hashcode is computed. And better maybe using Atomics instead of synchronized block, like
```
public int hashCode() {
  if (_computed.get() == false) {
    int hashCode = computeHashCode();
    if (_computed.get() == false) {
      _hashCode.compareAndSwap(INIT_VALUE, hashCode);
      _computed.set(true);
    }
  }
  return _hashCode.get();
}
```

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.41.13] - 2023-04-12
+Update FieldDef so that it will lazily cache the hashCode.
```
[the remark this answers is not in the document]
zhaih: Done, thanks!

zackthehuman (approved):

zackthehuman (approved):
Thanks for the long wait! Let's goooooo!

data/src/main/java/com/linkedin/data/template/FieldDef.java
PapaCharlie: FWIW here there's little guarantee the write will ever be visible to other threads. Have you considered making `_hashCode` at least `volatile`:
```java
private volatile Integer _hashCode;
```
It preserves the original desired behavior of not wanting a full lock but at least makes it memory-safe.

Alternatively, why not just compute it at construction and make it final? Presumably instances of `FieldDef` aren't dynamically generated, they come from parsed .pdl and .pdsc schemas right? You could just eat the cost exactly once and forget about it, especially since it will happen very early on and likely completely out of the query serving codepath

zhaih: Hi @PapaCharlie , thanks for chiming in:
For the flushing to memory problem: it's valid but since we'll not set it based on it's previous value, essentially all threads will set it to the same value if they cannot see it. So eventually that's not a safety issue but rather a performance issue (we compute it # of thread times rather than once). I like to compute it in ctor better (and I actually first implemented it in this way) but @zackthehuman is worrying about impacting the performance of the ctor so we changed it to lazy later on.

PR #895, opened 2023-04-18
Get line/col number in PDL schema encoder

jhandley:
Add ability to get the line and column numbers of schema elements written out by SchemaToPDL encoder.

### DIFF
--- a/data/src/main/java/com/linkedin/data/schema/SchemaToPdlEncoder.java
+++ b/data/src/main/java/com/linkedin/data/schema/SchemaToPdlEncoder.java
@@ -147,7 +147,7 @@ public class SchemaToPdlEncoder extends AbstractSchemaEncoder
       _writer = out;
       _writeLocations = Collections.emptyMap();
     }
-    _encodingStyle = EncodingStyle.INDENTED;
+    setEncodingStyle(EncodingStyle.INDENTED);
     _trackWriteLocations = returnContextLocations;
   }
 
@@ -159,6 +159,18 @@ public class SchemaToPdlEncoder extends AbstractSchemaEncoder
   public void setEncodingStyle(EncodingStyle encodingStyle)
   {
     _encodingStyle = encodingStyle;
+
+    // When counting column numbers, CompactPDLBuilder treats ',' as whitespace
+    if (_writer instanceof LineColumnNumberWriter)
+    {
+      if (_encodingStyle == EncodingStyle.COMPACT)
+      {
+        ((LineColumnNumberWriter) _writer).setIsWhitespaceFunction(c -> Character.isWhitespace(c) || c == ',');
+      } else
+      {
+        ((LineColumnNumberWriter) _writer).setIsWhitespaceFunction(Character::isWhitespace);
+      }
+    }
   }
 
   /**
@@ -172,18 +184,6 @@ public class SchemaToPdlEncoder extends AbstractSchemaEncoder
   {
     // Initialize a new builder for the preferred encoding style
     _builder = _encodingStyle.newBuilderInstance(_writer);
-
-    // When counting column numbers, CompactPDLBuilder treats ',' as whitespace
-    if (_writer instanceof LineColumnNumberWriter)
-    {
-      if (_encodingStyle == EncodingStyle.COMPACT)
-      {
-        ((LineColumnNumberWriter) _writer).setIsWhitespaceFunction(c -> Character.isWhitespace(c) || c == ',');
-      } else
-      {
-        ((LineColumnNumberWriter) _writer).setIsWhitespaceFunction(Character::isWhitespace);
-      }
-    }
 
     // Set and write root namespace/package
     if (schema instanceof NamedDataSchema)
--- a/data/src/main/java/com/linkedin/data/schema/SchemaToPdlEncoder.java
+++ b/data/src/main/java/com/linkedin/data/schema/SchemaToPdlEncoder.java
@@ -18,11 +18,14 @@ package com.linkedin.data.schema;
 
 import com.linkedin.data.DataList;
 import com.linkedin.data.DataMap;
+import com.linkedin.data.schema.grammar.PdlSchemaParser;
+import com.linkedin.util.LineColumnNumberWriter;
 import java.io.IOException;
 import java.io.StringWriter;
 import java.io.Writer;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.IdentityHashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
@@ -107,6 +110,10 @@ public class SchemaToPdlEncoder extends AbstractSchemaEncoder
   private String _namespace = "";
   private String _package = "";
 
+  private final boolean _trackWriteLocations;
+
+  private final Map<Object, PdlSchemaParser.ParseLocation> _writeLocations;
+
   /**
    * Construct a .pdl source code encoder.
    *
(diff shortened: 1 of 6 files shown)


data/src/main/java/com/linkedin/data/schema/SchemaToPdlEncoder.java
```diff
@@ -107,15 +110,42 @@ PdlBuilder newBuilderInstance(Writer writer)
...
+   *                              locations can be retrieved using {@link #getWriteLocations()} after parsing.
+   */
+  public SchemaToPdlEncoder(Writer out, boolean returnContextLocations)
+  {
+    if (returnContextLocations) {
+      _writeLocations = new IdentityHashMap<>();
+      // Wrap the Writer to track line/column numbers to report to elementWriteListener
+      LineColumnNumberWriter lineColumnNumberWriter = new LineColumnNumberWriter(out);
+      // When counting column numbers, CompactPDLBuilder treats ',' as whitespace
+      lineColumnNumberWriter.setIsWhitespaceFunction(c -> Character.isWhitespace(c) || c == ',');
```
zhihaodev, line 141: Should we set `|| c == ','` if encodingStyle is COMPACT?

data/src/main/java/com/linkedin/util/LineColumnNumberWriter.java
```diff
@@ -0,0 +1,193 @@
...
+   */
+  public void setIsWhitespaceFunction(Predicate<Character> isWhitespaceFunction) {
+    _isWhitespaceFunction = isWhitespaceFunction;
+  }
+
+  @Override
+  public void write(char[] cbuf, int off, int len) throws IOException {
+    _writer.write(cbuf, off, len);
+    for (; len > 0; len--) {
+      char c = cbuf[off++];
```
zhihaodev, line 83: Is boundary check needed, or is it taken care by line 81 already?

jhandley: This is taken care of by line 81. It will throw IndexOutOfBoundsException if len/offset are invalid.

data/src/main/java/com/linkedin/data/schema/SchemaToPdlEncoder.java
jhandley: Good call. Updated.

junchuanwang:
I am sorry that Pegasus might be super picky about coding style, do you mind to reformat according to existing curly brace style?

jhandley:
> I am sorry that Pegasus might be super picky about coding style, do you mind to reformat according to existing curly brace style?

Reformatted

data/src/main/java/com/linkedin/data/schema/SchemaToPdlEncoder.java
```diff
@@ -441,24 +502,29 @@ private void writeTyperef(TyperefDataSchema schema) throws IOException
...
         .writeComma()
         .writeSpace();
     writeReferenceOrInline(schema.getValues(), schema.isValuesDeclaredInline());
     _builder.write("]");
+    recordSchemaElementLocation(schema);
   }
 
   private void writeArray(ArrayDataSchema schema) throws IOException
   {
+    markSchemaElementStartLocation();
```
aman1309, line 522: instead adding these functions in every type write method can move it the switch case in encode and writeRecord field iterator. what is the purpose of it in each type?

data/src/main/java/com/linkedin/data/schema/SchemaToPdlEncoder.java
```diff
@@ -865,4 +935,25 @@ else if (_namespace.equals(schema.getNamespace()) && !_importsByLocalName.contai
...
+  }
+
+  private void recordSchemaElementLocation(Object schemaElement)
+  {
+    if (_trackWriteLocations)
+    {
+      LineColumnNumberWriter.CharacterPosition startPosition = ((LineColumnNumberWriter) _writer).popSavedPosition();
+      LineColumnNumberWriter.CharacterPosition endPosition =
+          ((LineColumnNumberWriter) _writer).getLastNonWhitespacePosition();
+      _writeLocations.put(schemaElement,
```
aman1309, line 954: there could be fields with same schema, no?

data/src/main/java/com/linkedin/data/schema/SchemaToPdlEncoder.java
```diff
@@ -107,15 +110,41 @@ PdlBuilder newBuilderInstance(Writer writer)
...
   {
-    _writer = out;
+    this(out, false);
+  }
+
+  /**
+   * Construct a .pdl source code encoder.
+   *
+   * @param out provides the encoded .pdl destination.
+   * @param returnContextLocations Enable recording the context locations of schema elements during parsing. The
```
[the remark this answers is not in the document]
jhandley: I see why that would make sense but I used this naming because it matches the terminology in PdlSchemaParser [here](https://github.com/linkedin/rest.li/blob/b8fadffc4b9d7369ef0da31dd93fab4a527665db/data/src/main/java/com/linkedin/data/schema/grammar/PdlSchemaParser.java#L125)

data/src/main/java/com/linkedin/data/schema/SchemaToPdlEncoder.java
jhandley: We don't call them for every type in the switch statement in encode. They aren't called for primitive types. Can certainly move this to encode but will not be as clean as just adding the two lines there since would have to call them only for non-primitives. Let me know what you prefer.

data/src/main/java/com/linkedin/data/schema/SchemaToPdlEncoder.java
jhandley: That would be a problem. What would an example be?

data/src/main/java/com/linkedin/data/schema/SchemaToPdlEncoder.java
```diff
@@ -140,6 +169,18 @@ public void encode(DataSchema schema) throws IOException
     // Initialize a new builder for the preferred encoding style
     _builder = _encodingStyle.newBuilderInstance(_writer);
 
+    // When counting column numbers, CompactPDLBuilder treats ',' as whitespace
```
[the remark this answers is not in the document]
jhandley: There is no guarantee that setEncodingStyle is called before encode. Seemed simpler to do this in one place otherwise would have to repeat this in the constructor too. This isn't a particularly expensive operation so seems ok to do it here. I'm ok moving it though if you feel strongly. Let me know.

data/src/main/java/com/linkedin/data/schema/SchemaToPdlEncoder.java
aman1309: For example take 2 fields with the same type.

data/src/main/java/com/linkedin/data/schema/SchemaToPdlEncoder.java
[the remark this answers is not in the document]
aman1309: we can just call setEncodingStyle() from constructor.

data/src/main/java/com/linkedin/data/schema/SchemaToPdlEncoder.java
```diff
@@ -107,15 +110,41 @@ PdlBuilder newBuilderInstance(Writer writer)
...
    * @param out provides the encoded .pdl destination.
    */
   public SchemaToPdlEncoder(Writer out)
   {
-    _writer = out;
+    this(out, false);
+  }
+
+  /**
+   * Construct a .pdl source code encoder.
```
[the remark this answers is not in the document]
jhandley: Done

data/src/main/java/com/linkedin/data/schema/SchemaToPdlEncoder.java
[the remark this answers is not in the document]
jhandley: Done

data/src/main/java/com/linkedin/data/schema/SchemaToPdlEncoder.java
jhandley: I'm trying to reproduce the case where two fields with the same type would cause a problem but I'm not able to. If I try to parse a PDL where the same field type is inlined twice like this I get an error that the schema is defined twice:
```
record ParentRecordSchema {
  field1: record SharedSchema {
    sharedSchemaField: int
  }
  field2: record SharedSchema {
    sharedSchemaField: int
  }
}
```
The only way this parses is if I reference the type by name for the second field:
```
record ParentRecordSchema {
  field1: record SharedSchema {
    sharedSchemaField: int
  }
  field2: SharedSchema
}
```
However, when the type is referenced by name like that, the record schema for SharedSchema is only defined once in the file so there is no problem just using the map without the list. In that case the map has a singe entry for each of ParentRecordSchema, SharedSchema, sharedSchemaField, field1 and field2.

Is there another case I'm missing?

aman1309: what if SharedSchema is in a different PDL file and is referenced in ParentRecordSchema. ie.
import com.....SharedSchema;
record ParentRecordSchema {
  field1: SharedSchema
  field2: SharedSchema
}

aman1309: or we have below schema
record ParentRecordSchema {
  field1: record SharedSchema {
    sharedSchemaField: int
  }
  field2: SharedSchema
  field3: SharedSchema
}

jhandley: Those are not a problem since the map from the schema to the location at which it is defined, not the location(s) where it is referenced. In all of these cases there will be one mapping from SharedSchema to the line/col where it is defined. For the fields, we map from the RecordSchema.Field to the location where the field is defined. If the type of the field is just a reference, we don't map that location.

junchuanwang (approved):

PR #896, opened 2023-04-20
HEAD requests content-length is not based on the body as per the http…

nizarm:
HEAD requests content-length is not based on the body as per the http spec. Fixing it to not override the content-length if the request is HEADER type

### DIFF
--- a/r2-core/src/main/java/com/linkedin/r2/transport/common/AbstractClient.java
+++ b/r2-core/src/main/java/com/linkedin/r2/transport/common/AbstractClient.java
@@ -89,7 +89,13 @@ public abstract class AbstractClient implements Client
     // This is needed as the legacy R2 server (before 2.8.0) does not support chunked transfer encoding.
     requestContext.putLocalAttr(R2Constants.IS_FULL_REQUEST, true);
     // here we add back the content-length header for the response because some client code depends on this header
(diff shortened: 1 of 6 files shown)


r2-core/src/main/java/com/linkedin/r2/transport/common/AbstractClient.java
```diff
@@ -89,7 +89,13 @@ public void restRequest(RestRequest request, RequestContext requestContext, Call
     // This is needed as the legacy R2 server (before 2.8.0) does not support chunked transfer encoding.
     requestContext.putLocalAttr(R2Constants.IS_FULL_REQUEST, true);
     // here we add back the content-length header for the response because some client code depends on this header
-    streamRequest(streamRequest, requestContext, Messages.toStreamCallback(callback, true));
+
+    boolean addContentLengthHeader = true;
+    if ("HEAD".equalsIgnoreCase(request.getMethod())) {
+      addContentLengthHeader = false;
+    }
```
PapaCharlie, line 96: Isn't there a way to get this as an enum? Feels like a string equality check is a little weak... Otherwise, this feels better anyway:
```java
boolean addContentLengthHeader = !"HEAD".equalsIgnoreCase(request.getMethod());
```

bohhyang (approved):

yzhan1 (approved):
Thanks for the fix!

ChrisCarini:

PR #898, opened 2023-04-21
HEAD requests content-length is not based on the body as per the http…

nizarm:
HEAD requests content-length is not based on the body as per the http spec. Fixing it to not override the content-length if the request is HEADER type


ChrisCarini:

PR #899, opened 2023-04-24
[Draft] Update use of java-doc related apis to use standard APIs so it can be Jdk 17 compatible

junchuanwang:
Replace the use of old `com.sun.javadoc` apis which are deprecated since JDK9 and removed since JDK 17.

Our CI/CD is using JDK8 for testing so that needs update too.


ChrisCarini:

PR #900, opened 2023-04-26
Improve synchronization on the R2 `RequestContext` local attributes.

jpstewart:
It is possible to cause a `ConcurrentModificationException` in the `RequestContext` because we synchronize it by using `Collections.synchronizedMap()`, which only synchronizes individual reads/writes. When making a `RequestContext` copy we must iterate over the collection, allowing other threads to interrupt it by modifying the map during the operation.

This fix synchronizes the whole copy operation as well. The default
mutex used for synchronizing in a `SynchronizedMap` is `this`, so
synchronizing on the map itself guards against modification.

### DIFF
--- a/r2-core/src/main/java/com/linkedin/r2/message/RequestContext.java
+++ b/r2-core/src/main/java/com/linkedin/r2/message/RequestContext.java
@@ -54,12 +54,14 @@ public class RequestContext
    */
   public RequestContext(RequestContext other)
   {
-    _localAttrs = Collections.synchronizedMap(new HashMap<>(other._localAttrs));
+    synchronized (other._localAttrs) {
+      _localAttrs = Collections.synchronizedMap(new HashMap<>(other._localAttrs));
+    }
   }
 
   private RequestContext(Map<String, Object> localAttrs)
   {
-    _localAttrs = localAttrs;
+    _localAttrs = Collections.synchronizedMap(localAttrs);
   }
 
   /**
@@ -117,7 +119,7 @@ public class RequestContext
   public boolean equals(Object o)
   {
     return (o instanceof RequestContext) &&
-            ((RequestContext)o)._localAttrs.equals(this._localAttrs);
+        ((RequestContext)o)._localAttrs.equals(this._localAttrs);
   }
 
   @Override
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.42.2] - 2023-05-11
+- Fix synchronization on `RequestContext` to prevent `ConcurrentModificationException`.
+
 ## [29.42.1] - 2023-05-11
 - Add support for returning location of schema elements from the PDL schema encoder.
   
@@ -5465,7 +5468,8 @@ patch operations can re-use these classes for generating patch messages.
 
 ## [0.14.1]
 
-[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.42.1...master
+[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.42.2...master
+[29.42.2]: https://github.com/linkedin/rest.li/compare/v29.42.1...v29.42.2
 [29.42.1]: https://github.com/linkedin/rest.li/compare/v29.42.0...v29.42.1
 [29.42.0]: https://github.com/linkedin/rest.li/compare/v29.41.12...v29.42.0
 [29.41.12]: https://github.com/linkedin/rest.li/compare/v29.41.11...v29.41.12
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-version=29.42.1
+version=29.42.2
 group=com.linkedin.pegasus
 org.gradle.configureondemand=true
 org.gradle.parallel=true


r2-core/src/main/java/com/linkedin/r2/message/RequestContext.java
```diff
@@ -43,7 +50,7 @@ public RequestContext()
     // this class publicly exposes the _localAttrs through getLocalAttrs() and
     // returned Map that must support null values because there is plenty of code
     // that is using this property.
-    _localAttrs = Collections.synchronizedMap(new HashMap<>());
```
rickzx, line 46: How about we just switch to `ConcurrentHashMap`. From the javadoc
>Similarly, Iterators, Spliterators and Enumerations return elements reflecting the state of the hash table at some point at or since the creation of the iterator/enumeration. They do not throw [ConcurrentModificationException](https://docs.oracle.com/javase/8/docs/api/java/util/ConcurrentModificationException.html). However, iterators are designed to be used by only one thread at a time.

jpstewart: My concerns with `ConccurentHashMap` are first that it will introduce more overhead into a very low fundamental part of the R2 stack, and second that it allows for a copy to be created that is not consistent with a single point in time. Since the iteration behavior is "return each value at most once", the copy could end up with two values that never coexisted at the same time in the original. This is a change in the contract the current behavior provides, and seeing as how we see quite a few of these errors in EXC tickets a lot of people are relying on concurrent edits to the attribute map and could be potentially impacted by a change in behavior.

r2-core/src/main/java/com/linkedin/r2/message/RequestContext.java
```diff
@@ -131,4 +140,204 @@ public String toString()
   {
     return _localAttrs.toString();
   }
+
+  private static class AttrsMap<K, V> implements Map<K, V>, Lock {
```
bbarkley, line 144: Curious why this implements Lock and exposes that

r2-core/src/main/java/com/linkedin/r2/message/RequestContext.java
```diff
@@ -131,4 +140,204 @@ public String toString()
...
+      } finally {
+        other.lock.unlock();
+      }
+    }
+
+    @Override
+    public boolean equals(Object other) {
+      lock();
+      try {
+        return map.equals(other);
```
bbarkley, line 177: Seems like this should be testing to see if the other object is AttrsMap, and locking it if so

r2-core/src/main/java/com/linkedin/r2/message/RequestContext.java
```diff
@@ -131,4 +140,204 @@ public String toString()
...
+      } finally {
+        unlock();
+      }
+    }
+
+    @Override
+    public Set<Entry<K, V>> entrySet() {
+      lock();
+      try {
+        return map.entrySet();
```
bbarkley, line 307: This exposes the entries to clients, which are able to make modifications and bypass your locking, leading to continued ConcurrentModificationExceptions. Same for keySet and values.

r2-core/src/main/java/com/linkedin/r2/message/RequestContext.java
bbarkley: Have you verified that ConcurrentHashMap has a higher overhead? I'd be somewhat surprised if it did, as reads are not blocking.

I'm not sure if the concern over the consistency is an issue. Currently if anyone is making changes to the map during an iteration the exception is thrown, so there is no consistent behavior during modifications. Using a concurrent map won't guarantee consistency, but we also don't know what the impact of allowing that will be.

r2-core/src/main/java/com/linkedin/r2/message/RequestContext.java
```diff
@@ -69,7 +80,7 @@ private RequestContext(Map<String, Object> localAttrs)
    */
   public Map<String, Object> getLocalAttrs()
   {
-    return _localAttrs;
+    return Collections.unmodifiableMap(_localAttrs);
```
bbarkley, line 83: If clients are making changes on the return value currently, this is a change in behavior which may break them.

bbarkley (changes requested):
The main concern here is the change in behavior that clients will see, as well as leaking the ability to still modify the map via key/value views.

r2-core/src/main/java/com/linkedin/r2/message/RequestContext.java
gu0keno0: Alternatively, if the main pain point is ```When making a RequestContext copy we must iterate over the collection``` how about just adding a thread-safe copy() method?

gu0keno0: Just to share what I got from looking at the stack traces. Looks like the common pattern is like the following:

```streamRequest(request, new RequestContext(requestContext)```

Then the copy constructor of RequestContext will need to copy the localAttrs syncMap, and if at the same time the original context is modified in another thread for whatever reason, the copy will hit the exception. @bbarkley 

And to make it worse, we offers an unsafe API of "getLocalAttrs()" which directly returns that internal map to users, and that would make it impossible(?) for us to fix the concurrency locally inside RequestContext .

Comment (gu0keno0) [reply r2-core/src/main/java/com/linkedin/r2/message/RequestContext.java:46]: @jpstewart from the doc https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#synchronizedMap(java.util.Map) , seems like we can at least fix the pattern above by fixing the copy constructor? We can add a synchronized loop there to copy the map.

Comment (jpstewart) [reply r2-core/src/main/java/com/linkedin/r2/message/RequestContext.java:46]: It looks like you are correct, I tested and verified this works. I expected it would not since the `SynchronizedMap` synchronizes on a mutex object held privately, but actually `Collections.synchronizedMap(m)` will use m itself as the mutex, so this works.

**Review by jpstewart — COMMENTED** [review]
Based on the finding that the mutex in the `SynchronizedMap` is `this`, this simply synchronizes around the whole iteration.

**Review by gu0keno0 — APPROVED** [review]

File: r2-core/src/main/java/com/linkedin/r2/message/RequestContext.java [hunk 6]
```diff
@@ -54,7 +54,9 @@ public RequestContext()
    */
   public RequestContext(RequestContext other)
   {
-    _localAttrs = Collections.synchronizedMap(new HashMap<>(other._localAttrs));
+    synchronized (other._localAttrs) {
+      _localAttrs = Collections.synchronizedMap(new HashMap<>(other._localAttrs));
+    }
   }
 
   private RequestContext(Map<String, Object> localAttrs)
```
bbarkley, line 62: For safety/clarity I think it would be better if the incoming map here was also wrapped in a synchronizedMap, and that behavior removed from the clone method. Unfortunately we can't rely on the type system to enforce that _localAttrs is a synchronized map.

bbarkley (approved):
Looks good, minor suggestion for clarity

r2-core/src/main/java/com/linkedin/r2/message/RequestContext.java
jpstewart: I don't think that will help, since if the other map is not a SynchronizedMap then it can still be directly modified during the copy, since the caller has direct access to the underlying map. Instead I propose changing the constructor RequestContext(Map localAttrs) to also wrap the _localAttrs in a SynchronizedMap. Then, since the constructors are the only places _localAttrs can be set it is provable it is synchronized.

gu0keno0 (approved):

r2-core/src/main/java/com/linkedin/r2/message/RequestContext.java
bbarkley: Sorry if my suggestion wasn't clear - what you propose is the same as my suggestion :)

PR #902, opened 2023-05-08
Rewrite Java Doc logic with JDK 17 compatible APIs

JoeJoe1989:
Currently RestLi is using com.sun.javadoc APIs to generate Java Doc. However these APIs are removed in JDK 17, making RestLi and its consumers incompatible with JDK 17. This PR aims to address the issue by rewriting Java Doc related logic with JDK 17 compatible APIs based on [migration doc](https://docs.oracle.com/en/java/javase/11/docs/api/jdk.javadoc/jdk/javadoc/doclet/package-summary.html#migration). 

Given that the current RestLi is built in JDK 8 while the replacement APIs were introduced in JDK 9 (and remain compatible through JDK 17), we would need to build RestLi in newer JDK version to leverage the replacement APIs. More specifically, JDK 11 is chosen for the build since it is the next LTS version after JDK 9. This would also require a major version bump of RestLi because of the use of APIs after JDK 8 and the resulting compatibility with JDK 8 consumers.

Build passed locally:
```
export JAVA_HOME=/Library/Java/JavaVirtualMachines/jdk11.0.13.8-msft.jdk/Contents/Home/
./gradle build
```
Also tested the local snapshot of RestLi against Java 11 and Java 17 consumers.


ChrisCarini:

PR #903, opened 2023-05-10
Support Unix sockets for HTTP/1

vigchan:
Extrapolating the code from HTTP/2 support

### DIFF
--- a/r2-netty/src/main/java/com/linkedin/r2/netty/client/http/HttpChannelPoolFactory.java
+++ b/r2-netty/src/main/java/com/linkedin/r2/netty/client/http/HttpChannelPoolFactory.java
@@ -29,7 +29,6 @@ import io.netty.channel.ChannelInitializer;
 import io.netty.channel.ChannelOption;
 import io.netty.channel.EventLoopGroup;
 import io.netty.channel.epoll.EpollDomainSocketChannel;
-import io.netty.channel.epoll.EpollEventLoopGroup;
 import io.netty.channel.group.ChannelGroup;
 import io.netty.channel.socket.nio.NioSocketChannel;
 import java.net.SocketAddress;
@@ -95,19 +94,13 @@ public class HttpChannelPoolFactory implements ChannelPoolFactory
     _tcpNoDelay = tcpNoDelay;
     _channelPoolWaiterTimeout = channelPoolWaiterTimeout;
 
-    if (!StringUtils.isEmpty(udsAddress)) {
-      _bootstrap = new Bootstrap().
-          group(eventLoopGroup).
-          channel(EpollDomainSocketChannel.class).
-          option(ChannelOption.CONNECT_TIMEOUT_MILLIS, connectTimeout).
-          handler(initializer);
-    } else {
-      _bootstrap = new Bootstrap().
-          group(eventLoopGroup).
-          channel(NioSocketChannel.class).
-          option(ChannelOption.CONNECT_TIMEOUT_MILLIS, connectTimeout).
-          handler(initializer);
-    }
+    Bootstrap bootstrap = !StringUtils.isEmpty(udsAddress) ?
+        new Bootstrap().channel(EpollDomainSocketChannel.class) : new Bootstrap().channel(NioSocketChannel.class);
+
+    _bootstrap = bootstrap
+        .group(eventLoopGroup)
+        .option(ChannelOption.CONNECT_TIMEOUT_MILLIS, connectTimeout)
+        .handler(initializer);
   }
 
   @Override
--- a/r2-netty/src/main/java/com/linkedin/r2/netty/client/http/HttpChannelPoolFactory.java
+++ b/r2-netty/src/main/java/com/linkedin/r2/netty/client/http/HttpChannelPoolFactory.java
@@ -28,12 +28,16 @@ import io.netty.channel.Channel;
 import io.netty.channel.ChannelInitializer;
 import io.netty.channel.ChannelOption;
 import io.netty.channel.EventLoopGroup;
+import io.netty.channel.epoll.EpollDomainSocketChannel;
+import io.netty.channel.epoll.EpollEventLoopGroup;
 import io.netty.channel.group.ChannelGroup;
 import io.netty.channel.socket.nio.NioSocketChannel;
 import java.net.SocketAddress;
 import java.util.concurrent.ScheduledExecutorService;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLParameters;
+import org.apache.commons.lang.StringUtils;
+
 
 /**
  * Factory class to produce {@link AsyncPool}&#060;{@link Channel}&#062; for Http Channels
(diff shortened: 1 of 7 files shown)


gu0keno0:
cc: @nizarm @jpstewart @rickzx @bohhyang This is a R2 related change for service mesh.

r2-netty/src/main/java/com/linkedin/r2/netty/client/http/HttpChannelPoolFactory.java
```diff
@@ -90,8 +96,19 @@ public HttpChannelPoolFactory(
     _tcpNoDelay = tcpNoDelay;
     _channelPoolWaiterTimeout = channelPoolWaiterTimeout;
 
-    _bootstrap = new Bootstrap().group(eventLoopGroup).channel(NioSocketChannel.class).
-        option(ChannelOption.CONNECT_TIMEOUT_MILLIS, connectTimeout).handler(initializer);
+    if (!StringUtils.isEmpty(_udsAddress)) {
+      _bootstrap = new Bootstrap().
+          group(new EpollEventLoopGroup()).
```
jpstewart, line 101: Could this event loop group be injected from the factory? Should it be different from `eventLoopGroup`?
In `ChannelPoolManagerFactoryImpl` it says the creator of the factory has responsibility for shutting down the event loop group; I don't see this one being shut down?

r2-netty/src/main/java/com/linkedin/r2/netty/client/http/HttpChannelPoolFactory.java
```diff
@@ -90,8 +96,19 @@ public HttpChannelPoolFactory(
     _tcpNoDelay = tcpNoDelay;
     _channelPoolWaiterTimeout = channelPoolWaiterTimeout;
 
-    _bootstrap = new Bootstrap().group(eventLoopGroup).channel(NioSocketChannel.class).
-        option(ChannelOption.CONNECT_TIMEOUT_MILLIS, connectTimeout).handler(initializer);
+    if (!StringUtils.isEmpty(_udsAddress)) {
```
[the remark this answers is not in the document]
jpstewart: Since `udsAddress` is only used in this constructor, this probably should just use the parameter and the field could be dropped.

r2-netty/src/main/java/com/linkedin/r2/netty/client/http/HttpChannelPoolFactory.java
```diff
@@ -96,9 +95,9 @@ public HttpChannelPoolFactory(
     _tcpNoDelay = tcpNoDelay;
     _channelPoolWaiterTimeout = channelPoolWaiterTimeout;
 
-    if (!StringUtils.isEmpty(_udsAddress)) {
+    if (!StringUtils.isEmpty(udsAddress)) {
       _bootstrap = new Bootstrap().
-          group(new EpollEventLoopGroup()).
+          group(eventLoopGroup).
           channel(EpollDomainSocketChannel.class).
```
jpstewart, line 101: Can this be simplified into a single bootstrap assignment, with a ternary for the socket channel class?

r2-netty/src/main/java/com/linkedin/r2/netty/client/http2/Http2ChannelPoolFactory.java
```diff
@@ -124,7 +124,7 @@ public Http2ChannelPoolFactory(
 
     if (!StringUtils.isEmpty(_udsAddress)) {
       _bootstrap = new Bootstrap().
-          group(new EpollEventLoopGroup()).
+          group(eventLoopGroup).
           channel(EpollDomainSocketChannel.class).
```
jpstewart, line 128: Same as above.

r2-netty/src/main/java/com/linkedin/r2/netty/client/http/HttpChannelPoolFactory.java
```diff
@@ -28,12 +28,16 @@
 import io.netty.channel.ChannelInitializer;
 import io.netty.channel.ChannelOption;
 import io.netty.channel.EventLoopGroup;
+import io.netty.channel.epoll.EpollDomainSocketChannel;
+import io.netty.channel.epoll.EpollEventLoopGroup;
```
jpstewart, line 32: Unused import.

r2-netty/src/main/java/com/linkedin/r2/netty/client/http2/Http2ChannelPoolFactory.java
jpstewart: In this case, where do we use the `EPollEventGroupLoop`? Is the change to inject that type where appropriate coming in a later PR, or being left to downstream consumers?

r2-netty/src/main/java/com/linkedin/r2/netty/client/http/HttpChannelPoolFactory.java
vigchan: This is a great point, that you called out about shutdown. Thank you. I have tried to address it two ways:
- Changes to the HttpClientFactory class, where the default would be `EpollEventLoopGroup`
- Added a change in `container` lib to pass in `EpolllEventLoopGroup` as well. This should hopefully address the problem. Please do let me know if I can handle this in a better way.

r2-netty/src/main/java/com/linkedin/r2/netty/client/http/HttpChannelPoolFactory.java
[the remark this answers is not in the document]
vigchan: This should hopefully be addressed now.

r2-netty/src/main/java/com/linkedin/r2/netty/client/http2/Http2ChannelPoolFactory.java
vigchan: It turns out that I do need to use `EpollEventLoopGroup`. Without this the Unix socket transport does not get instantiated properly. I have the container changes here -> https://linkedin.ghe.com/multiproduct/container/compare/uds-support?expand=1

jpstewart (approved):

jpstewart:
Looks like this just needs rebased and the version number and changelog updated.

vigchan:
> Looks like this just needs rebased and the version number and changelog updated.

Thank you so much. I think I updated the code now and hopefully there are no issues. I also have a test that keeps failing on local linux :
```
Gradle suite > Gradle test > com.linkedin.r2.transport.http.server.TestAsyncLockup > testAsyncLockup FAILED
    java.util.concurrent.TimeoutException at TestAsyncLockup.java:9
 ```
 Do I have to worry about this?

jpstewart:
> > Looks like this just needs rebased and the version number and changelog updated.
> 
> Thank you so much. I think I updated the code now and hopefully there are no issues. I also have a test that keeps failing on local linux :
> 
> ```
> Gradle suite > Gradle test > com.linkedin.r2.transport.http.server.TestAsyncLockup > testAsyncLockup FAILED
>     java.util.concurrent.TimeoutException at TestAsyncLockup.java:9
> ```
> 
> Do I have to worry about this?

I'll see if it also happens in the github test job. There are a number of flaky async tests.

jpstewart:
The tests are running fine on Github.

vigchan:
> The tests are running fine on Github.

Awesome! Thank you. Would you be able to merge the PR as well? I don't have the privileges to do so :)

jpstewart:
The branch currently needs rebased before merging.

PR #907, opened 2023-05-12
Implement rest.li xDS service discovery flow and DualRead mode loadbalancer

rickzx:
Implement rest.li xDS service discovery flow, including xDS client, xDS to D2 adaptor and event publisher. 
Implement DualRead mode load balancer.

Test:
- End-to-end test with internal test MP toki. PR link: (TBD)

TODO:
- Create unit tests
- Implement DualRead monitor system
- End-to-end test with LiX dynamic switch
- Integration test with @gu0keno0 's framework
- Handle resource un-subscription
- Documentations

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/LoadBalancer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/LoadBalancer.java
@@ -19,6 +19,7 @@ package com.linkedin.d2.balancer;
 import com.linkedin.common.callback.Callback;
 import com.linkedin.common.callback.FutureCallback;
 import com.linkedin.common.util.None;
+import com.linkedin.d2.balancer.dualread.DualReadModeProvider;
 import com.linkedin.d2.balancer.properties.ClusterProperties;
 import com.linkedin.d2.balancer.properties.ServiceProperties;
 import com.linkedin.d2.balancer.properties.UriProperties;
@@ -26,6 +27,7 @@ import com.linkedin.d2.discovery.event.PropertyEventThread.PropertyEventShutdown
 import com.linkedin.r2.message.Request;
 import com.linkedin.r2.message.RequestContext;
 import com.linkedin.r2.transport.common.bridge.client.TransportClient;
+import java.util.Map;
 import java.util.concurrent.ExecutionException;
 import org.apache.commons.lang3.tuple.Pair;
 
@@ -96,6 +98,11 @@ public interface LoadBalancer
   void start(Callback<None> callback);
 
   void shutdown(PropertyEventShutdownCallback shutdown);
+
+  default void updateDualReadStatus(Map<String, DualReadModeProvider.DualReadMode> dualReadStatus)
+  {
+    // Default to no-op
+  }
 
   // ################## Methods to deprecate Section ##################
 
--- a/d2/src/main/java/com/linkedin/d2/balancer/LoadBalancer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/LoadBalancer.java
@@ -19,12 +19,15 @@ package com.linkedin.d2.balancer;
 import com.linkedin.common.callback.Callback;
 import com.linkedin.common.callback.FutureCallback;
 import com.linkedin.common.util.None;
+import com.linkedin.d2.balancer.properties.ClusterProperties;
 import com.linkedin.d2.balancer.properties.ServiceProperties;
+import com.linkedin.d2.balancer.properties.UriProperties;
 import com.linkedin.d2.discovery.event.PropertyEventThread.PropertyEventShutdownCallback;
 import com.linkedin.r2.message.Request;
 import com.linkedin.r2.message.RequestContext;
 import com.linkedin.r2.transport.common.bridge.client.TransportClient;
 import java.util.concurrent.ExecutionException;
+import org.apache.commons.lang3.tuple.Pair;
 
 
 /**
@@ -82,6 +85,12 @@ public interface LoadBalancer
     {
       clientCallback.onError(e);
     }
+  }
+
+  default void getLoadBalancedClusterProperties(String clusterName,
+      Callback<Pair<ClusterProperties, UriProperties>> callback)
+  {
+    throw new UnsupportedOperationException();
   }
 
   void start(Callback<None> callback);
--- a/d2/src/main/java/com/linkedin/d2/balancer/LoadBalancer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/LoadBalancer.java
@@ -87,7 +87,7 @@ public interface LoadBalancer
     }
   }
 
-  default void getLoadBalancedClusterProperties(String clusterName,
+  default void getLoadBalancedClusterAndUriProperties(String clusterName,
       Callback<Pair<ClusterProperties, UriProperties>> callback)
   {
     throw new UnsupportedOperationException();
--- a/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
@@ -74,6 +74,9 @@ public class DualReadLoadBalancer implements LoadBalancerWithFacilities
   @Override
   public void start(Callback<None> callback)
   {
+    // Prefetch the global dual read mode
+    _dualReadStateManager.checkAndSwitchMode(null);
+
     _newLb.start(new Callback<None>()
     {
       @Override
@@ -94,9 +97,6 @@ public class DualReadLoadBalancer implements LoadBalancerWithFacilities
     // Call back will succeed as long as the old balancer is successfully started. New load balancer failure
     // won't block application start up.
     _oldLb.start(callback);
-
-    // Prefetch the global dual read mode
-    _dualReadStateManager.checkAndSwitchMode(null);
   }
 
   @Override
--- a/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
@@ -16,7 +16,6 @@
 
 package com.linkedin.d2.balancer.dualread;
 
-import com.google.common.util.concurrent.RateLimiter;
 import com.linkedin.common.callback.Callback;
 import com.linkedin.common.callback.Callbacks;
 import com.linkedin.common.util.None;
@@ -35,9 +34,7 @@ import com.linkedin.r2.message.Request;
 import com.linkedin.r2.message.RequestContext;
 import com.linkedin.r2.transport.common.TransportClientFactory;
 import com.linkedin.r2.transport.common.bridge.client.TransportClient;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.ScheduledExecutorService;
+import javax.annotation.Nonnull;
 import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -53,34 +50,24 @@ import org.slf4j.LoggerFactory;
  *
  * In OLD_LB_ONLY mode, it reads exclusively from the old load balancer.
  * In NEW_LB_ONLY mode, it reads exclusively from the new load balancer.
- * In DUAL_READ mode, it reads from both the old and the new load balancer, but relys on the data from old
+ * In DUAL_READ mode, it reads from both the old and the new load balancer, but relies on the data from old
  * load balancer only.
  */
-@SuppressWarnings("UnstableApiUsage")
 public class DualReadLoadBalancer implements LoadBalancerWithFacilities
 {
-  private static final Logger _log = LoggerFactory.getLogger(DualReadLoadBalancer.class);
+  private static final Logger LOG = LoggerFactory.getLogger(DualReadLoadBalancer.class);
   private final LoadBalancerWithFacilities _oldLb;
   private final LoadBalancerWithFacilities _newLb;
-  private final DualReadModeProvider _dualReadModeProvider;
-  private final ScheduledExecutorService _executorService;
-  private final RateLimiter _rateLimiter;
+  private final DualReadStateManager _dualReadStateManager;
 
-  // Stores service-level dual read mode
-  private final ConcurrentMap<String, DualReadModeProvider.DualReadMode> _serviceDualReadModes;
-  // Stores global dual read mode
-  private volatile DualReadModeProvider.DualReadMode _dualReadMode = DualReadModeProvider.DualReadMode.OLD_LB_ONLY;
   private boolean _isNewLbReady;
 
   public DualReadLoadBalancer(LoadBalancerWithFacilities oldLb, LoadBalancerWithFacilities newLb,
-      DualReadModeProvider dualReadModeProvider, ScheduledExecutorService executorService, int modeSwitchInterval)
+      @Nonnull DualReadStateManager dualReadStateManager)
   {
     _oldLb = oldLb;
     _newLb = newLb;
-    _dualReadModeProvider = dualReadModeProvider;
-    _executorService = executorService;
-    _rateLimiter = RateLimiter.create((double) 1 / modeSwitchInterval);
-    _serviceDualReadModes = new ConcurrentHashMap<>();
+    _dualReadStateManager = dualReadStateManager;
     _isNewLbReady = false;
   }
 
@@ -92,14 +79,14 @@ public class DualReadLoadBalancer implements LoadBalancerWithFacilities
       @Override
       public void onError(Throwable e)
       {
-        _log.warn("Failed to start new load balancer. Fall back to read from old balancer only", e);
+        LOG.warn("Failed to start new load balancer. Fall back to read from old balancer only", e);
         _isNewLbReady = false;
       }
 
       @Override
       public void onSuccess(None result)
       {
-        _log.info("New load balancer successfully started");
+        LOG.info("New load balancer successfully started");
         _isNewLbReady = true;
       }
     });
@@ -109,7 +96,7 @@ public class DualReadLoadBalancer implements LoadBalancerWithFacilities
     _oldLb.start(callback);
 
     // Prefetch the global dual read mode
-    _dualReadMode = _dualReadModeProvider.getDualReadMode();
+    _dualReadStateManager.checkAndSwitchMode(null);
   }
 
   @Override
@@ -122,31 +109,31 @@ public class DualReadLoadBalancer implements LoadBalancerWithFacilities
         _newLb.getClient(request, requestContext, clientCallback);
         break;
       case DUAL_READ:
-
         _newLb.getLoadBalancedServiceProperties(serviceName, new Callback<ServiceProperties>()
         {
           @Override
           public void onError(Throwable e)
           {
-            _log.error("Double read failure. Unable to read service properties from: " + serviceName, e);
+            LOG.error("Double read failure. Unable to read service properties from: " + serviceName, e);
           }
 
           @Override
           public void onSuccess(ServiceProperties result)
           {
             String clusterName = result.getClusterName();
+            _dualReadStateManager.updateCluster(clusterName, DualReadModeProvider.DualReadMode.DUAL_READ);
             _newLb.getLoadBalancedClusterAndUriProperties(clusterName, new Callback<Pair<ClusterProperties, UriProperties>>()
             {
               @Override
               public void onError(Throwable e)
               {
-                _log.error("Dual read failure. Unable to read cluster properties from: " + clusterName, e);
+                LOG.error("Dual read failure. Unable to read cluster properties from: " + clusterName, e);
               }
 
               @Override
               public void onSuccess(Pair<ClusterProperties, UriProperties> result)
               {
-                _log.debug("Dual read is successful. Get cluster and uri properties: " + result);
+                LOG.debug("Dual read is successful. Get cluster and uri properties: " + result);
               }
             });
           }
@@ -268,32 +255,6 @@ public class DualReadLoadBalancer implements LoadBalancerWithFacilities
     }
   }
 
-  /**
-   * Asynchronously check and update the dual read mode for the given D2 service
-   * @param d2ServiceName the name of the D2 service
-   */
-  public void checkAndSwitchMode(String d2ServiceName)
-  {
-    _executorService.submit(() ->
-    {
-      boolean shouldCheck = _rateLimiter.tryAcquire();
-      if (shouldCheck)
-      {
-        // Check and switch global dual read mode
-        _dualReadMode = _dualReadModeProvider.getDualReadMode();
-        _log.debug("Global dual read mode updated: " + _dualReadMode);
-
-        // Check and switch service-level dual read mode}
-        if (d2ServiceName != null)
-        {
-          DualReadModeProvider.DualReadMode mode = _dualReadModeProvider.getDualReadMode(d2ServiceName);
-          _serviceDualReadModes.put(d2ServiceName, mode);
-          _log.debug("Dual read mode for service " + d2ServiceName + " updated: " + mode);
-        }
-      }
-    });
-  }
-
   private boolean shouldReadFromOldLb()
   {
     DualReadModeProvider.DualReadMode dualReadMode = getDualReadMode();
@@ -308,8 +269,7 @@ public class DualReadLoadBalancer implements LoadBalancerWithFacilities
       return DualReadModeProvider.DualReadMode.OLD_LB_ONLY;
     }
 
-    checkAndSwitchMode(null);
-    return _dualReadMode;
+    return _dualReadStateManager.getDualReadMode();
   }
 
   private DualReadModeProvider.DualReadMode getDualReadMode(String d2ServiceName)
@@ -319,8 +279,7 @@ public class DualReadLoadBalancer implements LoadBalancerWithFacilities
       return DualReadModeProvider.DualReadMode.OLD_LB_ONLY;
     }
 
-    checkAndSwitchMode(d2ServiceName);
-    return _serviceDualReadModes.getOrDefault(d2ServiceName, _dualReadMode);
+    return _dualReadStateManager.getServiceDualReadMode(d2ServiceName);
   }
 
   @Override
@@ -328,7 +287,7 @@ public class DualReadLoadBalancer implements LoadBalancerWithFacilities
   {
     _newLb.shutdown(() ->
     {
-      _log.info("New load balancer successfully shut down");
+      LOG.info("New load balancer successfully shut down");
     });
 
     _oldLb.shutdown(callback);
--- a/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
@@ -1,0 +1,312 @@
+package com.linkedin.d2.balancer.dualread;
+
+import com.google.common.util.concurrent.RateLimiter;
+import com.linkedin.common.callback.Callback;
+import com.linkedin.common.callback.Callbacks;
+import com.linkedin.common.util.None;
+import com.linkedin.d2.balancer.Directory;
+import com.linkedin.d2.balancer.KeyMapper;
+import com.linkedin.d2.balancer.LoadBalancerWithFacilities;
+import com.linkedin.d2.balancer.properties.ClusterProperties;
+import com.linkedin.d2.balancer.properties.ServiceProperties;
+import com.linkedin.d2.balancer.properties.UriProperties;
+import com.linkedin.d2.balancer.util.ClusterInfoProvider;
+import com.linkedin.d2.balancer.util.LoadBalancerUtil;
+import com.linkedin.d2.balancer.util.hashing.HashRingProvider;
+import com.linkedin.d2.balancer.util.partitions.PartitionInfoProvider;
+import com.linkedin.d2.discovery.event.PropertyEventThread;
+import com.linkedin.r2.message.Request;
+import com.linkedin.r2.message.RequestContext;
+import com.linkedin.r2.transport.common.TransportClientFactory;
+import com.linkedin.r2.transport.common.bridge.client.TransportClient;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ScheduledExecutorService;
+import org.apache.commons.lang3.tuple.Pair;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+@SuppressWarnings("UnstableApiUsage")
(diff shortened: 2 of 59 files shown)


d2/src/test/java/com/linkedin/d2/xds/XdsToD2SampleClient.java
```diff
@@ -0,0 +1,96 @@
...
+      );
+    }
+
+    XdsChannelFactory xdsChannelFactory = new XdsChannelFactory(sslContext, xdsServer);
+    XdsClient xdsClient = new XdsClientImpl(node, xdsChannelFactory.createChannel(),
+        Executors.newSingleThreadScheduledExecutor());
+
+    XdsToD2PropertiesAdaptor adaptor = new XdsToD2PropertiesAdaptor(xdsClient);
+    adaptor.listenToService(serviceName);
+    adaptor.listenToCluster("TokiBackendGrpc");
```
rickzx, line 90: Will need to update this SampleClient later as a more generic example

gu0keno0:
cc: @jpstewart

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsLoadBalancerWithFacilitiesFactory.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+
+
+public class XdsLoadBalancerWithFacilitiesFactory implements LoadBalancerWithFacilitiesFactory
+{
+  @Override
+  public LoadBalancerWithFacilities create(D2ClientConfig config)
+  {
+    D2ClientJmxManager d2ClientJmxManager = new D2ClientJmxManager(config.d2JmxManagerPrefix, config.jmxManager);
```
gu0keno0, line 22: Do we need to give xDS data another Jmx prefix? How does it look like in Jmx UI?

d2/src/main/java/com/linkedin/d2/xds/balancer/DualReadXdsLoadBalancerFactory.java
```diff
@@ -0,0 +1,31 @@
...
+
+
+public class DualReadXdsLoadBalancerFactory implements LoadBalancerWithFacilitiesFactory
+{
+  private final LoadBalancerWithFacilitiesFactory _oldLbFactory;
+  private final LoadBalancerWithFacilitiesFactory _newLbFactory;
+  private final DualReadModeProvider _dualReadModeProvider;
+
+  public DualReadXdsLoadBalancerFactory(LoadBalancerWithFacilitiesFactory oldLbFactory,
+      LoadBalancerWithFacilitiesFactory newLbFactory, DualReadModeProvider dualReadModeProvider)
```
gu0keno0, line 18: Nit: for the context of this class, we can just create XdsLoadBalancerWithFacilitiesFactory as the newLbFactory and remove it from the constructor parameter?

d2/src/main/java/com/linkedin/d2/balancer/LoadBalancer.java
```diff
@@ -84,6 +87,12 @@ default void getLoadBalancedServiceProperties(String serviceName, Callback<Servi
     }
   }
 
+  default void getLoadBalancedClusterProperties(String clusterName,
```
bohhyang, line 90: "getLoadBalancedClusterAndUriProperties" to be more descriptive? same applies to the other load balancer classes in this PR.

d2/src/main/java/com/linkedin/d2/xds/balancer/DualReadXdsLoadBalancerFactory.java
```diff
@@ -0,0 +1,31 @@
...
+
+import com.linkedin.d2.balancer.D2ClientConfig;
+import com.linkedin.d2.balancer.LoadBalancerWithFacilities;
+import com.linkedin.d2.balancer.LoadBalancerWithFacilitiesFactory;
+import com.linkedin.d2.balancer.dualread.DualReadLoadBalancer;
+import com.linkedin.d2.balancer.dualread.DualReadModeProvider;
+import java.util.concurrent.Executors;
+
+
+public class DualReadXdsLoadBalancerFactory implements LoadBalancerWithFacilitiesFactory
```
bohhyang, line 11: If we want to make this factory general for any dual read situation in the future, we should remove the "Xds" part. However, I don't think we will have more dual read cases since restli clients should be deprecated by gRPC clients in near future. I think we can just name it particularly as "ZkAndXds", and name the new and old factory directly as xdsLbFactory and zkLbFactory.

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
```diff
@@ -0,0 +1,312 @@
...
+        _oldLb.getLoadBalancedClusterProperties(clusterName, callback);
+    }
+  }
+
+  @Override
+  public Directory getDirectory()
+  {
+    DualReadModeProvider.DualReadMode dualReadMode = getDualReadMode();
+    if (dualReadMode == DualReadModeProvider.DualReadMode.DUAL_READ
+        || dualReadMode == DualReadModeProvider.DualReadMode.OLD_LB_ONLY)
```
bohhyang, line 175: looks like this part can be in a small helper function "isNotReadingXdsOnly" since it's repeated quite many times.

d2/src/main/java/com/linkedin/d2/balancer/LoadBalancer.java
```diff
@@ -84,6 +87,12 @@ default void getLoadBalancedServiceProperties(String serviceName, Callback<Servi
     }
   }
```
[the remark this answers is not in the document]
rickzx: Adding all javadocs in the next commit

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
```diff
@@ -0,0 +1,312 @@
...
+        _oldLb.getLoadBalancedClusterProperties(clusterName, callback);
+        break;
+      case OLD_LB_ONLY:
+      default:
+        _oldLb.getLoadBalancedClusterProperties(clusterName, callback);
+    }
+  }
+
+  @Override
+  public Directory getDirectory()
```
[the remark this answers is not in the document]
rickzx: https://github.com/linkedin/rest.li/blob/3e5ae8ccdfb4477f3ba7e89d03d983978a3d0049/d2/src/main/java/com/linkedin/d2/balancer/Directory.java#L28

It returns a directory that can query all service and cluster names in ZK (just the name, not the data). I think we will need a separate INDIS registry observer API for this. I'll leave a TODO item for now.

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
```diff
@@ -0,0 +1,312 @@
...
+  public void checkAndSwitchMode(String d2ServiceName)
+  {
+    _executorService.submit(() ->
+    {
+      boolean shouldCheck = _rateLimiter.tryAcquire();
+      if (shouldCheck)
+      {
+        // Check and switch global dual read mode
+        _dualReadMode = _dualReadModeProvider.getDualReadMode();
+        _log.debug("Global dual read mode updated: " + _dualReadMode);
```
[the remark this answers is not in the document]
rickzx: The global `_dualReadMode` is updated the line above, right?

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
```diff
@@ -0,0 +1,312 @@
...
+
+        // Check and switch service-level dual read mode}
+        if (d2ServiceName != null)
+        {
+          DualReadModeProvider.DualReadMode mode = _dualReadModeProvider.getDualReadMode(d2ServiceName);
+          _serviceDualReadModes.put(d2ServiceName, mode);
```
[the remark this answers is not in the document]
rickzx: Yes, because all the update block is put inside a single-threaded executor. So we won't run into concurrency issues.

d2/src/main/proto/diskzk.proto
```diff
@@ -0,0 +1,41 @@
+syntax = "proto3";
+
+package indis;
+
+import "google/protobuf/struct.proto";
```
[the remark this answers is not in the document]
rickzx: Sounds good. Changed to XdsD2.proto

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancerMonitor.java
```diff
@@ -0,0 +1,76 @@
...
+  private Cache<String, CacheEntry<T>> buildCache()
+  {
+    return CacheBuilder.newBuilder()
+        .maximumSize(MAX_CACHE_SIZE)
+        // Invalidate cache entry after 30s to avoid overflow
+        .expireAfterWrite(CACHE_EXPIRATION_SECOND, TimeUnit.SECONDS)
+        .removalListener(notification ->
+        {
+          // If a cache entry is evicted due to TTL, print a WARN log because it only
+          // receives update from one source
```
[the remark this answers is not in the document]
rickzx: It's not always true. It's actually a good thing about guava Cache. It separates cache entry removal by causes: https://guava.dev/releases/19.0/api/docs/com/google/common/cache/RemovalCause.html.

At line 35, we manually remove the cache entry, so the removal cause is `EXPLICIT` and won't trigger the warning message. Only cache entry removal caused by `EXPIRED` due to TTL will trigger the warning

d2/src/main/java/com/linkedin/d2/balancer/properties/ServicePropertiesJsonSerializer.java
```diff
@@ -232,7 +232,7 @@ private ServiceProperties buildServicePropertiesFromMap(Map<String,Object> map)
     Map<String, Object> relativeStrategyProperties = mapGetOrDefault(map, PropertyKeys.RELATIVE_STRATEGY_PROPERTIES, Collections.emptyMap());
     boolean enableClusterSubsetting = map.containsKey(PropertyKeys.ENABLE_CLUSTER_SUBSETTING) ? PropertyUtil.coerce(
         map.get(PropertyKeys.ENABLE_CLUSTER_SUBSETTING), Boolean.class) : SubsettingStrategy.DEFAULT_ENABLE_CLUSTER_SUBSETTING;
-    int minClusterSubsetSize = map.containsKey(PropertyKeys.MIN_CLUSTER_SUBSET_SIZE) ? PropertyUtil.coerce(
+    Integer minClusterSubsetSize = map.containsKey(PropertyKeys.MIN_CLUSTER_SUBSET_SIZE) ? PropertyUtil.coerce(
```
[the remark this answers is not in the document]
rickzx: This is a really weird Java compilation thing. If i use primitive int, it always complain about Integer cannot be cast to int

d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancer.java
```diff
@@ -793,6 +794,67 @@ public void getLoadBalancedServiceProperties(String serviceName, boolean waitFor
     }
   }
 
+  @Override
+  public void getLoadBalancedClusterProperties(String clusterName,
```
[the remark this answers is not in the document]
rickzx: The getLoadBalancedServiceProperties is already existing, and is only used internally in rest.li. But both these API are good for double read, as it ensures listening on the resource

d2/src/main/java/com/linkedin/d2/xds/balancer/DualReadXdsLoadBalancerFactory.java
rickzx: Good point!

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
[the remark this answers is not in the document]
gu0keno0: I see, sounds good. INDIS Registry Observer supports "*" subscription, you can open a new stream and do a one-off read to get this effect.

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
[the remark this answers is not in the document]
gu0keno0: Oh yeah, you are right!

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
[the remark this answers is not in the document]
gu0keno0: AFAIK, updating the variable may or may not take effect immediately if we read it from another thread. "**volatile**" is made for this purpose, or not in Java?

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -0,0 +1,583 @@
...
+    _executorService.execute(() ->
+    {
+      _shutdown = true;
+      _log.info("Shutting down");
+      if (_adsStream != null) {
+        _adsStream.close(Status.CANCELLED.withDescription("shutdown").asException());
+      }
+    });
+
+    _executorService.shutdown();
```
gu0keno0, line 117: Nit: it may be better to let the owner to shutdown this executor. E.g. maybe let XdsLB class to own the thread, and let it close the executor when it is shutdown.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -0,0 +1,583 @@
...
+      if (_closed)
+      {
+        return;
+      }
+      _log.error("ADS stream closed with status {}: {}. Cause: {}", error.getCode(), error.getDescription(),
+          error.getCause());
+      _closed = true;
+      handleStreamClosed(error);
+      cleanUp();
+      if (_responseReceived || _retryBackoffPolicy == null) {
```
gu0keno0, line 554: Any particular reason to gate exponential backoff by _responseReceived? Or, should we just always do exp-backoff and reset it upon successful connection?

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -0,0 +1,583 @@
...
+{
+  private static final Logger _log = LoggerFactory.getLogger(XdsClientImpl.class);
+
+  private final Map<String, ResourceSubscriber> _d2NodeSubscribers = new HashMap<>();
+  private final Map<String, ResourceSubscriber> _d2NodeMapSubscribers = new HashMap<>();
+
+  private final Node _node;
+  private final ManagedChannel _managedChannel;
+  private final ScheduledExecutorService _executorService;
+  private final BackoffPolicy.Provider _backoffPolicyProvider = new ExponentialBackoffPolicy.Provider();
```
gu0keno0, line 59: Let's be explicit about the multiplier and the maximum: https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/ExponentialBackoffPolicy.java#L79 . So that we can fully control our client's behavior.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -0,0 +1,583 @@
...
+      ResourceUpdate resourceUpdate = entry.getValue();
+      ResourceSubscriber subscriber = getResourceSubscriberMap(type).get(resourceName);
+      if (subscriber != null)
+      {
+        subscriber.onData(resourceUpdate);
+      }
+    }
+  }
+
+  private void handleStreamClosed(Status error) {
```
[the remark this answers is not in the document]
bohhyang: nvm, saw the AdsStream logging it.

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadModeProvider.java
```diff
@@ -0,0 +1,17 @@
+package com.linkedin.d2.balancer.dualread;
+
+public interface DualReadModeProvider
+{
+  enum DualReadMode
+  {
+    OLD_LB_ONLY,
```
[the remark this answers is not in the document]
rickzx: I was hoping this LB will be a generic setting that can be used for any migration or dual-read purpose. The actual LB is `DualReadXdsLoadBalancerFactory`, and will set old lb = direct ZK and new lb = XDS.

I also updated Javadocs for all public classes. What do you think?

d2/src/main/proto/diskzk.proto
```diff
@@ -0,0 +1,41 @@
...
+
+option go_package = "golnkd.in/indis-registry-observer/diskzk/schema";
+
+message Stat {
```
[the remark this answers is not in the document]
rickzx: Yeah, as we discussed, it will be a challenge since rest.li is open-sourced.

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
```diff
@@ -0,0 +1,261 @@
...
+      public void onChanged(XdsClient.D2NodeUpdate update)
+      {
+        if (_serviceEventBus != null)
+        {
+          try
+          {
+            _serviceEventBus.publishInitialize(serviceName, toServiceProperties(update.getNodeData().getData()));
+          } catch (InvalidProtocolBufferException | PropertySerializationException e)
+          {
+            _log.error("Failed to parse D2 service properties from xDS update", e);
```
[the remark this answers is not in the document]
rickzx: Added the service name to the log. The json string is parsed directly from the proto message. If that step is throwing an exception, we're not able to print the json string at all.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -0,0 +1,583 @@
...
+import com.google.rpc.Code;
+import indis.XdsD2;
+import io.envoyproxy.envoy.service.discovery.v3.AggregatedDiscoveryServiceGrpc;
+import io.envoyproxy.envoy.service.discovery.v3.DeltaDiscoveryRequest;
+import io.envoyproxy.envoy.service.discovery.v3.DeltaDiscoveryResponse;
+import io.envoyproxy.envoy.service.discovery.v3.Resource;
+import io.grpc.ManagedChannel;
+import io.grpc.Status;
+import io.grpc.internal.BackoffPolicy;
+import io.grpc.internal.ExponentialBackoffPolicy;
```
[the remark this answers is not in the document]
rickzx: I think this one should be a stable API, because the behavior is well documented here: https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
rickzx: The responseReceived here is exactly trying to handle the backoff reset upon successful connection. The successful reconnection is defined as it has received a successful ADS response.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
rickzx: Yeah i agree. But it seems all the setters in this class are package-private (a lot of native gRPC code behaves like this, ugh..). What are you thoughts on copying this class over to our own code base?

rickzx: Oh actually, i think these parameters are part of the documented algorithm here: https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsTogglingLoadBalancerFactory.java
```diff
@@ -0,0 +1,180 @@
...
+    TogglingPublisher<UriProperties> uriToggle = new TogglingPublisher<>(uriPropertiesPublisher, fsUriStore, uriBus);
+
+    SimpleLoadBalancerState state =
+        new SimpleLoadBalancerState(executorService, uriBus, clusterBus, serviceBus, _clientFactories,
+            _loadBalancerStrategyFactories, _sslContext, _sslParameters, _isSSLEnabled, _partitionAccessorRegistry,
+            _sslSessionValidatorFactory, _deterministicSubsettingMetadataProvider, _canaryDistributionProvider);
+    _d2ClientJmxManager.setSimpleLoadBalancerState(state);
+
+    SimpleLoadBalancer balancer =
+        new SimpleLoadBalancer(state, _lbTimeout, _lbTimeoutUnit, executorService, _failoutConfigProviderFactory);
```
bohhyang, line 153: so in DualRead mode, the client will have two simpleLoadBalancerState (one for old zk, one for xds). Both actively create and refresh tracker clients when d2 service properties have updates. Since the xds one will not really be used to send a request under DualRead mode (only for comparing the data), can we have a check and only really create the simple load balancer and its state under NEW_LB_ONLY mode? especially given that we have seen issues with some clients having memory peaks when refreshing tracker clients.

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadModeProvider.java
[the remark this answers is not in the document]
bohhyang: looks good. thanks.

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancerMonitor.java
```diff
@@ -19,74 +19,195 @@
...
+    private final DualReadLoadBalancerJmx _dualReadLoadBalancerJmx;
+
+    public ClusterPropertiesDualReadMonitor(DualReadLoadBalancerJmx dualReadLoadBalancerJmx, Clock clock)
+    {
+      super(clock);
+      _dualReadLoadBalancerJmx = dualReadLoadBalancerJmx;
+    }
+
+    @Override
+    boolean compareEntries(CacheEntry<ClusterProperties> oldLbEntry, CacheEntry<ClusterProperties> newLbEntry)
```
bohhyang, line 138: "isEqual" is simpler. Same applies to the Service and Uri comparing functions too.

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadStateManager.java
```diff
@@ -0,0 +1,153 @@
...
+    LOG.debug("Dual read mode for service " + service + " updated: " + mode);
+  }
+
+  public void updateCluster(String cluster, DualReadModeProvider.DualReadMode mode)
+  {
+    _clusterDualReadModes.put(cluster, mode);
+    LOG.debug("Dual read mode for cluster " + cluster + " updated: " + mode);
+  }
+
+  public DualReadModeProvider.DualReadMode getDualReadMode()
```
bohhyang, line 69: nit: getGlobalDualReadMode to be clearer.

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
```diff
@@ -162,7 +167,10 @@ public void onChanged(XdsClient.D2NodeUpdate update)
         {
           try
           {
-            _clusterEventBus.publishInitialize(clusterName, toClusterProperties(update.getNodeData().getData()));
+            ClusterProperties clusterProperties = toClusterProperties(update.getNodeData().getData(),
+                update.getNodeData().getStat().getMzxid());
```
bohhyang, line 171: nice using the mzxid as the version!

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsTogglingLoadBalancerFactory.java
bohhyang: as discussed in the standup meeting, we can verify that the xDS LB state shares the existing connections that ZK LB state has, and minimal memory/CPU overhead is added (by enabling dual read on toki and checking its related metrics).

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
```diff
@@ -302,19 +276,138 @@ private ClusterProperties toClusterProperties(Struct clusterProperties, long ver
...
+    {
+      if (_eventEmitter == null)
+      {
+        _log.info("Service discovery event emitter in ZookeeperEphemeralStore is null. Skipping emitting events.");
+        return;
+      }
+
+      long timestamp = System.currentTimeMillis();
+
+      MapDifference<String, UriProperties> mapDifference = Maps.difference(_currentData, updates);
```
bohhyang, line 381: good to know!

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
```diff
@@ -302,19 +276,138 @@ private ClusterProperties toClusterProperties(Struct clusterProperties, long ver
...
+      _eventEmitter.emitSDStatusInitialRequestEvent(cluster, true, initialFetchDurationMillis, true);
+
     }
 
-    return _uriPropertiesMerger.merge(clusterName, allUriProperties);
+    private void emitSDStatusUpdateReceiptEvents(Map<String, UriProperties> updates)
+    {
+      if (_eventEmitter == null)
+      {
+        _log.info("Service discovery event emitter in ZookeeperEphemeralStore is null. Skipping emitting events.");
```
bohhyang, line 375: ZookeeperEphemeralStore should be XdsToD2PropertiesAdaptor, same applies to the other check for initial request.

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
```diff
@@ -302,19 +276,138 @@ private ClusterProperties toClusterProperties(Struct clusterProperties, long ver
...
+        {
+          _log.error("Failed to parse D2 uri properties from xDS update. Cluster name: " + _clusterName, e);
+        }
+      }
+    }
+
+    @Override
+    public void onError(Status error)
+    {
+      notifyAvailabilityChanges(false);
```
bohhyang, line 342: if the initial fetch failed, we need to emit the initialRequestEvent with failure, right?

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
```diff
@@ -302,19 +276,138 @@ private ClusterProperties toClusterProperties(Struct clusterProperties, long ver
...
+      {
+        emitSDStatusInitialRequestEvent(_clusterName);
+      }
+
+      if (_uriEventBus != null)
+      {
+        try
+        {
+          Map<String, UriProperties> updates = toUriProperties(update.getNodeDataMap());
+          emitSDStatusUpdateReceiptEvents(updates);
```
bohhyang, line 323: as we discussed in the standup, if this is from initial request response, we don't want to emit update receipt events to avoid the false spike in propagation latency (because the servers could be announced very long time ago before the client comes up).

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsTogglingLoadBalancerFactory.java
```diff
@@ -0,0 +1,181 @@
...
+/**
+ * A factory that creates a load balancer with xDS and file store toggles for service discovery
+ */
+public class XdsTogglingLoadBalancerFactory
+{
+  private static final Logger _log = LoggerFactory.getLogger(XdsTogglingLoadBalancerFactory.class);
+
+  private final long _lbTimeout;
+  private final TimeUnit _lbTimeoutUnit;
+  private final String _fsd2DirPath;
```
bohhyang, line 69: nit: _fsIndisDirPath

d2/src/main/java/com/linkedin/d2/balancer/util/downstreams/FSBasedDownstreamServicesFetcher.java
```diff
@@ -30,18 +33,34 @@
...
   {
+
     FileSystemDirectory fsDirectory = new FileSystemDirectory(_d2FsPath, _d2ServicePath);
-    callback.onSuccess(fsDirectory.getServiceNames());
+    Set<String> serviceNames = new HashSet<>(fsDirectory.getServiceNames());
+
+    if (_indisFsPath != null)
+    {
+      FileSystemDirectory indisFsDirectory = new FileSystemDirectory(_indisFsPath, _d2ServicePath);
+      serviceNames.addAll(indisFsDirectory.getServiceNames());
```
bohhyang, line 61: Maybe I missed something, would mixing the services in indis dir and the ones in d2 dir actually cause the d2 simple load balancer tries to warm up some service that only exists in the indis file? and vice versa?

Or even that happens, it doesn't matter (just fails and wait for the ZK/observer to connect)? If so, could you add a comment here to explain it?

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsTogglingLoadBalancerFactory.java
```diff
@@ -0,0 +1,181 @@
...
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLParameters;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * A factory that creates a load balancer with xDS and file store toggles for service discovery
+ */
+public class XdsTogglingLoadBalancerFactory
```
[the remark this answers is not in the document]
bohhyang: ping for this comment, since it does help devs navigate to this class (and understand its functionality better) if its name contains "FS".

d2/src/main/java/com/linkedin/d2/balancer/util/downstreams/FSBasedDownstreamServicesFetcher.java
rickzx: Sounds good, will use a separate warm up load balancer for the INDIS path

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancerJmx.java
```diff
@@ -0,0 +1,81 @@
+package com.linkedin.d2.balancer.dualread;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+
+public class DualReadLoadBalancerJmx implements DualReadLoadBalancerJmxMBean
```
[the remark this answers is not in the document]
rickzx: Yes, we only register it when dual read is enabled

bohhyang (approved):
the implementation LGTM! thank you for the big effort!

PR #910, opened 2023-05-26
INDIS: defining next generation announcements and wiring up to existing D2 announcements

ZoabKapoor:
This defines a next-generation service discovery announcement framework (called INDIS), separate from the Zookeeper framework, and wires it in to the existing D2 so that we double write to both systems

Testing done: unit test for `ZookeeperServer` modified to check that `markUp` and `markDown` calls `announce` and `deannounce` with the right parameters

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
@@ -20,6 +20,8 @@
 
 package com.linkedin.d2.balancer.servers;
 
+import com.linkedin.d2.discovery.event.IndisAnnouncer;
+import com.linkedin.d2.discovery.event.NoopIndisAnnouncer;
 import java.net.URI;
 import java.util.ArrayDeque;
 import java.util.Collections;
@@ -96,6 +98,8 @@ public class ZooKeeperAnnouncer implements D2ServiceDiscoveryEventHelper
 
   private ServiceDiscoveryEventEmitter _eventEmitter;
 
+  private final IndisAnnouncer _announcer;
+
   /**
    * Field that indicates if the user requested the server to be up or down. If it is requested to be up,
    * it will try to bring up the server again on ZK if the connection goes down, or a new store is set
@@ -154,11 +158,11 @@ public class ZooKeeperAnnouncer implements D2ServiceDiscoveryEventHelper
       boolean isDarkWarmupEnabled, String warmupClusterName, int warmupDuration, ScheduledExecutorService executorService)
   {
     this(server, initialIsUp, isDarkWarmupEnabled, warmupClusterName, warmupDuration, executorService,
-        new LogOnlyServiceDiscoveryEventEmitter()); // default to use log-only event emitter
+        new LogOnlyServiceDiscoveryEventEmitter(), new NoopIndisAnnouncer()); // default to use log-only event emitter
   }
 
   public ZooKeeperAnnouncer(ZooKeeperServer server, boolean initialIsUp,
-      boolean isDarkWarmupEnabled, String warmupClusterName, int warmupDuration, ScheduledExecutorService executorService, ServiceDiscoveryEventEmitter eventEmitter)
+      boolean isDarkWarmupEnabled, String warmupClusterName, int warmupDuration, ScheduledExecutorService executorService, ServiceDiscoveryEventEmitter eventEmitter, IndisAnnouncer announcer)
   {
     _server = server;
     // initialIsUp is used for delay mark up. If it's false, there won't be markup when the announcer is started.
@@ -174,6 +178,7 @@ public class ZooKeeperAnnouncer implements D2ServiceDiscoveryEventHelper
     _warmupDuration = warmupDuration;
     _executorService = executorService;
     _eventEmitter = eventEmitter;
+    _announcer = announcer;
 
     _server.setServiceDiscoveryEventHelper(this);
   }
--- a/d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
@@ -20,8 +20,6 @@
 
 package com.linkedin.d2.balancer.servers;
 
-import com.linkedin.d2.discovery.event.IndisAnnouncer;
-import com.linkedin.d2.discovery.event.NoopIndisAnnouncer;
 import java.net.URI;
 import java.util.ArrayDeque;
 import java.util.Collections;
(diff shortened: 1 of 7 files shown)


d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -268,6 +268,7 @@ public synchronized void markUp(final Callback<None> callback)
 
   private synchronized void doMarkUp(Callback<None> callback)
   {
+    _indisAnnouncer.emitAnnouncement(_cluster, _uri.getHost(), _uri.getPort(), IndisAnnouncer.HealthStatus.READY);
```
bohhyang, line 271: Here you don't have the uri properties yet so the announcement can't happen. It should happen in the ephemeralStore class where all the complex logic about checks, warmups, building uriProperties, etc, have been done (in the ZooKeeperAnnouncer and ZooKeeperServer).
(Here the markUp could happen to warmupCluster too, please take a look at the warmup-related code in this class. So using _cluster here is not right. If you do the announcement in the ephemeralStore class, the exact cluster/prop name will be passed into it, so you don't need to worry about it.)

ZoabKapoor: Makes sense, I have pulled the IndisAnnouncer down into ZookeeperServer which, as you say, actually has all the info we need to make an announcement. Thanks for the pointer!

jpstewart: What is the feasibility of implementing an alternative server to `ZooKeeperServer`? It doesn't feel right to have the `IndisAnnouncer`, which is itself an abstraction over any underlying technology, to be used from a ZooKeeper-specific concrete implementation of `LoadBalancerServer`.

bohhyang: we're just piggy-backing zookeeper logic for **triggering the announcement** for now just to keep the announcement behaviors the same as ZK in the transition phase (rolling out INDIS on top of ZK d2 data). After the transition phase, we'll deprecate the whole ZookeeperAnnouncer -- ZookeeperServer -- ZKStore system and build a INDIS readiness manager in container which manages the data and timing of the announcement and integrates with INDISAnnouncer.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperServer.java
```diff
@@ -131,7 +147,6 @@ public void onSuccess(None none)
           info(_log, sb);
         }
         _store.put(clusterName, new UriProperties(clusterName, partitionDesc, myUriSpecificProperties), callback);
```
bohhyang, line 149: we want to do the Kafka announce as close to the ZK announcement so that we can compare the propagation time between INDIS control plane and ZK-architecture fairly. So the announcement should happen either right before this line.

ZoabKapoor: I thought we had discussed this previously -- didn't we decide that we want the Indis announce separate from the ZK announcement callback? I had originally had the announcement next to the store.put, but moved it out because I thinkthe `store.put` is in a callback. 

What do you think? I feel like making the ZK and Kafka announcements independent may be more important than keeping the logic close together, which means the Kafka announcement has to be either before or after the callback

I could also move it from before to after, if that would help!

ZoabKapoor: Discussed offline with @bohhyang and we agreed that keeping the Kafka announcement independent from the ZK announcement is more important

I have added a comment mentioning why this code is separate from the ZK announcement so future readers will know

d2/src/test/java/com/linkedin/d2/balancer/servers/ZooKeeperServerTest.java
```diff
@@ -184,6 +192,13 @@ public void testZkServer() throws Exception
     assertEquals(properties.getUriSpecificProperties().size(), 1);
     assertEquals(properties.getUriSpecificProperties().get(URI_2), _uri2SpecificProperties);
 
+    UriProperties propertiesForUri2 = new UriProperties(properties.getClusterName(),
```
ZoabKapoor, line 195: When we announce, we only have access to the properties for *this URI specifically* which will normally be different from the entire properties we load from D2 (as this will contain data for all URIs in the cluster)

One consequence of this is that whichever entity _consumes_ the announcements and deannouncements will need to also manage merging the `UriProperties` for a cluster from the fragments passed by each announcement

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperServer.java
```diff
@@ -47,18 +49,32 @@
...
+  private IndisAnnouncer _indisAnnouncer = new NoopIndisAnnouncer();
+
   private volatile ZooKeeperEphemeralStore<UriProperties> _store;
   private D2ServiceDiscoveryEventHelper _eventHelper = null;
 
   public ZooKeeperServer()
   {
   }
 
+  public ZooKeeperServer(IndisAnnouncer indisAnnouncer) {
```
bohhyang, line 63: how could we construct this class without initializing the ZooKeeperEphemeralStore?

ZoabKapoor: In the `D2ServerFactory` in container we use the no-argument `ZookeeperServer` constructor, which is where we will be we initializing the IndisAnnouncer implementation. I think the reason it works is because somewhere in the code they use `setStore` to assign the store before the D2Server actually starts

d2/src/test/java/com/linkedin/d2/balancer/servers/ZooKeeperServerTest.java
ZoabKapoor: Sounds good, glad to hear this won't be an issue!

d2/src/main/java/com/linkedin/d2/discovery/event/IndisAnnouncer.java
```diff
@@ -0,0 +1,19 @@
...
+ * It will be implemented internally, and we provide a default implementation {@link NoopIndisAnnouncer}
+ * that open-source restli users can use (or provide their own implementation for their own service discovery system)
+ */
+public interface IndisAnnouncer {
+  void start();
+  void stop();
+  void announce(String cluster, String protocol, String host, int port, String path, Map<Integer, PartitionData> partitionDataMap,
+      Map<String, Object> uriSpecificProperties, UriProperties d2UriProperties);
+
+  void deannounce(String cluster, String protocol, String host, int port, String path);
```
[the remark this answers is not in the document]
ZoabKapoor: Yes, exactly. I don't think we have access to these parameters in the `markDown` function, so we can't easily pass them, but in our implementation in container we can use the existing data from cache rather than starting from scratch. 

This is the relevant change in container to enable that: https://linkedin.ghe.com/multiproduct/container/pull/135/commits/0ffefa26f9863c0adac20a24be06a66c98cc7aab

PapaCharlie (commented):
Publishing this review but upon further discussion with Zoab, this may not be necessary

ZoabKapoor:
As discussed in team meeting today, we are abandoning this approach in favor of `extending ZookeeperServer` internally and putting custom logic in there. The only thing that needs to be done in rest.li is factoring some common logic out into a `protected` function. That is a different enough change that I will create a new PR for it

PR #911, opened 2023-06-02
Add log message in RestClient for ScatterGatherStrategy map URIs empty case

nickibi:
Add log message in RestClient for ScatterGatherStrategy map URIs empty case.

Based on the code path this situation should not happen. 
Adding log message for debugging purpose if such rare case occurs.

### DIFF
--- a/restli-client/src/main/java/com/linkedin/restli/client/RestClient.java
+++ b/restli-client/src/main/java/com/linkedin/restli/client/RestClient.java
@@ -81,6 +81,8 @@ import java.util.Random;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 import javax.activation.MimeTypeParseException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import static com.linkedin.r2.disruptor.DisruptContext.*;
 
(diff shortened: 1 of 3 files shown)


restli-client/src/main/java/com/linkedin/restli/client/RestClient.java
```diff
@@ -1108,6 +1111,9 @@ public void onSuccess(ProtocolVersion protocolVersion)
         if (mappingResults == null || mappingResults.getMappedKeys().isEmpty())
         {
           // Strategy returns null URIMappingResult or empty mapped hosts, assuming no scatter is needed
+          LOG.error("ScatterGatherStrategy cannot map URIs for request: " + request
+              + ", requestContext: " + requestContext
+              + ", ScatterGatherStrategy needScatterGatherStrategy value: " +  strategy.needScatterGather(request));
```
kk415kk, line 1116: Is it possible for `strategy` to be null?

kk415kk (approved):
LGTM. Thanks for adding this!

junchuanwang (approved):

restli-client/src/main/java/com/linkedin/restli/client/RestClient.java
nickibi: I don't think strategy can be null. Look at line 1100, if strategy is null, it would not reach here.

PR #915, opened 2023-06-22
Refactor ZookeeperServer: put uri property building logic to a method

ZoabKapoor:
## What does this do?

This refactors ZookeeperServer, moving the functionality to generate a `UriProperties` object into a protected helper function. This makes the code slightly cleaner and also enables any class that extends `ZookeeperServer` to have access to every parameter that `ZookeeperServer` writes to ZK.

## Testing done
`ZooKeeperServerTest` continues to pass, and I've confirmed that the function gets full code coverage. We could also add unit tests specifically for this, though I've heard that [testing protected methods is an anti-pattern](https://stackoverflow.com/questions/5601730/should-private-protected-methods-be-under-unit-test)? 

Screenshot 2023-06-22 at 12 02 04 PM

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperServer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperServer.java
@@ -96,19 +96,6 @@ public class
       @Override
       public void onSuccess(None none)
       {
-        Map<URI, Map<Integer, PartitionData>> partitionDesc = new HashMap<>();
-        partitionDesc.put(uri, partitionDataMap);
-
-        Map<URI, Map<String, Object>> myUriSpecificProperties;
-        if (uriSpecificProperties != null && !uriSpecificProperties.isEmpty())
-        {
-          myUriSpecificProperties = new HashMap<>();
-          myUriSpecificProperties.put(uri, uriSpecificProperties);
-        }
-        else
-        {
-          myUriSpecificProperties = Collections.emptyMap();
-        }
 
         if (_log.isInfoEnabled())
         {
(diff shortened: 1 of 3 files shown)


d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperServer.java
```diff
@@ -372,6 +359,22 @@ public void shutdown()
...
+      uriToUriSpecificProperties = new HashMap<>();
+      uriToUriSpecificProperties.put(uri, uriSpecificProperties);
+    }
+    else
+    {
+      uriToUriSpecificProperties = Collections.emptyMap();
+    }
+    return new UriProperties(clusterName, partitionDesc, uriToUriSpecificProperties);
+  }
   private void storeGet(final String clusterName, final Callback<UriProperties> callback)
```
bohhyang, line 378: blank line between the functions

bohhyang (approved):
lgtm. Only a nit: the commit message "Refactor ZookeeperServer" sounds like a huge scary change. Maybe be specific, like "Put uri property building logic to a method", is better.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperServer.java
```diff
@@ -372,6 +359,23 @@ public void shutdown()
     }
   }
 
+  protected UriProperties constructUriPropertiesForNode(final String clusterName, final URI uri, final Map<Integer, PartitionData> partitionDataMap, final Map<String, Object> uriSpecificProperties) {
```
PapaCharlie, line 362: Can you break this long line so it doesn't go past 120 chars?

PR #921, opened 2023-07-08
Fix list tag in Dynamic_Discovery.html

RonWang:
Currently the page is not displaying the list
```
When getClient is called on the simple load balancer, it: <ul> <li>First tries to extract the service name from the URI that was provided. </li> <li> It then makes sure that it’s listening to that service in the LoadBalancerState. </li> <li> It then makes sure that it’s listening to the service’s cluster in the LoadBalancerState </li> <li> If either the service or cluster is unknown, it will throw a ServiceUnavailableException. </li> <li> It will then iterate through the prioritized schemes (prpc, http, etc) for the cluster. </li> <li> For each scheme, it will get all URIs in the service’s cluster for that scheme, and ask the service’s load balancer strategy to load balance them.</li> <li> If the load balancer strategy returns a client, it will be returned, otherwise the next scheme will be tried.</li> <li>If all schemes are exhausted, and no client was found, a ServiceUnavailableException will be thrown. </li> </ul>
```

This commit fixes it now:
```
When getClient is called on the simple load balancer, it:

First tries to extract the service name from the URI that was provided.
It then makes sure that it’s listening to that service in the LoadBalancerState.
It then makes sure that it’s listening to the service’s cluster in the LoadBalancerState
If either the service or cluster is unknown, it will throw a ServiceUnavailableException.
It will then iterate through the prioritized schemes (prpc, http, etc) for the cluster.
For each scheme, it will get all URIs in the service’s cluster for that scheme, and ask the service’s load balancer strategy to load balance them.
If the load balancer strategy returns a client, it will be returned, otherwise the next scheme will be tried.
If all schemes are exhausted, and no client was found, a ServiceUnavailableException will be thrown.
```


ChrisCarini:

PR #923, opened 2023-07-12
[rest.li] Update SimpleLoadBalancer to use for loop instead of Map

vshwnth2:
[rest.li] Update SimpleLoadBalancer to use for loop instead of Map

When running ODP analysis, we observed that SimpleLoadBalancer.getPotentialClients map collect uses around 4% CPU. Much of this seems Collector.toMap's `uniqKeysMapAccumulator`. In this scenario, this is performing unnecessary extra computation -- the keys of the Map we are creating come directly from a Set.

The updated implementation avoids the `stream(...).collect(Collectors.toMap(...)` and directly uses a for loop to populate the Map

Screenshot 2023-07-11 at 8 55 38 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
+## [29.48.0] - 2023-11-13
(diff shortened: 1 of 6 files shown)


PapaCharlie (approved):
LGTM! Thank you for the change

PapaCharlie:
You will need to rerun the checks, they can be occasionally flaky

PR #924, opened 2023-07-19
updated PegasusPlugin to still register restModel tasks even if IDL files dir is empty

aman1309:
Details: Allow creating rest model gradle tasks even if there are no IDL files are present at project evaluation time if `rest.idl.processEmptyIdlDir` is set to true for the project.

###Testing Done 
build

### DIFF
--- a/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
+++ b/gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
@@ -562,6 +562,7 @@ public class PegasusPlugin implements Plugin<Project>
   private static final Pattern TEST_DIR_REGEX = Pattern.compile("^(integ)?[Tt]est");
   private static final String SNAPSHOT_NO_PUBLISH = "rest.model.noPublish";
   private static final String SNAPSHOT_FORCE_PUBLISH = "rest.model.forcePublish";
+  private static final String IDL_AUTO_GENERATION_ENABLED = "rest.idl.autoGeneration.enabled";
   private static final String IDL_NO_PUBLISH = "rest.idl.noPublish";
   private static final String IDL_FORCE_PUBLISH = "rest.idl.forcePublish";
   private static final String SKIP_IDL_CHECK = "rest.idl.skipCheck";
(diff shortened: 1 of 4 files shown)


gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -1991,10 +1991,6 @@ protected void configureRestClientGeneration(Project project, SourceSet sourceSe
   {
     // idl directory for api project
     File idlDir = project.file(getIdlPath(project, sourceSet));
-    if (SharedFileUtils.getSuffixedFiles(project, idlDir, IDL_FILE_SUFFIX).isEmpty())
-    {
-      return;
```
junchuanwang, line 1996: will there be downsides for packing an empty jar?

why will you need a jar that is empty? Can your logic assume it empty when it doesn't exist?

aman1309: no downsides or compatibility issues. just some negligible space will be used for PDL only projects to publish empty jars.
We will never use the empty jar. only the jars with IDL/ restmodels will be used by clients.

junchuanwang:
Discussed offline, so the empty jar published seems to be an unfortunate and inevitable side-effect if we use these changes to make these restModel tasks configured.

Publishing empty jar should be backward compatible in restli sense.

I give ship as we assessed this won't impact existing user.

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
aman1309: updated PR as per offline discussion @junchuanwang @mchen07

gradle-plugins/src/main/java/com/linkedin/pegasus/gradle/PegasusPlugin.java
```diff
@@ -562,6 +562,7 @@ public class PegasusPlugin implements Plugin<Project>
   private static final Pattern TEST_DIR_REGEX = Pattern.compile("^(integ)?[Tt]est");
   private static final String SNAPSHOT_NO_PUBLISH = "rest.model.noPublish";
   private static final String SNAPSHOT_FORCE_PUBLISH = "rest.model.forcePublish";
+  private static final String IDL_AUTO_GENERATION_ENABLED = "rest.idl.autoGeneration.enabled";
```
junchuanwang, line 565: just one minor comment otherwise looks good to me.

Why this word "autoGeneration" ? I cannot understand only by the term itself. I suggest either change to something more intuitive, or add a comment here.

aman1309: updated the config to processEmptyIdlDir

junchuanwang (approved):

PR #926, opened 2023-07-28
Dynamically switch jmx/sensor names based on dual read mode and source type

bohhyang:
### Background
Currently Jmx/sensors (load balancer, load balancer state, file stores, clusterInfo, serviceProperties, LoadBalancerStrategy, etc) are having the same name in ZK and xDS read flow. The current code intended to give [D2ClientJmxManager.java](https://github.com/linkedin/rest.li/pull/907/files#diff-9993e287812781a541ea17425748df60fe6966b7a77ea69eba72fcf8a5890102) a different prefix in xDS flow ("-xDS"), which had a bug in container and wasn't working. Even if worked, it would still break the jmx/sensors because depending on the dynamically changing dual read mode value, the primary source (ZK in OLD_LB_ONLY and DUAL_READ modes; xDS in NEW_LB_ONLY mode) should register the jmx/sensor names same as today, so that users can still monitor the same metrics. And, the secondary source should register different names to avoid conflicting with the primary.

### Changes
1. added watchers to dual read mode and dynamically register jmx/sensor names for primary/secondary sources under the new dual read mode.
2. Include the source type name ("-xDS", "-ZK") in the secondary source's jmx/sensor names.
3. bumped major version. Will create a container PR to include this change to container. Since the previous 29.43.xxx versions will have broken Jmx/sensor names if switched to observer-only read mode, I'll deprecate those versions once this is released.

### Test Done
1. added unit tests for add/remove watchers and registering jmx/sensor names.
2. QEI deployment with Toki and switch dual read modes with lix:
Under dual read or zk-only modes, the ZK source is primary, and xDS source registered the names with "-xDS":
Screenshot 2023-08-02 at 5 24 03 PM

Under observer-only mode, app log shows the primary names are unregistered and re-registered (done by xDS source), and the "-ZK" names are registered (by ZK source):
Screenshot 2023-08-06 at 4 58 19 PM
, and the jmx/sensor names also show the same:
Screenshot 2023-08-06 at 5 25 08 PM

Note: the "-xDS" names are left there (not unregistered), since they won't hurt during the dual read testing and will be cleaned when the app adopts xDS load balancer (by config, so dual read load balancer is not used at all) and redeploys.

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadStateManager.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadStateManager.java
@@ -26,6 +26,7 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ScheduledExecutorService;
+import javax.annotation.Nonnull;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -89,7 +90,7 @@ public class DualReadStateManager
     _dualReadMode = mode;
     if (updated) {
       LOG.info("Global dual read mode updated: {}", mode);
-      notifyGlobalWatchers();
+      notifyGlobalWatchers(_dualReadMode);
     }
   }
 
@@ -98,7 +99,7 @@ public class DualReadStateManager
     DualReadModeProvider.DualReadMode oldMode = _serviceDualReadModes.put(service, mode);
     if (oldMode != mode) {
       LOG.info("Dual read mode for service {} updated: {}", service, mode);
-      notifyServiceWatchers(service);
+      notifyServiceWatchers(service, mode);
     }
   }
 
@@ -107,7 +108,7 @@ public class DualReadStateManager
     DualReadModeProvider.DualReadMode oldMode = _clusterDualReadModes.put(cluster, mode);
     if (oldMode != mode) {
       LOG.info("Dual read mode for cluster {} updated: {}", cluster, mode);
-      notifyClusterWatchers(cluster);
+      notifyClusterWatchers(cluster, mode);
     }
   }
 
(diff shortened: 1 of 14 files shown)


jpstewart (approved):

d2/src/main/java/com/linkedin/d2/jmx/D2ClientJmxManager.java
```diff
@@ -129,17 +129,17 @@ public void onServicePropertiesRemoval(LoadBalancerStateItem<ServiceProperties>
...
       {
-        return String.format("%s-ServiceProperties", serviceName);
+        return String.format("%s-%s-ServiceProperties", _prefix, serviceName);
       }
 
       private String getLoadBalancerStrategyJmxName(String serviceName, String scheme)
       {
-        return serviceName + "-" + scheme + "-LoadBalancerStrategy";
+        return String.format("%s-%s-%s-LoadBalancerStrategy", _prefix, serviceName, scheme);
       }
```
PapaCharlie, line 143: !!!!!
This will break existing alerts! You need to preserve the old metric, otherwise it could break the monitoring for certain apps.

PapaCharlie (changes requested):

d2/src/main/java/com/linkedin/d2/jmx/D2ClientJmxManager.java
bohhyang: as we discussed, used the prefix only in xDS LB metric names to keep the existing ZK LB metric names intact for backward compatibility (even though that was a bug before which should've used the prefix).

shivamgupta1 (approved):

ZoabKapoor (approved):

d2/src/test/java/com/linkedin/d2/jmx/D2ClientJmxManagerTest.java
```diff
@@ -17,165 +17,440 @@
...
 import org.testng.annotations.Test;
 
 import static org.mockito.Mockito.*;
 
 
 public class D2ClientJmxManagerTest {
-  @Mock
-  SimpleLoadBalancerState _simpleLoadBalancerState;
-  @Mock
-  JmxManager _jmxManager;
```
bohhyang, line 45: I've put these mocks into a fixture because direct mock member vars inside the class can't be used in a data provider. For example, when I create different D2ClientJmxManager with different params but use the same _jmxManager mock:

`@DataProvider
public Object[][] dataProvider()
{
   return new Object[][]
   {
      {new D2ClientJmxManager(..., _jmxManager)}
      {new D2ClientJmxManager(..., _jmxManager, ...ZK, null)}
      {new D2ClientJmxManager(..., _jmxManager, ....xDS, null)}
   };
}`
Using this data provider in a test method (such as `testSetSimpleLBStateListenerUpdateClusterInfo`), the test will fail with errors saying the jmxManager is null.

d2/src/main/java/com/linkedin/d2/jmx/D2ClientJmxManager.java
bohhyang: updated the names based on primary and secondary source according to the dual read mode.

bohhyang:
discussed with @PapaCharlie over slack, it's better to separate out the watchers for different property types into separate maps or member vars, so it's less error-prong for future devs to add watchers for new types of properties. I did so and also moved the watcher management logic to a separate manager class for cleanlyness.

d2/src/test/java/com/linkedin/d2/jmx/D2ClientJmxManagerTest.java
PapaCharlie: Seems fine by me, it's easier to pass in one object than 10 !

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadStateManager.java
```diff
@@ -198,4 +208,70 @@ public DualReadModeProvider getDualReadModeProvider()
   {
     return _dualReadModeProvider;
   }
+
```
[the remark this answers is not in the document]
bohhyang: added

PapaCharlie (approved):

PR #929, opened 2023-08-11
Rewrite the JavaDoc logic in Java 11 APIs and use multi-release jar to ensure backward compatibility

JoeJoe1989:
**Background**

Currently, RestLi uses Java 8 APIs, such as com.sun.javadoc, to generate JavaDoc. However, these APIs have been removed in Java 17, making RestLi incompatible with Java 17. This PR aims to resolve the issue by rewriting the JavaDoc related logic using [Java 11 APIs](https://docs.oracle.com/en/java/javase/11/docs/api/jdk.javadoc/jdk/javadoc/doclet/package-summary.html#migration), which are compatibility with JDK 17.

Furthermore, since RestLi has a mix of Java 8 and Java 11 consumers and we are using Java 11 APIs, we will utilize a multi-release JAR to ensure backward compatibility with Java 8 consumers. On a high level, a multi-release JAR compiles the code using Java 11 and sets the target version to either Java 11 or Java 8, depending on whether Java 11 APIs are used or not. As a result, the JAR contains different versions of bytecode and is compatible with different versions of consumers.

**Changes**

1. Rewrite the JavaDoc logic in Java 11 APIs.
2. Use multi-release jar to ensure compatibility with Java 8, Java 11 and Java 17 consumers.
3. Change the GitHub workflow action to use Java 11 to compile RestLi.

**Tests Done**

Build in Java 11:
```
export JAVA_HOME=/Library/Java/JavaVirtualMachines/jdk11.0.13.8-msft.jdk/Contents/Home/
./gradlew build
```

I tested the build process of 16 consumers who are using Java 8, Java 11, and Java 17 against the local snapshot of RestLi.

I also ran a wc-test of container against RestLi and noticed some IDL changes caused by the difference between Java 8 and Java 11 APIs in terms of handling comments in unexpected formats in source files. For example, there is an extra comma after the parameter name `count`:

`* @param count, the number of the recommendations`

Java 8 APIs are not able to handle this case, so no JavaDoc is generated for `count`. However, Java 11 APIs can handle it and generate JavaDoc as follows:

`"doc" : ", the number of the recommendations"`"

Consumers who encounter such IDL changes during build can commit the changes as normal. Ideally, they should fix the comments in unexpected formats in source files.


gradle-plugins/src/integTest/groovy/com/linkedin/pegasus/gradle/PegasusPluginCacheabilityTest.groovy
```diff
@@ -96,6 +96,6 @@ class PegasusPluginCacheabilityTest extends Specification {
     preparedSchema.exists()
 
     where:
-    gradleVersion << [ '4.0', '5.2.1', '5.6.4', '6.9', '7.0.2' ]
+    gradleVersion << [ '5.2.1', '5.6.4', '6.9', '7.0.2' ]
```
junchuanwang, line 99: why moved `4.0`?

gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=29.45.1
+version=29.46.0
```
junchuanwang, line 1: can you mention how many consumer have you tesetd in the test description? Now I only see "Tested the build process of Java 8, Java 11 and Java 17 consumers against the local snapshot of RestLi."

restli-tools/src/main/java11/com/linkedin/restli/tools/idlgen/DocletDocsProvider.java
```diff
@@ -0,0 +1,267 @@
+/*
+   Copyright (c) 2012 LinkedIn Corp.
```
junchuanwang, line 2: as a matter of fact, since this is newly written, let's change the license and public doc

restli-tools/src/main/java11/com/linkedin/restli/tools/idlgen/DocletDocsProvider.java
```diff
@@ -0,0 +1,267 @@
...
+import javax.lang.model.element.AnnotationMirror;
+import javax.lang.model.element.AnnotationValue;
+import javax.lang.model.element.Element;
+import javax.lang.model.element.ExecutableElement;
+import javax.lang.model.element.TypeElement;
+import javax.lang.model.element.VariableElement;
+
+
+/**
+ * Specialized {@link DocsProvider} whose documentation comes from the Javadoc Doclet {@link RestLiDoclet}.
```
junchuanwang, line 58: similarily, just refer to the java 8 counterpart and mention this new file is written in java 11 api, referring the logic on those files

gradle-plugins/src/integTest/groovy/com/linkedin/pegasus/gradle/PegasusPluginCacheabilityTest.groovy
JoeJoe1989: Since we will be compiling RestLi in Java 11, we need to use Gradle 5.0 or higher, which is the minimum version required to support Java 11. As a result, I removed Gradle 4.0. This change was confirmed with @njuliyu from the Gradle Team, who added this test.

gradle.properties
JoeJoe1989: Added.

restli-tools/src/main/java11/com/linkedin/restli/tools/idlgen/DocletDocsProvider.java
JoeJoe1989: Done.

restli-tools/src/main/java11/com/linkedin/restli/tools/idlgen/DocletDocsProvider.java
JoeJoe1989: Done.

PR #931, opened 2023-08-26
Downgrade major version to 29

zackthehuman:
Bumping the major version to 30 is technically the correct thing to do from a semver perspective when a public API is removed. However, it also has a huge impact that outweighs the API breakage in terms of propogating the change to all consumers. We feel like this tradeoff is acceptable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,12 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.45.0] - 2023-08-25
+
+- Downgrade major version back to 29. Technically this is not semver-compatible
+  but we feel that the impact should be less severe than the impact of bumping
+  the major version.
+
 ## [30.0.0] - 2023-08-15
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,12 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.45.0] - 2023-08-25
+
+- Downgrade major version back to 29. Technically this is not semver-compatible
+  but we feel that the impact should be less sever than the impact of bumping
```
mchen07, line 20: less severe?

zackthehuman: Good catch. Fixed.

ZoabKapoor (approved):

ZoabKapoor (approved):
Looks good, thanks for doing this! As mentioned in the release notes, this is definitely not SemVer compatible as it's a version decrement, but it seems like the safest way to undo the `30.0.0` change which was made in error as we don't want to publish multiple restli major versions simultaneously

PR #933, opened 2023-09-19
Support d2 symlink node in indis flow

bohhyang:
## Background 
We need to support D2 symlink (d2 single master/primary feature, [wiki](https://iwww.corp.linkedin.com/wiki/cf/display/ENGS/Multi-Colo+Single+Master+Fast+Failover#MultiColoSingleMasterFastFailover-D2SingleMasterSupport)) in INDIS flow. 

The current ZK flow handles symlinks implicitly in SymlinkAwareZookeeper, which captures the data of a symlink node, immediately get the data of the actual master node, and send the fetched data to the symlink node callback. It can't be shared with INDIS flow.  

For symlink clusters, the event bus subscribers, like "ClusterLoadBalancerSubscriber", "UriLoadBalancerSubscriber" subscribe to the symlinks ($FooClusterMaster), instead of the actual master node (FooCluster-prod-ltx1), we need to publish to event bus under the symlink names.
For other clusters, we need to publish under its original name. Note that these clusters could be either:
1) regular clusters requested normally.
2) clusters that were pointed by a symlink previously, but no longer the case after the symlink points to other clusters.

## Changes
Subscribe to and handle d2 symlink node data in xDSClient and the xDS-to-d2 data adaptor.

## Test Done
build and unit tests.
Verified with Toki qei deployment that the normal service, cluster, and uri nodes can be fetched successfully. 
Will need to test fetching the symlink node with d2-proxy after release (due to SI-34767, the snapshot can't build with d2-proxy).

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
@@ -260,15 +260,21 @@ public class XdsToD2PropertiesAdaptor
             ClusterProperties clusterProperties = toClusterProperties(update.getNodeData().getData(),
                 update.getNodeData().getStat().getMzxid());
             // For symlink clusters, ClusterLoadBalancerSubscriber subscribed to the symlinks, instead of the actual node, in event bus,
-            // so we need to publish under the symlink names.
+            // so we need to publish under the symlink names. Also, rarely and possibly, the original cluster could have subscribers
+            // too when calls are made directly to the original cluster, so we publish for it too.
             // For other clusters, publish under its original name. Note that these clusters could be either:
             // 1) regular clusters requested normally.
             // 2) clusters that were pointed by a symlink previously, but no longer the case after the symlink points to other clusters.
             // For case #2: the symlinkAndActualNode map will no longer has an entry for this cluster (removed in
             // D2SymlinkNodeResourceWatcher::onChanged), thus the updates will be published under the original cluster name
-            // (like "FooCluster-prod-ltx1"), which has no subscribers anyway, so no harm to publish.
-            String clusterNameToPublish = StringUtils.defaultString(getSymlink(clusterName), clusterName);
-            _clusterEventBus.publishInitialize(clusterNameToPublish, clusterProperties);
+            // (like "FooCluster-prod-ltx1"), which has no symlink subscribers anyway, so no harm to publish.
+            _clusterEventBus.publishInitialize(clusterName, clusterProperties);
+            String symlinkName = getSymlink(clusterName);
+            if (symlinkName != null)
+            {
+              _clusterEventBus.publishInitialize(symlinkName, clusterProperties);
+            }
+
             if (_dualReadStateManager != null)
             {
               _dualReadStateManager.reportData(clusterName, clusterProperties, true);
@@ -431,7 +437,8 @@ public class XdsToD2PropertiesAdaptor
     }
 
     // For symlink clusters, UriLoadBalancerSubscriber subscribed to the symlinks, instead of the actual node, in event bus,
-    // so we need to publish under the symlink names.
+    // so we need to publish under the symlink names. Also, rarely but possibly, the original cluster could have subscribers
+    // too when calls are made directly to the original cluster, so we publish for it too.
     // For other clusters, publish under its original name. Note that these clusters could be either:
     // 1) regular clusters requested normally.
     // 2) clusters that were pointed by a symlink previously, but no longer the case after the symlink points to other clusters.
@@ -460,8 +467,13 @@ public class XdsToD2PropertiesAdaptor
           _currentData = updates;
           UriProperties mergedUriProperties = _uriPropertiesMerger.merge(_clusterName, _currentData.values());
 
-          String clusterNameToPublish = StringUtils.defaultString(getSymlink(_clusterName), _clusterName);
-          _uriEventBus.publishInitialize(clusterNameToPublish, mergedUriProperties);
+          _uriEventBus.publishInitialize(_clusterName, mergedUriProperties);
+          String symlinkName = getSymlink(_clusterName);
+          if (symlinkName != null)
+          {
+            _uriEventBus.publishInitialize(symlinkName, mergedUriProperties);
+          }
+
           if (_dualReadStateManager != null)
           {
             _dualReadStateManager.reportData(_clusterName, mergedUriProperties, true);
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
@@ -128,11 +128,6 @@ public class XdsToD2PropertiesAdaptor
     _clusterEventBus = clusterEventBus;
   }
 
-  private boolean isSymlinkNode(String nodeNameOrPath)
-  {
-    return nodeNameOrPath != null && nodeNameOrPath.indexOf(SYMLINK_NODE_IDENTIFIER) >= 0;
-  }
-
   public void registerXdsConnectionListener(XdsConnectionListener listener)
   {
     _xdsConnectionListeners.add(listener);
@@ -142,7 +137,7 @@ public class XdsToD2PropertiesAdaptor
   {
     // if cluster name is a symlink, watch for D2SymlinkNode instead
     String resourceName = D2_CLUSTER_NODE_PREFIX + clusterName;
-    if (!checkAndListenToSymlink(clusterName, resourceName))
+    if (shouldNotListenToSymlink(clusterName, resourceName))
     {
       _watchedClusterResources.computeIfAbsent(clusterName, k ->
       {
@@ -158,7 +153,7 @@ public class XdsToD2PropertiesAdaptor
   {
     // if cluster name is a symlink, watch for D2SymlinkNode instead
     String resourceName = D2_URI_NODE_PREFIX + clusterName;
-    if (!checkAndListenToSymlink(clusterName, resourceName))
+    if (shouldNotListenToSymlink(clusterName, resourceName))
     {
       _watchedUriResources.computeIfAbsent(clusterName, k ->
       {
@@ -170,7 +165,18 @@ public class XdsToD2PropertiesAdaptor
     }
   }
 
-  private boolean checkAndListenToSymlink(String symlinkName, String fullResourceName)
+  public void listenToService(String serviceName)
+  {
+    _watchedServiceResources.computeIfAbsent(serviceName, k ->
+    {
+      XdsClient.D2NodeResourceWatcher watcher = getServiceResourceWatcher(serviceName);
+      _xdsClient.watchXdsResource(D2_SERVICE_NODE_PREFIX + serviceName, XdsClient.ResourceType.D2_NODE,
+          watcher);
+      return watcher;
+    });
+  }
+
+  private boolean shouldNotListenToSymlink(String symlinkName, String fullResourceName)
   {
     boolean isSymlink = isSymlinkNode(symlinkName);
     if (isSymlink)
@@ -187,18 +193,12 @@ public class XdsToD2PropertiesAdaptor
       });
     }
 
-    return isSymlink;
+    return !isSymlink;
   }
 
-  public void listenToService(String serviceName)
+  private boolean isSymlinkNode(String nodeNameOrPath)
   {
-    _watchedServiceResources.computeIfAbsent(serviceName, k ->
-    {
-      XdsClient.D2NodeResourceWatcher watcher = getServiceResourceWatcher(serviceName);
-      _xdsClient.watchXdsResource(D2_SERVICE_NODE_PREFIX + serviceName, XdsClient.ResourceType.D2_NODE,
-          watcher);
-      return watcher;
-    });
+    return nodeNameOrPath != null && nodeNameOrPath.indexOf(SYMLINK_NODE_IDENTIFIER) >= 0;
   }
 
   XdsClient.D2NodeResourceWatcher getServiceResourceWatcher(String serviceName)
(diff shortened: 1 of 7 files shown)


d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
```diff
@@ -239,6 +295,71 @@ XdsClient.D2NodeMapResourceWatcher getUriResourceWatcher(String clusterName)
...
+      // If existing actual node differs from new node, remove the existing entry in actualNodeToSymlink map.
+      if (v != null && !v.equals(actualNodeName))
+      {
+        _actualNodeToSymlink.remove(v);
+      }
+      return actualNodeName;
+    });
+
+    _actualNodeToSymlink.put(actualNodeName, symlinkName);
+  }
```
PapaCharlie, line 348: I think this is too likely to become racy. A read that comes in between `_symlinkToActualNode.compute` and `_actualNodeToSymlink.put` can read an incorrect value.

IMO, this would be safer:
```java
private final Object _symlinkToActualNodeLock = new Object();
private final HashBiMap<String, String> _symlinkToActualNode = HashBiMap.create();

private void updateSymlinkAndActualNodeMaps(String symlinkName, String actualNodeName) {
  synchronized (_symlinkToActualNodeLock) {
    _symlinkToActualNode.put(symlinkName, actualNodeName);
  }
}

private String getSymlink(String actualNodeName) {
  synchronized (_symlinkToActualNodeLock) {
    return _symlinkToActualNode.inverse().get(actualNodeName);
  }
}
```
Alternatively, if `HashBiMap` from Guava isn't available in the rest.li code, you can still achieve the same thing with the two maps but you still must lock both

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
```diff
@@ -239,6 +295,71 @@ XdsClient.D2NodeMapResourceWatcher getUriResourceWatcher(String clusterName)
...
+
+  private String removeNodePathPrefix(String path, String prefix)
```
PapaCharlie, line 350: Nit: can be static :)

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
```diff
@@ -344,7 +475,7 @@ public void onChanged(XdsClient.D2NodeMapUpdate update)
     @Override
     public void onError(Status error)
     {
-      if (_init.compareAndSet(false, true))
+      if (_isInit.get())
```
PapaCharlie, line 478: This means that `onError` can now execute multiple times instead of once. Is that intentional? What was the reasoning behind flipping this init logic?

PapaCharlie (changes requested):

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
bohhyang: right, what we really want to flag is the initial success of fetching the data, until then, onError shouldn't be flipping the flag.

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
bohhyang: yeah, I thought about that too. But the xDSClientImpl is designed to have only one thread processing the response, which means only one resource update will happen at a time (only one watcher's onChanged will be called a time). So a read of the actualNodeToSymlink (inside the cluster resource watcher) will never come in when an update of the maps is ongoing.

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
```diff
@@ -154,6 +176,31 @@ public void listenToService(String serviceName)
     });
   }
 
+  private boolean shouldNotListenToSymlink(String symlinkName, String fullResourceName)
```
shivamgupta1, line 179: A method called `shouldNotListenToSymlink` should not have side-effects (like registering the watcher).

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
```diff
@@ -154,6 +176,31 @@ public void listenToService(String serviceName)
...
+    boolean isSymlink = isSymlinkNode(symlinkName);
+    if (isSymlink)
+    {
+      // use full resource name ("/d2/clusters/$FooClusterMater", "/d2/uris/$FooClusterMaster") as the key
+      // instead of just the symlink name ("$FooClusterMaster") to differentiate clusters and uris symlink resources.
+      _watchedSymlinkResources.computeIfAbsent(fullResourceName, k ->
+      {
+        // use symlink name "$FooClusterMaster" to create the watcher
+        XdsClient.D2SymlinkNodeResourceWatcher watcher = getSymlinkResourceWatcher(symlinkName);
+        _xdsClient.watchXdsResource(fullResourceName, XdsClient.ResourceType.D2_SYMLINK_NODE,
```
shivamgupta1, line 190: Nit: better to use `k` rather than `fullResourceName` here since it's inside the mapping function.

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
```diff
@@ -123,24 +135,34 @@ public void registerXdsConnectionListener(XdsConnectionListener listener)
 
   public void listenToCluster(String clusterName)
   {
-    _watchedClusterResources.computeIfAbsent(clusterName, k ->
+    // if cluster name is a symlink, watch for D2SymlinkNode instead
+    String resourceName = D2_CLUSTER_NODE_PREFIX + clusterName;
+    if (shouldNotListenToSymlink(clusterName, resourceName))
```
shivamgupta1, line 140: Confusing, `shouldNotListen...` being true shouldn't go to the if block. This can be done:
```
if (isSymlink)
    listen to symlink
else
    listen to cluster
```
Same for URIs.

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
shivamgupta1: `HashBiMap` seems like an elegant solution, but `_actualNodeToSymlink.put(actualNodeName, symlinkName);` could also be moved inside the compute method lambda?

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
bohhyang: right, but the "listen to symlink" logic is the same for cluster and uri, so I put it into a helper function. I'll separate the isSymlink and the listenToSymlink to two functions then.

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
```diff
@@ -64,8 +65,16 @@ public class XdsToD2PropertiesAdaptor
...
+  private final ConcurrentMap<String, XdsClient.D2SymlinkNodeResourceWatcher> _watchedSymlinkResources;
   private final ConcurrentMap<String, XdsClient.D2NodeResourceWatcher> _watchedServiceResources;
   private final ConcurrentMap<String, XdsClient.D2NodeMapResourceWatcher> _watchedUriResources;
+  // Mapping from a symlink name, like "$FooClusterMaster", to the actual node name it's pointing to, like
+  // "FooCluster-prod-ltx1".
+  // Note that this works for both cluster symlink "/d2/clusters/$FooClusterMaster" and uri-parent symlink
+  // "/d2/uris/$FooClusterMaster".
+  private final ConcurrentMap<String, String> _symlinkToActualNode;
+  // An inverse mapping of the above map.
+  private final ConcurrentMap<String, String> _actualNodeToSymlink;
```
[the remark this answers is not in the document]
bohhyang: not really, the symlink is auto-generated based on the actual cluster name by d2-config-service in a 1-to-1 mapping. And the script that updates it (singlemaster mp) only changes the fabric suffix part.

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
bohhyang: ah, yeah, why not.

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
bohhyang: for sure

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
```diff
@@ -239,6 +300,68 @@ XdsClient.D2NodeMapResourceWatcher getUriResourceWatcher(String clusterName)
...
+    }
+  }
+
+  private String getSymlink(String actualNodeName) {
+    synchronized (_symlinkAndActualNodeLock) {
+      return _symlinkAndActualNode.inverse().get(actualNodeName);
+    }
+  }
+
+  private static String removeNodePathPrefix(String path, String prefix)
```
shivamgupta1, line 352: Can `SymlinkUtil` be leveraged here?

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
bohhyang: I ended up making it thread-safe with HashBiMap, just in case we need to add threads to xDSClient in the future.

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
```diff
@@ -207,7 +259,16 @@ public void onChanged(XdsClient.D2NodeUpdate update)
           {
             ClusterProperties clusterProperties = toClusterProperties(update.getNodeData().getData(),
                 update.getNodeData().getStat().getMzxid());
-            _clusterEventBus.publishInitialize(clusterName, clusterProperties);
+            // For symlink clusters, ClusterLoadBalancerSubscriber subscribed to the symlinks, instead of the actual node, in event bus,
```
shivamgupta1, line 262: Curious, how does it know when to make a request for the $ prefixed symlink rather than a regular cluster?

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
bohhyang: Here we need to find the index of "$" and remove all preceding path. SymlinkUtil finds the end of the symlink:
 /**
   * return the index to split the first symlink out of param path.
   *
   * If no symlink is found, -1 is returned.
   *
   * Examples:
   *  firstSymlinkIndex("/$symlink/foo/bar") returns 9
   *  firstSymlinkIndex("/foo/bar/$symlink") returns 17
   *  firstSymlinkIndex("/notASymlink") returns -1
   *

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
bohhyang: the service properties on ZK has a field "clusterName" which is "$xxx" for symlink-enabled cluster.

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
```diff
@@ -207,7 +259,16 @@ public void onChanged(XdsClient.D2NodeUpdate update)
...
+            // so we need to publish under the symlink names.
+            // For other clusters, publish under its original name. Note that these clusters could be either:
+            // 1) regular clusters requested normally.
+            // 2) clusters that were pointed by a symlink previously, but no longer the case after the symlink points to other clusters.
+            // For case #2: the symlinkAndActualNode map will no longer has an entry for this cluster (removed in
+            // D2SymlinkNodeResourceWatcher::onChanged), thus the updates will be published under the original cluster name
+            // (like "FooCluster-prod-ltx1"), which has no subscribers anyway, so no harm to publish.
+            String clusterNameToPublish = StringUtils.defaultString(getSymlink(clusterName), clusterName);
```
shivamgupta1, line 270: For symlink enabled clusters, can the clients never request a specific fabric (eg. `FooCluster-prod-ltx1`)? Looked at one of the symlink-enabled clusters (`$JobServicesMaster`), and the service `/d2/services/jobIngestionBlacklistRules` ZK data instead has `clusterName` set to `JobsServices-prod-ltx1` directly. From the code change, it appears that if `$JobServicesMaster` points to `JobsServices-prod-ltx1` and there's a subscriber for the former, then the latter will never be published. Maybe both should be published instead?
```
{
  "path": "/jobIngestionBlacklistRules",
  "serviceName": "jobIngestionBlacklistRules",
  "transportClientProperties": {
    "http.idleTimeout": "10500000",
    "allowedClientOverrideKeys": "http.protocolVersion,http.requestTimeout",
    "http.poolSize": "200",
    "http.maxHeaderSize": "16834",
    "http.maxResponseSize": "10000000",
    "http.sslIdleTimeout": "10500000",
    "http.requestTimeout": "10000"
  },
  "backupRequests": [],
  "serviceMetadataProperties": {
    "isDefaultService": "true",
    "restli.protocol": "2.0.0"
  },
  "clusterName": "JobsServices-prod-ltx1",
```

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
```diff
@@ -154,6 +187,25 @@ public void listenToService(String serviceName)
...
+  private static boolean isSymlinkNode(String nodeNameOrPath)
+  {
+    return nodeNameOrPath != null && nodeNameOrPath.indexOf(SYMLINK_NODE_IDENTIFIER) >= 0;
+  }
+
+  private void listenToSymlink(String name, String fullResourceName)
+  {
+    // use full resource name ("/d2/clusters/$FooClusterMater", "/d2/uris/$FooClusterMaster") as the key
+    // instead of just the symlink name ("$FooClusterMaster") to differentiate clusters and uris symlink resources.
+    _watchedSymlinkResources.computeIfAbsent(fullResourceName, k ->
```
shivamgupta1, line 199: The results are stored without the cluster and uri prefixes in the bi-map later:
```

  // Mapping between a symlink name, like "$FooClusterMaster" and the actual node name it's pointing to, like
  // "FooCluster-prod-ltx1".
  // (Note that this name does NOT include the full path so that it works for both cluster symlink
  // "/d2/clusters/$FooClusterMaster" and uri-parent symlink "/d2/uris/$FooClusterMaster").
  private final HashBiMap<String, String> _symlinkAndActualNode = HashBiMap.create();
```

Should the symlinks for uris and clusters happen to be different, they will overwrite each other and be quite error prone. For consistency, the symlink resource names should be kept the same here as well, or they should be kept separately in `_symlinkAndActualNode` too.

bohhyang: symlinks for uris and clusters are always the same. The tooling always update the URI and cluster symlinks together.

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
bohhyang: clients make calls to d2 service jobIngestionBlacklistRulesMaster instead for symlink-enabled service.
That said, users indeed could send calls directly to non "Master" d2 services too. We should publish both. Thanks for the deep thinking, investigation, and nice catch!

shivamgupta1: Thanks, would be good to add a test-case specifically for this too.

bohhyang: updated

PapaCharlie (approved):

shivamgupta1 (approved):

PR #936, opened 2023-09-26
add service/cluster-not-found count to simple load balancer jmx. And add entry-out-of-sync count to dual read monitoring.

bohhyang:
## Background
We had issues where Observer accidentally deleted all cluster properties, but we weren't able to see that the client is failing to get clusters on any ingraph metric but only via digging into logs.
For dual read, we've been only monitoring when the data is mismatched, but not when data is received on side (like ZK) but not on the other side (like xDS).

## Changes
1. Added service and cluster not found counter to simple load balancer jmx, which gets incremented when timeout happens at fetching the resource.
2. Added entry out-of-sync count to dual read monitoring, which gets incremented when data is either mis-match or received on one side but not on the other (until a new version is received on the first side), and gets decremented when the data matches.

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancerJmx.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancerJmx.java
@@ -29,6 +29,11 @@ public class DualReadLoadBalancerJmx implements DualReadLoadBalancerJmxMBean
   private final AtomicInteger _clusterPropertiesEvictCount = new AtomicInteger();
   private final AtomicInteger _uriPropertiesEvictCount = new AtomicInteger();
 
+  private final AtomicInteger _servicePropertiesOutOfSyncCount = new AtomicInteger();
+  private final AtomicInteger _clusterPropertiesOutOfSyncCount = new AtomicInteger();
+  private final AtomicInteger _uriPropertiesOutOfSyncCount = new AtomicInteger();
+
+
   @Override
   public int getServicePropertiesErrorCount()
   {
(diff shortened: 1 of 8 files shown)


d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancerJmx.java
```diff
@@ -29,6 +29,11 @@ public class DualReadLoadBalancerJmx implements DualReadLoadBalancerJmxMBean
   private final AtomicInteger _clusterPropertiesEvictCount = new AtomicInteger();
   private final AtomicInteger _uriPropertiesEvictCount = new AtomicInteger();
 
+  private final AtomicInteger _servicePropertiesOutOfSyncCount = new AtomicInteger();
```
shivamgupta1, line 32: Don't seem to need any Atomic methods for these, should `volatile` be good enough?

shivamgupta1 (commented):
LGTM otherwise

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancerJmx.java
bohhyang: this is being written by ZK flow and indis flow, so we do need it.

shivamgupta1 (approved):

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancerJmx.java
shivamgupta1: That's right, only `volatile` will not work since increment/decrement are composite operations. Here's a handy guide: https://www.baeldung.com/java-volatile-variables-thread-safety#:~:text=The%20volatile%20keyword%20is%20useful,that%20the%20operation%20is%20atomic.

PR #937, opened 2023-09-27
Fix ByteBuffer errors to be compatible with Java 8 runtimes

JoeJoe1989:
**Background**

The root cause is that, after compiling ResLi in Java 11, the new Java 11 implementation was preferred over the Java 8 implementation. However, since the new implementation was introduced after Java 8, a NoSuchMethodError occurred for Java 8 consumers. To resolve the issue, we chose an implementation that exists in both Java 8 and Java 11.

**Changes**

This PR will apply the same fix to URIDecoderUtils.java as was done in the prior PR to [URIDecoderUtils.java](https://github.com/linkedin/rest.li/blob/master/restli-common/src/main/java/com/linkedin/restli/internal/common/URIDecoderUtils.java#L204).

**Tests Done**
 
Followed the instructions provided by the team who reported the issue and tested locally using curli command.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.46.3] - 2023-09-26
+- Fix ByteBuffer errors to be compatible with Java 8 runtimes.
+
 ## [29.46.2] - 2023-09-25
 - add service/cluster-not-found count to simple load balancer jmx. And add entry-out-of-sync count to dual read monitoring.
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.46.2] - 2023-09-26
+- Fix ByteBuffer errors
```
zackthehuman, line 18: Let's be more specific about this. I suggest:

> Fix usage of `ByteBuffer` to be compatible with Java 8 runtimes.

zackthehuman (changes requested):

CHANGELOG.md
JoeJoe1989: Updated.

zackthehuman (approved):

zackthehuman (approved):

PR #939, opened 2023-10-04
simplify symlink subscription in xds flow

bohhyang:
## Summary 
Current Observer only has a symlink for cluster config /d2/clusters/$xxxMaster, but not a counterpart symlink for uri parent /d2/uris/$xxxMaster. In fact, since cluster symlink and uri parent symlink always point to the same actual node name, and it's a redundancy and a burden for the symlink-update tool to maintain two symlinks for the same actual node name. 
We optimize xds flow to use the cluster symlink to listen to the actual nodes for both cluster and uri parent.
Once all apps migrate off ZK flow, we don't need to maintain the uri parent symlink anymore.

## Test Done
updated unit tests.

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
@@ -354,15 +354,7 @@ public class XdsToD2PropertiesAdaptor
 
   private static String getNodeName(String path)
   {
-    int idx = path.lastIndexOf(PATH_SEPARATOR);
-    if (idx == -1)
-    {
-      return path;
-    }
-    else
-    {
-      return path.substring(idx + 1);
-    }
+    return path.substring(path.lastIndexOf(PATH_SEPARATOR) + 1);
   }
 
   private void notifyAvailabilityChanges(boolean isAvailable)
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
@@ -45,7 +45,6 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
-import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
```diff
@@ -355,16 +352,16 @@ private String getSymlink(String actualNodeName) {
...
+  private static String getNodeName(String path)
   {
-    int idx = path.indexOf(prefix);
+    int idx = path.lastIndexOf(PATH_SEPARATOR);
     if (idx == -1)
     {
       return path;
     }
     else
     {
```
PapaCharlie, line 363: Kind of a nit, but since you're doing `idx + 1`, you don't need to check for `idx == -1` anymore. This can be simplified to:
```java
private static String getNodeName(String path)
{
  return path.substring(path.lastIndexOf(PATH_SEPARATOR) + 1);
}
```

bohhyang: ah, yeah, updated.

PapaCharlie (approved):

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
```diff
@@ -313,20 +313,17 @@ XdsClient.D2SymlinkNodeResourceWatcher getSymlinkResourceWatcher(String symlinkN
...
-        }
+        String actualNodeName = getNodeName(actualResourceName);
+        updateSymlinkAndActualNodeMap(symlinkName, actualNodeName);
+        // listen to the actual nodes
+        // Note: since cluster symlink and uri parent symlink always point to the same actual node name, and it's a
+        // redundancy and a burden for the symlink-update tool to maintain two symlinks for the same actual node name,
+        // we optimize here to use the cluster symlink to listen to the actual nodes for both cluster
+        // and uri parent.
+        listenToCluster(actualNodeName);
+        listenToUris(actualNodeName);
```
shivamgupta1, line 326: Just a thought, would we also want to delete the existing symlink properties from the property bus, and let the requests fail in case the `listenToUris(actualNodeName);` or cluster calls fail? Because if we don't remove those properties and fetching new properties from xDS fails due to some reason, traffic would still keep on flowing to the old leader (~master~), which is not one might want.

bohhyang: that's true. But I'd prefer to keep parity with zk flow (which will also point to the old leader) for now to ease INDIS migration monitoring. We could create a backlog item to improve this after the migration.

shivamgupta1: Makes sense.

PR #943, opened 2023-10-12
Add metrics about xds connection status and count

bohhyang:
Add jmx metrics for xds connection lost, closed, reconnected counts, and its current status of being connected or not.
Will create a container PR to add the corresponding ingraph sensor for the same.

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
@@ -148,6 +148,8 @@ public class XdsClientImpl extends XdsClient
       _log.warn("ADS stream not ready within {} milliseconds", _readyTimeoutMillis);
       // notify subscribers about the error and wait for the stream to be ready by keeping it open.
       notifyStreamError(Status.DEADLINE_EXCEEDED);
+      // note: no need to start a retry task explicitly since xds stream internally will keep on retrying to connect
+      // to one of the sub-channels (unless an error or complete callback is called).
     }, _readyTimeoutMillis, TimeUnit.MILLISECONDS);
     _adsStream.start();
     _log.info("ADS stream started, connected to server: {}", _managedChannel.authority());
@@ -205,8 +207,6 @@ public class XdsClientImpl extends XdsClient
         _readyTimeoutFuture = null; // set it to null to avoid repeat notifications to subscribers.
         _xdsClientJmx.incrementReconnectionCount();
         notifyStreamReconnect();
-        // note: no need to start a retry task explicitly since xds stream internally will keep on retrying to connect
-        // to one of the sub-channels (unless an error or complete callback is called).
       }
       _xdsClientJmx.setIsConnected(true);
     }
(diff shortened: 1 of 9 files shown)


d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -70,18 +71,22 @@ public class XdsClientImpl extends XdsClient
...
 
   public XdsClientImpl(Node node, ManagedChannel managedChannel, ScheduledExecutorService executorService,
-      long readyTimeoutMillis) {
+      long readyTimeoutMillis)
+  {
     _readyTimeoutMillis = readyTimeoutMillis;
     _node = node;
     _managedChannel = managedChannel;
     _executorService = executorService;
+    _xdsClientJmx = new XdsClientJmx();
```
shivamgupta1, line 89: Nit: As a general rule, the use of `new` should be avoided and injection through the constructor should be favored. There could be another constructor that takes the jmx bean instance as input, and perhaps the existing constructor could instantiate it for now and call the new constructor. This could also help with writing some UTs in case we ever do that. Not really required as of now though, feel free to ignore.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -166,29 +177,38 @@ String getXdsServerAuthority()
...
+      if (_readyTimeoutFuture != null)
+      {
         // timeout task will be cancelled only if it hasn't already executed.
         boolean cancelledTimeout = _readyTimeoutFuture.cancel(false);
         _log.info("ADS stream ready, cancelled timeout task: {}", cancelledTimeout);
         _readyTimeoutFuture = null; // set it to null to avoid repeat notifications to subscribers.
+        _xdsClientJmx.incrementReconnectionCount();
         notifyStreamReconnect();
+        // note: no need to start a retry task explicitly since xds stream internally will keep on retrying to connect
+        // to one of the sub-channels (unless an error or complete callback is called).
```
shivamgupta1, line 209: Maybe this should go here? https://github.com/linkedin/rest.li/blob/559f619ba0b68873ec35b8e612206b6a9af1a375/d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java#L150
Thanks for adding it btw.

shivamgupta1 (commented):
LGTM otherwise, one minor comment about a comment.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: whoops, yeah, wrong place. Corrected.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: makes sense. I'll keep for now though for simplicity.

shivamgupta1 (approved):

PR #944, opened 2023-10-31
Use Node instead of D2Node and D2URIMap instead of NodeMap for xDS flow

PapaCharlie:
The initial intent of using D2Node was to pre-deserialize the JSON, but because there is no way for the rest.li client to actually leverage this, it simply re-serializes the JSON then deserializes it again. This causes memory pressure along with validation errors as numerical typesa get mungled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.47.0] - 2023-11-13
+- Use Node(Map) instead of D2node(Map)
+
 ## [29.46.9] - 2023-11-02
 - Update FieldDef so that it will lazily cache the hashCode.
 
@@ -5557,7 +5560,8 @@ patch operations can re-use these classes for generating patch messages.
 
 ## [0.14.1]
 
-[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.46.9...master
+[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.47.0...master
+[29.47.0]: https://github.com/linkedin/rest.li/compare/v29.46.9...v29.47.0
 [29.46.9]: https://github.com/linkedin/rest.li/compare/v29.46.8...v29.46.9
 [29.46.8]: https://github.com/linkedin/rest.li/compare/v29.46.7...v29.46.8
 [29.46.7]: https://github.com/linkedin/rest.li/compare/v29.46.6...v29.46.7
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and what APIs have changed, if applicable.
 ## [Unreleased]
 
 ## [29.47.0] - 2023-11-13
-- Use Node(Map) instead of D2node(Map)
+- Use Node instead of D2Node and D2URIMap instead of NodeMap for xDS flow
 
 ## [29.46.9] - 2023-11-02
 - Update FieldDef so that it will lazily cache the hashCode.
--- a/d2/src/main/java/com/linkedin/d2/balancer/properties/UriPropertiesJsonSerializer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/properties/UriPropertiesJsonSerializer.java
@@ -24,13 +24,12 @@ import com.linkedin.d2.discovery.PropertyBuilder;
 import com.linkedin.d2.discovery.PropertySerializationException;
 import com.linkedin.d2.discovery.PropertySerializer;
 import indis.XdsD2;
-import java.util.Collections;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import java.net.URI;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class UriPropertiesJsonSerializer implements PropertySerializer<UriProperties>, PropertyBuilder<UriProperties>
 {
@@ -129,32 +128,6 @@ public class UriPropertiesJsonSerializer implements PropertySerializer<UriProper
           partitions.put(partition.getKey(), new PartitionData(partition.getValue()));
         }
         partitionDesc.put(URI.create(entry.getKey()), partitions);
-      }
-
-
-      Map<URI, Map<Integer, PartitionData>> partitionDescFromWeights = new HashMap<>(uri.getWeightsCount());
-      for (Map.Entry<String, Double> entry : uri.getWeightsMap().entrySet())
-      {
-        partitionDescFromWeights.put(URI.create(entry.getKey()), Collections.singletonMap(
-            DefaultPartitionAccessor.DEFAULT_PARTITION_ID,
-            new PartitionData(entry.getValue())
-        ));
-      }
-
-      // if both partitionDesc and weights exist, check consistency
-      if (!partitionDesc.isEmpty() && !partitionDescFromWeights.isEmpty())
-      {
-        if (!partitionDesc.equals(partitionDescFromWeights))
-        {
-          _log.error("Inconsistency detected between partitionDesc and weights {} {}",
-              partitionDesc, partitionDescFromWeights);
-        }
-      }
-
-      // always trust partitionDesc over weights
-      if (partitionDesc.isEmpty())
-      {
-        partitionDesc = partitionDescFromWeights;
       }
 
       return new UriProperties(
--- a/d2/src/main/java/com/linkedin/d2/balancer/properties/UriPropertiesJsonSerializer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/properties/UriPropertiesJsonSerializer.java
@@ -16,18 +16,18 @@
 
 package com.linkedin.d2.balancer.properties;
 
-
+import com.google.protobuf.ByteString;
 import com.linkedin.d2.balancer.properties.util.PropertyUtil;
 import com.linkedin.d2.balancer.util.JacksonUtil;
 import com.linkedin.d2.balancer.util.partitions.DefaultPartitionAccessor;
 import com.linkedin.d2.discovery.PropertyBuilder;
 import com.linkedin.d2.discovery.PropertySerializationException;
 import com.linkedin.d2.discovery.PropertySerializer;
-import java.util.Collections;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.net.URI;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -103,6 +103,29 @@ public class UriPropertiesJsonSerializer implements PropertySerializer<UriProper
 
   @Override
   public UriProperties fromBytes(byte[] bytes, long version) throws PropertySerializationException
+  {
+    UriProperties uriProperties = fromBytes(bytes);
+    uriProperties.setVersion(version);
+    return uriProperties;
+  }
+
+  @Override
+  public UriProperties fromBytes(ByteString bytes) throws PropertySerializationException
+  {
+    try
+    {
+      @SuppressWarnings("unchecked")
+      Map<String, Object> untyped = JacksonUtil.getObjectMapper().readValue(bytes.newInput(), HashMap.class);
+      return fromMap(untyped);
+    }
+    catch (Exception e)
+    {
+      throw new PropertySerializationException(e);
+    }
+  }
+
+  @Override
+  public UriProperties fromBytes(ByteString bytes, long version) throws PropertySerializationException
   {
     UriProperties uriProperties = fromBytes(bytes);
     uriProperties.setVersion(version);
--- a/d2/src/main/java/com/linkedin/d2/balancer/properties/UriPropertiesJsonSerializer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/properties/UriPropertiesJsonSerializer.java
@@ -16,7 +16,6 @@
 
 package com.linkedin.d2.balancer.properties;
 
-import com.google.protobuf.Struct;
 import com.linkedin.d2.balancer.properties.util.PropertyUtil;
 import com.linkedin.d2.balancer.util.JacksonUtil;
 import com.linkedin.d2.balancer.util.partitions.DefaultPartitionAccessor;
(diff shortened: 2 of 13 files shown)


d2/src/test/java/com/linkedin/d2/xds/TestXdsToD2PropertiesAdaptor.java
```diff
@@ -74,26 +77,27 @@ public void testListenToNormalCluster()
...
     XdsToD2PropertiesAdaptorFixture fixture = new XdsToD2PropertiesAdaptorFixture();
     fixture.getSpiedAdaptor().listenToCluster(SYMLINK_NAME);
 
-    verify(fixture._xdsClient).watchXdsResource(eq(CLUSTER_SYMLINK_RESOURCE_NAME), eq(XdsClient.ResourceType.D2_SYMLINK_NODE), any());
+    verify(fixture._xdsClient).watchXdsResource(eq(CLUSTER_SYMLINK_RESOURCE_NAME), eq(XdsClient.ResourceType.NODE), any());
 
-    XdsClient.D2SymlinkNodeResourceWatcher symlinkNodeWatcher =
-        (XdsClient.D2SymlinkNodeResourceWatcher) fixture._symlinkWatcherArgumentCaptor.getValue();
+    XdsClient.SymlinkNodeResourceWatcher symlinkNodeWatcher =
+        (XdsClient.SymlinkNodeResourceWatcher) fixture._clusterWatcherArgumentCaptor.getValue();
```
ZoabKapoor, line 93: Why do we change `fixture._symlinkWatcherArgumentCaptor` to `fixture._clusterWatcherArgumentCaptor` here?

ZoabKapoor: Ah, I see, we removed `_symlinkWatcherArgumentCaptor` since symlinks and Nodes can be treated the same way under the new scheme

ZoabKapoor (commented):
Overall looks good, 1 question before merging: will this work with the existing Observer implementation today? It seems like we are expecting `type.googleapis.com/indis.NodeMap` instead of `"type.googleapis.com/indis.D2NodeMap"` (and similar for the other types), but as far as I know the Observer only serves these types today:

```
envoy.config.cluster.v3.Cluster
envoy.config.core.v3.TypedExtensionConfig
envoy.config.endpoint.v3.ClusterLoadAssignment
envoy.config.listener.v3.Listener
envoy.config.route.v3.RouteConfiguration
envoy.config.route.v3.ScopedRouteConfiguration
envoy.config.route.v3.VirtualHost
envoy.extensions.transport_sockets.tls.v3.Secret
envoy.service.runtime.v3.Runtime
indis.D2Node
indis.D2NodeMap
indis.D2SymlinkNode
indis.Stat
```

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
```diff
@@ -306,20 +300,20 @@ public void onReconnect()
...
-    return new XdsClient.D2SymlinkNodeResourceWatcher()
+    return new XdsClient.SymlinkNodeResourceWatcher()
     {
       @Override
-      public void onChanged(String resourceName, XdsClient.D2SymlinkNodeUpdate update)
+      public void onChanged(String resourceName, XdsClient.NodeUpdate update)
       {
         // Update maps between symlink name and actual node name
-        String actualResourceName = update.getNodeData().getMasterClusterNodePath();
+        String actualResourceName = update.getNodeData().getData().toString(StandardCharsets.UTF_8);
```
ZoabKapoor, line 316: Won't this get the `bytes data = 2` field from `Node`, isn't that the whole node data and not just the resource name?

Or if there is some semantic meaning to `data` (like it's bytes for only the resource name) we should probably document that, either here or in the proto

d2/src/main/java/com/linkedin/d2/balancer/properties/UriPropertiesJsonSerializer.java
```diff
@@ -109,6 +110,66 @@ public UriProperties fromBytes(byte[] bytes, long version) throws PropertySerial
     return uriProperties;
   }
 
+  public UriProperties fromProto(XdsD2.D2URI uri) throws PropertySerializationException
```
logstashbugreporter, line 113: ideally, we can directly use XdsD2.D2URI w/o constructing a new object UriProperties, however, I understand that this restli solution will be deprecated eventually, so it is not a priority to optimize this for now.

d2/src/main/proto/XdsD2.proto
```diff
@@ -40,5 +40,22 @@ message D2SymlinkNode {
 }
 
 message D2NodeMap {
+  // Deprecated in favor of D2UriMap
+  option deprecated = true;
   map<string, D2Node> nodes = 1;
 }
+
+message D2URI {
```
logstashbugreporter, line 48: please add comments to this new data structure.

logstashbugreporter (commented):
LGTM,  some minor comments inlined, which we talked about offline.

d2/src/test/java/com/linkedin/d2/xds/TestXdsToD2PropertiesAdaptor.java
PapaCharlie: This has actually be reverted actualyl

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
PapaCharlie: This has been updated, but for the record the node's data _is_ the symlink. Symlink nodes don't have JSON in them, they just have a string

d2/src/main/java/com/linkedin/d2/discovery/stores/zk/SymlinkAwareZooKeeper.java
```diff
@@ -601,6 +602,19 @@ public String fromBytes(byte[] bytes) throws PropertySerializationException
       }
     }
 
+    @Override
+    public String fromBytes(ByteString bytes) throws PropertySerializationException
+    {
+      try
+      {
+        return bytes.toString("UTF-8");
```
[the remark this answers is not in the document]
PapaCharlie: This is now gone in the latest version

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
```diff
@@ -24,9 +24,8 @@
 
 public abstract class XdsClient
```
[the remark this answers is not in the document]
PapaCharlie: Technically it's not backwards incompatible, we're using new interfaces. Plus the proto interfaces are bundled in the same jar as `XdsClient` so it's impossible for there to be a mismatch. The only issue is in the observer support but the observer will simply support both the old and new types during the transition

d2/src/main/java/com/linkedin/d2/balancer/properties/UriPropertiesJsonSerializer.java
PapaCharlie: Yeah to do so would imply refactoring pretty much the entirety of the D2 load balancing code to use the new class, along with the ZK monitoring code since it needs to translate the ZK JSON into the new proto D2URI class. Definitely not worth it at this stage

d2/src/main/proto/XdsD2.proto
PapaCharlie: ACK

PapaCharlie: Done

d2/src/main/proto/XdsD2.proto
```diff
@@ -40,5 +40,55 @@ message D2SymlinkNode {
...
+//		"clusterName": "Toki"
+//	}
+message D2URI {
+  // The ZK stat of the node that contained this announcement.
+  Stat stat = 1;
+
+  // The name of the cluster this announcement belongs to. This is inferred from the original "clusterName" field.
+  string cluster_name = 2;
+
+  message PartitionData {
```
logstashbugreporter, line 78: why we need a nested message?  other fields(partition_desc and uri_specific_properties) are just a map?

d2/src/main/proto/XdsD2.proto
```diff
@@ -40,5 +40,55 @@ message D2SymlinkNode {
...
+
+// D2URI is a proto representation of com.linkedin.d2.balancer.properties.UriProperties. Note that a D2 UriProperties is
+// actually an announcement from a single machine for potentially multiple URIs. This class is a misnomer but it seems
+// important to be consistent with the old naming strategy as this new proto bridges the old type. The string keys in
+// the maps of the message below will be the URIs of this announcement. Here is a sample ZK announcement for additional
+// clarity:
+//	{
+//		"weights": {
+//			"https://foo.stg.linkedin.com:18792/Toki/resources": 1.0
+//		},
```
logstashbugreporter, line 56: why we need "weights" field? from this example,  partitionDesc seems like a superset of weights?

d2/src/main/proto/XdsD2.proto
```diff
@@ -40,5 +40,55 @@ message D2SymlinkNode {
 }
 
 message D2NodeMap {
+  // Deprecated in favor of D2UriMap
+  option deprecated = true;
   map<string, D2Node> nodes = 1;
 }
+
+// D2URI is a proto representation of com.linkedin.d2.balancer.properties.UriProperties. Note that a D2 UriProperties is
```
logstashbugreporter, line 48: why we have announcement from a single machine for potentially multiple URIs?  is this i001, and i002 multiple instances on a single machine? 
the original name UriProperties is a plural, we would need plural to be consistent with UriProperties?:)

logstashbugreporter (commented):
thanks for adding more comments. some quick comments on [XdsD2.proto] are inlined:

d2/src/main/proto/XdsD2.proto
PapaCharlie: You can't have a nested map type like `map<string, map<int, double>>` in protobuf, you need an explicit message type for the value

d2/src/main/proto/XdsD2.proto
logstashbugreporter: I might have been confused by this example:).  as an example of D2URI object, it should not have separate weight field?

PapaCharlie: Yes but it's a legacy feature. I linked the full contents of a URI node but "weights" is ignored here, it's actually mentioned in the comment on `partition_desc`

d2/src/main/proto/XdsD2.proto
PapaCharlie: This is a legacy feature, I'm not sure if it's actually ever used, but we can't take that chance. As far as the naming goes, it's the properties of 1 "URI" since it's one URI ZK node, but the code itself suggests there can be more than one uri within the URI node. It's terrible naming all around but the naming of this message is consistent with the rest of the code

d2/src/main/proto/XdsD2.proto
```diff
@@ -40,5 +40,55 @@ message D2SymlinkNode {
...
+    map<int32, double> weights = 1;
+  }
+  // The partitions for each URI contained in this announcement. This is inferred from the original "partitionDesc" and
+  // "weights" fields. If "partitionDesc" is present in the original ZK node, it is always used regardless of "weights".
+  // Otherwise, "weights" is assumed to be for partition 0, as specified in UriProperties.
+  map<string, PartitionData> partition_desc = 3;
+
+  // Additional metadata for each URI in this announcement. This is inferred from the original "uriSpecificProperties"
+  // fields.
+  map<string, google.protobuf.Struct> uri_specific_properties = 4;
```
shivamgupta1, line 89: This is `Map<URI, Map<String, Object>>` in the java code. Can we similarly make it `map<string, map<string, struct>` here? You may need to declare the nested map separately though.

shivamgupta1 (commented):
Can we test this locally with corresponding changes on the observer side too, and add a testing done section? May also need to update the changelog.

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
```diff
@@ -105,18 +105,18 @@ public String getVersion()
     }
   }
 
-  static final class D2NodeMapUpdate implements ResourceUpdate
+  static final class D2URIMapUpdate implements ResourceUpdate
   {
     String _version;
-    Map<String, XdsD2.D2Node> _nodeDataMap;
+    Map<String, XdsD2.D2URI> _nodeDataMap;
```
logstashbugreporter, line 111: nit pick, should we also update the variable name?  _uriDataMap?

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
```diff
@@ -26,7 +26,7 @@ public abstract class XdsClient
 {
   private static final String D2_NODE_TYPE_URL = "type.googleapis.com/indis.D2Node";
```
logstashbugreporter, line 27: do we need to change NODE_TYPE to URI_TYPE?

logstashbugreporter (commented):
LGTM, some nitpick clarification questions inlined:

d2/src/main/proto/XdsD2.proto
PapaCharlie: `map<string, struct>` is effectively the same as `struct`, isn't it?

d2/src/main/proto/XdsD2.proto
shivamgupta1: It would be more comforting to know exactly for what purpose this feature existed before getting rid of it.

shivamgupta1: Nvm, just checked many of the URIs using some scripting and none of them seems to be using multiple URIs for one ephemeral node.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -256,19 +256,19 @@ private void handleD2SymlinkNodeResponse(DiscoveryResponseData data)
...
       {
-        XdsD2.D2NodeMap d2NodeMap = resource.getResource().unpack(XdsD2.D2NodeMap.class);
-        Map<String, XdsD2.D2Node> nodeData = d2NodeMap.getNodesMap();
-        updates.put(resourceName, new D2NodeMapUpdate(resource.getVersion(), nodeData));
+        XdsD2.D2URIMap uriMap = resource.getResource().unpack(XdsD2.D2URIMap.class);
+        Map<String, XdsD2.D2URI> nodeData = uriMap.getNodesMap();
+        updates.put(resourceName, new D2URIMapUpdate(resource.getVersion(), nodeData));
       } catch (InvalidProtocolBufferException e)
       {
         _log.warn("Failed to unpack D2NodeMap response", e);
```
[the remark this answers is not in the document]
PapaCharlie: Done

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
PapaCharlie: Sure makes sense

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
PapaCharlie: No the D2Node type was still used for clusters and services

d2/src/main/proto/XdsD2.proto
```diff
@@ -30,15 +31,80 @@ message Stat {
...
+  int64 version = 1;
+
+  // The time at which this announcement was last updated. When coming from ZK this will be the node's mtime.
+  google.protobuf.Timestamp modified_time = 2;
+
+  // The name of the cluster this announcement belongs to. This is inferred from the original "clusterName" field.
+  string cluster_name = 3;
+
+  // The URI for this announcement, i.e. the host, port and context path that requests should be sent to.
+  string uri = 4;
```
bohhyang, line 97: this assumes there is only one uri in a zk uri node data, but in rare cases there could be multiple. I have a concern with creating this D2URI structure instead of using the exact same ZK uri node data, please see my comment on the counterpart indis observer PR. thanks.

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.47.0] - 2023-11-13
+- Use Node(Map) instead of D2node(Map)
```
bohhyang, line 18: nit: to be more specific so people know it's about indis change, how about "Use Node and D2UriMap for observer flow"?

bohhyang (approved):

CHANGELOG.md
PapaCharlie: Oops, yeah let me update the PR title as well

d2/src/main/proto/XdsD2.proto
PapaCharlie: For posterity, we double checked that there is no way that there can be more than one announcement. Additionally, the new data type should be much, much cheaper to deserialize and provides enough benefit to warrant the switch from raw ZK data to the pre-processed version

bohhyang (approved):

PR #945, opened 2023-11-10
Fix dual-read potential risk by using ThreadPool

brycezhongqing:
## Background
During the `dual_read` mode for service discovery, the `new loadBalance` may impact the `old loadBalance`. As observed, when `new_Lb.getProperites` is executed, it can potentially block `oldLb.getClient` due to their synchronous progress. If `new_Lb.getProperities` times out, it leads to the blocking of `oldLb`, resulting in a timeout for the request. To address this, _**it is essential to incorporate a fallback mechanism**_.
image

## Solution:
- There we use the `ExecutorService like ThreadPoolExecutor` to decouple(async) `new loadBalance` and `old loadBalance`. For each `new loadBalance getProperities task` we will request a thread to process.
image

There are some keypoints :
[1]  We need to create a `ExecutorService`
[2] We need to write the `fallback` logic
[3] We shouldn't create a large `task queue` and we need to discard the task which could not be processed in time.
[4] We should add the relative metrics and log.
[5] We need promise the code change is backward compatible
[6] We need to provide the api to `custimize` the `ExecutorService` like `ThreadPoolExecutor`

## Test
[✅]./gradlew build
[✅] test in toki(a test service)


d2/src/main/java/com/linkedin/d2/xds/LoadBalanceTaskPool/NewLoadBalancerTaskthreadPool.java
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  public void execute(Runnable task) {
+    _threadPoolExecutor.execute(task);
+  }
+
+  public void shutdown() {
+    _threadPoolExecutor.shutdown();
+  }
+}
```
PapaCharlie, line 29: I don't think its' a good idea to have a custom wrapper for what's just a normal threadpool. This is especially true in the context of LinkedIn code where we have to use threadpools provided by Offspring for monitoring purposes. You can delete this class and inject a threadpool in DualReadZkAndXdsLoadBalancerFactory instead.

d2/src/main/java/com/linkedin/d2/xds/balancer/DualReadZkAndXdsLoadBalancerFactory.java
```diff
@@ -33,17 +34,20 @@ public class DualReadZkAndXdsLoadBalancerFactory implements LoadBalancerWithFaci
   private final LoadBalancerWithFacilitiesFactory _zkLbFactory;
   private final LoadBalancerWithFacilitiesFactory _xdsLbFactory;
   private final DualReadStateManager _dualReadStateManager;
+  private final NewLoadBalancerTaskthreadPool _newNewLoadBalancerTaskThreadPool;
 
   public DualReadZkAndXdsLoadBalancerFactory(@Nonnull DualReadStateManager dualReadStateManager)
```
PapaCharlie, line 39: See my comment above, you will have to inject a threadpool into this constructor. The problem is that it's backwards incompatible to add a constructor parameter, so you will have to do it like this:
```java
  @Deprecated
  public DualReadZkAndXdsLoadBalancerFactory(@Nonnull DualReadStateManager dualReadStateManager) {
    this(dualReadStateManager, Executors.newSingleThreadExecutor());
  }

  public DualReadZkAndXdsLoadBalancerFactory(@Nonnull DualReadStateManager dualReadStateManager, ExecutorService executor) {
    ...
```

d2/src/main/java/com/linkedin/d2/xds/LoadBalanceTaskPool/NewBalanceGetPropertiesTask.java
```diff
@@ -0,0 +1,85 @@
...
+                  @Override
+                  public void onSuccess(Pair<ClusterProperties, UriProperties> result) {
+                    //TODO change back to debug
+                    LOG.info("Dual read is successful. Get cluster and uri properties: " + result);
+                  }
+                });
+          }
+        });
+      }
+    });
```
PapaCharlie, line 68: I'm definitely in favor of leaving this as a lambda inside `DualReadLoadBalancer`. If you leave this as a public class, it makes the API for this package unclear and confusing. This is an implementation detail that should not be exposed to users of this package. If you want, you can make it a function of `DualReadLoadBalancer` then invoke it like this:
```java
_newNewLoadBalancerTaskThreadPool.execute(() -> getLoadBalancedServiceProperties(serviceName));
...

private void getLoadBalancedServiceProperties(String serviceName) {
    _newLb.getLoadBalancedServiceProperties(serviceName, new Callback<ServiceProperties>() {
      @Override
      public void onError(Throwable e) {
        LOG.error("Double read failure. Unable to read service properties from: " + serviceName, e);
      }

      @Override
      public void onSuccess(ServiceProperties result) {
        String clusterName = result.getClusterName();
        _dualReadStateManager.updateCluster(clusterName, DualReadModeProvider.DualReadMode.DUAL_READ);
        _newLb.getLoadBalancedServiceProperties(serviceName, new Callback<ServiceProperties>() {
          @Override
          public void onError(Throwable e) {
            LOG.error("Double read failure. Unable to read service properties from: " + serviceName, e);
          }

          @Override
          public void onSuccess(ServiceProperties result) {
            String clusterName = result.getClusterName();
            _dualReadStateManager.updateCluster(clusterName, DualReadModeProvider.DualReadMode.DUAL_READ);
            _newLb.getLoadBalancedClusterAndUriProperties(clusterName,
                new Callback<Pair<ClusterProperties, UriProperties>>() {
                  @Override
                  public void onError(Throwable e) {
                    LOG.error("Dual read failure. Unable to read cluster properties from: " + clusterName, e);
                  }

                  @Override
                  public void onSuccess(Pair<ClusterProperties, UriProperties> result) {
                    //TODO change back to debug
                    LOG.info("Dual read is successful. Get cluster and uri properties: " + result);
                  }
                });
          }
        });
}
```

d2/src/main/java/com/linkedin/d2/xds/LoadBalanceTaskPool/NewLoadBalancerTaskthreadPool.java
```diff
@@ -0,0 +1,29 @@
...
+public class NewLoadBalancerTaskthreadPool {
+  private ThreadPoolExecutor _threadPoolExecutor;
+  private int _corePoolSize = 2;
+  private int _maximumPoolSize = 3;
+  private long _keepAliveTime = 200;
+  private int _queueSize = 1000;
+
+  public NewLoadBalancerTaskthreadPool() {
+    RejectedExecutionHandler rejectedExecutionHandler = new NewBalanceTaskRejectedPolicy();
+    _threadPoolExecutor = new ThreadPoolExecutor(_corePoolSize, _maximumPoolSize, _keepAliveTime, TimeUnit.MILLISECONDS,
```
shivamgupta1, line 18: It's usually necessary to use LinkedIn's executor services (for eg. the one created from `ScheduledExecutorServiceFactory` – although that one is quite old and I remember seeing some newer ones) since they provide IC handling and instrumentation for sending metrics to inGraphs, etc.

It'll need to be wired in from `container`, maybe via `D2ClientConfig`, as is done for other d2 executors.

d2/src/main/java/com/linkedin/d2/xds/LoadBalanceTaskPool/NewBalanceGetPropertiesTask.java
shivamgupta1: +1, those are good arguments.

d2/src/main/java/com/linkedin/d2/xds/LoadBalanceTaskPool/NewLoadBalancerTaskthreadPool.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+
+public class NewLoadBalancerTaskthreadPool {
+  private ThreadPoolExecutor _threadPoolExecutor;
+  private int _corePoolSize = 2;
+  private int _maximumPoolSize = 3;
+  private long _keepAliveTime = 200;
+  private int _queueSize = 1000;
```
whutwhu, line 14: It would be better if these parameters could be managed through cfg2, possibly from the global configuration, allowing services to customize them or use default values if not specified. Once the migration from the oldLB to the newLB is complete, we can clean up all of these configurations, including the threadpool.

d2/src/main/java/com/linkedin/d2/xds/balancer/DualReadZkAndXdsLoadBalancerFactory.java
brycezhongqing: Thanks. That makes sense. For backward compatible, what I have done:
- use `D2ClientConfig` , which could be wired with `container`
- If there is no `ThreadPool` setted by `container`, the `DualReadLoadBalancer` will create a default `ThreadPool`

d2/src/main/java/com/linkedin/d2/xds/LoadBalanceTaskPool/NewBalanceGetPropertiesTask.java
brycezhongqing: Updated.

d2/src/main/java/com/linkedin/d2/xds/LoadBalanceTaskPool/NewLoadBalancerTaskthreadPool.java
```diff
@@ -0,0 +1,29 @@
+package com.linkedin.d2.xds.LoadBalanceTaskPool;
+
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.RejectedExecutionHandler;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+
+public class NewLoadBalancerTaskthreadPool {
```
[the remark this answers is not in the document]
brycezhongqing: Thanks. That make senses. Updated

d2/src/main/java/com/linkedin/d2/xds/LoadBalanceTaskPool/NewLoadBalancerTaskthreadPool.java
brycezhongqing: Thanks. That makes sense. Could I learn which `executor services` u mentiond ? As I see lot's of services will directly use `ThreadPool` to process tasks.

What I have done :
- use `D2ClientConfig` , which could be wired with `container`
- If there is no `ThreadPool` setted by `container`, the `DualReadLoadBalancer` will create a default `ThreadPool`

d2/src/main/java/com/linkedin/d2/xds/balancer/DualReadZkAndXdsLoadBalancerFactory.java
```diff
@@ -33,17 +34,20 @@ public class DualReadZkAndXdsLoadBalancerFactory implements LoadBalancerWithFaci
...
   {
     _zkLbFactory = new ZKFSLoadBalancerWithFacilitiesFactory();
     _xdsLbFactory = new XdsLoadBalancerWithFacilitiesFactory();
     _dualReadStateManager = dualReadStateManager;
+    _newNewLoadBalancerTaskThreadPool = new NewLoadBalancerTaskthreadPool();
```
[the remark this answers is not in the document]
brycezhongqing: What I have done :
- use `D2ClientConfig` , which could be wired with `container`
- If there is no `ThreadPool` setted by `container`, the `DualReadLoadBalancer` will create a default `ThreadPool`

d2/src/main/java/com/linkedin/d2/xds/LoadBalanceTaskPool/NewLoadBalancerTaskthreadPool.java
brycezhongqing: Thanks. That make sense.

What I have done :
- use `D2ClientConfig` , which could be wired with `container`
- If there is no `ThreadPool` setted by `container`, the `DualReadLoadBalancer` will create a default `ThreadPool`

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
```diff
@@ -109,35 +111,8 @@ public void getClient(Request request, RequestContext requestContext, Callback<T
...
-
-              @Override
-              public void onSuccess(Pair<ClusterProperties, UriProperties> result)
-              {
-                LOG.debug("Dual read is successful. Get cluster and uri properties: " + result);
-              }
-            });
-          }
-        });
+        _newNewLoadBalancerTaskThreadPool.execute(
```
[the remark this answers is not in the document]
brycezhongqing: Thanks. That make senses. Updated.

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
```diff
@@ -282,6 +280,29 @@ private DualReadModeProvider.DualReadMode getDualReadMode(String d2ServiceName)
...
+      this._loadBalancerThreadPool = loadBalancerThreadPool;
+    } else {
+      LOG.info("LoadBalancerTaskThreadPool is null, using default thread pool");
+      this._loadBalancerThreadPool =
+          new ThreadPoolExecutor(2, 3, 200, TimeUnit.MILLISECONDS, new ArrayBlockingQueue<>(1000), (r, executor) -> {
+            LOG.error("LoadBalancerTaskThreadPool rejected execution, isNewReady: " + _isNewLbReady);
+          });
+    }
+  }
+
```
PapaCharlie, line 305: I don't think this is a good idea. First of all, you have this backwards: if you wanted a default threadpool, you should have the default initialization in the `get` method, not the `set` method, otherwise your code can still throw a null pointer exception.

That being said, these types of critical fields should ideally never be uninitialized. In this case, you should inject the threadpool in the constructor. You will need to create a new constructor and deprecate the old one

PapaCharlie (changes requested):

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
```diff
@@ -282,6 +280,29 @@ private DualReadModeProvider.DualReadMode getDualReadMode(String d2ServiceName)
     return _dualReadStateManager.getServiceDualReadMode(d2ServiceName);
   }
 
+
+  /**
+   * Get the thread pool for load balancer tasks. If not set, a default thread pool will be used.
+   */
+  public ThreadPoolExecutor getLoadBalancerThreadPool() {
```
PapaCharlie, line 287: After adding the threadpool to the constructor you can get rid of these two getters and setters

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
```diff
@@ -643,6 +645,11 @@ public D2ClientBuilder setDualReadStateManager(DualReadStateManager dualReadStat
     return this;
   }
 
+  public D2ClientBuilder setLoadBalancerThreadPool(ThreadPoolExecutor loadBalancerThreadPool) {
```
PapaCharlie, line 648: It's better to make this a more generic signature here, otherwise it limits the threadpool implementations that can be used. In this case the minimum interface you need is `ExecutorService`, which is the basic interface for a threapool in Java.

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
```diff
@@ -109,35 +111,31 @@ public void getClient(Request request, RequestContext requestContext, Callback<T
...
-            _newLb.getLoadBalancedClusterAndUriProperties(clusterName, new Callback<Pair<ClusterProperties, UriProperties>>()
-            {
+        getLoadBalancerThreadPool().execute(
+            () -> _newLb.getLoadBalancedServiceProperties(serviceName, new Callback<ServiceProperties>() {
               @Override
-              public void onError(Throwable e)
-              {
-                LOG.error("Dual read failure. Unable to read cluster properties from: " + clusterName, e);
+              public void onError(Throwable e) {
+                LOG.error("Double read failure. Unable to read service properties from: " + serviceName, e);
```
PapaCharlie, line 118: Since you're modifying this code, can you fix this string concatenation in the logger?
```java
LOG.error("Double read failure. Unable to read service properties from: {}", serviceName, e);
```

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
```diff
@@ -109,35 +111,31 @@ public void getClient(Request request, RequestContext requestContext, Callback<T
...
-              {
-                LOG.debug("Dual read is successful. Get cluster and uri properties: " + result);
+              public void onSuccess(ServiceProperties result) {
+                String clusterName = result.getClusterName();
+                _dualReadStateManager.updateCluster(clusterName, DualReadModeProvider.DualReadMode.DUAL_READ);
+                _newLb.getLoadBalancedClusterAndUriProperties(clusterName,
+                    new Callback<Pair<ClusterProperties, UriProperties>>() {
+                      @Override
+                      public void onError(Throwable e) {
+                        LOG.error("Dual read failure. Unable to read cluster properties from: " + clusterName, e);
```
PapaCharlie, line 129: Same here, remove the string concatenation and use the log formatting:
```java
LOG.error("Dual read failure. Unable to read cluster properties from: {}",  clusterName, e);
```

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
```diff
@@ -109,35 +111,31 @@ public void getClient(Request request, RequestContext requestContext, Callback<T
...
+                      }
+
+                      @Override
+                      public void onSuccess(Pair<ClusterProperties, UriProperties> result) {
+                        LOG.debug("Dual read is successful. Get cluster and uri properties: " + result);
```
PapaCharlie, line 134: Same:
```java
LOG.debug("Dual read is successful. Get cluster and uri properties: {}", result);
```

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
```diff
@@ -291,5 +312,6 @@ public void shutdown(PropertyEventThread.PropertyEventShutdownCallback callback)
     });
 
     _oldLb.shutdown(callback);
+    _loadBalancerThreadPool.shutdown();
```
PapaCharlie, line 315: Technically you're getting this as a constructor parameter, so it's not your responsibility to shut it down. You would only have to shut it down if your had initialized it yourself

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
PapaCharlie: In this case, you want to do something like this:
```java
  @Deprecated
  public DualReadLoadBalancer(LoadBalancerWithFacilities oldLb, LoadBalancerWithFacilities newLb,
      @Nonnull DualReadStateManager dualReadStateManager)
  {
    // Using a direct executor here means the code is executed directly, blocking the caller. This means the old behavior is preserved.
    this(oldLb, newLb, dualReadStateManager, false, MoreExecutors.directExecutor());
    LOG.warn("Deprecated DualReadLoadBalancer constructor used without a threadpool");
  }

  public DualReadLoadBalancer(LoadBalancerWithFacilities oldLb, LoadBalancerWithFacilities newLb,
      @Nonnull DualReadStateManager dualReadStateManager, ExecutorService threadpool)
  {
    _oldLb = oldLb;
    _newLb = newLb;
    _dualReadStateManager = dualReadStateManager;
    _isNewLbReady = false;
    _threadpool = threadpool;
  }
```

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
```diff
@@ -53,13 +56,12 @@
...
  */
-public class DualReadLoadBalancer implements LoadBalancerWithFacilities
-{
+public class DualReadLoadBalancer implements LoadBalancerWithFacilities {
   private static final Logger LOG = LoggerFactory.getLogger(DualReadLoadBalancer.class);
   private final LoadBalancerWithFacilities _oldLb;
   private final LoadBalancerWithFacilities _newLb;
   private final DualReadStateManager _dualReadStateManager;
-
+  private ThreadPoolExecutor _loadBalancerThreadPool;
```
PapaCharlie, line 64: Nit: the name of the class is `DualReadLoadBalancer`, so it's redundant to call this field `_loadBalancer...`. The type of the field will change to `ExecutorService` according to my other comments, so it's best to just call it `_executor`

PapaCharlie (changes requested):

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
brycezhongqing: Updated

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
brycezhongqing: Updated

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
brycezhongqing: Upated~

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
brycezhongqing: After offline dicussion with @PapaCharlie . We thought we need to deprecated the old `contructor` and create a new one with backward compatible

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
brycezhongqing: Make senses. Remove the shutdown logic

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
brycezhongqing: Make sense. Rename the field to `executor`

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
brycezhongqing: Make sense. Update it to more generic signature :
`ExecutorService`

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
brycezhongqing: Yeap. Updated

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
```diff
@@ -53,22 +55,36 @@
  * In DUAL_READ mode, it reads from both the old and the new load balancer, but relies on the data from old
  * load balancer only.
  */
-public class DualReadLoadBalancer implements LoadBalancerWithFacilities
-{
+public class DualReadLoadBalancer implements LoadBalancerWithFacilities {
```
PapaCharlie, line 58: Ditto on the brace formatting, you have to leave it on the new line

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
```diff
@@ -53,22 +55,36 @@
...
+
   public DualReadLoadBalancer(LoadBalancerWithFacilities oldLb, LoadBalancerWithFacilities newLb,
-      @Nonnull DualReadStateManager dualReadStateManager)
+      @Nonnull DualReadStateManager dualReadStateManager, ExecutorService executor)
   {
     _oldLb = oldLb;
     _newLb = newLb;
     _dualReadStateManager = dualReadStateManager;
     _isNewLbReady = false;
+    if(executor == null){
```
PapaCharlie, line 80: Missing space before and after parens, plus style formatting:
```java
    if (executor == null)
    {
      // Using a direct executor here means the code is executed directly,
      // blocking the caller. This means the old behavior is preserved.
      _executor = MoreExecutors.newDirectExecutorService();
      LOG.warn("Deprecated DualReadLoadBalancer constructor used without a threadpool executor");
    }
    else
    {
      _executor = executor;
    }
```

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
```diff
@@ -109,35 +125,31 @@ public void getClient(Request request, RequestContext requestContext, Callback<T
...
+                      public void onError(Throwable e) {
+                        LOG.error("Dual read failure. Unable to read cluster properties from: {}", clusterName, e);
+                      }
+
+                      @Override
+                      public void onSuccess(Pair<ClusterProperties, UriProperties> result) {
+                        LOG.debug("Dual read is successful. Get cluster and uri properties: {}", result);
+                      }
+                    });
               }
```
PapaCharlie, line 151: Ditto here, make sure to match the code style for braces. I know it's kind of annoying...

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
```diff
@@ -643,6 +644,11 @@ public D2ClientBuilder setDualReadStateManager(DualReadStateManager dualReadStat
     return this;
   }
 
+  public D2ClientBuilder setLoadBalancerExecutor(ExecutorService loadBalancerExecutor) {
+    _config.loadBalancerExecutor = loadBalancerExecutor;
+    return this;
+  }
+
```
PapaCharlie, line 651: It's unfortunate but you need to match the local code style here and use this formatting for braces:
```java
  public D2ClientBuilder setLoadBalancerExecutor(ExecutorService loadBalancerExecutor)
  {
    _config.loadBalancerExecutor = loadBalancerExecutor;
    return this;
  }
```

PapaCharlie (changes requested):

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
brycezhongqing: Will keep with origin format. Thanks

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
brycezhongqing: Updated

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
brycezhongqing: Updated

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
```diff
@@ -201,7 +201,8 @@ public D2Client build()
                   _config.serviceDiscoveryEventEmitter,
                   _config.dualReadStateManager,
                   _config.xdsExecutorService,
-                  _config.xdsStreamReadyTimeout
+                  _config.xdsStreamReadyTimeout,
+                  _config.loadBalancerExecutor
```
bohhyang, line 205: to be more descriptive, "dualReadNewLbExecutor" would be better?

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
```diff
@@ -59,16 +61,35 @@ public class DualReadLoadBalancer implements LoadBalancerWithFacilities
   private final LoadBalancerWithFacilities _oldLb;
   private final LoadBalancerWithFacilities _newLb;
   private final DualReadStateManager _dualReadStateManager;
-
+  private ExecutorService _executor;
```
bohhyang, line 64: nit: newLbExecutor

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
```diff
@@ -59,16 +61,35 @@ public class DualReadLoadBalancer implements LoadBalancerWithFacilities
...
     _oldLb = oldLb;
     _newLb = newLb;
     _dualReadStateManager = dualReadStateManager;
     _isNewLbReady = false;
+    if(executor == null)
+    {
+      // Using a direct executor here means the code is executed directly,
+      // blocking the caller. This means the old behavior is preserved.
+      _executor = MoreExecutors.newDirectExecutorService();
+      LOG.warn("Deprecated DualReadLoadBalancer constructor used without a threadpool executor");
```
bohhyang, line 87: this error message is emitted on line 72. This one can be alerting about the executor passed in being null instead.

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
```diff
@@ -109,35 +130,36 @@ public void getClient(Request request, RequestContext requestContext, Callback<T
...
-            _dualReadStateManager.updateCluster(clusterName, DualReadModeProvider.DualReadMode.DUAL_READ);
-            _newLb.getLoadBalancedClusterAndUriProperties(clusterName, new Callback<Pair<ClusterProperties, UriProperties>>()
+        _executor.execute(
+            () -> _newLb.getLoadBalancedServiceProperties(serviceName, new Callback<ServiceProperties>()
             {
               @Override
               public void onError(Throwable e)
               {
-                LOG.error("Dual read failure. Unable to read cluster properties from: " + clusterName, e);
+                LOG.error("Double read failure. Unable to read service properties from: {}", serviceName, e);
```
bohhyang, line 139: nit" "Double" should be "Dual" instead

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.46.10] - 2023-11-13
```
bohhyang, line 17: since later you'll make container side change that requires this restli change, you'll need to bump the minor version here, so that users who adopt that container version but still use an older restli version will fail to build, forcing them to bump restli version too.

bohhyang (changes requested):

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
brycezhongqing: Thanks. Updated

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
brycezhongqing: Thanks. Updated~

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
brycezhongqing: Thanks for reminding. Updated

CHANGELOG.md
brycezhongqing: Thanks for the suggestion. Have updated the `minor` version

bohhyang (approved):
LGTM. Please resolve the conflict with master (Paul's change is also bumping the minor version, so you'll need to bump to 29.48.0)

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
PapaCharlie: Haha I opened an issue about this naming because it used to be `loadBalancerExecutor` and said that it was redundant, since the name of the class is `DualReadLoadBalancer`.

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
```diff
@@ -59,16 +61,35 @@ public class DualReadLoadBalancer implements LoadBalancerWithFacilities
...
-
+  private ExecutorService _executor;
   private boolean _isNewLbReady;
 
+  @Deprecated
   public DualReadLoadBalancer(LoadBalancerWithFacilities oldLb, LoadBalancerWithFacilities newLb,
       @Nonnull DualReadStateManager dualReadStateManager)
+  {
+    this(oldLb, newLb, dualReadStateManager, null);
+    LOG.warn("Deprecated DualReadLoadBalancer constructor used without a threadpool");
```
PapaCharlie, line 72: Agree with Bohan, this can be removed now

PapaCharlie (changes requested):

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
brycezhongqing: Agree. Removed.

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
```diff
@@ -59,16 +61,34 @@ public class DualReadLoadBalancer implements LoadBalancerWithFacilities
...
+  }
+
+  public DualReadLoadBalancer(LoadBalancerWithFacilities oldLb, LoadBalancerWithFacilities newLb,
+      @Nonnull DualReadStateManager dualReadStateManager, ExecutorService newLbExecutor)
   {
     _oldLb = oldLb;
     _newLb = newLb;
     _dualReadStateManager = dualReadStateManager;
     _isNewLbReady = false;
+    if(newLbExecutor == null)
```
PapaCharlie, line 81: Nit: space after the `if`

PapaCharlie (approved):

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
brycezhongqing: Thanks. Fix that.

PapaCharlie (approved):

PR #946, opened 2023-11-27
Remove usage of Optional from SimpleLoadBalancer

PapaCharlie:
According to benchmark results, this is creating significant garbage. Because this is not exposed as an external API, directly returning null instead of an Optional is safe and will be significantly less expensive.


bohhyang (approved):

bohhyang:
reminder for changing the changelog and version

PR #950, opened 2023-12-06
correct where to increment the clusterNotFound count and adjust quarantine log level

bohhyang:
##Summary##
Currently the clusterNotFound count is incremented when there is no Uris in the cluster in addition to when the cluster config is not found. This change moves it to a place which only increment for cluster config not found.

Also, changing the log level about long-quarantined hosts to Info instead of Error, because 1) often it's just a temporary small capacity which's not causing availability issue. 2) it floods the upstream app's log triggering alerts and/or creating EXC tickets that the upstream app owner is not responsible for. Clarified the log message to show the issue is the downstream host's health. 

##Test done##
build and test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.48.4] - 2023-12-06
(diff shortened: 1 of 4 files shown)


brycezhongqing (approved):
LGTM

PR #951, opened 2023-12-06
Rename next to nextPageToken to keep it uniform with grpc

jagriti-madan:
./gradlew idea
./gradlew build -x test

### DIFF
--- a/restli-server/src/main/java/com/linkedin/restli/server/CursorCollectionResult.java
+++ b/restli-server/src/main/java/com/linkedin/restli/server/CursorCollectionResult.java
@@ -20,16 +20,4 @@ public class CursorCollectionResult<T extends RecordTemplate> extends Collection
   {
     super(elements, null, pagination);
   }
-
-  /**
-   * Constructor
-   *
-   * @param elements       List of elements in the current page.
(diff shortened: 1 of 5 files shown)


mchen07 (approved):

restli-server/src/main/java/com/linkedin/restli/server/CursorCollectionResult.java
```diff
@@ -20,4 +20,16 @@ public CursorCollectionResult(final List<T> elements, CursorPagination paginatio
...
   }
+
+  /**
+   * Constructor
+   *
+   * @param elements       List of elements in the current page.
+   * @param total          Provides the total elements
+   * @param pagination     The cursor pagination metadata.
+   */
+  public CursorCollectionResult(final List<T> elements, final Integer total, CursorPagination pagination)
```
mchen07, line 31: Will this mix offset-based pagination with cursor-based pagination?

jagriti-madan: The idea is to be able to return total number of entities if a partner wants it in cursor based pagination. 
Should we rather have a different parameter for it then?

PR #952, opened 2023-12-12
add debug log to dual read caches

bohhyang:
## Summary 
We've seen the Cluster, Service, Uri properties OutOfSync metrics in DualReadLoadBalancer sensor being high for some apps (voyager-api-groups-dash, registration-mt, etc). The metric seems to increment on every update of the property (see attached screenshot of UriPropertiesOutOfSyncCount metric). It bumps up periodically, likely during downstream deployments.
Screenshot 2023-12-12 at 9 15 17 AM

This change added debug logs to when entries are added/compared in the dual read caches to further debug the issue.

Another change is for SI-36476 where D2ClientJmxManager tries to find the right jmx name for d2 services being removed at shutdown (the name depends on dual read mode), causing the executor for getting the dual read mode throws errors since it's already shut down. Added a check to skip getting dual read mode if the executor is shut down (the last read dual read mode will be used).

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancerMonitor.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancerMonitor.java
@@ -103,12 +103,6 @@ public abstract class DualReadLoadBalancerMonitor<T>
     _rateLimitedLogger.debug("Current entries on dual read caches: {}", entriesLogMsg);
   }
 
-  private String getEntriesMessage(boolean fromNewLb, CacheEntry<T> oldE, CacheEntry<T> newE)
-  {
-    return String.format("Old LB: {}, New LB: {}.",
-        fromNewLb? oldE : newE, fromNewLb? newE : oldE);
-  }
-
   abstract void incrementEntryOutOfSyncCount();
 
   abstract void decrementEntryOutOfSyncCount();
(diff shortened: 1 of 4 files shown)


d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancerMonitor.java
```diff
@@ -71,29 +71,37 @@ public void reportData(String propertyName, T property, String propertyVersion,
...
-    if (entry != null && Objects.equals(entry._version, propertyVersion))
+    if (entryToCompare != null && Objects.equals(entryToCompare._version, propertyVersion))
     {
-      CacheEntry<T> newEntry = new CacheEntry<>(propertyVersion, getTimestamp(), property);
-      if (!isEqual(entry, newEntry))
+      if (!isEqual(entryToCompare, newEntry))
       {
         _rateLimitedLogger.warn("Received mismatched properties from dual read. Old LB: {}, New LB: {}",
-            fromNewLb ? entry : newEntry, fromNewLb ? newEntry : entry);
+            fromNewLb ? entryToCompare : newEntry, fromNewLb ? newEntry : entryToCompare);
```
PapaCharlie, line 82: Too bad this logic is repeated in multiple places... I can't think of a good way to simplify it though, so I guess it's fine to leave it as is unless you can think of something?

PapaCharlie (approved):

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadStateManager.java
```diff
@@ -183,6 +183,11 @@ private void reportUriPropertiesData(String propertyName, UriProperties property
    */
   public void checkAndSwitchMode(String d2ServiceName)
   {
+    if (_executorService.isShutdown())
+    {
+      return;
```
shivamgupta1, line 188: add info log here?

bohhyang: sure. added

brycezhongqing (approved):
LGTM

PR #955, opened 2023-12-14
Translate optional map/arrays with defaults to Avro without breaking

dg-builder:


data-avro/src/main/java/com/linkedin/data/avro/DefaultDataToAvroConvertCallback.java
```diff
@@ -137,9 +137,19 @@ protected Object translateField(List<Object> path, Object fieldValue, RecordData
           if (_options.getOptionalDefaultMode() != OptionalDefaultMode.TRANSLATE_TO_NULL &&
               field.getDefault() != null)
           {
-            throw new IllegalArgumentException(
-                message(path,
-                    "cannot translate absent optional field (to have null value) because this field is optional and has a default value"));
+            DataSchema.Type fieldDefaultValueType = field.getType().getType();
+            // punt on other default values for now (too complex to handle)
```
mchen07, line 141: If we are not handling all cases, what usecases do we unblock?

dg-builder: The Python Gradle plugin (LI internal) always translates to Avro. I will share the exact PR in Slack.

data-avro/src/main/java/com/linkedin/data/avro/DefaultDataToAvroConvertCallback.java
```diff
@@ -137,9 +137,19 @@ protected Object translateField(List<Object> path, Object fieldValue, RecordData
...
+            if (fieldDefaultValueType == DataSchema.Type.UNION || fieldDefaultValueType == DataSchema.Type.RECORD
```
mchen07, line 142: In #133, it already ruled out UNION case? Also for TYPEREF, we should look at dereferencedType?

PR #956, opened 2023-12-14
fix publishing uri and cluster properties for symlink clusters

bohhyang:
## Summary
gcn-39955
After switched to observer-only mode, job-postings-mt was not able to make downstream calls to d2 services under a symlink cluster. We found that although the uri nodes were fetched correctly for symlink cluster, it was merged with the real cluster name, instead of the symlink cluster name (e.g: FooCluster-prod-ltx1 instead of $FooClusterMaster). The cluster name in the merged uri properties is used (by UriLoadBalancerSubscriber) to find d2 services associated with it, which fails since the services are managed under the symlink cluster name.

This change:
1. merges the uri properties under the symlink cluster name. 
2. report the uri properties to dual read monitoring under the symlink cluster name. 
3. removed the additional publishing under the real cluster name when it's publish for symlink cluster.
4. added jmx metrics to count the number of symlink clusters, the number of total hosts (tracker clients) for each d2 service in default partition and all partitions. 

## Test Done
Unit tests. 

Manual testing with QEI d2-proxy with observer-only config (<property name="indisConfigProvider.dualReadMode" value="NEW_LB_ONLY"/>):
curli -v "d2://cart" --d2-proxy-url "http://localhost:21360/d2/" -X OPTIONS --force-insecure-d2
< Content-Type: application/json
< x-linkedin-processing-colo: ei-ltx1
< x-linkedin-processing-machine: ltx1-app4245.stg.linkedin.com
{
  "models": {
    "com.linkedin.payments.ContactInfo": {
      "namespace": "com.linkedin.payments",
      "name": "ContactInfo",
      "doc": "Contact Information.",
      "fields": [
        {
          "type": "string",
          "name": "firstName",
          "doc": "First name."
        },
………
      ],
      "type": "record"
    }
  },
  "resources": {
    "com.linkedin.payments.client.cart": {
      "schema": "com.linkedin.payments.Cart",
      "path": "/cart",
      "namespace": "com.linkedin.payments.client",
      "name": "cart",
      "doc": "Cart restli resource. Handles all the operations to the cart.\n\ngenerated from: com.linkedin.oms.bps.rest.impl.CartResource",
      "collection": {
..............
        "supports": [
          "create",
          "update"
        ],
        "methods": [
          {
            "method": "create",
            "javaMethodName": "create"
          },
          {
            "method": "update",
            "javaMethodName": "update"
          }
        ]
      },
      "resourceClass": "com.linkedin.oms.bps.rest.impl.CartResource"
    }
  }
}

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/StateUpdater.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/StateUpdater.java
@@ -159,8 +159,7 @@ public class StateUpdater
   int getTotalHostsInAllPartitions()
   {
     return _partitionLoadBalancerStateMap.values().stream()
-        .map(partitionState -> partitionState.getTrackerClients().size())
-        .mapToInt(Integer::intValue)
+        .mapToInt(partitionState -> partitionState.getTrackerClients().size())
         .sum();
   }
 
--- a/d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/StateUpdater.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/StateUpdater.java
@@ -154,6 +154,17 @@ public class StateUpdater
   }
 
   /**
+   * Return the total tracker clients in all partitions regardless of their statuses.
+   */
+  int getTotalHostsInAllPartitions()
+  {
+    return _partitionLoadBalancerStateMap.values().stream()
+        .map(partitionState -> partitionState.getTrackerClients().size())
+        .mapToInt(Integer::intValue)
+        .sum();
+  }
+
+  /**
    * Return the first valid partition id. This is mainly used for monitoring at least one valid partition.
    */
   int getFirstValidPartitionId()
--- a/d2/src/main/java/com/linkedin/d2/discovery/stores/zk/SymlinkUtil.java
+++ b/d2/src/main/java/com/linkedin/d2/discovery/stores/zk/SymlinkUtil.java
@@ -64,4 +64,9 @@ public class SymlinkUtil
   {
     return (firstSymlinkIndex(path) < 0) ? false : true;
   }
+
+  public static boolean isSymlinkNodeOrPath(String nodeNameOrPath)
+  {
+    return nodeNameOrPath != null && nodeNameOrPath.indexOf(SYMLINK_PREFIX) >= 0;
+  }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.48.7] - 2023-12-13
+- fix publishing uri and cluster properties for symlink clusters
+
 ## [29.48.6] - 2023-12-12
 - Rename next to nextPageToken in standardized models for cursor based pagination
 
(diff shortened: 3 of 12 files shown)


brycezhongqing (approved):
Just double confirm. From former comment, there is a  corner case:
> Also, rarely and possibly, the original cluster could have subscribers too when calls are made directly to the original cluster, so we publish for it too.

So when the service use `symlink cluster` :
the original logic is :

- publish symlinkName
- publish raw clusterName

But now we only publish symlinkName, right ?

bohhyang:
> Just double confirm. From former comment, there is a corner case:
> 
> > Also, rarely and possibly, the original cluster could have subscribers too when calls are made directly to the original cluster, so we publish for it too.
> 
> So when the service use `symlink cluster` : the original logic is :
> 
> * publish symlinkName
> * publish raw clusterName
> 
> But now we only publish symlinkName, right ?

yes, that's right. It's actually impossible to make calls directly to the original/real cluster. And the two publishes (even with different cluster names) are publishing the same data, which can have unknown risky behaviors in updating the load balancer state (and its listeners).

d2/src/main/java/com/linkedin/d2/discovery/stores/zk/SymlinkUtil.java
```diff
@@ -64,4 +64,9 @@ public static boolean containsSymlink(String path)
   {
     return (firstSymlinkIndex(path) < 0) ? false : true;
   }
+
+  public static boolean isSymlinkNodeOrPath(String nodeNameOrPath)
+  {
+    return nodeNameOrPath != null && nodeNameOrPath.indexOf(SYMLINK_PREFIX) >= 0;
```
brycezhongqing, line 70: How about change to 
```
return nodeNameOrPath != null && nodeNameOrPath.contains(SYMLINK_PREFIX);
```

d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/StateUpdater.java
```diff
@@ -153,6 +153,17 @@ PartitionState getPartitionState(int partitionId)
     return _partitionLoadBalancerStateMap.get(partitionId);
   }
 
+  /**
+   * Return the total tracker clients in all partitions regardless of their statuses.
+   */
+  int getTotalHostsInAllPartitions()
```
brycezhongqing, line 159: Just double check how about
```
 return _partitionLoadBalancerStateMap.values().stream()
        .mapToInt(partitionState -> partitionState.getTrackerClients().size())
        .sum();
```

brycezhongqing (commented):
Overall LGTM. Just some little comments

d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/StateUpdater.java
bohhyang: ah, yeah, why not. Thx.

d2/src/main/java/com/linkedin/d2/discovery/stores/zk/SymlinkUtil.java
brycezhongqing: Oh. Thanks for ur explanation.You are right, if the `SYMLINK_PREFIX` is the `char` , it's prefer to use `indexOf`

PR #958, opened 2023-12-19
add warn logs about invalid property versions

bohhyang:
## Summary
As the title. In case bad uri property version is produced for any edge cases (connection loss, etc.), we want to warn that.

## Test Done
build and test

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/properties/UriPropertiesMerger.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/properties/UriPropertiesMerger.java
@@ -39,12 +39,10 @@ public class UriPropertiesMerger implements ZooKeeperPropertyMerger<UriPropertie
 
     String clusterName = propertyName;
     long maxVersion = -1;
-    long propertyVersion;
 
     for (UriProperties property : propertiesToMerge)
     {
-      propertyVersion = property.getVersion();
-      maxVersion = Long.max(maxVersion, propertyVersion);
+      maxVersion = Long.max(maxVersion, property.getVersion());
       for (Map.Entry<URI, Map<Integer, PartitionData>> entry : property.getPartitionDesc().entrySet())
       {
         partitionData.put(entry.getKey(), entry.getValue());
--- a/d2/src/main/java/com/linkedin/d2/balancer/properties/UriPropertiesMerger.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/properties/UriPropertiesMerger.java
@@ -23,9 +23,14 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 
 public class UriPropertiesMerger implements ZooKeeperPropertyMerger<UriProperties>
 {
+  private static final Logger _log = LoggerFactory.getLogger(UriPropertiesMerger.class);
+
   @Override
   public UriProperties merge(String propertyName, Collection<UriProperties> propertiesToMerge)
   {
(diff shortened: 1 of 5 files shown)


d2/src/main/java/com/linkedin/d2/balancer/properties/UriPropertiesMerger.java
```diff
@@ -34,10 +39,16 @@ public UriProperties merge(String propertyName, Collection<UriProperties> proper
...
     String clusterName = propertyName;
     long maxVersion = -1;
+    long propertyVersion;
 
     for (UriProperties property : propertiesToMerge)
     {
-      maxVersion = Long.max(maxVersion, property.getVersion());
+      propertyVersion = property.getVersion();
+      maxVersion = Long.max(maxVersion, propertyVersion);
+      if (maxVersion == -1)
```
brycezhongqing, line 48: Should we change the logic to 
```
if (propertyVersion <= -1) ?
```
As if the `propertiesToMerges` is (1, -1, -10), there is not warning.

d2/src/main/java/com/linkedin/d2/balancer/properties/UriPropertiesMerger.java
```diff
@@ -23,9 +23,14 @@
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 
 public class UriPropertiesMerger implements ZooKeeperPropertyMerger<UriProperties>
 {
+  private static final Logger _log = LoggerFactory.getLogger(UriPropertiesMerger.class);
```
ZoabKapoor, line 32: I was going to comment that I thought it's convention to use `LOG` for a `static final` logger, but I see that the [existing codebase](https://github.com/search?q=repo%3Alinkedin%2Frest.li%20%22private%20static%20final%20Logger%22&type=code) is inconsistent 😂 So I think this is fine

ZoabKapoor (approved):

d2/src/main/java/com/linkedin/d2/balancer/properties/UriPropertiesMerger.java
bohhyang: that's ok since the -1, -10 will warned already in [ZooKeeperEphemeralStore.java](https://github.com/linkedin/rest.li/pull/958/files#diff-77b01f0ad40699c361791ff221b474e25279aba3277103a4d7abe7ad5d1a2d3b). Here we just warn for the max version result.

brycezhongqing: Make sense. Thanks

brycezhongqing (approved):
LGTM

d2/src/main/java/com/linkedin/d2/balancer/properties/UriPropertiesMerger.java
```diff
@@ -48,6 +55,10 @@ public UriProperties merge(String propertyName, Collection<UriProperties> proper
       }
     }
 
+    if (maxVersion == -1)
+    {
+      _log.warn("Merged Uri properties has invalid version -1. It should be > -1.");
+    }
```
shivamgupta1: Should these checks be outside the `UriPropertiesMerger` code and somewhere in the dual read comparison code, since this class isn't really concerned and/or responsible about such validations?

It might also result in a lot of logs where not really required.

d2/src/main/java/com/linkedin/d2/balancer/properties/UriPropertiesMerger.java
bohhyang: yeah, makes sense, just fixed!

d2/src/main/java/com/linkedin/d2/balancer/properties/UriPropertiesMerger.java
bohhyang: the one in [ZooKeeperEphemeralStore.java](https://github.com/linkedin/rest.li/pull/958/files#diff-77b01f0ad40699c361791ff221b474e25279aba3277103a4d7abe7ad5d1a2d3b) logs about the invalid data and property name. Here is where the merge-version logic happens, so it's responsible for the merged final version.
The dual read monitor is not responsible for this validation, and the property there is generic. But this version issue only applies to uris. 
I've add a warn log to the xds flow uri property as well.

> It might also result in a lot of logs where not really required.
Any of this warn message is important since it means some invalid zk (and future Kafka) data and need our attention to fix.

PR #960, opened 2023-12-21
bump minor version

bohhyang:
As in the title. Container PR will require the new pegasus code. We need to bump the minor version so that MPs which bump container but pin pegasus to a lower version can get build errors (patch version difference won't work, need minor version bump) hinting to bump the corresponding pegasus version too. Otherwise, their build will succeed but the app will crash at runtime since the old pegasus doesn’t have the new code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
+- bump minor version
 
 ## [29.48.9] - 2023-12-20
 - No-op proxy detector for xDS gRPC channel builder.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.49.0] - 2023-12-21
+- bump minor version
```
zackthehuman, line 18: Let's add a little more context:

> Bump minor version due to internal LinkedIn tooling requirement. No functional changes.

zackthehuman (changes requested):

CHANGELOG.md
bohhyang: makes sense. Updated.

zackthehuman (approved):

PR #961, opened 2023-12-22
Use a separate indis warmup executor service

bohhyang:
## Summary

As reported, some apps ramped to Indis timed out at warmup causing d2 client startup timeout, the warmup executor was a single-thread executor shared by ZK and Indis flow, leading to Indis warmup blocks ZK warmup.

While the RC fix will be reducing the rate limiting on Indis Observer, here we add a separate executor for indis warmup to prevent blocking/affecting ZK flow in the future.

## Test Done
build and test

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
@@ -541,6 +541,18 @@ public class D2ClientBuilder
     return this;
   }
 
+  public D2ClientBuilder setStartUpExecutorService(ScheduledExecutorService executorService)
+  {
+    _config.startUpExecutorService = executorService;
+    return this;
+  }
+
+  public D2ClientBuilder setIndisStartUpExecutorService(ScheduledExecutorService executorService)
+  {
+    _config.indisStartUpExecutorService = executorService;
+    return this;
+  }
+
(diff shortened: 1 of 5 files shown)


d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
```diff
@@ -104,6 +104,13 @@ public D2Client build()
       executorsToShutDown.add(_config.startUpExecutorService);
     }
 
+    if (_config.indisStartUpExecutorService == null)
```
shivamgupta1, line 107: Might also need to create a setter `setIndisStartUpExecutor` to allow setting it from container.

bohhyang: we currently don't have one for ZK, just maybe there is no use case for it, but sure, will add.

shivamgupta1: That's probably true. The xDS requests are gRPC calls though, so it might be better to add IC support for the executors to avoid unwanted log messages.

bohhyang: sounds good. done

shivamgupta1 (approved):

PR #962, opened 2023-12-22
Make warmup to respect dual read mode, and separate warmup configs for indis

bohhyang:
## Summary 
Current warmup starts for both ZK and xDS regardless of dual read mode, and the logic of fetching dual read mode (service data) blocks the warmup timeout which fails app startup.

This change: 
1. make warmup to respect the dual read mode per d2 service.
2. move the blocking service data fetching logic into a separate task and set a timeout, so that it won't blocks the warmup timeout callback to be called (so startup still has a chance to complete).
3. add separate configs for Indis warmup timeout and concurrent requests. We could tune them in the future as needed depending on how rate limiting on Indis observer will go. And wait for indis warmup (thus use the indis warmup timeout config) under observer-only mode during startup (instead of waiting for zk warmup). 
4. add checks to dual read load balancer monitoring to skip putting duplicate data into the report caches (which incorrectly increase outOfSync metric).

## Testing Done
1. added unit tests
2. tested with d2-proxy warming up some 200 services under different dual read modes, see the corresponding container PR for logs.

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
@@ -97,15 +97,15 @@ public class DualReadLoadBalancer implements LoadBalancerWithFacilities
     // Prefetch the global dual read mode
     DualReadModeProvider.DualReadMode mode = _dualReadStateManager.getGlobalDualReadMode();
 
-    _newLb.start(getStartUpCallback(true,
-        null //mode == DualReadModeProvider.DualReadMode.NEW_LB_ONLY ? callback : null
+    // if in new-lb-only mode, new lb needs to start successfully to call the callback. Otherwise, the old lb does.
+    // Use a separate executor service to start the new lb, so both lbs can start concurrently.
+    _newLbExecutor.execute(() -> _newLb.start(getStartUpCallback(true,
+        mode == DualReadModeProvider.DualReadMode.NEW_LB_ONLY ? callback : null)
     ));
 
-    // Call back will succeed as long as the old balancer is successfully started. New load balancer failure
-    // won't block application start up.
     _oldLb.start(getStartUpCallback(false,
-        callback //mode == DualReadModeProvider.DualReadMode.NEW_LB_ONLY ? null : callback
-        ));
+        mode == DualReadModeProvider.DualReadMode.NEW_LB_ONLY ? null : callback
+    ));
   }
 
   private Callback<None> getStartUpCallback(boolean isForNewLb, Callback<None> callback)
--- a/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
@@ -95,28 +95,50 @@ public class DualReadLoadBalancer implements LoadBalancerWithFacilities
   public void start(Callback<None> callback)
   {
     // Prefetch the global dual read mode
-    _dualReadStateManager.checkAndSwitchMode(null);
+    DualReadModeProvider.DualReadMode mode = _dualReadStateManager.getGlobalDualReadMode();
 
-    _newLb.start(new Callback<None>()
-    {
-      @Override
-      public void onError(Throwable e)
-      {
-        LOG.warn("Failed to start new load balancer. Fall back to read from old balancer only", e);
-        _isNewLbReady = false;
-      }
-
-      @Override
-      public void onSuccess(None result)
-      {
-        LOG.info("New load balancer successfully started");
-        _isNewLbReady = true;
-      }
-    });
+    _newLb.start(getStartUpCallback(true,
+        null //mode == DualReadModeProvider.DualReadMode.NEW_LB_ONLY ? callback : null
+    ));
 
     // Call back will succeed as long as the old balancer is successfully started. New load balancer failure
     // won't block application start up.
-    _oldLb.start(callback);
+    _oldLb.start(getStartUpCallback(false,
+        callback //mode == DualReadModeProvider.DualReadMode.NEW_LB_ONLY ? null : callback
+        ));
+  }
+
+  private Callback<None> getStartUpCallback(boolean isForNewLb, Callback<None> callback)
+  {
+    return new Callback<None>() {
+      @Override
+      public void onError(Throwable e) {
+        LOG.warn("Failed to start {} load balancer.", isForNewLb ? "new" : "old", e);
+        if (isForNewLb)
+        {
+          _isNewLbReady = false;
+        }
+
+        if (callback != null)
+        {
+          callback.onError(e);
+        }
+      }
+
+      @Override
+      public void onSuccess(None result) {
+        LOG.info("{} load balancer successfully started", isForNewLb ? "New" : "Old");
+        if (isForNewLb)
+        {
+          _isNewLbReady = true;
+        }
+
+        if (callback != null)
+        {
+          callback.onSuccess(None.none());
+        }
+      }
+    };
   }
 
   @Override
--- a/d2/src/main/java/com/linkedin/d2/balancer/util/WarmUpLoadBalancer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/util/WarmUpLoadBalancer.java
@@ -72,6 +72,7 @@ public class WarmUpLoadBalancer extends LoadBalancerWithFacilitiesDelegator
   private final ScheduledExecutorService _executorService;
   private final DownstreamServicesFetcher _downstreamServicesFetcher;
   private final DualReadStateManager _dualReadStateManager;
+  private final boolean _isIndis; // whether warming up for Indis (false means warming up for ZK)
   private volatile boolean _shuttingDown = false;
 
   /**
@@ -86,12 +87,12 @@ public class WarmUpLoadBalancer extends LoadBalancerWithFacilitiesDelegator
                             int warmUpTimeoutSeconds, int concurrentRequests)
   {
     this(balancer, serviceWarmupper, executorService, d2FsDirPath, d2ServicePath, downstreamServicesFetcher,
-        warmUpTimeoutSeconds, concurrentRequests, null);
+        warmUpTimeoutSeconds, concurrentRequests, null, false);
   }
 
   public WarmUpLoadBalancer(LoadBalancerWithFacilities balancer, WarmUpService serviceWarmupper, ScheduledExecutorService executorService,
       String d2FsDirPath, String d2ServicePath, DownstreamServicesFetcher downstreamServicesFetcher,
-      int warmUpTimeoutSeconds, int concurrentRequests, DualReadStateManager dualReadStateManager)
+      int warmUpTimeoutSeconds, int concurrentRequests, DualReadStateManager dualReadStateManager, boolean isIndis)
   {
     super(balancer);
     _serviceWarmupper = serviceWarmupper;
@@ -104,6 +105,7 @@ public class WarmUpLoadBalancer extends LoadBalancerWithFacilitiesDelegator
     _outstandingRequests = new ConcurrentLinkedDeque<>();
     _usedServices = new HashSet<>();
     _dualReadStateManager = dualReadStateManager;
+    _isIndis = isIndis;
   }
 
   @Override
@@ -166,30 +168,6 @@ public class WarmUpLoadBalancer extends LoadBalancerWithFacilitiesDelegator
           return;
         }
 
-        if (_dualReadStateManager != null)
-        {
-          DualReadModeProvider dualReadModeProvider = _dualReadStateManager.getDualReadModeProvider();
-          serviceNames.forEach(serviceName ->
-          {
-            DualReadModeProvider.DualReadMode dualReadMode = dualReadModeProvider.getDualReadMode(serviceName);
-            _dualReadStateManager.updateService(serviceName, dualReadMode);
-
-            getLoadBalancedServiceProperties(serviceName, new Callback<ServiceProperties>()
-            {
-              @Override
-              public void onError(Throwable e)
-              {
-                LOG.warn("Failed to warm up dual read mode for service: " + serviceName, e);
-              }
-
-              @Override
-              public void onSuccess(ServiceProperties result)
-              {
-                _dualReadStateManager.updateCluster(result.getClusterName(), dualReadMode);
-              }
-            });
-          });
-        }
         WarmUpTask warmUpTask = new WarmUpTask(serviceNames, timeoutCallback);
 
         // get the min value because it makes no sense have an higher concurrency than the number of request to be made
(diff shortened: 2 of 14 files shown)


d2/src/main/java/com/linkedin/d2/balancer/util/WarmUpLoadBalancer.java
```diff
@@ -242,8 +220,50 @@ void execute()
...
+          serviceName);
 
+      // warm up dual read mode for the service and its belonging cluster. This is needed so that when the actual data
+      // of service/cluster/uri is received, they can be reported to dual read monitoring under dual read mode.
+      // NOTE: this will actually fetch the service data SYNCHRONOUSLY.
+      if (_dualReadStateManager != null)
+      {
+        DualReadModeProvider.DualReadMode finalDualReadMode = dualReadMode;
+        // NOTE: this call blocks!
+        getLoadBalancedServiceProperties(serviceName, new Callback<ServiceProperties>()
```
shivamgupta1, line 249: Since this is blocking, this will break the current warmup flow in which executions finish instantaneously allowing multiple requests to be made simultaneously from the single-threaded executor.

d2/src/main/java/com/linkedin/d2/balancer/util/WarmUpLoadBalancer.java
```diff
@@ -242,8 +220,50 @@ void execute()
...
+      {
+        DualReadModeProvider dualReadModeProvider = _dualReadStateManager.getDualReadModeProvider();
+        dualReadMode = dualReadModeProvider.getDualReadMode(serviceName);
+        _dualReadStateManager.updateService(serviceName, dualReadMode);
+
+        // dual read mode can be fine-tuned per service, check if the specific mode for this service should start
+        // the warm-up. If not, skip this service.
+        if (!isModeToWarmUp(dualReadMode, _isIndis))
+        {
+          return;
```
shivamgupta1, line 235: It shouldn't be possible to return from here since this would kill one of the warmup chains ("concurrent requests"). Eg. if there are 10 services per batch, one of them would get killed right here and the next batch would have only 9, since the 10th one was supposed to be executed following this one's callback completion.

d2/src/main/java/com/linkedin/d2/balancer/util/WarmUpLoadBalancer.java
bohhyang: moved the whole fetching dual read mode logic to a separate executor task.

d2/src/main/java/com/linkedin/d2/balancer/util/WarmUpLoadBalancer.java
```diff
@@ -104,105 +107,160 @@ public WarmUpLoadBalancer(LoadBalancerWithFacilities balancer, WarmUpService ser
...
+        Future<?> prepareTaskFuture = _executorService.submit(() -> prepareWarmUp());
+        try
+        {
+          prepareTaskFuture.get(_warmUpTimeoutSeconds, TimeUnit.SECONDS);
+        }
+        catch (TimeoutException e)
+        {
+          LOG.info("{} hit timeout: {}s. The WarmUp will continue in background", _printName, _warmUpTimeoutSeconds);
+          callback.onSuccess(None.none());
+        }
```
shivamgupta1, line 136: While this is a great idea, it'll end up blocking the caller of the callback (which could be just about any thread? not sure). This is a concern considering the variety of load-balancers this class is used to wrap with.

The root-cause for the timeout not triggering seems to be the scheduled executor behavior of executing all tasks in order, including the scheduled ones. This clearly doesn't work here. So what we might need to do here is make the tasks themselves timeout aware (as in each task checks for whether it has exceeded the warmup timeout) and have them trigger the callback. This can be discussed further.

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
```diff
@@ -95,28 +95,50 @@ public DualReadLoadBalancer(LoadBalancerWithFacilities oldLb, LoadBalancerWithFa
...
-    _dualReadStateManager.checkAndSwitchMode(null);
+    DualReadModeProvider.DualReadMode mode = _dualReadStateManager.getGlobalDualReadMode();
 
-    _newLb.start(new Callback<None>()
-    {
+    // if in new-lb-only mode, new lb needs to start successfully to call the callback. Otherwise, the old lb does.
+    // Use a separate executor service to start the new lb, so both lbs can start concurrently.
+    _newLbExecutor.execute(() -> _newLb.start(getStartUpCallback(true,
+        mode == DualReadModeProvider.DualReadMode.NEW_LB_ONLY ? callback : null)
+    ));
```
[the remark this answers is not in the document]
bohhyang: No, we are using NEW_LB_ONLY mode, which is the indis_registry_observer treatment in the lix. It's one of the modes for dual read load balancer: OLD_LB_ONLY (ZK only), DUAL_READ, NEW_LB_ONLY (observer only).

d2/src/main/java/com/linkedin/d2/balancer/util/WarmUpLoadBalancer.java
```diff
@@ -104,105 +107,160 @@ public WarmUpLoadBalancer(LoadBalancerWithFacilities balancer, WarmUpService ser
...
+        catch (Exception e)
+        {
+          LOG.error("{} failed to fetch dual read mode, continuing warmup.", _printName, e);
+        }
+        finally
+        {
+          continueWarmUp(callback);
```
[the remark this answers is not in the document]
bohhyang: no, the warmup should continue in background after timeout, which is the current behavior.

d2/src/main/java/com/linkedin/d2/balancer/util/WarmUpLoadBalancer.java
```diff
@@ -104,105 +107,160 @@ public WarmUpLoadBalancer(LoadBalancerWithFacilities balancer, WarmUpService ser
...
+        try
+        {
+          prepareTaskFuture.get(_warmUpTimeoutSeconds, TimeUnit.SECONDS);
+        }
+        catch (TimeoutException e)
+        {
+          LOG.info("{} hit timeout: {}s. The WarmUp will continue in background", _printName, _warmUpTimeoutSeconds);
+          callback.onSuccess(None.none());
+        }
+        catch (Exception e)
```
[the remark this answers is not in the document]
bohhyang: actually it's supposed to catch any kind of exception happened in the warmup (just like line 259), to make sure warmup won't fail startup since warmup is not required to complete successfully during startup.

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
```diff
@@ -95,28 +95,50 @@ public DualReadLoadBalancer(LoadBalancerWithFacilities oldLb, LoadBalancerWithFa
...
+  private Callback<None> getStartUpCallback(boolean isForNewLb, Callback<None> callback)
+  {
+    return new Callback<None>() {
       @Override
-      public void onError(Throwable e)
-      {
-        LOG.warn("Failed to start new load balancer. Fall back to read from old balancer only", e);
-        _isNewLbReady = false;
+      public void onError(Throwable e) {
+        LOG.warn("Failed to start {} load balancer.", isForNewLb ? "new" : "old", e);
```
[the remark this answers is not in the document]
bohhyang: unfortunately it won't work since both new and old lbs are wrapped in a WarmUpLoadBalancer, so both of them will be shown as WarmUpLoadBalancer which is not meaningful.

d2/src/main/java/com/linkedin/d2/balancer/util/WarmUpLoadBalancer.java
bohhyang: good point. Although only restli client uses WarmUpLoadBalancer (grpc client doesn't), it's not ideal to block the thread that calls the start callback, especially to the property event bus thread that updates the simple load balancer. I've put the work to a timeout-aware, non-blocking task.

d2/src/main/java/com/linkedin/d2/balancer/util/WarmUpLoadBalancer.java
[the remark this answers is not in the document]
shivamgupta1: It should already be continuing, since a timeout won't cancel the running task? It will already be running in the background?

d2/src/main/java/com/linkedin/d2/balancer/util/WarmUpLoadBalancer.java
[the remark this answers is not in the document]
shivamgupta1: For that one usually catches `RuntimeException`, but it appears that that could just be an opinion.

d2/src/main/java/com/linkedin/d2/balancer/util/WarmUpLoadBalancer.java
```diff
@@ -114,91 +115,117 @@ public WarmUpLoadBalancer(LoadBalancerWithFacilities balancer, WarmUpService ser
...
+        if (e instanceof TimeoutException)
         {
           LOG.info("{} hit timeout: {}s. The WarmUp will continue in background", _printName, _warmUpTimeoutSeconds);
           callback.onSuccess(None.none());
         }
-        catch (Exception e)
+        else
         {
           LOG.error("{} failed to fetch dual read mode, continuing warmup.", _printName, e);
         }
```
shivamgupta1, line 130: Should call `callback.onError(e)` here?

d2/src/main/java/com/linkedin/d2/balancer/util/WarmUpLoadBalancer.java
```diff
@@ -114,91 +115,117 @@ public WarmUpLoadBalancer(LoadBalancerWithFacilities balancer, WarmUpService ser
...
+            {
+              LOG.info("{} skipping service: {} based on its dual read mode: {}",
+                  _printName, serviceName, dualReadMode);
+            }
+            return res;
+          }).collect(Collectors.toList());
+
+          _servicesToWarmUp.forEach(serviceName -> {
+            // check timeout before continue
+            if (!hasTimedOut.get()
```
shivamgupta1, line 189: The variable doesn't need to be atomic since it's only accessed within the executor and is actually local. This operation (read and then update) is in itself non-atomic and thus thread-unsafe even if an atomic was required.

d2/src/main/java/com/linkedin/d2/balancer/util/WarmUpLoadBalancer.java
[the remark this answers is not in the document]
shivamgupta1: Nevermind, I understood it now. The preparation phase would have completed and that's a different callback now.

d2/src/main/java/com/linkedin/d2/balancer/util/WarmUpLoadBalancer.java
shivamgupta1: The solution to this problem looks good.

d2/src/main/java/com/linkedin/d2/balancer/util/WarmUpLoadBalancer.java
bohhyang: not really, since this is just the preparation phase (fetching the dual read mode), the actual warmup phase can still warm up more data if it haven't hit timeout before letting the startup complete. (In fact, if startup callback's onError is called, the startup will fail, which shouldn't be caused by warmup since warmup is not required.)

d2/src/main/java/com/linkedin/d2/balancer/util/WarmUpLoadBalancer.java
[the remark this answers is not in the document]
bohhyang: I understand. But to ensure no exception is raised up (including checked and runtime exceptions) from failures that we don't care, I think it's ok to catch the general exception/throwable (commonly seen in many mps).

d2/src/test/java/com/linkedin/d2/balancer/util/WarmUpLoadBalancerTest.java
```diff
@@ -316,22 +358,124 @@ public void testHitTimeout() throws URISyntaxException, InterruptedException, Ex
...
+      throws InterruptedException, ExecutionException, TimeoutException
   {
-    _tmpdir = LoadBalancerUtil.createTempDirectory("d2FileStore");
-    _FSBasedDownstreamServicesFetcher = new FSBasedDownstreamServicesFetcher(_tmpdir.getAbsolutePath(), MY_SERVICES_FS);
+    int warmUpTimeout = 4;
+    createDefaultServicesIniFiles();
+    setDualReadMode(mode);
+
+    // 3 dual read fetches take 1.5s, 3 warmups take at most 3 * (500 +/- 10) ms. Total at most is 3030 ms.
+    TestLoadBalancer balancer = new TestLoadBalancer(500, 500);
```
shivamgupta1, line 383: Is it possible to keep this in single digit millis? Tests taking long to execute add up quickly to result in a slower build.

bohhyang: I thought about this too. But the warmup timeout config takes in seconds, so to test the timeout case, the dual read fetching got be more than at least 1s.

shivamgupta1: In that case, it might be best to add another constructor with `Supplier<Long> timeSupplier` as a param, which is set to `() -> SystemClock.instance().currentTimeMillis()` for the existing constructors. This should make the tests more reliable while avoiding the penalty of wasted time.

zackthehuman: This. Please make any time source pluggable through a constructor. Otherwise you will always have issues calling into some static clock with tests.

bohhyang: but there will be multiple calls to get the time (2 critical ones --- one in fetching dual read, one in actual warmup work. And many other ones for logs) which need different returns. It will be hard to code what time to return for which call.

zackthehuman: Time always moves forward (I hope) so would a supplied that returns a monotonically increasing value work?

bohhyang: > a supplied that returns a monotonically increasing value

Could you list a sample snippet? just to make sure I understand what it means.

zackthehuman: Something like this:

```java
new Supplier<Long>() {
  long time = 0;

  @Override
  public Long get() {
    return time += 100;
  }
}
```

Every time the supplier is called it will "move time forward" by 100 milliseconds. Would something like this work for the tests?

bohhyang: updated. Thanks!

d2/src/test/java/com/linkedin/d2/util/TestDataHelper.java
```diff
@@ -216,4 +221,23 @@ public void verifyZeroEmissionOfSDStatusUpdateReceiptEvents() {
       assertTrue(_receiptMarkDownClusters.isEmpty());
     }
   }
+
+  // A time supplier that proceed with speedMillis but could freeze on special calls specified in freezedCalls.
+  // This is for convenience when the code being tested has calls where the time shouldn't move forward (no
+  // time-consuming work is done before this call).
+  public static Supplier<Long> getTimeSupplier(long speedMillis, int... freezedCalls)
```
shivamgupta1, line 228: Looking good.

d2/src/test/java/com/linkedin/d2/balancer/util/WarmUpLoadBalancerTest.java
shivamgupta1: Thanks, keep the tests nice and fast.

shivamgupta1 (approved):

PR #966, opened 2024-01-11
Add a WIRE_COMPATIBLE compatibility checker level

gcki:
Adding new enumeration values has special consideration in the compatibility checker. Even though such changes are compatible on the wire level (don't break deserialization for old clients), they are still considered backwards incompatible by the checker. The documentation explains why: `However, it’s still not possible to guarantee backward compatibility, even with the “$UNKNOWN” symbol available. It’s possible that clients did not handle the “$UNKNOWN” symbol in the best possible way, and even if they did it may be that they cannot do anything other than fail if they encounter a enum symbol they do not recognize.`

This patch adds a new parameter to the compatibility checker to handle this special scenario. Adding new enumeration values will still fail under the BACKWARDS and EQUIVALENT levels, but will be allowed under the WIRE_COMPATIBLE level.

### DIFF
--- a/restli-tools/src/main/java/com/linkedin/restli/tools/idlcheck/CompatibilityInfo.java
+++ b/restli-tools/src/main/java/com/linkedin/restli/tools/idlcheck/CompatibilityInfo.java
@@ -31,7 +31,7 @@ public class CompatibilityInfo
     COMPATIBLE,
 
     /**
-     * Serialization and deserialization is backwards compatible, but the change may be mishandled by clients
+     * Old readers can deserialize changes serialized by new writers, but may not be able to handle them correctly.
      * Currently only used for adding new enum values.
      **/
     WIRE_COMPATIBLE
--- a/restli-tools/src/main/java/com/linkedin/restli/tools/idlcheck/CompatibilityInfo.java
+++ b/restli-tools/src/main/java/com/linkedin/restli/tools/idlcheck/CompatibilityInfo.java
@@ -28,7 +28,13 @@ public class CompatibilityInfo
   public enum Level
   {
     INCOMPATIBLE,
-    COMPATIBLE
+    COMPATIBLE,
+
+    /**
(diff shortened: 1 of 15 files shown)


junchuanwang:
can you update changelog and gradle.properties as well?

restli-tools/src/main/java/com/linkedin/restli/tools/idlcheck/CompatibilityInfo.java
```diff
@@ -28,7 +28,13 @@ public class CompatibilityInfo
   public enum Level
   {
     INCOMPATIBLE,
-    COMPATIBLE
+    COMPATIBLE,
+
+    /**
+     * Serialization and deserialization is backwards compatible, but the change may be mishandled by clients
```
junchuanwang, line 34: by reading this comment "Serialization and deserialization is backwards compatible" I still don't know what WIRE_COMPATIBLE means, I think you meant pegasus serializer's serialized result can be deserialized by old pegasus deserializer

gcki: Yes, but also by new pegasus deserializers (even though this was never a problem).
The combinations: new writer, old reader, and new reader, old writer both work with new enum values. I will clarify the comment

gcki: @junchuanwang is WIRE_COMPATIBLE a good name or do you have a different suggestion?

zackthehuman: `WIRE_COMPATIBLE` seems OK to me. It means that the change is compatible "over the wire" but not necessarily understood by the client.

junchuanwang (approved):

zackthehuman (approved):

PR #967, opened 2024-01-12
Fix Dualread monitoring log

bohhyang:
As in the title, fix a string formatting in dualread monitoring log to show the properties being compared.

## Test Done
Added unit tests.


brycezhongqing (approved):
LGTM

PR #970, opened 2024-01-22
Adding `translated.from` annotation when using SchemaTranslator

li-ukumar:
### What
Adding an annotation in translated schemas, translated through SchemaTranslator

### Why
This will help in figuring out whether a schema is translated or source of truth.
This annotation can help in determining different behavior for SoT and translated schemas

### Test
Updated Unit tests

### DIFF
--- a/data-avro/src/main/java/com/linkedin/data/avro/SchemaTranslator.java
+++ b/data-avro/src/main/java/com/linkedin/data/avro/SchemaTranslator.java
@@ -18,12 +18,15 @@ package com.linkedin.data.avro;
 
 
 import com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper;
+import com.linkedin.avroutil1.compatibility.ConfigurableSchemaComparator;
+import com.linkedin.avroutil1.compatibility.SchemaComparisonConfiguration;
 import com.linkedin.avroutil1.compatibility.SchemaParseConfiguration;
 import com.linkedin.data.DataMap;
 import com.linkedin.data.DataMapBuilder;
 import com.linkedin.data.schema.DataSchema;
 import com.linkedin.data.schema.DataSchemaResolver;
 import com.linkedin.data.schema.DataSchemaTraverse;
+import com.linkedin.data.schema.NamedDataSchema;
 import com.linkedin.data.schema.RecordDataSchema;
 import com.linkedin.data.schema.SchemaFormatType;
 import com.linkedin.data.schema.SchemaParser;
@@ -36,6 +39,7 @@ import com.linkedin.data.schema.validation.ValidationOptions;
 import com.linkedin.data.template.DataTemplateUtil;
 
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.IdentityHashMap;
 import java.util.Map;
@@ -53,6 +57,7 @@ public class SchemaTranslator
   private static final Logger log = LoggerFactory.getLogger(SchemaTranslator.class);
 
   public static final String DATA_PROPERTY = "com.linkedin.data";
+  public static final String TRANSLATED_FROM_SOURCE_OPTION = "li.data.translated.from";
   public static final String SCHEMA_PROPERTY = "schema";
   public static final String OPTIONAL_DEFAULT_MODE_PROPERTY = "optionalDefaultMode";
   public static final String AVRO_FILE_EXTENSION = ".avsc";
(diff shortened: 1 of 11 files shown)


data-avro/src/main/java/com/linkedin/data/avro/SchemaTranslator.java
```diff
@@ -166,9 +171,12 @@ public static DataSchema avroToDataSchema(String avroSchemaInJson, AvroToDataSch
             {
               avroSchemaFromEmbedded.addProp(DATA_PROPERTY, embededSchemaPropertyVal);
             }
-            if (!avroSchemaFromEmbedded.equals(avroSchemaFromJson))
-            {
-              throw new IllegalArgumentException("Embedded schema does not translate to input Avro schema: " + avroSchemaInJson);
+            // Compare using configuration equivalent to STRICT, except ignore TRANSLATED_FROM_SOURCE_OPTION
+            if (!ConfigurableSchemaComparator.equals(avroSchemaFromEmbedded, avroSchemaFromJson,
+                new SchemaComparisonConfiguration(true, true, true, false, true, true,
+                    Collections.singleton((TRANSLATED_FROM_SOURCE_OPTION))))) {
```
mchen07, line 177: This is a bit hard to read. Why can you provide a constructor in `ConfigurableSchemaComparator` to allow configure what options to be excluded? Also why `DATA_PROPERTY` above is not handled simiarly?

data-avro/src/main/java/com/linkedin/data/avro/SchemaTranslator.java
```diff
@@ -341,6 +354,26 @@ public static String dataToAvroSchemaJson(DataSchema dataSchema, DataToAvroSchem
...
+   * Adds TRANSLATED_FROM_SOURCE_OPTION property to named data schemas if not already present.
+   * @param dataSchema the data schema to add the property to
+   * @return the data schema with the property added if applicable.
+   */
+  private static DataSchema addTranslatedPropToNamedDataSchema(DataSchema dataSchema) {
+    // Add translated from annotation if this is a named dataSchema
+    if (dataSchema instanceof NamedDataSchema) {
+      NamedDataSchema namedDataSchema = (NamedDataSchema) dataSchema;
+      // Add annotation if not already present
+      if (!namedDataSchema.getProperties().containsKey(TRANSLATED_FROM_SOURCE_OPTION)) {
```
mchen07, line 367: what should you do if existing translated_from prop is different from updated one?

data-avro/src/main/java/com/linkedin/data/avro/SchemaTranslator.java
```diff
@@ -341,6 +354,26 @@ public static String dataToAvroSchemaJson(DataSchema dataSchema, DataToAvroSchem
...
+        Map<String, Object> properties = new HashMap<>(namedDataSchema.getProperties());
+        properties.put(TRANSLATED_FROM_SOURCE_OPTION, namedDataSchema.getFullName());
+        namedDataSchema.setProperties(properties);
+      }
+      dataSchema = namedDataSchema;
+    }
+    return dataSchema;
```
mchen07, line 374: this will have side effect of changing passed `DataSchema`?

data-avro/src/main/java/com/linkedin/data/avro/SchemaTranslator.java
li-ukumar: It doesn't update the source in that case. So, if schema `A` translates to `B`, and `B` is translated to `C`,
the translatedFrom option in C will be `A`

data-avro/src/main/java/com/linkedin/data/avro/SchemaTranslator.java
```diff
@@ -53,6 +58,7 @@ public class SchemaTranslator
   private static final Logger log = LoggerFactory.getLogger(SchemaTranslator.class);
 
   public static final String DATA_PROPERTY = "com.linkedin.data";
+  public static final String TRANSLATED_FROM_SOURCE_OPTION = "schema.translated.from.src";
```
mchen07, line 61: will `translated.from` enough?

data-avro/src/main/java/com/linkedin/data/avro/SchemaTranslator.java
mchen07: My original question is talking about an error case, I am guessing that should never happen and we should throw error? 
But your response provides me something  different from my understanding, I thought that annotation is just for direct source, seems not?

li-ukumar: We don't need to stop users from translating a translated schema again, as it is allowed right now in SchemaTranslator right now, and stopping it might fail users.

However, we can use this annotation to fail other flows like PDL -> Proto or Avro -> proto.

data-avro/src/main/java/com/linkedin/data/avro/SchemaTranslator.java
li-ukumar: This check will never fail for `DATA_PROPERTY` because the code above this check adds it just before the check.
```java
Object embededSchemaPropertyVal = avroSchemaFromJson.getObjectProp(DATA_PROPERTY);
            if (embededSchemaPropertyVal != null)
            {
              avroSchemaFromEmbedded.addProp(DATA_PROPERTY, embededSchemaPropertyVal);
            }
```

Updated the SchemaComparisonConfiguration

data-avro/src/main/java/com/linkedin/data/avro/SchemaTranslator.java
li-ukumar: Yes. any unique key will work.

I'll update to `translated.from`

data-avro/src/main/java/com/linkedin/data/avro/SchemaTranslator.java
mchen07: Based on this logic, `translated.from` annotation is used to indicate the root source, not directly translated source, is this clearly communicated?

mchen07:
> Temporarily modified the tests to ignore then new option
Do you mean that this is a temporary solution or you have followup fix?

li-ukumar:
> > Temporarily modified the tests to ignore then new option
> > Do you mean that this is a temporary solution or you have followup fix?

It was the description for the initial PR. The unit tests are updated now.

Updated the `Description` .

ChrisCarini:

PR #973, opened 2024-01-30
Fix r2-netty illegal state exception due to premature channel recycling

TylerHorth:
Currently, the r2-netty client recycles connections once the response body has been fully received. However, it is possible for a server to return a response before fully consuming the request body. In this case, the channel will be recycled before the request body has made it through the pipeline. If a subsequent request comes in before the prior request is complete, an [illegal state exception](https://gist.github.com/TylerHorth/623489b9805f35803c710dcd80fd9eae) is thrown by the netty http object encoder.

I created a unit test to reproduce the issue and updated the lifecycle handler to recycle the channel only after both the request and response are complete.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.50.1] - 2024-01-31
+- Fix r2-netty illegal state exception due to premature channel recycling.
+
 ## [29.50.0] - 2024-01-31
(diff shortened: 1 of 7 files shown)


jcleezer:
Fix looks good!

zackthehuman (approved):
Looks good to me!

zackthehuman:
Needs rebase due to the minor version being bumped. See #974.

zackthehuman (approved):

PR #979, opened 2024-02-13
Default back to pick_first policy

PapaCharlie:
Newer versions of gRPC (v1.59 and above) offer configuration for the `pick_first` policy that enables shuffling the address list returned by DNS. This code change wires in the ability to provide said configuration. Note that I have tested this locally with the new config and can confirm that it works as expected, i.e. it does not create a connection to each downstream and does randomly pick the host.

Also bump `setup-java` as there is a known issue with v2: https://github.com/actions/setup-java/issues/506


bohhyang (approved):

bohhyang:
would be good to share the result of lsof on your local testing

shivamgupta1:
- Can we verify how this behaves with lower grpc versions that don't support this feature? 
- Should this be made configurable (via configs) so that we don't have to migrate owners to newer library versions in case this is also buggy?
- This should be tested locally to also check that it's shuffling the addresses correctly and that it doesn't have the same issue of establishing unneeded connections.

PapaCharlie:
> Can we verify how this behaves with lower grpc versions that don't support this feature?

Yeah sure, let me give it a shot, just to double check.

> Should this be made configurable (via configs) so that we don't have to migrate owners to newer library versions in case this is also buggy?

If you explicitly specify a load balancing policy, it disables this feature altogether. To go back to the original behavior, you can provide `"pick_first"` as the policy, and it won't configure it. Is that good enough?

PapaCharlie:
Actually @shivamgupta1, will it matter here?
> Can we verify how this behaves with lower grpc versions that don't support this feature?

Since rest.li is explicitly pulling up grpc to 1.59, it should pull everyone that depends on it up as well, unless they have an explicit dependency on specific versions. That being said, I looked back through the history of this file, and before this version it's hardcoded to ignore all configs, so it should have no impact. I tried pinning gRPC to a lower version and it compiled, but didn't work (as expected).

> This should be tested locally to also check that it's shuffling the addresses correctly and that it doesn't have the same issue of establishing unneeded connections

Can confirm. After restarting a local instance of `d2-proxy` multiple times, and/or killing the observer host it was talking to, it rebalances to a random host in the fleet.

PapaCharlie:
```
% lsof -p $(</export/content/lid/logs/d2-proxy/i001/d2-proxy.pid) | grep 32123
java    23785 pchesnai 1106u     IPv6            6299507       0t0        TCP pchesnai-ld1.linkedin.biz:35236->ltx1-app5542.stg.linkedin.com:32123 (ESTABLISHED)
```

shivamgupta1:
>Since rest.li is explicitly pulling up grpc to 1.59, it should pull everyone that depends on it up as well, unless they have an explicit dependency on specific versions. That being said, I looked back through the history of this file, and before this version it's hardcoded to ignore all configs, so it should have no impact. I tried pinning gRPC to a lower version and it compiled, but didn't work (as expected).

If it works this way, then Depex will try to bump pegasus versions, and those builds will fail since internally the gRPC version isn't ELR'ed for general usage yet. Which also means that once checked in, all restli fixes etc. would be blocked until the gRPC team certifies the new version.

>If you explicitly specify a load balancing policy, it disables this feature altogether. To go back to the original behavior, you can provide "pick_first" as the policy, and it won't configure it. Is that good enough?

The comment was to do the same via cfg2 and not code.

bohhyang:
add the policy to d2 client config makes sense, so we can control it without code change.  

@shivamgupta1 do you know if grpc team is working on certifying 1.59?

PapaCharlie:
> If it works this way, then Depex will try to bump pegasus versions, and those builds will fail since internally the gRPC version isn't ELR'ed for general usage yet. Which also means that once checked in, all restli fixes etc. would be blocked until the gRPC team certifies the new version.

Yeah we can't check this in until they validate that version. I think they're working on it now though. I will double check.

> The comment was to do the same via cfg2 and not code. It may not be a blocker but is something to consider.

I am definitely going to add a config for this in container. I don't want to have to deal with this again.

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
```diff
@@ -43,6 +43,7 @@
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import org.apache.commons.lang3.StringUtils;
```
bohhyang, line 46: unused import

d2/src/main/java/com/linkedin/d2/xds/XdsChannelFactory.java
```diff
@@ -61,20 +66,28 @@ public XdsChannelFactory(SslContext sslContext, String xdsServerUri)
...
       String xdsServerUri,
-      @Nullable String defaultLoadBalancingPolicy)
+      @Nullable String defaultLoadBalancingPolicy,
+      @Nullable Map<String, ?> loadBalancingPolicyConfig)
   {
     _sslContext = sslContext;
     _xdsServerUri = xdsServerUri;
+    if (defaultLoadBalancingPolicy == null && loadBalancingPolicyConfig != null)
+    {
+      throw new IllegalArgumentException("Cannot specify loadBalancingPolicyConfig without specifying " +
```
bohhyang, line 86: Doesn't seem to be a fatal error that needs to throw exception and crash the app. Can it just log a warning instead? the config will just be ignored in this case.

bohhyang (commented):
please also update the PR description since the PR is taking a different approach now.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -614,6 +625,10 @@ private void handleResponse(DiscoveryResponseData response)
       {
         return;
       }
+      if (!_responseReceived && response.getControlPlaneIdentifier() != null)
+      {
+        _log.info("Successfully established stream with ADS server: {}", response.getControlPlaneIdentifier());
```
shivamgupta1, line 630: Can we also test that this is working correctly?

d2/src/main/java/com/linkedin/d2/xds/XdsChannelFactory.java
```diff
@@ -79,6 +103,15 @@ public ManagedChannel createChannel()
     {
       _log.info("Applying custom load balancing policy for xDS channel: {}", _defaultLoadBalancingPolicy);
       builder = builder.defaultLoadBalancingPolicy(_defaultLoadBalancingPolicy);
```
shivamgupta1, line 105: This would be `null` by default? Does that work?

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
PapaCharlie: Yeah it was there the entire time, my editor just removed it, so I brought it back to avoid a diff in that file. But I can remove it anyway

d2/src/main/java/com/linkedin/d2/xds/XdsChannelFactory.java
PapaCharlie: Sure

d2/src/main/java/com/linkedin/d2/xds/XdsChannelFactory.java
PapaCharlie:
```java
    if (_defaultLoadBalancingPolicy != null)
    {
      _log.info("Applying custom load balancing policy for xDS channel: {}", _defaultLoadBalancingPolicy);
      builder = builder.defaultLoadBalancingPolicy(_defaultLoadBalancingPolicy);
```
This code block is only entered if it _isn't_ null.

By default it uses `pick_first` though

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
PapaCharlie: I did test it, it works 👍

shivamgupta1 (approved):

bohhyang (approved):

PR #981, opened 2024-02-16
Use tracingId in xDS flow for SD tracking events

bohhyang:
## Summary
1. Use the new tracingId field in xDS d2 proto for SD tracking events (previously, we were hard-coding the tracing id to zookeeper node path, which is invalid for Kafka data). 
2. Correct the version field in the tracking event (previously hard-coding to 0 for ZK data).
3. fix uri specific properties formatting in xds data. Before this fix, xDS uri properties put an empty map inside merged uriSpecificProperties for uris that don't have such property, leading to uri mismatches as described [here](https://docs.google.com/document/d/1phtGwYaQ6yuuqumS8lMZ8_Y5EkY11_R9bXZHzp0BX9E/edit#bookmark=id.i8523geqptmk).

## Test Done
1. Existing tests cover the uri publish logic, updated to cover the tracking events.
2. Manual tested with d2-proxy. Was able to make calls to downstream apps for both symlink and non-symlink clusters. And verified the merged uri properties are exactly the same now:

- ZK uri data to compare against:
[$EntitlementsBackendMaster.txt](https://github.com/linkedin/rest.li/files/14348799/EntitlementsBackendMaster.txt)

- xDS uri data before the fix:
[$EntitlementsBackendMaster_before.txt](https://github.com/linkedin/rest.li/files/14348804/EntitlementsBackendMaster_before.txt)

- xDS uri data after the fix:
[$EntitlementsBackendMaster_after.txt](https://github.com/linkedin/rest.li/files/14348809/EntitlementsBackendMaster_after.txt)

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
@@ -39,10 +39,12 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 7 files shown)


d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
```diff
@@ -433,51 +430,77 @@ public void onChanged(XdsClient.D2URIMapUpdate update)
...
+
+    private UriProperties toUriProperties(String uriName, XdsD2.D2URI xdsUri)
+    {
+      UriProperties uriProperties = null;
+      try {
+        uriProperties = _uriPropertiesJsonSerializer.fromProto(xdsUri);
+        if (uriProperties.getVersion() < 0)
         {
-          LOG.error("Failed to parse D2 uri properties from xDS update. Cluster name: " + _clusterName, e);
+          LOG.warn("xDS data: {} for uri: {} has invalid version: {}", xdsUri, uriName, uriProperties.getVersion());
```
brycezhongqing, line 479: Could we add clusterName to the log ?

brycezhongqing (approved):
LGTM

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
bohhyang: sure

PR #983, opened 2024-02-24
Fix excessive logs in uri data/version mismatch and dual read failure

bohhyang:
## Summary 
1. fix [SI-37741](https://jira01.corp.linkedin.com:8443/browse/SI-37741) - set this log to debug level instead of warn: the log that reports uri data being the same but the version is different. 
2. fix [SI-37667](https://jira01.corp.linkedin.com:8443/browse/SI-37667) - rate limit dual read failure logs, since its volume can be huge on high QPS apps.

## Test Done
existing tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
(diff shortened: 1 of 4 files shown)


brycezhongqing (approved):
LGTM

PR #985, opened 2024-02-29
Change UriProperties logging to debug

jefflai2:
Logging UriProperties diffs can cause frequent and very large log lines at the warn level due to large entries.


d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancerMonitor.java
```diff
@@ -97,14 +97,7 @@ public void reportData(String propertyName, T property, String propertyVersion,
                   fromNewLb ? "New" : "Old", propertyClassName, propertyName, existingEntry._version,
                   propertyVersion, property);
 
-          if (isUriProp)
-          {
-            LOG.debug(msg);
-          }
```
bohhyang, line 103: This already logging to debug level. No need to further change it. And it's critical to warn Service and Cluster properties mismatch, so we only log to debug level for Uri properties.

bohhyang (changes requested):

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancerMonitor.java
```diff
@@ -136,7 +129,7 @@ else if (!isDataEqual && isVersionEqual)
     { // data is not the same but version is the same, a mismatch!
       incrementPropertiesErrorCount();
       incrementEntryOutOfSyncCount(); // increment the out-of-sync count for the entry received later
-      warnByPropType(isUriProp,
+      logByPropType(isUriProp,
           String.format("Received mismatched %s for %s. %s", propertyClassName, propertyName, entriesLogMsg));
```
bohhyang, line 133:
```suggestion
String msg = String.format("Received mismatched %s for %s. %s", propertyClassName, propertyName, entriesLogMsg);
if (isUriProp)
{
    LOG.debug(msg);
}
else
{
    warnByPropType(isUriProp, msg);    
}
```

jefflai2: in the isUriProp case, this should still probably be rateLimitedLogger right?

jefflai2: also, with this version, this is actually equivalent to my proposed change. The isUriProp path in warnByLogType will be dead code since it's check prior to every invocation and DEBUG'd

bohhyang:
How about keep using warnByPropType (instead of LOG.debug) but tune the rate to a long time?

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancerMonitor.java
jefflai2: resolved offline, we will tune the rate limit for now instead of changing log level and explore better handling of long log lines

bohhyang (approved):
Thanks for looking into this!

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancerMonitor.java
the-coder-o: `String msg = String.format("Received mismatch %s for %s. %s", propertyClassName, propertyName, entriesLogMsg);

if (isUriProp) {
    LOG.debug(msg);
} else {
    warnByPropType(isUriProp, msg);
}
`

PR #991, opened 2024-03-20
fix indis null guard log

brycezhongqing:
## Summary
As the title said, fix indis null guard log issue for `dual-read` mode.

## Followup

create a Depex ticket to bump up this version for all services.

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancer.java
@@ -915,7 +915,6 @@ public class SimpleLoadBalancer implements LoadBalancer, HashRingProvider, Clien
       Pair<ClusterProperties, UriProperties> pair = getClusterAndUriPropertiesFromCache(clusterName, pairCallback);
       if (pair != null)
       {
-        _log.info("getClusterAndUriProperties for {} timed out, used cached value instead.", clusterName);
         pairCallback.onSuccess(pair);
       }
     };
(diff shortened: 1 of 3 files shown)


bohhyang (approved):

d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancer.java
```diff
@@ -939,6 +938,7 @@ private void handleTimeoutFromGetClusterAndUriProperties(Throwable e, String clu
         getClusterAndUriPropertiesFromCache(clusterName, clusterAndUriPropertiesCallback);
     if (pair != null)
     {
+      _log.info("getClusterAndUriProperties for {} timed out, used cached value instead.", clusterName);
       clusterAndUriPropertiesCallback.onSuccess(pair);
     }
     else
```
shivamgupta1, line 944: The callback already seems to be invoked by `die` inside `getClusterAndUriPropertiesFromCache`, so looks like it'll be called twice. Perhaps not a good idea to pass the callback and not call it for both cases anyway.

brycezhongqing: Hi, @shivamgupta1 As u could find in  `getClusterAndUriPropertiesFromCache` , the callback is only called when it could not get info from the cache.
The we need to call it from outside, when it is successful.

shivamgupta1: The case being referred to here is the `else` one below (Github doesn't allow me to select it while commenting): `clusterAndUriPropertiesCallback.onError`

brycezhongqing: Got u. Yes. We should not call it twice, as the `die` will trigger the `onError`.

PR #992, opened 2024-03-22
Address the multiple onError calls in dual read and enhance unit test

brycezhongqing:
## Summary
As the title said
- Fix the multiple onError calls in dual-read null guard logic
- Fefactor and add more strict unit test to make each `onError` method and `onSuccess` method will only be called one time.
- Catch up more potential exception for dual-read
- Try to fix some legacy unit test flaky failed issue

## Test done (Unit test)
- [✅] Add more unit test condition to make sure each `onError` and `onSuccess` method will only be called one time.
-  [✅] Add more unit test condition to make sure handleTime... method will be called when timeout

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
@@ -116,10 +116,6 @@ public class DualReadLoadBalancer implements LoadBalancerWithFacilities
       _rateLimitedLogger.debug("newLb executor rejected new task for start. "
           + "It is shut down or its queue size has reached max limit");
     }
-    catch (Exception e)
-    {
-      _rateLimitedLogger.debug("Failed to execute newLb task for start. ", e);
-    }
 
     _oldLb.start(getStartUpCallback(false,
         mode == DualReadModeProvider.DualReadMode.NEW_LB_ONLY ? null : callback
@@ -210,10 +206,6 @@ public class DualReadLoadBalancer implements LoadBalancerWithFacilities
           _rateLimitedLogger.debug("newLb executor rejected new task for getClient. "
               + "It is shut down or its queue size has reached max limit");
         }
-        catch (Exception e)
-        {
-          _rateLimitedLogger.debug("Failed to execute newLb task for getClient. ", e);
-        }
 
         _oldLb.getClient(request, requestContext, clientCallback);
         break;
@@ -241,10 +233,6 @@ public class DualReadLoadBalancer implements LoadBalancerWithFacilities
           _rateLimitedLogger.debug("newLb executor rejected new task for getLoadBalancedServiceProperties. "
               + "It is shut down or its queue size has reached max limit");
         }
-        catch (Exception e)
-        {
-          _rateLimitedLogger.debug("Failed to execute newLb task for getLoadBalancedServiceProperties. ", e);
-        }
         _oldLb.getLoadBalancedServiceProperties(serviceName, clientCallback);
         break;
       case OLD_LB_ONLY:
@@ -271,10 +259,6 @@ public class DualReadLoadBalancer implements LoadBalancerWithFacilities
         {
           _rateLimitedLogger.debug("newLb executor rejected new task for getLoadBalancedClusterAndUriProperties. "
               + "It is shut down or its queue size has reached max limit");
-        }
-        catch (Exception e)
-        {
-          _rateLimitedLogger.debug("Failed to execute newLb task for getLoadBalancedClusterAndUriProperties. ", e);
         }
         _oldLb.getLoadBalancedClusterAndUriProperties(clusterName, callback);
         break;
--- a/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
@@ -116,6 +116,10 @@ public class DualReadLoadBalancer implements LoadBalancerWithFacilities
       _rateLimitedLogger.debug("newLb executor rejected new task for start. "
           + "It is shut down or its queue size has reached max limit");
     }
+    catch (Exception e)
+    {
+      _rateLimitedLogger.debug("Failed to execute newLb task for start. ", e);
+    }
 
     _oldLb.start(getStartUpCallback(false,
         mode == DualReadModeProvider.DualReadMode.NEW_LB_ONLY ? null : callback
@@ -206,6 +210,10 @@ public class DualReadLoadBalancer implements LoadBalancerWithFacilities
           _rateLimitedLogger.debug("newLb executor rejected new task for getClient. "
               + "It is shut down or its queue size has reached max limit");
         }
+        catch (Exception e)
+        {
+          _rateLimitedLogger.debug("Failed to execute newLb task for getClient. ", e);
+        }
 
         _oldLb.getClient(request, requestContext, clientCallback);
         break;
@@ -233,6 +241,10 @@ public class DualReadLoadBalancer implements LoadBalancerWithFacilities
           _rateLimitedLogger.debug("newLb executor rejected new task for getLoadBalancedServiceProperties. "
               + "It is shut down or its queue size has reached max limit");
         }
+        catch (Exception e)
+        {
+          _rateLimitedLogger.debug("Failed to execute newLb task for getLoadBalancedServiceProperties. ", e);
+        }
         _oldLb.getLoadBalancedServiceProperties(serviceName, clientCallback);
         break;
       case OLD_LB_ONLY:
@@ -259,6 +271,10 @@ public class DualReadLoadBalancer implements LoadBalancerWithFacilities
         {
           _rateLimitedLogger.debug("newLb executor rejected new task for getLoadBalancedClusterAndUriProperties. "
               + "It is shut down or its queue size has reached max limit");
+        }
+        catch (Exception e)
+        {
+          _rateLimitedLogger.debug("Failed to execute newLb task for getLoadBalancedClusterAndUriProperties. ", e);
         }
         _oldLb.getLoadBalancedClusterAndUriProperties(clusterName, callback);
         break;
--- a/d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancer.java
@@ -840,7 +840,7 @@ public class SimpleLoadBalancer implements LoadBalancer, HashRingProvider, Clien
     }
     else
     {
-      _log.error("getServiceProperties for {} timed out, but no value in cache!", serviceName);
+      _log.warn("getServiceProperties for {} timed out, but no value in cache!", serviceName);
     }
   }
 
@@ -929,6 +929,7 @@ public class SimpleLoadBalancer implements LoadBalancer, HashRingProvider, Clien
     }
   }
 
+  @VisibleForTesting
   public void handleTimeoutFromGetClusterAndUriProperties(String clusterName,
       Callback<Pair<ClusterProperties, UriProperties>> clusterAndUriPropertiesCallback)
   {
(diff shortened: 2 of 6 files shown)


d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancer.java
```diff
@@ -943,8 +941,7 @@ private void handleTimeoutFromGetClusterAndUriProperties(Throwable e, String clu
     }
     else
     {
-      clusterAndUriPropertiesCallback.onError(
-          new ServiceUnavailableException(clusterName, "PEGA_1011. " + e.getMessage(), e));
+      _log.error("getClusterAndUriProperties for {} timed out, but no value in cache!", clusterName);
```
shivamgupta1, line 944: There's already a warn log in `getClusterAndUriPropertiesFromCache`:
```java
      warn(_log, "unable to find cluster: ", clusterName);
```

Can we remove that log and change the level of this one to warn?

d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancer.java
```diff
@@ -931,7 +929,7 @@ public void getLoadBalancedClusterAndUriProperties(String clusterName, boolean w
     }
   }
 
-  private void handleTimeoutFromGetClusterAndUriProperties(Throwable e, String clusterName,
+  public void handleTimeoutFromGetClusterAndUriProperties(String clusterName,
```
shivamgupta1, line 932: Does this only need to have default visibility for tests? Can we also add the `@VisibleForTesting` annotation?

d2/src/test/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerTest.java
```diff
@@ -439,117 +439,81 @@ public void testClusterInfoProviderUnregisterClusterListener()
...
+    // Make sure the onSuccess is called with SERVICE_PROPERTIES only once.
+    verify(callback).onSuccess(eq(SERVICE_PROPERTIES));
 
     // case3: listenToService without timeout
     serviceRegistry.put(SERVICE_NAME, SERVICE_PROPERTIES);
-    loadBalancer = setupLoadBalancer(null, serviceRegistry, clusterRegistry, uriRegistry);
+    doNothing().when(state).listenToService(any(), any());
+    loadBalancer = spy(new SimpleLoadBalancer(state, 5, TimeUnit.SECONDS, _d2Executor));
     loadBalancer.listenToServiceAndCluster(SERVICE_NAME, callback);
+    verify(callback).onSuccess(eq(SERVICE_PROPERTIES));
```
shivamgupta1, line 475: This seems incorrect. Calling Mockito `verify` once doesn't clear that invocation, unless it's forcibly cleared. If the success callback was indeed called, this would've needed to be `verify(callback, times(2)).onSuccess(eq(SERVICE_PROPERTIES));`. So this means that `doNothing().` has caused the callback to actually not be invoked.

d2/src/test/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerTest.java
```diff
@@ -439,117 +439,81 @@ public void testClusterInfoProviderUnregisterClusterListener()
...
     MockStore<UriProperties> uriRegistry = new MockStore<>();
-    LoadBalancerTestState state = spy(new LoadBalancerTestState());
-    state.listenToService = false;
-    SimpleLoadBalancer loadBalancer = spy(setupLoadBalancer(state, serviceRegistry, clusterRegistry, uriRegistry));
+    SimpleLoadBalancerState state =
+        spy(new SimpleLoadBalancerState(new SynchronousExecutorService(), uriRegistry, clusterRegistry, serviceRegistry,
+                                        new HashMap<>(), new HashMap<>()));
+    doAnswer(invocation ->
+             {
+               Thread.sleep(100);
```
shivamgupta1, line 452: Looks like this will block the test thread? Can we try to avoid that and if that's unavoidable, can we set a lower sleep duration, like 10ms?

d2/src/test/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerTest.java
```diff
@@ -439,117 +439,81 @@ public void testClusterInfoProviderUnregisterClusterListener()
...
-        assertTrue(e instanceof ServiceUnavailableException);
-      }
-
-      @Override
-      public void onSuccess(ServiceProperties result)
-      {
-        throw new RuntimeException("onSuccess should not be called");
-      }
-    };
+    FutureCallback<ServiceProperties> callback = spy(new FutureCallback<>());
```
shivamgupta1, line 457: Why not use a mock directly rather than building a spy?

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancer.java
```diff
@@ -116,6 +116,10 @@ public void start(Callback<None> callback)
       _rateLimitedLogger.debug("newLb executor rejected new task for start. "
           + "It is shut down or its queue size has reached max limit");
     }
+    catch (Exception e)
+    {
+      _rateLimitedLogger.debug("Failed to execute newLb task for start. ", e);
+    }
```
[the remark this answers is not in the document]
bohhyang: We don't want to write restli code according to particular class passed in from container. It's better to guard it in general.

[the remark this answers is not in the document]
shivamgupta1: `RejectedExecutionHandler` and `RejectedExecutionException` are only `ThreadPoolExecutor` constructs, so this code is already assuming that a `ThreadPoolExecutor` is being used. If the reference to the executor was to `ThreadPoolExecutor` rather than `ExecutorService`, then it would've made sense to handle this exception. Otherwise, not.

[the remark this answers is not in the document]
bohhyang: It depends on the actual handler used to handle the rejection, doesn't it? According to [ThreadPoolExecutor's doc](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html) : "Four predefined handler policies are provided:
In the default [ThreadPoolExecutor.AbortPolicy](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.AbortPolicy.html), the handler throws a runtime [RejectedExecutionException](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/RejectedExecutionException.html) upon rejection."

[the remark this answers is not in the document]
shivamgupta1: > We don't want to write restli code according to a particular class passed in from container. It's better to guard it in general.

The point is that this statement is likely incorrect, since `RejectedExecutionException` only needs to be handled when the reference is to `ThreadPoolExecutor`. In this case, in Restli, the object reference is to `ExecutorService` instead, which doesn't throw this exception.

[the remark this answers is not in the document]
bohhyang: oh, I see. Yes, you're right. That makes sense.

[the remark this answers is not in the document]
shivamgupta1: @brycezhongqing did discuss this with me earlier and it's not necessary to make that change though (it's up to him). Just that we can avoid adding more exception handling for now perhaps.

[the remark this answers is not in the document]
brycezhongqing: Yes. We could not catch the `ThreadPoolExecutor` task exception here.  I will raise a new PR focus on removing the extra exception and add more exception inside the task of `ThreadPoolExecutor`

d2/src/test/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerTest.java
brycezhongqing: I updated the code to `async` mock. Now it's ok to set 10ms sleep duration

d2/src/test/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerTest.java
brycezhongqing: Please see the latest code. It needs the `spy` for using `callback.get()` method. Thanks for ur suggestion.

d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancer.java
brycezhongqing: Yes. Added

d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancer.java
brycezhongqing: Discussed with @shivamgupta1 . As this log is in `getClusterAndUriProperties for timeout` scenario, so it's best to keep this to check distinguish this special scenario. But it's ok to change the level from `error` to `warn`.

d2/src/test/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerTest.java
brycezhongqing: Yes. Thanks for reminding. Have updated the code to `initialize` a new `callback` each time.

d2/src/test/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerTest.java
bohhyang: Spy an object is needed when mocking some method's result (while other methods do the real behavior). Simply calling .get method on the callback shouldn't need it.

d2/src/test/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerTest.java
```diff
@@ -439,117 +439,112 @@ public void testClusterInfoProviderUnregisterClusterListener()
...
+    callback.get();
+    verify(callback).onSuccess(eq(Pair.of(CLUSTER_PROPERTIES, URI_PROPERTIES)));
+
 
     // case3: getLoadBalancedClusterAndUriProperties without timeout
-    loadBalancer = setupLoadBalancer(null, serviceRegistry, clusterRegistry, uriRegistry);
+    state =
+        spy(new SimpleLoadBalancerState(new SynchronousExecutorService(), uriRegistry, clusterRegistry, serviceRegistry,
+                                        new HashMap<>(), new HashMap<>()));
+    loadBalancer = spy(new SimpleLoadBalancer(state, 5, TimeUnit.SECONDS, _d2Executor));
```
bohhyang, line 540: these state, load balancer, and the callback below are not seem to be spied for mocking any method's result (like mocking with doReturn, thenReturn, doAnswer, etc? they could just use real objects, right? 

Same applied to some other objects in other tests.

brycezhongqing: Hi, @bohhyang 
#1. state needs spy for
```java
 doAnswer(invocation ->
           {
             Thread.sleep(10);
             return null;
           }).when(state).listenToService(any(), any());
```

#2 If I don't use `spy` or `mock` for `callback` and `loadBalancer`. The test will be failed for
```java
org.mockito.exceptions.misusing.NotAMockException: 
Argument passed to verify() is of type FutureCallback and is not a mock!
Make sure you place the parenthesis correctly!
See the examples of correct verifications:
    verify(mock).someMethod();
    verify(mock, times(10)).someMethod();
    verify(mock, atLeastOnce()).someMethod();
```

bohhyang (approved):

PR #994, opened 2024-03-27
SchemaTranslator: support non-null defaults for primitive and collection optional fields 

dg-builder:
Map, array, and primitive non-null defaults are now supported for optional fields.

Added unit tests to cover this case.


data-avro/src/test/java/com/linkedin/data/avro/TestSchemaTranslator.java
```diff
@@ -2645,36 +2656,6 @@ public Object[][] toAvroSchemaErrorData()
                 IllegalArgumentException.class,
                 "cannot translate union value"
             },
-            {
-                // inconsistent default,
```
dg-builder, line 2649: we now support this case.

aman1309 (approved):

PR #999, opened 2024-04-29
Implement new DualReadMonitor for UriProperties

PapaCharlie:
## Summary
Previous dual read matches URI properties only when the versions are the same (same ZK data served from ZK vs Observer), but with Kafka data served, the version will never be the same. And the OutOfSync metric was never useful since only one version of uri properties is cached for each cluster, when uri data changes frequently, the OutOfSync increases and never comes down.
This new URI matching now produces a similarity metric which represents the fraction of matching uris present in the ZK response and the observer response, just the same as how observer is monitoring the cluster uri similarity. The similarity metric may go up during downstream deployment/updates, but eventually it should be 1. The goal is to verify that client side response processing logic in INDIS is not causing uri data to drift (given that Kafka data matches ZK data on observer already). 

The Cluster and Service monitors remain unchanged.  
(This new monitor should be as expensive as the previous one since the previous one called the `.equals` method on `UriProperties`, which compares every single URI). 

## Test Done
Added unit tests. Will do manual qei testing with d2-proxy in the counterpart container PR for adding the config.

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancerJmx.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancerJmx.java
@@ -16,8 +16,11 @@
 
 package com.linkedin.d2.balancer.dualread;
 
+import java.util.HashMap;
+import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
+import javax.annotation.Nullable;
 
 
 public class DualReadLoadBalancerJmx implements DualReadLoadBalancerJmxMBean
@@ -32,6 +35,8 @@ public class DualReadLoadBalancerJmx implements DualReadLoadBalancerJmxMBean
   private final AtomicInteger _clusterPropertiesOutOfSyncCount = new AtomicInteger();
 
   private final AtomicReference<Double> _uriPropertiesSimilarity = new AtomicReference<>(0d);
+
+  private final Map<String, UriPropertiesDualReadMonitor.ClusterMatchRecord> _clusters = new HashMap<>();
 
 
   @Override
@@ -97,6 +102,11 @@ public class DualReadLoadBalancerJmx implements DualReadLoadBalancerJmxMBean
     return _uriPropertiesSimilarity.get();
   }
 
+  @Override
+  public @Nullable UriPropertiesDualReadMonitor.ClusterMatchRecord getClusterMatchRecord(String clusterName) {
+    return _clusters.get(clusterName);
+  }
+
   public void incrementServicePropertiesErrorCount()
   {
     _servicePropertiesErrorCount.incrementAndGet();
@@ -140,5 +150,11 @@ public class DualReadLoadBalancerJmx implements DualReadLoadBalancerJmxMBean
   public void setUriPropertiesSimilarity(double similarity)
   {
     _uriPropertiesSimilarity.set(similarity);
+  }
+
+  public void setClusterMatchRecord(String clusterName,
+      UriPropertiesDualReadMonitor.ClusterMatchRecord clusterMatchRecord)
+  {
+    _clusters.put(clusterName, clusterMatchRecord);
   }
 }
--- a/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancerJmx.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancerJmx.java
@@ -17,21 +17,21 @@
 package com.linkedin.d2.balancer.dualread;
 
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 
 
 public class DualReadLoadBalancerJmx implements DualReadLoadBalancerJmxMBean
 {
   private final AtomicInteger _servicePropertiesErrorCount = new AtomicInteger();
   private final AtomicInteger _clusterPropertiesErrorCount = new AtomicInteger();
-  private final AtomicInteger _uriPropertiesErrorCount = new AtomicInteger();
 
   private final AtomicInteger _servicePropertiesEvictCount = new AtomicInteger();
   private final AtomicInteger _clusterPropertiesEvictCount = new AtomicInteger();
-  private final AtomicInteger _uriPropertiesEvictCount = new AtomicInteger();
 
   private final AtomicInteger _servicePropertiesOutOfSyncCount = new AtomicInteger();
   private final AtomicInteger _clusterPropertiesOutOfSyncCount = new AtomicInteger();
-  private final AtomicInteger _uriPropertiesOutOfSyncCount = new AtomicInteger();
+
+  private final AtomicReference<Double> _uriPropertiesSimilarity = new AtomicReference<>(0d);
 
 
   @Override
@@ -46,10 +46,11 @@ public class DualReadLoadBalancerJmx implements DualReadLoadBalancerJmxMBean
     return _clusterPropertiesErrorCount.get();
   }
 
+  @Deprecated
   @Override
   public int getUriPropertiesErrorCount()
   {
-    return _uriPropertiesErrorCount.get();
+    return 0;
   }
 
   @Override
@@ -64,25 +65,36 @@ public class DualReadLoadBalancerJmx implements DualReadLoadBalancerJmxMBean
     return _clusterPropertiesEvictCount.get();
   }
 
+  @Deprecated
   @Override
   public int getUriPropertiesEvictCount()
   {
-    return _uriPropertiesEvictCount.get();
+    return 0;
   }
 
   @Override
-  public int getServicePropertiesOutOfSyncCount() {
+  public int getServicePropertiesOutOfSyncCount()
+  {
     return _servicePropertiesOutOfSyncCount.get();
   }
 
   @Override
-  public int getClusterPropertiesOutOfSyncCount() {
+  public int getClusterPropertiesOutOfSyncCount()
+  {
     return _clusterPropertiesOutOfSyncCount.get();
   }
 
+  @Deprecated
   @Override
-  public int getUriPropertiesOutOfSyncCount() {
-    return _uriPropertiesOutOfSyncCount.get();
+  public int getUriPropertiesOutOfSyncCount()
+  {
+    return 0;
+  }
+
+  @Override
+  public double getUriPropertiesSimilarity()
+  {
+    return _uriPropertiesSimilarity.get();
   }
 
   public void incrementServicePropertiesErrorCount()
@@ -95,11 +107,6 @@ public class DualReadLoadBalancerJmx implements DualReadLoadBalancerJmxMBean
     _clusterPropertiesErrorCount.incrementAndGet();
   }
 
-  public void incrementUriPropertiesErrorCount()
-  {
-    _uriPropertiesErrorCount.incrementAndGet();
-  }
-
   public void incrementServicePropertiesEvictCount()
   {
     _servicePropertiesEvictCount.incrementAndGet();
@@ -108,11 +115,6 @@ public class DualReadLoadBalancerJmx implements DualReadLoadBalancerJmxMBean
   public void incrementClusterPropertiesEvictCount()
   {
     _clusterPropertiesEvictCount.incrementAndGet();
-  }
-
-  public void incrementUriPropertiesEvictCount()
-  {
-    _uriPropertiesEvictCount.incrementAndGet();
   }
 
   public void incrementServicePropertiesOutOfSyncCount()
@@ -125,11 +127,6 @@ public class DualReadLoadBalancerJmx implements DualReadLoadBalancerJmxMBean
     _clusterPropertiesOutOfSyncCount.incrementAndGet();
   }
 
-  public void incrementUriPropertiesOutOfSyncCount()
-  {
-    _uriPropertiesOutOfSyncCount.incrementAndGet();
-  }
-
   public void decrementServicePropertiesOutOfSyncCount()
   {
     _servicePropertiesOutOfSyncCount.decrementAndGet();
@@ -140,8 +137,8 @@ public class DualReadLoadBalancerJmx implements DualReadLoadBalancerJmxMBean
     _clusterPropertiesOutOfSyncCount.decrementAndGet();
   }
 
-  public void decrementUriPropertiesOutOfSyncCount()
+  public void setUriPropertiesSimilarity(double similarity)
   {
-    _uriPropertiesOutOfSyncCount.decrementAndGet();
+    _uriPropertiesSimilarity.set(similarity);
   }
 }
--- a/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancerMonitor.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancerMonitor.java
@@ -22,7 +22,6 @@ import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.RemovalCause;
 import com.linkedin.d2.balancer.properties.ClusterProperties;
 import com.linkedin.d2.balancer.properties.ServiceProperties;
-import com.linkedin.util.RateLimitedLogger;
 import com.linkedin.util.clock.Clock;
 import java.time.Instant;
 import java.time.ZoneId;
@@ -41,26 +40,27 @@ import org.slf4j.LoggerFactory;
  *
  * When a new service discovery data is reported, it will check if the cache of the other data source
  * has data for the same property name. If there is, it will compare whether the two data are equal.
+ *
+ * Note that there are only two implementations of this class, one for {@link ServiceProperties} and one for
+ * {@link ClusterProperties}, and not one for {@link com.linkedin.d2.balancer.properties.UriProperties}. This is because
+ * the URI properties need to be compared holistically at the cluster level.
  */
 public abstract class DualReadLoadBalancerMonitor<T>
 {
   private static final Logger LOG = LoggerFactory.getLogger(DualReadLoadBalancerMonitor.class);
   public final static String DEFAULT_DATE_FORMAT = "YYYY/MM/dd HH:mm:ss.SSS";
   public final static String VERSION_FROM_FS = "-1";
-  private static final long ERROR_REPORT_PERIOD = 600 * 1000; // Limit error report logging to every 10 minutes
   private static final int MAX_CACHE_SIZE = 10000;
   private final Cache<String, CacheEntry<T>> _oldLbPropertyCache;
   private final Cache<String, CacheEntry<T>> _newLbPropertyCache;
-  private final RateLimitedLogger _rateLimitedLogger;
   private final Clock _clock;
   private final DateTimeFormatter _format;
 
 
-  public DualReadLoadBalancerMonitor(Clock clock)
+  private DualReadLoadBalancerMonitor(Clock clock)
   {
     _oldLbPropertyCache = buildCache();
     _newLbPropertyCache = buildCache();
-    _rateLimitedLogger = new RateLimitedLogger(LOG, ERROR_REPORT_PERIOD, clock);
     _clock = clock;
     _format = DateTimeFormatter.ofPattern(DEFAULT_DATE_FORMAT);
   }
--- a/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancerMonitor.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancerMonitor.java
@@ -22,7 +22,6 @@ import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.RemovalCause;
 import com.linkedin.d2.balancer.properties.ClusterProperties;
 import com.linkedin.d2.balancer.properties.ServiceProperties;
-import com.linkedin.d2.balancer.properties.UriProperties;
 import com.linkedin.util.RateLimitedLogger;
 import com.linkedin.util.clock.Clock;
 import java.time.Instant;
@@ -72,7 +71,6 @@ public abstract class DualReadLoadBalancerMonitor<T>
     Cache<String, CacheEntry<T>> cacheToAdd = fromNewLb ? _newLbPropertyCache : _oldLbPropertyCache;
     CacheEntry<T> existingEntry = cacheToAdd.getIfPresent(propertyName);
     String propertyClassName = property.getClass().getSimpleName();
-    boolean isUriProp = property instanceof UriProperties;
 
     if (existingEntry != null && existingEntry._data.equals(property))
     {
@@ -92,19 +90,10 @@ public abstract class DualReadLoadBalancerMonitor<T>
         // different.
         if (!isReadFromFS(existingEntry._version, propertyVersion))
         {
-          String msg = String.format("Received same data of different versions in %s LB for %s: %s."
-                      + " Old version: %s, New version: %s, Data: %s",
-                  fromNewLb ? "New" : "Old", propertyClassName, propertyName, existingEntry._version,
-                  propertyVersion, property);
-
-          if (isUriProp)
-          {
-            LOG.debug(msg);
-          }
-          else
-          {
-            warnByPropType(isUriProp, msg);
-          }
+          LOG.warn("Received same data of different versions in {} LB for {} {}"
+                  + " Old version: {} New version: {} Data: {}",
+              fromNewLb ? "New" : "Old", propertyClassName, propertyName, existingEntry._version,
+              propertyVersion, property);
         }
         // still need to put in the cache, don't skip
       }
@@ -136,23 +125,14 @@ public abstract class DualReadLoadBalancerMonitor<T>
     { // data is not the same but version is the same, a mismatch!
       incrementPropertiesErrorCount();
       incrementEntryOutOfSyncCount(); // increment the out-of-sync count for the entry received later
-      warnByPropType(isUriProp,
-          String.format("Received mismatched %s for %s. %s", propertyClassName, propertyName, entriesLogMsg));
+      LOG.warn("Received mismatched {} for {}. {}", propertyClassName, propertyName, entriesLogMsg);
       cacheToCompare.invalidate(propertyName);
     }
     else {
       if (isDataEqual)
       {
-        String msg = String.format("Received same data of %s for %s but with different versions: %s",
+        LOG.warn("Received same data of {} for {} but with different versions: {}",
             propertyClassName, propertyName, entriesLogMsg);
-        if (isUriProp)
-        {
-          LOG.debug(msg);
-        }
-        else
-        {
-          warnByPropType(isUriProp, msg);
-        }
       }
       cacheToAdd.put(propertyName, newEntry);
       incrementEntryOutOfSyncCount();
@@ -207,18 +187,6 @@ public abstract class DualReadLoadBalancerMonitor<T>
   {
     // uri prop version from FS is "-1|x", where x is the number of uris, so we use startsWith here
     return v1.startsWith(VERSION_FROM_FS) || v2.startsWith(VERSION_FROM_FS);
-  }
-
-  private void warnByPropType(boolean isUriProp, String msg)
-  {
-    if (isUriProp)
-    {
-      _rateLimitedLogger.warn(msg);
-    }
-    else
-    {
-      LOG.warn(msg);
-    }
   }
 
   @VisibleForTesting
@@ -317,39 +285,6 @@ public abstract class DualReadLoadBalancerMonitor<T>
     void onEvict()
     {
       _dualReadLoadBalancerJmx.incrementServicePropertiesEvictCount();
-    }
-  }
-
-  public static final class UriPropertiesDualReadMonitor extends DualReadLoadBalancerMonitor<UriProperties>
-  {
-    private final DualReadLoadBalancerJmx _dualReadLoadBalancerJmx;
-
-    public UriPropertiesDualReadMonitor(DualReadLoadBalancerJmx dualReadLoadBalancerJmx, Clock clock)
-    {
-      super(clock);
-      _dualReadLoadBalancerJmx = dualReadLoadBalancerJmx;
-    }
-
-    @Override
-    void incrementEntryOutOfSyncCount() {
-      _dualReadLoadBalancerJmx.incrementUriPropertiesOutOfSyncCount();
-    }
-
-    @Override
-    void decrementEntryOutOfSyncCount() {
-      _dualReadLoadBalancerJmx.decrementUriPropertiesOutOfSyncCount();
-    }
-
-    @Override
-    void incrementPropertiesErrorCount()
-    {
-      _dualReadLoadBalancerJmx.incrementUriPropertiesErrorCount();
-    }
-
-    @Override
-    void onEvict()
-    {
-      _dualReadLoadBalancerJmx.incrementUriPropertiesEvictCount();
     }
   }
 }
--- a/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadStateManager.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadStateManager.java
@@ -60,15 +60,23 @@ public class DualReadStateManager
 
   private final DualReadLoadBalancerJmx _dualReadLoadBalancerJmx;
 
-  private final UriPropertiesDualReadMonitor _uriPropertiesDualReadMonitor;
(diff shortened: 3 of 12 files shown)


d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancerMonitor.java
```diff
@@ -72,7 +71,6 @@ public void reportData(String propertyName, T property, String propertyVersion,
     Cache<String, CacheEntry<T>> cacheToAdd = fromNewLb ? _newLbPropertyCache : _oldLbPropertyCache;
     CacheEntry<T> existingEntry = cacheToAdd.getIfPresent(propertyName);
     String propertyClassName = property.getClass().getSimpleName();
-    boolean isUriProp = property instanceof UriProperties;
```
bohhyang, line 75: could add a comment to this class and note DualReadLoadBalancerMonitor should never be extended for matching URI properties to make sure it won't be mis-used for URI at all. Otherwise, we could end up having log volume again with the URI logs.

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
```diff
@@ -0,0 +1,122 @@
...
+
+    if (cluster._oldLb == null || cluster._newLb == null)
+    {
+      cluster._uris = (cluster._oldLb == null) ? cluster._oldLb.Uris().size() : cluster._newLb.Uris().size();
+      _totalUris += cluster._uris;
+      return;
+    }
+
+    cluster._uris = cluster._oldLb.Uris().size();
+    Set<URI> newLbUris = new HashSet<>(cluster._newLb.Uris());
```
bohhyang, line 62: this will create a copy of all the uris, and if the uris are many and change frequently (like in seas apps), this will cause memory pressure on the client side, right?

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  private static <K, V> boolean compareMaps(
+      String type, String cluster, URI uri, Function<UriProperties, Map<URI, Map<K, V>>> extractor,
+      UriProperties oldLb, UriProperties newLb
+  )
+  {
+    Map<K, V> oldData = extractor.apply(oldLb).get(uri);
+    Map<K, V> newData = extractor.apply(newLb).get(uri);
+    if (Objects.equals(oldData, newData) || isEmptyMap(oldData) == isEmptyMap(newData))
```
bohhyang, line 108: when it's not matching, if it's because of map size is different, it's fine -- probably just a timing in the middle of a deployment. But if it's because of data mismatch (like what we had before with observer and client side formatting bugs), we could do a rated log (10min) so devs can look into the log to debug issues.

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
```diff
@@ -0,0 +1,122 @@
...
+      {
+        cluster._matched++;
+      }
+    }
+
+    cluster._uris += newLbUris.size();
+
+    _totalUris += cluster._uris;
+    _matchedUris += cluster._matched;
+    _dualReadLoadBalancerJmx.setUriPropertiesSimilarity(_matchedUris / _totalUris);
```
bohhyang, line 81: at the end, no matter it's matched or not, we could log both oldLb and newLb properties to debug level for debugging when needed, similar to what we had "LOG.debug("Current entries of {} for {}: {}", propertyClassName, propertyName, entriesLogMsg);", where entriesLogMsg is 
"String.format("\nOld LB: %s\nNew LB: %s", oldLb, newLb);"

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
```diff
@@ -0,0 +1,122 @@
...
+      return;
+    }
+
+    cluster._matched = 0;
+
+    if (cluster._oldLb == null || cluster._newLb == null)
+    {
+      cluster._uris = (cluster._oldLb == null) ? cluster._oldLb.Uris().size() : cluster._newLb.Uris().size();
+      _totalUris += cluster._uris;
+      return;
```
bohhyang, line 58: same here, we could log a debug level message similar to what we had, "LOG.debug("Added new URI properties for {} for {} LB.", propertyClassName, propertyName, fromNewLb ? "New" : "Old");", for debugging when needed.

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    cluster._uris += newLbUris.size();
+
+    _totalUris += cluster._uris;
+    _matchedUris += cluster._matched;
+    _dualReadLoadBalancerJmx.setUriPropertiesSimilarity(_matchedUris / _totalUris);
+  }
+
+  private static class Cluster
```
bohhyang, line 84: nit: ClusterMatchResult? or ClusterMatchRecord?

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancerMonitor.java
PapaCharlie: Sure, can do. I also made it guaranteed at compile time.

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
PapaCharlie: No this is a shallow copy, it doesn't copy the actual `URI` object itself, it just copies the pointer to the object.

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
PapaCharlie: Sure, `ClusterMatchRecord` makes sense

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
PapaCharlie: Yeah makes sense

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
PapaCharlie: Isn't that at the debug level already? Do you mean it should warn at the `INFO` level every 10 minutes?

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
PapaCharlie: Yeah makes sense

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
bohhyang: ah, that's right

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
bohhyang: Yeah, the log "URI {} for {}/{} mismatched between old and new LB.\nOld LB: {}\nNew LB: {}" should be logged to info level with the rate limited logger.

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
```diff
@@ -74,21 +81,29 @@ public void reportData(String clusterName, UriProperties property, boolean fromN
       }
     }
 
+    if (cluster._matched != cluster._uris)
+    {
+      RATE_LIMITED_LOGGER.info("Mismatched cluster properties for {} (match score {}):\nOld LB: {}\nNew LB: {}",
+          clusterName, cluster._matched / cluster._uris, cluster._oldLb, cluster._newLb);
+    }
```
bohhyang, line 88: Could this log be at the end of the method? since cluster._uris still need to include the new lb's uri count at line 90, and also it makes more sense to read the log at 95 first (when debug level log enabled) then this message.

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
bohhyang: actually, since the uris will go mismatch during deployment/updates, it's better to keep this log at debug level as it includes the whole URI properties. The info log about similarity score is good enough.

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
bohhyang: I'm updating this if it sounds good to you.

PapaCharlie: Makes sense

brycezhongqing (approved):
LGTM. Thanks for the improvement

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancerJmx.java
```diff
@@ -16,22 +16,27 @@
...
   private final AtomicInteger _clusterPropertiesEvictCount = new AtomicInteger();
-  private final AtomicInteger _uriPropertiesEvictCount = new AtomicInteger();
 
   private final AtomicInteger _servicePropertiesOutOfSyncCount = new AtomicInteger();
   private final AtomicInteger _clusterPropertiesOutOfSyncCount = new AtomicInteger();
-  private final AtomicInteger _uriPropertiesOutOfSyncCount = new AtomicInteger();
+
+  private final AtomicReference<Double> _uriPropertiesSimilarity = new AtomicReference<>(0d);
+
+  private final Map<String, UriPropertiesDualReadMonitor.ClusterMatchRecord> _clusters = new HashMap<>();
```
PapaCharlie, line 39: This needs to be a `ConcurrentHashMap`

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadStateManager.java
```diff
@@ -60,17 +60,24 @@ public class DualReadStateManager
...
   private final DualReadLoadBalancerMonitor.ServicePropertiesDualReadMonitor _servicePropertiesDualReadMonitor;
   private final DualReadLoadBalancerMonitor.ClusterPropertiesDualReadMonitor _clusterPropertiesDualReadMonitor;
+  private final UriPropertiesDualReadMonitor _uriPropertiesDualReadMonitor;
+  private final boolean _monitorUriProperties;
 
-
+  @Deprecated
   public DualReadStateManager(DualReadModeProvider dualReadModeProvider, ScheduledExecutorService executorService)
+  {
+    this(dualReadModeProvider, executorService, false);
```
PapaCharlie, line 71: Should this default to `true`? I guess we still need to benchmark it

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
```diff
@@ -0,0 +1,177 @@
...
+
+    @Override
+    public boolean equals(Object obj) {
+      if (this == obj)
+        return true;
+      if (obj == null)
+        return false;
+      if (getClass() != obj.getClass())
+        return false;
```
PapaCharlie, line 162: The go/javacode guidelines says:
> Always use braces around control statements

```suggestion
      if (this == obj) {
        return true;
      }
      if (obj == null) {
        return false;
      }
      if (getClass() != obj.getClass()) {
        return false;
      }
```

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    _totalUris += cluster._uris;
+    _matchedUris += cluster._matched;
+
+    updateJmxMetrics(clusterName, cluster);
+  }
+
+  private void updateJmxMetrics(String clusterName, ClusterMatchRecord cluster) {
+    _dualReadLoadBalancerJmx.setClusterMatchRecord(clusterName, cluster);
```
PapaCharlie, line 96: `ClusterMatchRecord` isn't thread safe. It's being modified by `reportData` in one thread, and read by the caller of `getClusterMatchRecord` in another thread. I would recommend something like this near the top of `reportData`:
```java
    ClusterMatchRecord cluster = _clusters.compute(clusterName, (k, v) -> {
      if (v == null) {
        return new ClusterMatchRecord();
      } else {
        return v.clone();
      }
    });
```
With the corresponding `clone` method added to `ClusterMatchRecord` of course.

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadStateManager.java
bohhyang: right, that's why the default should be disabled (false), right? so that we can enabled it on some app only for testing.

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
bohhyang: yeah, or we could simply call _dualReadLoadBalancerJmx.setClusterMatchRecord(clusterName, **cluster.clone()**) ?

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadStateManager.java
PapaCharlie: Yeah makes sense

d2/src/main/java/com/linkedin/d2/balancer/dualread/DualReadLoadBalancerJmx.java
bohhyang: yeah. updated.

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
bohhyang: updated.

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
bohhyang: added the lock

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
```diff
@@ -0,0 +1,200 @@
...
+import javax.annotation.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class UriPropertiesDualReadMonitor {
+  private static final Logger LOG = LoggerFactory.getLogger(UriPropertiesDualReadMonitor.class);
+
+  private final ConcurrentHashMap<String, ClusterMatchRecord> _clusters = new ConcurrentHashMap<>();
+  private final ConcurrentHashMap<String, Object> _clusterLocks = new ConcurrentHashMap<>();
```
PapaCharlie, line 24: You don't need to do this, you can use the raw `ClusterMatchRecord` as the lock. It makes things simpler too since you can freely throw `synchronized` on every method of `ClusterMatchRecord` to ensure nothing weird happens. `toString`, `equals` and `hashCode` all access the fields, so they need to be locked too. Java's synchronized locks are reentrant, so this works no problem:
```java
ClusterMatchRecord cluster = _clusters.computeIfAbsent(clusterName, k -> new ClusterMatchRecord());
synchronized (cluster) {
  // In go this would fail since it attempts to acquire the lock again. In Java, this works
  System.out.println(cluster.toString());
}
```

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
```diff
@@ -0,0 +1,200 @@
...
+  int getTotalUris() {
+    return _totalUris;
+  }
+
+  @VisibleForTesting
+  int getMatchedUris() {
+    return _matchedUris;
+  }
+
+  public static class ClusterMatchRecord implements Cloneable {
```
PapaCharlie, line 140: I wouldn't even bother with `Cloneable` here. I never really liked it and we don't know what it actually does. For example, does it do a deep clone? If so, it's making full copies of the `UriProperties` for no reason. IMO just have a constructor:
```java
private ClusterMatchRecord(UriProperties oldLb, UriProperties newLb, int uris, int matched) {...}
```
You can have a no-arg constructor too, for convenience:
```
private ClusterMatchRecord() {}
```
This one doesn't need to invoke `this(...)` since all the fields always start out as the 0-value.

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
```diff
@@ -0,0 +1,200 @@
...
+
+
+    public ClusterMatchRecord clone() {
+      try {
+        return (ClusterMatchRecord) super.clone();
+      } catch (CloneNotSupportedException e) {
+        LOG.debug("Failed to clone: {}", this, e);
+        return new ClusterMatchRecord();
+      }
+    }
```
PapaCharlie, line 164: See above, can just be:
```suggestion
    public synchronized ClusterMatchRecord clone() {
      return new ClusterMatchRecord(_oldLb, _newLb, _uris, _matched);
    }
```

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
```diff
@@ -0,0 +1,200 @@
...
+
+    @Override
+    public String toString() {
```
PapaCharlie, line 167: Should be `synchronized`

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
```diff
@@ -0,0 +1,200 @@
...
+      return "ClusterMatchRecord{ " +
+          "\nTotal Uris: " + _uris + ", Matched: " + _matched +
+          "\nOld LB: " + _oldLb +
+          "\nNew LB: " + _newLb +
+          '}';
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
PapaCharlie, line 176: Should be `synchronized`

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
```diff
@@ -0,0 +1,200 @@
...
+      ClusterMatchRecord o = (ClusterMatchRecord) obj;
+
+      return Objects.equals(_oldLb, o._oldLb)
+          && Objects.equals(_newLb, o._newLb)
+          && _uris == o._uris
+          && _matched == o._matched;
+    }
+
+    @Override
+    public int hashCode() {
```
PapaCharlie, line 196: Should be `synchronized`

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
bohhyang: since clone conflicts with the super class method, and IDE complains about not calling super class's clone, I just created a copy method instead

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
bohhyang: yeah, done

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
bohhyang: sure (.clone() doc says it does a shallow copy btw, but yeah, not using it now)

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
bohhyang: I hesitated about it since the cluster gets updated in the block (like cluster.uris = xxx), but that should be ok. Updated.

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
PapaCharlie: Sounds good to me 👍

d2/src/test/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitorTest.java
```diff
@@ -114,6 +124,75 @@ public void testReportData() {
...
+        },
+        {
+            Arrays.asList(
+                new ImmutablePair<>(URI_PROPERTIES_1, false),
+                new ImmutablePair<>(URI_PROPERTIES_1, true),
+                new ImmutablePair<>(URI_PROPERTIES_URI_1_AND_2, true),
+            new ImmutablePair<>(URI_PROPERTIES_URI_1_AND_2, false)
+            )
+        },
+        {
```
PapaCharlie, line 155: Inconsistent formatting?

d2/src/test/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitorTest.java
```diff
@@ -114,6 +124,75 @@ public void testReportData() {
...
+                new ImmutablePair<>(URI_PROPERTIES_1, true),
+                new ImmutablePair<>(URI_PROPERTIES_1, false),
+                new ImmutablePair<>(URI_PROPERTIES_URI_1_AND_2, true),
+                new ImmutablePair<>(URI_PROPERTIES_1, true)
+            )
+        }
+    };
+  }
+  @SuppressWarnings("ResultOfMethodCallIgnored")
+  @Test(dataProvider = "reportDataInMultiThreadsDataProvider")
```
PapaCharlie, line 168: A good way to sniff out race conditions is to run tests like this multiple times. IIRC the `@Test` decorator has a parameter for the count

d2/src/test/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitorTest.java
```diff
@@ -114,6 +124,75 @@ public void testReportData() {
...
+                new ImmutablePair<>(URI_PROPERTIES_URI_1_AND_2, false),
+                new ImmutablePair<>(URI_PROPERTIES_1, true),
+                new ImmutablePair<>(URI_PROPERTIES_1, false),
+                new ImmutablePair<>(URI_PROPERTIES_URI_1_AND_2, true),
+                new ImmutablePair<>(URI_PROPERTIES_1, true)
+            )
+        }
+    };
+  }
+  @SuppressWarnings("ResultOfMethodCallIgnored")
```
PapaCharlie, line 167: Nit: newline above this

d2/src/test/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitorTest.java
```diff
@@ -114,6 +124,75 @@ public void testReportData() {
...
+  }
+
+  private void reportAndVerifyState(UriPropertiesDualReadMonitor monitor, UriProperties prop, boolean fromNewLb) {
+    monitor.reportData(CLUSTER_1, prop, fromNewLb);
+    // if reportData on the same cluster are NOT synchronized, the total uris and matched uris counts could become < 0.
+    // e.g: when total uris = 1, matched uris = 1, if reporting URI_PROPERTIES_URI_1_AND_2 for new and old Lbs are
+    // executed concurrently, total uris will decrement by 1 twice and become -1.
+    // We verify that doesn't happen no matter what order the data is reported between the old and new Lbs.
+    assertTrue(monitor.getTotalUris() >= 0);
+    assertTrue(monitor.getMatchedUris() >= 0);
```
PapaCharlie, line 193: Can you figure out _exactly_ what these values should be and check that instead of just checking `0`?

bohhyang: It can be deterministic as it depends on which next task is picked up, right?

d2/src/test/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitorTest.java
bohhyang: makes sense, added.

d2/src/test/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitorTest.java
bohhyang: yeah, forgot to hit the auto-formatting before submission. Updated.

d2/src/test/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitorTest.java
```diff
@@ -0,0 +1,273 @@
...
+  }
+
+  private ClusterMatchRecord createMatchRecord(UriProperties oldLb, UriProperties newLb, int uris, int matched) {
+    ClusterMatchRecord cluster = new ClusterMatchRecord();
+    cluster._oldLb = oldLb;
+    cluster._newLb = newLb;
+    cluster._uris = uris;
+    cluster._matched = matched;
+    return cluster;
+  }
```
PapaCharlie, line 230: Maybe make that constructor `@VisibleForTesting`?

d2/src/test/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitorTest.java
```diff
@@ -0,0 +1,273 @@
...
+    ScheduledExecutorService executor = fixture.getExecutor();
+
+    CountDownLatch done = fixture.getDoneSignal(oldLbProps.size() + newLbProps.size());
+    // randomly report properties from old and new Lbs but properties in same lb are reported in order
+    executor.execute(() -> runNext(fixture, oldLbProps, false));
+    executor.execute(() -> runNext(fixture, newLbProps, true));
+
+    try {
+      boolean completed = done.await(1000, TimeUnit.MILLISECONDS);
+      if (completed) {
```
PapaCharlie, line 181: Technically, if `!completed` the test should fail right? You're better off doing a bar `done.await()` and adding the timeout flag in the `@Test` annotation. That way you don't have to explicitly handle the failure case. Similarly, I would add `throws Exception` to the test signature and avoid the explicit `catch (InterruptedException e)`. It should make the test easier to read

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
```diff
@@ -0,0 +1,208 @@
...
+      // add the remaining unmatched URIs in newLbUris to the uri count
+      cluster._uris += newLbUris.size();
+
+      if (cluster._matched != cluster._uris) {
+        String msg = String.format("Mismatched cluster properties for %s (match score: %f, total uris: %d):"
+                + "\nOld LB: %s\nNew LB: %s",
+            clusterName, (double) cluster._matched / (double) cluster._uris, cluster._uris, cluster._oldLb,
+            cluster._newLb);
+        RATE_LIMITED_LOGGER.info(msg);
+        LOG.debug(msg);
```
PapaCharlie, line 93: This is the kind of stuff that really should be avoided... It _always_ creates the string, even if it gets sent to DEBUG. This can create a lot of garbage. That's why the loggers take var args and lazily create the string only if logging at that level is enabled. Maybe you can make `RateLimitedLogger#logAllowed` public and do something like:
```
private static void infoOrDebugIfLimited(String msg, Object... args) {
  if (RATE_LIMITED_LOGGER.logAllowed()) {
    LOG.info(msg, args);
  } else {
    LOG.debug(msg, args);
  }
}
```
And that block becomes:
```suggestion
        infoOrDebugIfLimited(
            "Mismatched cluster properties for {} (match score: {}, total uris: {}):\nOld LB: {}\nNew LB: {}"
            clusterName, (double) cluster._matched / (double) cluster._uris, cluster._uris, cluster._oldLb,
            cluster._newLb);
```

bohhyang: makes a lot sense. Updated

d2/src/test/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitorTest.java
bohhyang: yeah, updated

d2/src/test/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitorTest.java
bohhyang: Actually it does time out rarely, so I added the “completed” check. I didn’t see anything wrong in the code though, do you?

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
PapaCharlie: We use the same trick now in the observer as well, and I really like it. Instead of losing the log statement altogether, it just gets hidden. If you turn on debug logging, you will see everything, but without duplicates.

bohhyang: yeah, it's nice.

d2/src/main/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitor.java
```diff
@@ -5,109 +5,120 @@
...
   private final DualReadLoadBalancerJmx _dualReadLoadBalancerJmx;
 
-  public UriPropertiesDualReadMonitor(DualReadLoadBalancerJmx dualReadLoadBalancerJmx) {
+  public UriPropertiesDualReadMonitor(DualReadLoadBalancerJmx dualReadLoadBalancerJmx)
+  {
     _dualReadLoadBalancerJmx = dualReadLoadBalancerJmx;
   }
 
-  public void reportData(String clusterName, UriProperties property, boolean fromNewLb) {
+  public synchronized void reportData(String clusterName, UriProperties property, boolean fromNewLb)
```
bohhyang, line 37: synchronizing the whole method will not allow update on different clusters happen at the same time. For apps that read hundreds of clusters, this may slow down the thread.

We can just add synchronized setters for _totalUris and _matchedUris.

bohhyang: well, since it's just for dual read, I guess it's fine.

bohhyang (approved):
LGTM

d2/src/test/java/com/linkedin/d2/balancer/dualread/UriPropertiesDualReadMonitorTest.java
bohhyang: turns out it was because the test has multi-thread accessing some non-thread safe vars in the monitor. Made the getter synchronized.

PR #1000, opened 2024-05-15
Handle HTTP/2 stream errors in tracker client and LB state updator

shivamgupta1:
# Summary
H/2 stream errors were seen in recent incidents and they aren't handled by the Tracker client currently. Handle `Http2Exception.StreamException` to fix this.

# Testing Done
Added UTs

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/clients/TrackerClientFactory.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/clients/TrackerClientFactory.java
@@ -54,9 +54,7 @@ public class TrackerClientFactory
 
   private static final int LOG_RATE_MS = 20000;
 
-  /**
-   * @see #createTrackerClient(URI, UriProperties, ServiceProperties, String, TransportClient, Clock)
-   */
+  @Deprecated
   public static TrackerClient createTrackerClient(URI uri,
                                            UriProperties uriProperties,
                                            ServiceProperties serviceProperties,
@@ -64,6 +62,17 @@ public class TrackerClientFactory
                                            TransportClient transportClient)
   {
     return createTrackerClient(uri, uriProperties, serviceProperties, loadBalancerStrategyName, transportClient, SystemClock.instance());
+  }
+
+  public static TrackerClient createTrackerClient(URI uri,
+      UriProperties uriProperties,
+      ServiceProperties serviceProperties,
+      String loadBalancerStrategyName,
+      TransportClient transportClient,
+      boolean loadBalanceStreamException)
+  {
+    return createTrackerClient(uri, uriProperties, serviceProperties, loadBalancerStrategyName, transportClient,
+        SystemClock.instance(), loadBalanceStreamException);
   }
 
   /**
@@ -77,12 +86,25 @@ public class TrackerClientFactory
    * @param clock Clock used for internal call tracking.
    * @return TrackerClient
    */
+  @Deprecated
   public static TrackerClient createTrackerClient(URI uri,
                                                   UriProperties uriProperties,
                                                   ServiceProperties serviceProperties,
                                                   String loadBalancerStrategyName,
                                                   TransportClient transportClient,
                                                   Clock clock)
+  {
+    return createTrackerClient(uri, uriProperties, serviceProperties, loadBalancerStrategyName, transportClient, clock,
+        false);
+  }
+
+  public static TrackerClient createTrackerClient(URI uri,
+      UriProperties uriProperties,
+      ServiceProperties serviceProperties,
+      String loadBalancerStrategyName,
+      TransportClient transportClient,
+      Clock clock,
+      boolean loadBalanceStreamException)
   {
     TrackerClient trackerClient;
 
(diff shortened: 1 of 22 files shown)


d2/src/test/java/com/linkedin/d2/balancer/strategies/relative/StateUpdaterTest.java
```diff
@@ -524,6 +523,34 @@ Object[][] getTrackerClients()
...
+        .setClusterGenerationId(DEFAULT_CLUSTER_GENERATION_ID)
+        .setTrackerClientStateMap(trackerClients,
+            Arrays.asList(StateUpdater.MAX_HEALTH_SCORE, StateUpdater.MAX_HEALTH_SCORE),
+            Arrays.asList(TrackerClientState.HealthState.HEALTHY, TrackerClientState.HealthState.HEALTHY),
+            Arrays.asList(30, 30))
+        .build();
+
+    ConcurrentMap<Integer, PartitionState> partitionLoadBalancerStateMap = new ConcurrentHashMap<>();
+    partitionLoadBalancerStateMap.put(DEFAULT_PARTITION_ID, state);
+    setup(new D2RelativeStrategyProperties().setHighErrorRate(0.5), partitionLoadBalancerStateMap);
```
shivamgupta1, line 545: The test utilizes Relative strategy, but the `StateUpdater` code is shared b/w relative and degrader, so it should be okay.

d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/StateUpdater.java
```diff
@@ -440,9 +440,11 @@ private static double getErrorRate(Map<ErrorType, Integer> errorTypeCounts, int
     Integer closedChannelExceptionCount = errorTypeCounts.getOrDefault(ErrorType.CLOSED_CHANNEL_EXCEPTION, 0);
     Integer serverErrorCount = errorTypeCounts.getOrDefault(ErrorType.SERVER_ERROR, 0);
     Integer timeoutExceptionCount = errorTypeCounts.getOrDefault(ErrorType.TIMEOUT_EXCEPTION, 0);
+    Integer streamErrorCount = errorTypeCounts.getOrDefault(ErrorType.STREAM_ERROR, 0);
     return callCount == 0
         ? 0
-        : (double) (connectExceptionCount + closedChannelExceptionCount + serverErrorCount + timeoutExceptionCount) / callCount;
+        : (double) (connectExceptionCount + closedChannelExceptionCount + serverErrorCount + timeoutExceptionCount +
+            streamErrorCount) / callCount;
```
bohhyang, line 447: since this could impact all LB behaviors, we need to be careful and put this change behind a config. cc: @nizarm

bohhyang (dismissed):
DegraderLoadBalancerStrategyV3 doesn't use StateUpdater. It uses DegraderTrackerClientUpdater to update the client state, which uses DegraderControl which then uses [DegraderImpl](https://jarvis.corp.linkedin.com/codesearch/result/?name=DegraderImpl.java&path=rest.li%2Fdegrader%2Fsrc%2Fmain%2Fjava%2Fcom%2Flinkedin%2Futil%2Fdegrader&reponame=linkedin%2Frest.li#516). 

Please adjust the logic there (put behind the config as suggested above) and you may add a UT for DegraderImpl to verify the isHigh result, which is critical to increase computed drop rate (hence quarantine).

d2/src/main/java/com/linkedin/d2/balancer/clients/TrackerClientFactory.java
```diff
@@ -66,6 +64,17 @@ public static TrackerClient createTrackerClient(URI uri,
...
 
+  public static TrackerClient createTrackerClient(URI uri,
+      UriProperties uriProperties,
+      ServiceProperties serviceProperties,
+      String loadBalancerStrategyName,
+      TransportClient transportClient,
+      boolean loadBalanceStreamException)
+  {
+    return createTrackerClient(uri, uriProperties, serviceProperties, loadBalancerStrategyName, transportClient,
+        SystemClock.instance(), loadBalanceStreamException);
```
shivamgupta1, line 75: Can we just propagate `D2ClientConfig` everywhere to avoid having to do this each time a change needs to be made to tracker client, state updater, degrader, etc.

bohhyang:
The way the config (in D2ClientConfig) is passed in is:
1) Relative LB: D2ClientBuilder -> RelativeLoadBalancerStrategyFactory
2) Degrader LB:
       1. ZK: ZKFSLoadBalancerWithFacilitiesFactory -> ZKFSTogglingLoadBalancerFactoryImpl -> SimpleLoadBalancerState -> TrackerClientFactory -> DegraderTrackerClientImpl -> DegraderImpl.Config
       2. xDS: XdsLoadBalancerWithFacilitiesFactory -> XdsFsTogglingLoadBalancerFactory -> SimpleLoadBalancerState -> Same as ZK above.

d2/src/main/java/com/linkedin/d2/balancer/clients/TrackerClientFactory.java
bohhyang: It will be weird for inner components like DegraderImpl.Config, DegraderTrackerClientImpl, etc. to hold a D2ClientConfig, which is the upper/larger layer that's supposed to control everything in d2 client.

shivamgupta1: In that case, can a new config object be created for the LB components such as tracker clients, state updater, etc.? It is essential for new code to be open for extensions.

bohhyang: adding a StateUpdater.LoadBalanceConfig makes sense. But adding one to TrackerClientFactory will be hard since the createTrackerClient method is static.

bohhyang: Added TODOs for optimizing config management later. 
Since we don't have bandwidth to address this now, and this PR is for an Incident action item, it doesn't worth holding off the merge for this optimization now.

bohhyang (approved):
aligned that we'll leave the config optimization as a TODO for future. Merging this in as an urgent incident action item.

PR #1001, opened 2024-05-23
Return RetriableRequestException for Netty Max Active Stream error

xt-dev:
This change adds logic in `HttpBridge` to return `RetriableRequestException` rather than `RemoteInvocationException` when StreamException from Netty is returned with cause of `Maximum active streams violated for this endpoint`. This allows the downstream client - `RetryClient` to retry the request with different hosts when stream hits the maximum of a given host.

[wc-test passed](https://crt.prod.linkedin.com/#/testing/executions/b4c75c5a-b5c9-4a4d-93da-2ebae16484e6)

### DIFF
--- a/r2-core/src/main/java/com/linkedin/r2/transport/http/common/HttpBridge.java
+++ b/r2-core/src/main/java/com/linkedin/r2/transport/http/common/HttpBridge.java
@@ -19,6 +19,7 @@ package com.linkedin.r2.transport.http.common;
 
 
 import com.linkedin.r2.RemoteInvocationException;
+import com.linkedin.r2.RetriableRequestException;
 import com.linkedin.r2.message.Request;
 import com.linkedin.r2.message.rest.RestException;
 import com.linkedin.r2.message.rest.RestRequest;
@@ -42,6 +43,9 @@ import java.util.Map;
  */
 public class HttpBridge
 {
+  private static final String NETTY_MAX_ACTIVE_STREAM_ERROR_MESSAGE =
+      "Maximum active streams violated for this endpoint";
+
   /**
    * Wrap application callback for incoming RestResponse with a "generic" HTTP callback.
    *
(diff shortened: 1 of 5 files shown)


r2-core/src/main/java/com/linkedin/r2/transport/http/common/HttpBridge.java
```diff
@@ -149,11 +153,21 @@ public void onResponse(TransportResponse<StreamResponse> response)
...
-          response =
-              TransportResponseImpl.error(new RemoteInvocationException("Failed to get response from server for URI "
-                                                                            + uri,
-                                                                        response.getError()),
-                                          response.getWireAttributes());
+          Throwable responseError = response.getError();
+          // If the error is due to the netty max active stream error, return a RetriableRequestException
+          if (shouldReturnRetriableRequestException(responseError))
+          {
+            StreamException streamException = (StreamException) responseError;
```
TylerHorth, line 160: Is this cast necessary?

r2-core/src/main/java/com/linkedin/r2/transport/http/common/HttpBridge.java
```diff
@@ -149,11 +153,21 @@ public void onResponse(TransportResponse<StreamResponse> response)
...
+            response = TransportResponseImpl.error(
+                new RetriableRequestException("Failed to get response from server for URI " + uri, streamException),
```
TylerHorth, line 162: Can we avoid duplicating the exception message?

```java
RemoteInvocationException wrapException(String message, Throwable exception) {
  if (shouldReturnRetriableRequestException(exception)) {
    return new RetriableRequestException(message, exception);
  } else {
    return new RemoteInvocationException(message, exception);
  }
}
```

r2-core/src/main/java/com/linkedin/r2/transport/http/common/HttpBridge.java
```diff
@@ -209,6 +223,12 @@ public void onResponse(TransportResponse<StreamResponse> response)
     };
   }
 
+  private static boolean shouldReturnRetriableRequestException(Throwable responseError)
+  {
+    return responseError instanceof StreamException && responseError.getMessage()
```
TylerHorth, line 228: Is this the right StreamException? Shouldn't it be [Http2Exception.StreamException](https://github.com/netty/netty/blob/70d6a3f40d7e8fd3f5ced7600ed209c58944f673/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Exception.java#L267)?

shivamgupta1 (approved):
thanks

TylerHorth (approved):
LGTM

TylerHorth (approved):

PR #1004, opened 2024-06-14
Add warn logs for receiving null properties

bohhyang:
## Summary
As the title. Add these logs for detecting potential issues and debugging. If the null guard in property event bus works as expected and xDS server sends valid data, these logs should NEVER appear (except the harmless one in FileStore).

## Test Done
unit tests

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/simple/ClusterLoadBalancerSubscriber.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/simple/ClusterLoadBalancerSubscriber.java
@@ -77,6 +77,7 @@ class ClusterLoadBalancerSubscriber extends
     }
     else
     {
+      _log.warn("Received a null cluster properties for {}", listenTo);
       // still insert the ClusterInfoItem when discoveryProperties is null, but don't create accessor
       _simpleLoadBalancerState.getClusterInfo().put(listenTo,
         new ClusterInfoItem(_simpleLoadBalancerState, null, null, null));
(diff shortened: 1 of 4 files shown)


d2/src/main/java/com/linkedin/d2/balancer/simple/ClusterLoadBalancerSubscriber.java
```diff
@@ -77,6 +77,7 @@ protected void handlePut(final String listenTo, final ClusterProperties discover
     }
     else
     {
+      _log.warn("Received a null cluster properties for {}", listenTo);
```
PapaCharlie, line 80: Last time we added logs around, they ended up being very noisy. Can you double check this isn't the case for this logger?

bohhyang: this log should not be logged unless there is a bug. And if there is a bug, we would want to see this log and take actions. So I think this log should be a must-show. I can make it an error instead?

bohhyang: I guess the Uri one could use a rate limiter in case of huge clusters having many uris. The service and cluster data won't change frequently anyway.

PapaCharlie: A rate limit seems like a good idea. I agree we want to fail loudly, but we don't want to fail so loudly it breaks people's logs to the point where we're scrambling to put in a fix! This has happened a few times already

PapaCharlie (approved):

PR #1007, opened 2024-06-24
Cancel xds stream ready timeout when the stream is closed. Correct xds connection status metric.

bohhyang:
## Summary
See context at [slack discussion](https://linkedin-randd.slack.com/archives/C02LGUWBLU8/p1718914811474299?thread_ts=1718903018.359899&cid=C02LGUWBLU8). 

This change:
1. cleans up xds stream ready timeout future when that stream is closed (instead of being ready). 
2. align xds connection status metric with what the connection listeners receive (in the previous edge case, the connected is indeed good, but the listeners are mis-notified that it's not connected).  
3. cleans up connection retry future when that reconnection stream is ready.
4. After all, if similar issues happen in future where xds stream is actually ready and taking in updates but the event bus is incorrectly enabled with backup store, updated XdsToD2PropertiesAdaptor will notify that xds is reconnected when it receives an update.

## Tests
N/A

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
@@ -214,7 +214,7 @@ public class XdsClientImpl extends XdsClient
       // to one of the sub-channels (unless an error or complete callback is called).
     }, _readyTimeoutMillis, TimeUnit.MILLISECONDS);
     _adsStream.start();
-    _log.info("Staring ADS stream, connecting to server: {}", _managedChannel.authority());
+    _log.info("Starting ADS stream, connecting to server: {}", _managedChannel.authority());
   }
 
   @Override
@@ -270,8 +270,6 @@ public class XdsClientImpl extends XdsClient
     _readyTimeoutFuture = null; // set it to null to avoid repeat notifications to subscribers.
     if (_retryRpcStreamFuture != null)
     {
-      _log.info("ADS stream reconnected after {} milliseconds",
-          _retryRpcStreamFuture.getDelay(TimeUnit.MILLISECONDS));
       _retryRpcStreamFuture = null;
       _xdsClientJmx.incrementReconnectionCount();
     }
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
@@ -257,26 +257,25 @@ public class XdsClientImpl extends XdsClient
       _log.warn("Unexpected state, ready called on null or backed off ADS stream!");
       return;
     }
-    // confirm ready state to neglect spurious callbacks; we'll get another callback whenever it is ready again.
-    if (_adsStream.isReady())
+    // Confirm ready state to neglect spurious callbacks; we'll get another callback whenever it is ready again.
+    // Also confirm ready timeout future is not null to avoid notifying multiple times.
+    if (!_adsStream.isReady() || _readyTimeoutFuture == null)
     {
-      // if the ready timeout future is non-null, a reconnect notification hasn't been sent yet.
-      if (_readyTimeoutFuture != null)
-      {
-        // timeout task will be cancelled only if it hasn't already executed.
-        boolean cancelledTimeout = _readyTimeoutFuture.cancel(false);
-        _log.info("ADS stream ready, cancelled timeout task: {}", cancelledTimeout);
-        _readyTimeoutFuture = null; // set it to null to avoid repeat notifications to subscribers.
-        if (_retryRpcStreamFuture != null)
-        {
-          _log.info("ADS stream reconnected after {} milliseconds",
-              _retryRpcStreamFuture.getDelay(TimeUnit.MILLISECONDS));
-          _retryRpcStreamFuture = null;
-          _xdsClientJmx.incrementReconnectionCount();
-        }
-        notifyStreamReconnect();
-      }
+      return;
     }
+
+    // timeout task will be cancelled only if it hasn't already executed.
+    boolean cancelledTimeout = _readyTimeoutFuture.cancel(false);
+    _log.info("ADS stream ready, cancelled timeout task: {}", cancelledTimeout);
+    _readyTimeoutFuture = null; // set it to null to avoid repeat notifications to subscribers.
+    if (_retryRpcStreamFuture != null)
+    {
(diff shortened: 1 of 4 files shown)


d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -267,10 +267,15 @@ private void readyHandler()
         boolean cancelledTimeout = _readyTimeoutFuture.cancel(false);
         _log.info("ADS stream ready, cancelled timeout task: {}", cancelledTimeout);
         _readyTimeoutFuture = null; // set it to null to avoid repeat notifications to subscribers.
-        _xdsClientJmx.incrementReconnectionCount();
+        if (_retryRpcStreamFuture != null)
```
brycezhongqing, line 270: Nit : Could we early returns to reduce nesting
```java
   if (!_adsStream.isReady())
    {
      return;
    }

    // if the ready timeout future is non-null, a reconnect notification hasn't been sent yet.
    if (_readyTimeoutFuture != null)
    {
      // timeout task will be cancelled only if it hasn't already executed.
      boolean cancelledTimeout = _readyTimeoutFuture.cancel(false);
      _log.info("ADS stream ready, cancelled timeout task: {}", cancelledTimeout);
      _readyTimeoutFuture = null; // set it to null to avoid repeat notifications to subscribers.
      if (_retryRpcStreamFuture != null)
      {
        _log.info("ADS stream reconnected after {} milliseconds",
                  _retryRpcStreamFuture.getDelay(TimeUnit.MILLISECONDS));
        _retryRpcStreamFuture = null;
        _xdsClientJmx.incrementReconnectionCount();
      }
      notifyStreamReconnect();
    }
```

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
```diff
@@ -284,6 +290,12 @@ public void onChanged(XdsClient.NodeUpdate update)
             _clusterEventBus.publishInitialize(clusterName, null);
           }
         }
+        else
+        {
+          // Received xds resource update while service event bus is not set. Notify that the xds source becomes
```
brycezhongqing: Nit this is the`Cluster Event bus`

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
```diff
@@ -542,6 +554,12 @@ private void mergeAndPublishUris(String clusterName)
       {
         _uriEventBus.publishInitialize(clusterName, mergedUriProperties);
       }
+      else
+      {
+        // Received xds resource update while service event bus is not set. Notify that the xds source becomes
```
brycezhongqing: Nit: This is th `Uri EventBus`

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: yeah, updated

brycezhongqing (approved):

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -257,21 +257,25 @@ private void readyHandler()
...
+    // timeout task will be cancelled only if it hasn't already executed.
+    boolean cancelledTimeout = _readyTimeoutFuture.cancel(false);
+    _log.info("ADS stream ready, cancelled timeout task: {}", cancelledTimeout);
+    _readyTimeoutFuture = null; // set it to null to avoid repeat notifications to subscribers.
+    if (_retryRpcStreamFuture != null)
+    {
+      _log.info("ADS stream reconnected after {} milliseconds",
+          _retryRpcStreamFuture.getDelay(TimeUnit.MILLISECONDS));
+      _retryRpcStreamFuture = null;
+      _xdsClientJmx.incrementReconnectionCount();
```
shivamgupta1, line 276: All these conditions seem to be precluded by the `isInBackoff` check above at line 255? Also, `getDelay` seems to return the still remaining delay and not the elapsed time.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -257,21 +257,25 @@ private void readyHandler()
...
-        notifyStreamReconnect();
-      }
-      _xdsClientJmx.setIsConnected(true);
+      return;
+    }
+
+    // timeout task will be cancelled only if it hasn't already executed.
+    boolean cancelledTimeout = _readyTimeoutFuture.cancel(false);
+    _log.info("ADS stream ready, cancelled timeout task: {}", cancelledTimeout);
+    _readyTimeoutFuture = null; // set it to null to avoid repeat notifications to subscribers.
```
shivamgupta1, line 270: Although not too bad, this might be some unneeded refactoring? (don't fix what ain't broken)

bohhyang: yeah, it's just return-early, as suggested by @brycezhongqing above.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: > precluded by the isInBackoff check above at line 255
not really. This is not for checking backoff but just to clean _retryRpcStreamFuture for the case when the retry future is executed (and the reconnect stream is ready), but the future itself is not cleaned (previously reconnectionCount metric was incremented for initial connection as well, which was not a reconnect).

> getDelay seems to return the still remaining delay
whoops, that's right. Removed, thanks.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -214,7 +214,7 @@ private void startRpcStreamLocal() {
       // to one of the sub-channels (unless an error or complete callback is called).
     }, _readyTimeoutMillis, TimeUnit.MILLISECONDS);
     _adsStream.start();
-    _log.info("ADS stream started, connected to server: {}", _managedChannel.authority());
+    _log.info("Staring ADS stream, connecting to server: {}", _managedChannel.authority());
```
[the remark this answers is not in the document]
bohhyang: fixed. thanks.

PR #1009, opened 2024-07-11
Allow both @extension and @grpcExtension extensions in schema validation

SpencerKwok:
### Summary

To help bridge rest.li with gRpc as part of the gRpc migration, this PR allows @extension and @grpcExtension extensions to co-exist in the same schema. More information can be found in [SI-39662](https://jira01.corp.linkedin.com:8443/browse/SI-39662).

### Tests

I added a unit test to check if @extension and @grpcExtension extensions can co-exist in the same schema. I depend on the previous existing tests to check for backwards compatibility.

### DIFF
--- a/restli-tools/src/test/extensions/validCase/BazExtensions.pdl
+++ b/restli-tools/src/test/extensions/validCase/BazExtensions.pdl
@@ -1,0 +1,10 @@
+/**
+ * Valid extension schema:
+ * The co-existence of @extension and @grpcExtension is allowed
+ */
+record BazExtensions includes Baz {
+  @extension.using = "finder: test"
+  @grpcExtension.rpc = "get"
+  @grpcExtension.versionSuffix = "V2"
+  testField: array[DummyKeyWithGrpc]
+}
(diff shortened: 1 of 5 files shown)


restli-tools/src/test/extensions/validCase/BazExtensions.pdl
```diff
@@ -0,0 +1,10 @@
+/**
+ * Valid extension schema:
+ * The co-existence of @extension and @grpcExtension is allowed
+ */
+record BazExtensions includes Baz {
+  @extension.using = "finder: test"
+  @grpcExtension.rpc = "get"
```
aman1309, line 7: what would be the expectation for translating this annotation to proto during migration?

aman1309 (approved):
LGTM

restli-tools/src/test/extensions/validCase/BazExtensions.pdl
SpencerKwok: The proto schema would look something like [this](https://jarvis.corp.linkedin.com/codesearch/result/?name=BarExtensions.proto&path=entity-relationship-infra%2Fgenerator%2Fsrc%2Ftest%2FprotoExtensions%2Fcom%2Flinkedin%2Fentityrelationship%2Fgenerator&reponame=linkedin-multiproduct%2Fentity-relationship-infra#50).

PR #1010, opened 2024-07-11
Increase verbosity of testExtensionSchemaValidation tests

SpencerKwok:
### Summary

When testExtensionSchemaValidation tests fail, the reason behind those failures are not logged. This PR replaces `Assert` with `SoftAssert` to always output the reason for failure in the unit tests. Below are some example outputs:

Before change:
```
Failure message for com.linkedin.restli.tools.data.TestExtensionSchemaValidationCmdLineApp > testExtensionSchemaValidation[0](validCase, true, null):
    The following asserts failed:
        expected [true] but found [false]
```

After change:
```
Failure message for com.linkedin.restli.tools.data.TestExtensionSchemaValidationCmdLineApp > testExtensionSchemaValidation[0](validCase, true, null):
    The following asserts failed:
        expected [true] but found [false],
        expected [null] but found [versionSuffix value: 'V3' does not match the versionSuffix value which was defined in resourceKey/grpcService annotation]
```

In addition, this PR updates `ValidateExtensionSchemaTask` to output schema validation errors to the build log.

More information can be found at [SI-40347](https://jira01.corp.linkedin.com:8443/browse/SI-40347).

### Tests
I manually modified valid test cases to be invalid and invalid test cases to be valid to check for error message improvements.

### DIFF
--- a/restli-tools/src/test/java/com/linkedin/restli/tools/data/TestExtensionSchemaValidationCmdLineApp.java
+++ b/restli-tools/src/test/java/com/linkedin/restli/tools/data/TestExtensionSchemaValidationCmdLineApp.java
@@ -86,20 +86,19 @@ public class TestExtensionSchemaValidationCmdLineApp
   {
       String resolverPath = testPegasusDir;
       String inputPath = testExtensionDir + File.separator + inputDir;
+
+      SoftAssert softAssert = new SoftAssert();
       try
       {
         ExtensionSchemaValidationCmdLineApp.parseAndValidateExtensionSchemas(resolverPath, new File(inputPath));
-        SoftAssert softAssert = new SoftAssert();
         softAssert.assertTrue(isValid);
         softAssert.assertEquals(null, errorMessage);
-        softAssert.assertAll();
       }
       catch (Exception e)
       {
-        SoftAssert softAssert = new SoftAssert();
         softAssert.assertTrue(!isValid);
         softAssert.assertEquals(e.getMessage(), errorMessage);
-        softAssert.assertAll();
       }
+      softAssert.assertAll();
   }
 }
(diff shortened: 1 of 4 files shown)


restli-tools/src/test/java/com/linkedin/restli/tools/data/TestExtensionSchemaValidationCmdLineApp.java
```diff
@@ -89,12 +89,17 @@ public void testExtensionSchemaValidation(String inputDir, boolean isValid, Stri
...
       }
       catch (Exception e)
       {
-        Assert.assertTrue(!isValid);
-        Assert.assertEquals(e.getMessage(), errorMessage);
+        SoftAssert softAssert = new SoftAssert();
+        softAssert.assertTrue(!isValid);
+        softAssert.assertEquals(e.getMessage(), errorMessage);
+        softAssert.assertAll();
       }
```
zackthehuman, line 103: It would make more sense to me to create a single `SoftAssert` and then `assertAll()` at the end:

```
SoftAssert softAssert = new SoftAssert();

try
{
  ExtensionSchemaValidationCmdLineApp.parseAndValidateExtensionSchemas(resolverPath, new File(inputPath));
  softAssert.assertTrue(isValid);
  softAssert.assertEquals(null, errorMessage);
}
catch (Exception e)
{
  softAssert.assertTrue(!isValid);
  softAssert.assertEquals(e.getMessage(), errorMessage);
}

softAssert.assertAll();
```

Otherwise, looks good to me.

zackthehuman (changes requested):

restli-tools/src/test/java/com/linkedin/restli/tools/data/TestExtensionSchemaValidationCmdLineApp.java
SpencerKwok: Ah yes, I can make that change

zackthehuman (approved):
Looks good to me!

PR #1011, opened 2024-07-18
Respect glob collection subscriptions on reconnect

PapaCharlie:
The RetryTask did not resubscribe to glob collections after a reconnect. Instead it goes back to map subscriptions. This change honors the glob collection config in the retry task.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,11 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.58.5] - 2024-09-04
+- Respect glob collection subscriptions on reconnect
+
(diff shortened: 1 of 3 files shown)


shivamgupta1 (commented):
LGTM, please update testing details. Recommend to at least test locally for correctness.

bohhyang (commented):
nice catch. LGTM. Could deploy indis-canary with this on QEI, find the connected EI observer in the log, use admin portal to let that observer kill all connections, then inspect the log to see the requests sent to the new observer that gets connected.

PapaCharlie:
Tested locally that this works as expected. After a reconnect, it resubscribes with glob collections

shivamgupta1 (approved):

PR #1014, opened 2024-08-09
Remove some dual read related tests and deprecate dual read LB

bohhyang:
1. Fixed a warm-up test about deleting files after shutdown. 
2. Some dual read related tests for WarmUpLoadBalancer has been flaky. Since dual read will be deprecated soon, disable these tests to smooth the release. 
3. As we get close to complete INDIS migration, marking dual read LB as deprecated to note that no new reference to it should be created anymore.


brycezhongqing (approved):
LGTM

PR #1020, opened 2024-09-13
Add watchAllResource subscriber

brycezhongqing:
As the title.

# Test Done
[unit test] Added
[manual test]Test has been finished in restli-resource-explorer #337

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.58.7] - 2024-09-13
+Add WildcardResourceSubscriber for restli-resource-explorer subscribe to all resources
+
 ## [29.58.6] - 2024-09-08
 - Allow for null paging inside Collection response envelopes
 
@@ -5728,7 +5731,8 @@ patch operations can re-use these classes for generating patch messages.
 
 ## [0.14.1]
 
-[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.58.6...master
+[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.58.7...master
+[29.58.7]: https://github.com/linkedin/rest.li/compare/v29.58.6...v29.58.7
 [29.58.6]: https://github.com/linkedin/rest.li/compare/v29.58.5...v29.58.6
 [29.58.5]: https://github.com/linkedin/rest.li/compare/v29.58.4...v29.58.5
 [29.58.4]: https://github.com/linkedin/rest.li/compare/v29.58.3...v29.58.4
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and what APIs have changed, if applicable.
 ## [Unreleased]
 
 ## [29.58.7] - 2024-09-13
-Add WildcardResourceSubscriber for restli-resource-explorer subscribe to all resources
+- Add WildcardResourceSubscriber which could subscribe to all resources, like NODE and URIMap resources.
 
 ## [29.58.6] - 2024-09-08
 - Allow for null paging inside Collection response envelopes
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
@@ -103,7 +103,7 @@ public abstract class XdsClient
      * Defining a private constructor means only classes that are defined in this file can extend this class (see
      * {@link ResourceWatcher}).
      */
-    private WildcardResourceWatcher(ResourceType type)
+    WildcardResourceWatcher(ResourceType type)
     {
       _type = type;
     }
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
@@ -355,11 +355,23 @@ public abstract class XdsClient
    */
   public abstract void watchAllXdsResources(WildcardResourceWatcher watcher);
 
+  /**
+   * Initiates the RPC stream to the xDS server.
+   */
   public abstract void startRpcStream();
 
+  /**
+   * Shuts down the xDS client.
+   */
   public abstract void shutdown();
 
+  /**
+   * Returns the authority of the xDS server.
+   */
   public abstract String getXdsServerAuthority();
 
+  /**
+   * Returns the JMX bean for the xDS client.
+   */
   public abstract XdsClientJmx getXdsClientJmx();
 }
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
@@ -95,6 +95,71 @@ public abstract class XdsClient
     }
   }
 
+  public static abstract class WildcardResourceWatcher
+  {
+    private final ResourceType _type;
+
+    /**
+     * Defining a private constructor means only classes that are defined in this file can extend this class (see
+     * {@link ResourceWatcher}).
+     */
+    private WildcardResourceWatcher(ResourceType type)
+    {
+      _type = type;
+    }
+
+    final ResourceType getType()
+    {
+      return _type;
+    }
+
+    /**
+     * Called when the resource discovery RPC encounters some transient error.
+     */
+    public abstract void onError(Status error);
+
+    /**
+     * Called when the resource discovery RPC reestablishes connection.
+     */
+    public abstract void onReconnect();
+
+    abstract void onChanged(String resourceName, ResourceUpdate update);
+
+    public abstract void onRemoval(String resourceName);
+  }
+
+  public static abstract class WildcardNodeResourceWatcher extends WildcardResourceWatcher
+  {
+    public WildcardNodeResourceWatcher()
+    {
+      super(ResourceType.NODE);
+    }
+
+    public abstract void onChanged(String resourceName, NodeUpdate update);
+
+    @Override
+    final void onChanged(String resourceName, ResourceUpdate update)
+    {
+      onChanged(resourceName, (NodeUpdate) update);
+    }
+  }
+
+  public static abstract class WildcardD2URIMapResourceWatcher extends WildcardResourceWatcher
+  {
+    public WildcardD2URIMapResourceWatcher()
+    {
+      super(ResourceType.D2_URI_MAP);
+    }
+
+    public abstract void onChanged(String resourceName, D2URIMapUpdate update);
+
+    @Override
+    final void onChanged(String resourceName, ResourceUpdate update)
+    {
+      onChanged(resourceName, (D2URIMapUpdate) update);
+    }
+  }
+
   public interface ResourceUpdate
   {
     boolean isValid();
@@ -109,7 +174,7 @@ public abstract class XdsClient
       _nodeData = nodeData;
     }
 
-    XdsD2.Node getNodeData()
+    public XdsD2.Node getNodeData()
     {
       return _nodeData;
     }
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.58.7] - 2024-09-13
+Add WildcardResourceSubscriber for restli-resource-explorer subscribe to all resources
```
ZoabKapoor, line 18: Nit: should there be a `-` at the beginning of this line? 

Also, I don't think we necessarily need to mention `restli-resource-explorer` specifically here, any user could use `WildcardResourceSubscriber` to subscribe to all resources, right? Plus then we don't need to name an internal service

ZoabKapoor, line 18: Nit: might need a `-` at the start of this line. Also, I don't know if we need to mention `restli-resource-explorer` by name, no need to put the name of an internal app in the code + in theory any app may want to use `WildcardResourceSubscriber`

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
```diff
@@ -261,13 +326,20 @@ static ResourceType fromTypeUrl(String typeUrl)
...
+
+  /**
+   * Subscribes the given {@link WildcardResourceWatcher} to all the resources of the corresponding type. The watcher
+   * will be notified whenever a resource is added or removed. Repeated calls to this function with the same watcher
+   * will always notify the given watcher of the current data.
+   */
+  public abstract void watchAllXdsResources(WildcardResourceWatcher watcher);
 
-  abstract void startRpcStream();
+  public abstract void startRpcStream();
```
ZoabKapoor, line 338: Why do we need to make these methods public now? Looks like they weren't previously

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -77,6 +77,7 @@ public class XdsClientImpl extends XdsClient
       Arrays.stream(ResourceType.values())
           .filter(e -> e.typeUrl() != null)
           .collect(Collectors.toMap(Function.identity(), e -> new HashMap<>())));
+  private final Map<ResourceType, WildcardResourceSubscriber> _wildcardSubscribers = new HashMap<>();
```
ZoabKapoor, line 80: Nit: I think you could use `Maps.newEnumMap` here to match the above map.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -168,6 +169,32 @@ void watchXdsResource(String resourceName, ResourceWatcher watcher)
...
+        subscriber = new WildcardResourceSubscriber(watcher.getType());
+        _wildcardSubscribers.put(watcher.getType(), subscriber);
+
+        if (_adsStream == null && !isInBackoff())
+        {
+          startRpcStreamLocal();
+        }
+        if (_adsStream != null)
+        {
+          _adsStream.sendDiscoveryRequest(watcher.getType(), Collections.singletonList("*"));
```
ZoabKapoor, line 190: Question for my knowledge: if the base `ResourceSubscriber` is already capable of doing glob collection, why wouldn't having a regular resource subscriber and then calling `watchXdsResource("*", ResourceWatcher myWatcher)` work to subscribe to everything?

(vs implementing a specific `WildcardResourceSubscriber`)

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -690,24 +741,131 @@ void onRemoval()
...
-              .collect(Collectors.toSet());
-          type = ResourceType.D2_URI;
+              .collect(Collectors.toCollection(HashSet::new));
+          rewrittenType = ResourceType.D2_URI;
+        }
+        else
+        {
+          rewrittenType = type;
+        }
+        if (getWildcardResourceSubscriber(type) != null)
```
ZoabKapoor, line 864: It took me a while to understand what's going on here -- looks like we introduce `rewrittenType` because we need the original `type` for the call to `getWildcardResourceSubscriber(type)`, right?

I think one option could be to put this `if` above `if (_subscribeToUriGlobCollection && type == ResourceType.D2_URI_MAP)`, so it gets the original type before reassigning it, that might be cleaner. Either way, it might also be worth adding a comment about the need to look up by the original type before transforming the `D2UriMap` request into a `D2URI` glob request

d2/src/test/java/com/linkedin/d2/xds/TestXdsClientImpl.java
```diff
@@ -443,6 +444,37 @@ public void testHandleD2URICollectionResponseWithRemoval()
     Assert.assertEquals(actualData.getURIMap(), D2_URI_MAP_UPDATE_WITH_DATA1.getURIMap());
   }
 
+  @Test
+  public void testWildCardResourceSubscription()
```
ZoabKapoor, line 448: Can we add testing to ensure that the wildcard subscription also subscribes to multiple different resources?

I think in this case we only test one service `SERVICE_RESOURCE_NAME` and one cluster `CLUSTER_RESOURCE_NAME`; I know we had a bug recently where we didn't handle multiple clusters correctly, but the tests missed it

ZoabKapoor (dismissed):
Nice job! Mostly nits here, but a couple of bigger comments I thought would be good to discuss

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
brycezhongqing: Because `watchXdsResource` is a public class, so all the abstract method should be `public`

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
brycezhongqing: Good idea. Updated

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
brycezhongqing: As I said in the description, we need to provide the `watchAllXdsResources` method directly instead of mixing up with the former resource subsciber logic.

The `watchAllXdsResources` will not use the glob collection (which may be caused some potential risk like before). And is easy to be used to subscribe all the resources directlly.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
brycezhongqing: Actually, move before `if (_subscribeToUriGlobCollection && type == ResourceType.D2_URI_MAP)` makes no sense, as the resources is used to create new list 

```java
 resources = resources.stream()
              .map(GlobCollectionUtils::globCollectionUrlForClusterResource)
              .collect(Collectors.toCollection(HashSet::new));
          rewrittenType = ResourceType.D2_URI;
```
So I prefer add comment there.

d2/src/test/java/com/linkedin/d2/xds/TestXdsClientImpl.java
brycezhongqing: Actually, I thought we only need to verify the `WildcardResourceSubscriber` could subscribe the `Node` and `URIMap` resource.

The bug you mentioned actually always from INDIS server side, from the client side it should be good.

CHANGELOG.md
brycezhongqing: Yeah. Thanks for suggestions. Update

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
ZoabKapoor: I see, so the XdsClient previously was wrong and we're fixing it here?

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
```diff
@@ -95,6 +95,71 @@ final void onChanged(ResourceUpdate update)
...
+    public abstract void onError(Status error);
+
+    /**
+     * Called when the resource discovery RPC reestablishes connection.
+     */
+    public abstract void onReconnect();
+
+    abstract void onChanged(String resourceName, ResourceUpdate update);
+
+    public abstract void onRemoval(String resourceName);
```
PapaCharlie, line 128: Let's add some docs on this too

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
```diff
@@ -95,6 +95,71 @@ final void onChanged(ResourceUpdate update)
...
+  }
+
+  public static abstract class WildcardNodeResourceWatcher extends WildcardResourceWatcher
+  {
+    public WildcardNodeResourceWatcher()
+    {
+      super(ResourceType.NODE);
+    }
+
+    public abstract void onChanged(String resourceName, NodeUpdate update);
```
PapaCharlie, line 138: Can you document this method?

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
```diff
@@ -95,6 +95,71 @@ final void onChanged(ResourceUpdate update)
...
+  }
+
+  public static abstract class WildcardD2URIMapResourceWatcher extends WildcardResourceWatcher
+  {
+    public WildcardD2URIMapResourceWatcher()
+    {
+      super(ResourceType.D2_URI_MAP);
+    }
+
+    public abstract void onChanged(String resourceName, D2URIMapUpdate update);
```
PapaCharlie, line 154: Same

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
PapaCharlie: Yes exactly

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
PapaCharlie: Yeah basically it needs separate code path where one subscriber can listen to multiple resources at once. The glob collections are a special case because a glob collection actually represents a map, and the `XdsClient` exposes updates to the collection as such. This way consumers of the `XdsClient` don't have to know whether glob collections are enabled.

d2/src/test/java/com/linkedin/d2/xds/TestXdsClientImpl.java
```diff
@@ -443,6 +443,37 @@ public void testHandleD2URICollectionResponseWithRemoval()
...
```
PapaCharlie, line 447: Can you add a test to validate that calling `watchAllXdsResources` multiple times (even if it's the same subscriber) also replays the updates, just like for `watchXdsResource`? That way it's consistent

PapaCharlie (changes requested):

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
brycezhongqing: Updated

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
brycezhongqing: Done

d2/src/test/java/com/linkedin/d2/xds/TestXdsClientImpl.java
brycezhongqing: After dicussion, it's ok for current unit test

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
```diff
@@ -261,13 +347,20 @@ static ResourceType fromTypeUrl(String typeUrl)
...
+  public abstract void startRpcStream();
 
-  abstract void shutdown();
+  public abstract void shutdown();
 
-  abstract String getXdsServerAuthority();
+  public abstract String getXdsServerAuthority();
 
-  abstract public XdsClientJmx getXdsClientJmx();
+  public abstract XdsClientJmx getXdsClientJmx();
```
PapaCharlie, line 364: Now that these are `public`, we should also document these

brycezhongqing: For sure. Added

PR #1022, opened 2024-09-20
Fix invalid handling of glob collections for wildcard subscribers

PapaCharlie:
As the title says, things weren't wired through correctly for glob collections, namely they were being ignored because they were triggering the check for whtehre the cluster is being watched at all. This fixes that (and the fact that the initial subscription did not respect the `useGlobCollections` flag). This change has been unit tested and tested through in restli-resource-explorer, and it works for both glob and non-glob.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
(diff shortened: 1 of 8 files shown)


brycezhongqing (approved):
LGTM

PR #1028, opened 2024-10-17
Add support for initial resource versions in xds client

shivamgupta1:
Setting the `initial_resource_versions` property in a delta resource request allows the xDS server to avoid re-sending data the client may already have from a previous successful connection to another server.

See https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/discovery/v3/discovery.proto#service-discovery-v3-deltadiscoveryrequest 

Changes:
- Track the incoming resource versions and send the version map to the server in the event of a re-connection.


ChrisCarini:

PR #1029, opened 2024-10-17
Restore the old constructor to avoid incompatible issue

brycezhongqing:
As title

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperConnectionManager.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperConnectionManager.java
@@ -184,6 +184,75 @@ public class ZooKeeperConnectionManager extends ConnectionManager
     }
   }
 
+  /**
+   * deprecated, use {@link  ConnectionManager#markDownAllServers(Callback)} instead.
+   */
+  @Deprecated
+  public void markDownAllServers(final Callback<None> callback)
+  {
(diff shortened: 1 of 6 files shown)


zackthehuman (approved):
Thanks for adding this back!

zackthehuman:
This should fix the regression from https://github.com/linkedin/rest.li/pull/1027 with the constructors.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperConnectionManager.java
```diff
@@ -184,6 +184,75 @@ protected None convertResponse(None none) throws Exception
     }
   }
 
+  /**
+   * deprecated, use {@link  ConnectionManager#markDownAllServers(Callback)} instead.
+   */
+  @Deprecated
+  public void markDownAllServers(final Callback<None> callback)
```
bohhyang, line 191: also needs "@Override", right? same for markUpAllServers

brycezhongqing: Actually. It no needs to add override, the java will get this automatically. But it's good to point it out

bohhyang: Yeah, just better for code readers.

bohhyang (approved):

PR #1032, opened 2024-10-24
Disable dark traffic dispatching during dark warmup

shivamgupta1:
Description
----
Sending dark traffic during dark warmup is risky since it can result in the calls being sent back to the same hosts, resulting in a cycle.

Changes:
- Use `ZookeeperAnnouncer` in dark cluster utils to ensure that no dark requests are sent while any of the announcers is in warmup mode (and the app is receiving some dark traffic).

Testing
-------
Added UT.

### DIFF
--- a/darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterManagerImpl.java
+++ b/darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterManagerImpl.java
@@ -16,6 +16,7 @@
 
 package com.linkedin.darkcluster.impl;
 
+import com.linkedin.d2.balancer.servers.ZooKeeperAnnouncer;
 import com.linkedin.darkcluster.api.DarkGateKeeper;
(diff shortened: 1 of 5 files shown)


bohhyang (approved):

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterManagerImpl.java
```diff
@@ -99,11 +115,17 @@ public DarkClusterManagerImpl(@Nonnull String sourceClusterName,
...
   }
 
   @Override
   public boolean handleDarkRequest(RestRequest originalRequest, RequestContext originalRequestContext)
   {
+    for (ZooKeeperAnnouncer announcer : _announcers) {
+      if (announcer.isWarmingUp()) {
+          return false;
+      }
+    }
```
brycezhongqing, line 128:
```suggestion
    if (_announcers.stream().anyMatch(ZooKeeperAnnouncer::isWarmingUp))
    {
      return false;
    }
```

PR #1033, opened 2024-10-25
check and warn for invalid partition weight

bohhyang:
## Summary
Announcer weight has been a double with no limit rules. Technically, a maximum of 100, and at most 2 digits after the decimal point should be sufficient to serve all load distribution needs in reality. With INDIS observer serving Envoy clients which require an integer uri weight, unlimited double number will cause conversion complexity that should never be encountered in real use case. In this PR, we add a max weight config which defines the max value and the number of decimal places allowed. We also added another config for the action to take when the weight rule is breached. It can be configured to ignore, log warn msg, throw exception, or rectify the value.

Note that this change will warn invalid weights set dynamically thru ZookeeperAnnouncerJmx as well.

Also added jmx methods to return the count of weight breaches for max value and decimal places, and the current markUp status of the announcer for regular cluster and dark warmup cluster respectively

See [container PR](https://github.com/linkedin-multiproduct/container/pull/1195) for LI-specific use case and tests. 

## Test Done
See container PR.

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
@@ -102,6 +102,9 @@ public class ZooKeeperAnnouncer implements D2ServiceDiscoveryEventHelper
    * For example, 100.00 means the max weight allowed is 100 and the max number of decimal places is 2.
    */
   private final BigDecimal _maxWeight;
+  /**
+   * The action to take when d2 weight breaches validation rules.
+   */
   private ActOnWeightBreach _actOnWeightBreach = ActOnWeightBreach.IGNORE;
   private volatile Map<String, Object> _uriSpecificProperties;
 
@@ -151,15 +154,14 @@ public class ZooKeeperAnnouncer implements D2ServiceDiscoveryEventHelper
   // Field to store the dark warm-up time duration in seconds, defaults to zero
   private final int _warmupDuration;
 
-  /**
-   * The action to take when d2 weight breaches validation rules.
-   */
   public enum ActOnWeightBreach {
     // Ignore and no op.
     IGNORE,
-    //
+    // only log warnings
     WARN,
+    // throw exceptions
     THROW,
+    // rectify the invalid weight (e.g: cap to the max, round to the nearest valid decimal places)
     RECTIFY
   }
 
@@ -245,7 +247,11 @@ public class ZooKeeperAnnouncer implements D2ServiceDiscoveryEventHelper
     _executorService = executorService;
     _eventEmitter = eventEmitter;
     _maxWeight = maxWeight;
-    _actOnWeightBreach = actOnWeightBreach;
+    
+    if (actOnWeightBreach != null)
+    {
+      _actOnWeightBreach = actOnWeightBreach;
+    }
 
     if (server instanceof ZooKeeperServer)
     {
--- a/d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
@@ -246,6 +246,8 @@ public class ZooKeeperAnnouncer implements D2ServiceDiscoveryEventHelper
     _warmupDuration = warmupDuration;
     _executorService = executorService;
     _eventEmitter = eventEmitter;
+
+    // take in max weight as a string to be precise at decimal places when creating a BigDecimal.
     _maxWeight = maxWeight == null ? null : new BigDecimal(maxWeight);
     if (actOnWeightBreach != null)
     {
@@ -873,6 +875,7 @@ public class ZooKeeperAnnouncer implements D2ServiceDiscoveryEventHelper
             throw getMaxWeightBreachException(weight, entry.getKey());
           case RECTIFY:
             entry.setValue(new PartitionData(_maxWeight.intValue()));
+            weight = _maxWeight;
             _log.warn("Capped weight {} in Partition {} to the max weight allowed: {}.", weight.doubleValue(),
                 entry.getKey(), _maxWeight.intValue());
             break;
--- a/d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
@@ -110,7 +110,7 @@ public class ZooKeeperAnnouncer implements D2ServiceDiscoveryEventHelper
   /**
    * The action to take when d2 weight breaches validation rules.
    */
-  private ActionOnWeightBreach _actionOnWeightBreach = ActionOnWeightBreach.IGNORE;
+  private final ActionOnWeightBreach _actionOnWeightBreach;
 
   private final AtomicInteger _maxWeightBreachedCount = new AtomicInteger(0);
   private final AtomicInteger _weightDecimalPlacesBreachedCount = new AtomicInteger(0);
@@ -267,10 +267,7 @@ public class ZooKeeperAnnouncer implements D2ServiceDiscoveryEventHelper
     _eventEmitter = eventEmitter;
 
     _maxWeight = maxWeight;
-    if (actionOnWeightBreach != null)
-    {
-      _actionOnWeightBreach = actionOnWeightBreach;
-    }
+    _actionOnWeightBreach = actionOnWeightBreach != null ? actionOnWeightBreach : ActionOnWeightBreach.IGNORE;
 
     if (server instanceof ZooKeeperServer)
     {
--- a/d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
@@ -848,6 +848,7 @@ public class ZooKeeperAnnouncer implements D2ServiceDiscoveryEventHelper
    */
   public boolean isWarmingUp() {
     return _isWarmingUp;
+  }
 
   private Map<Integer, PartitionData> validatePartitionData(Map<Integer, PartitionData> partitionData) {
     if (_maxWeight == null || _actOnWeightBreach == ActOnWeightBreach.IGNORE) {
--- a/d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
@@ -20,6 +20,7 @@
 
 package com.linkedin.d2.balancer.servers;
 
+import java.math.BigDecimal;
 import java.net.URI;
 import java.util.ArrayDeque;
 import java.util.Collections;
@@ -93,6 +94,7 @@ public class ZooKeeperAnnouncer implements D2ServiceDiscoveryEventHelper
   private final AtomicLong _markDownStartAtRef = new AtomicLong(Long.MAX_VALUE);
 
   private volatile Map<Integer, PartitionData> _partitionDataMap;
+  private final Integer _maxWeight; // max weight for the weight in PartitionData, if null, no max weight is set.
   private volatile Map<String, Object> _uriSpecificProperties;
 
   private ServiceDiscoveryEventEmitter _eventEmitter;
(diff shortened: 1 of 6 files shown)


d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -826,4 +822,25 @@ private ImmutablePair<String, String> getZnodePathAndData(String cluster) {
...
+      }
+
+      // check decimal places
+      if (BigDecimal.valueOf(weight).scale() > 2) {
+        _log.warn("", new IllegalArgumentException(String.format("Weight %f in Partition %d has more than 2 decimal "
+            + "places. It will be rounded to 2 decimal places in the future.", weight, entry.getKey())));
+        // TODO: round weight to 2 decimal places.
+      }
+    }
+  }
```
brycezhongqing, line 845:
```suggestion
 private void validatePartitionData(Map<Integer, PartitionData> partitionData) {
    partitionData.forEach((partitionId, data) -> {
      final double weight = data.getWeight();
      // check max weight
      if (_maxWeight != null && weight > _maxWeight) {
        _log.warn("", new IllegalArgumentException(String.format("[ACTION NEEDED] Weight %f in Partition %d is greater"
                                                                     + " than the max weight allowed: %d. Please correct the weight. It will be force-capped to the max weight "
                                                                     + "in the future.", weight, partitionId, _maxWeight)));
        // TODO: throw exception or cap weight to max weight
      }

      // check decimal places
      if (BigDecimal.valueOf(weight).scale() > 2) {
        _log.warn("", new IllegalArgumentException(String.format("Weight %f in Partition %d has more than 2 decimal "
                                                                     + "places. It will be rounded to 2 decimal places in the future.", weight, partitionId)));
        // TODO: round weight to 2 decimal places.
      }
    });
  }

```

bohhyang: Updated the code with configurable actions, and we need to modify the entry value when the bad value needs to be rectified. So we need to keep the for-loop.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -833,4 +854,59 @@ private ImmutablePair<String, String> getZnodePathAndData(String cluster) {
...
+      // check max weight
+      if (weight.compareTo(_maxWeight) > 0) {
+        switch (_actOnWeightBreach) {
+          case WARN:
+            _log.warn("", getMaxWeightBreachException(weight, entry.getKey()));
+            break;
+          case THROW:
+            throw getMaxWeightBreachException(weight, entry.getKey());
+          case RECTIFY:
+            entry.setValue(new PartitionData(_maxWeight.intValue()));
```
brycezhongqing, line 875: Interesting. This value will be overridded by following code, right ? Let's say the weight is `101.123`, the current logic result will be `101.12`

bohhyang: yeah, if the value breaches decimal places, it will be rounded. If it breaches the max, it will be capped. I updated the log files in the container PR with sample behaviors. Please take a look, thanks.

ZoabKapoor: I think @brycezhongqing might be saying that if the value both breaches decimal places **and** breaches the max, will only the rounding take effect but not the capping?

Because we do `entry.setValue` but we don't update the `weight` variable

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -833,4 +854,59 @@ private ImmutablePair<String, String> getZnodePathAndData(String cluster) {
   public boolean isWarmingUp() {
     return _isWarmingUp;
   }
+
+  private Map<Integer, PartitionData> validatePartitionData(Map<Integer, PartitionData> partitionData) {
```
ZoabKapoor, line 858: If we make this function package-private instead of fully private, we can probably add some tests for it in `TestZookeeperAnnouncer`, which I feel like might be useful, especially in validating the behavior of the `RECTIFY` action

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -833,4 +854,59 @@ private ImmutablePair<String, String> getZnodePathAndData(String cluster) {
...
+        switch (_actOnWeightBreach) {
+          case WARN: // both WARN and THROW only log the warning. Don't throw exception for decimal places.
+          case THROW:
+            _log.warn("", new IllegalArgumentException(String.format("Weight %f in Partition %d has more than %d"
+                + " decimal places. It will be rounded in the future.", weight.doubleValue(), entry.getKey(),
+                _maxWeight.scale())));
+            break;
+          case RECTIFY:
+            double newWeight = weight.setScale(_maxWeight.scale(), RoundingMode.HALF_UP).doubleValue();
+            entry.setValue(new PartitionData(newWeight));
```
ZoabKapoor, line 895: I think this will override the `entry.setValue` in the previous `switch` if the provided weight is both bigger than the max weight and has more decimal places

ZoabKapoor (changes requested):
Overall looks good, I had one doubt about the implementation logic and also wondering if we can add a test for `validatePartitionData`.

Also, are we deciding not to emit a metric initially & instead rely on logs to detect use cases with invalid partition data? Or does emitting the metric not need any restli change?

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
brycezhongqing: Yes. @ZoabKapoor 
So @bohhyang If a weight value both breaches both `max value and decimal places`, there will be a potential risk for current logic.

bohhyang: oh, that's right. Thanks for catching!

brycezhongqing (approved):

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -139,7 +166,18 @@ public class ZooKeeperAnnouncer implements D2ServiceDiscoveryEventHelper
   private final AtomicLong _warmupClusterMarkDownStartAtRef = new AtomicLong(Long.MAX_VALUE);
 
   // Field to store the dark warm-up time duration in seconds, defaults to zero
-  private int _warmupDuration;
+  private final int _warmupDuration;
+
+  public enum ActOnWeightBreach {
+    // Ignore and no op.
+    IGNORE,
```
PapaCharlie, line 173: Isn't this a little redundant to having `_maxWeight` null? Just a thought

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -195,26 +233,18 @@ public ZooKeeperAnnouncer(LoadBalancerServer server, boolean initialIsUp,
...
 
   public ZooKeeperAnnouncer(LoadBalancerServer server, boolean initialIsUp,
       boolean isDarkWarmupEnabled, String warmupClusterName, int warmupDuration, ScheduledExecutorService executorService, ServiceDiscoveryEventEmitter eventEmitter)
+  {
+    this(server, initialIsUp, isDarkWarmupEnabled, warmupClusterName, warmupDuration, executorService, eventEmitter, null, ActOnWeightBreach.IGNORE);
+  }
+
+  public ZooKeeperAnnouncer(LoadBalancerServer server, boolean initialIsUp,
+      boolean isDarkWarmupEnabled, String warmupClusterName, int warmupDuration, ScheduledExecutorService executorService,
+      ServiceDiscoveryEventEmitter eventEmitter, String maxWeight, ActOnWeightBreach actOnWeightBreach)
```
PapaCharlie, line 247: Why not take the `maxWeight` as a `BigDecimal` directly and let callers figure out what they want?

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -833,4 +894,64 @@ private ImmutablePair<String, String> getZnodePathAndData(String cluster) {
...
+          case WARN:
+            _log.warn("", getMaxWeightBreachException(weight, entry.getKey()));
+            break;
+          case THROW:
+            throw getMaxWeightBreachException(weight, entry.getKey());
+          case RECTIFY:
+            entry.setValue(new PartitionData(_maxWeight.intValue()));
+            weight = _maxWeight;
+            _log.warn("Capped weight {} in Partition {} to the max weight allowed: {}.", weight.doubleValue(),
+                entry.getKey(), _maxWeight.intValue());
```
PapaCharlie, line 919: Dumb question but what if the weight is negative? Feels like for that one we can freely throw an exception in `setWeight`, regardless of the mode

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -833,4 +894,64 @@ private ImmutablePair<String, String> getZnodePathAndData(String cluster) {
...
+        }
+      }
+    }
+    return res;
+  }
+
+  private IllegalArgumentException getMaxWeightBreachException(BigDecimal weight, int partition) {
+    return new IllegalArgumentException(String.format("[ACTION NEEDED] Weight %f in Partition %d is greater"
+        + " than the max weight allowed: %d. Please correct the weight. It will be force-capped to the max weight "
+        + "in the future.", weight.doubleValue(), partition, _maxWeight.intValue()));
```
PapaCharlie, line 955: Technically you don't know if `BigDecimal` is an integer (even though in practice it'll only ever be 100). I think you can just have `than the max weight allowed: %s` and let the `toString` do the formatting right?

PapaCharlie (changes requested):

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -139,7 +166,18 @@ public class ZooKeeperAnnouncer implements D2ServiceDiscoveryEventHelper
   private final AtomicLong _warmupClusterMarkDownStartAtRef = new AtomicLong(Long.MAX_VALUE);
 
   // Field to store the dark warm-up time duration in seconds, defaults to zero
-  private int _warmupDuration;
+  private final int _warmupDuration;
+
+  public enum ActOnWeightBreach {
```
PapaCharlie, line 171: I think this should be called `ActionOnWeightBreach`, for grammar reasons :)

PapaCharlie (changes requested):

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
bohhyang: sure, fixed.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
bohhyang: yeah, good catch. fixed.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
bohhyang: makes sense. Updated to throw.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
bohhyang: synced offline. I was thinking of using string to help callers less likely to mis-configure the decimal places on max weight (if people write new BigDecimal(100), it will have 1 decimal places, while new BigDecimal("100.00") will precisely define 2 decimal places.). But actually string won't help much either if people don't pay close attention to how to define it. Updated to BigDecimal.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
bohhyang: not really, as IGNORE will increment the metrics.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
bohhyang: makes sense. added.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -93,6 +98,23 @@ public class ZooKeeperAnnouncer implements D2ServiceDiscoveryEventHelper
...
+   * 2. The maximum number of decimal places allowed. Use 0s on decimal places to indicate it.
+   * For example, 100.00 means the max weight allowed is 100 and the max number of decimal places is 2.
+   * CAUTION: BigDecimal yields accurate scale when constructed with a string of the number, instead of a double/float.
+   * E.g: new BigDecimal("100.00") instead of new BigDecimal(100.00).
+   */
+  private final BigDecimal _maxWeight;
+  /**
+   * The action to take when d2 weight breaches validation rules.
+   */
+  private ActionOnWeightBreach _actionOnWeightBreach = ActionOnWeightBreach.IGNORE;
```
PapaCharlie, line 113: If you write the code like this in the constructor, this can be upgraded to `final` instead:
```java
    _actionOnWeightBreach = (actionOnWeightBreach != null) ? actionOnWeightBreach : IGNORE;
```
Just makes it a little cleaner

PapaCharlie (approved):

ZoabKapoor (approved):

PR #1037, opened 2025-01-09
Allow subscribing to a single D2URI

PapaCharlie:
This new endpoint on `XdsClient` allows one to watch a single URI using the glob collection interface. The watcher will be notified whenever the URI changes or is deleted, which allows for lightweight watches of a singular URI instead of needing to watch the entire cluster.

I was able to test this against a local Observer deployment, and validated that the URI updates and deletions are correctly propagated.

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
@@ -95,21 +95,20 @@ public abstract class XdsClient
     }
   }
 
-  public static abstract class D2UriResourceWatcher
+  public static abstract class D2UriResourceWatcher extends ResourceWatcher
   {
-    public abstract void onChanged(XdsD2.D2URI d2Uri);
+    public D2UriResourceWatcher()
+    {
+      super(ResourceType.D2_URI);
+    }
 
-    public abstract void onDelete();
+    public abstract void onChanged(D2URIUpdate update);
 
-    /**
-     * Called when the resource discovery RPC encounters some transient error.
-     */
-    public abstract void onError(Status error);
-
-    /**
-     * Called when the resource discovery RPC reestablishes connection.
-     */
-    public abstract void onReconnect();
+    @Override
+    final void onChanged(ResourceUpdate update)
+    {
+      onChanged((D2URIUpdate) update);
+    }
   }
 
   public static abstract class WildcardResourceWatcher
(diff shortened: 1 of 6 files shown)


bohhyang (commented):
Please if my understanding aligns with your thought: we don't want to store this single uri resource into event bus, so that it will not to be put into load balancer, nor saved to FS, as it's not needed.

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
```diff
@@ -95,6 +95,23 @@ final void onChanged(ResourceUpdate update)
     }
   }
 
+  public static abstract class D2UriResourceWatcher
```
bohhyang, line 98: we should make this extends ResourceWatcher for better code structure, so that:
1. watchD2Uri is not needed, and can just reuse watchXdsResource(globCollectionUrn(cluster, uri), watcher).
2. D2UriSubscriber is not needed, and can just use ResourceSubscriber.

bohhyang:
just a reminder for updating the version and changelog

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -858,6 +925,75 @@ void onRemoval(String resourceName)
...
+      if (d2Uri == null)
+      {
+        for (D2UriResourceWatcher watcher : _watchers)
+        {
+          watcher.onDelete();
+        }
+      }
+      else
+      {
+        metricsProvider.trackLatency(System.currentTimeMillis() - d2Uri.getModifiedTime().getSeconds() * 1000);
```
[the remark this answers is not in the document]
PapaCharlie: Lol oops

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
PapaCharlie: Sure, makes sesne

bohhyang:
lgtm. Please test along with the container PR on a test app, showing the announcement monitoring metric result.

gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=29.63.2
+version=29.63.3
```
bohhyang, line 1: this should be 29.64.0, so that apps that only bumped container but not restli will get a build error.

bohhyang (changes requested):

gradle.properties
PapaCharlie: OK makes sense

bohhyang (approved):

PR #1038, opened 2025-01-15
Add XdsDirectory to get d2 service and cluster names from INDIS

bohhyang:
## Summary
Support getting d2 service and cluster name lists from INDIS with Xds load balancer. This is needed by some apps (like d2-proxy) that need to read the names to migrate to INDIS read. 

Note that observer hasn't supported the resource D2ClusterOrServiceName yet, so build and test are failing.

## Test
UT. QEI test with d2-proxy. The names is fetched and served correctly.

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
@@ -138,7 +138,8 @@ public abstract class XdsClient
     public abstract void onRemoval(String resourceName);
 
     /**
-     * Just a signal to notify that all resources (including both changed and removed ones) have been processed.
+     * Just a signal to notify that all resources (including both changed and removed ones) in all response chunks (if
+     * any) have been processed.
      * Default implementation does nothing.
      */
     public void onAllResourcesProcessed()
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
@@ -17,6 +17,7 @@
 package com.linkedin.d2.xds;
 
 import com.google.common.base.MoreObjects;
+import com.google.common.base.Strings;
 import com.linkedin.d2.jmx.XdsClientJmx;
 import indis.XdsD2;
 import io.grpc.Status;
@@ -301,7 +302,8 @@ public abstract class XdsClient
     @Override
     public boolean isValid()
     {
-      return _nameData != null && (!_nameData.getClusterName().isEmpty() || !_nameData.getServiceName().isEmpty());
+      return _nameData != null
+          && (!Strings.isNullOrEmpty(_nameData.getClusterName()) || !Strings.isNullOrEmpty(_nameData.getServiceName()));
     }
 
     @Override
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
@@ -206,7 +206,7 @@ public abstract class XdsClient
     @Override
     final void onChanged(String resourceName, ResourceUpdate update)
     {
-      onChanged(resourceName, update);
+      onChanged(resourceName, (D2ClusterOrServiceNameUpdate) update);
     }
   }
 
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
@@ -180,6 +180,27 @@ public abstract class XdsClient
     }
   }
 
+  public static abstract class WildcardD2ClusterOrServiceNameResourceWatcher extends WildcardResourceWatcher
+  {
+    public WildcardD2ClusterOrServiceNameResourceWatcher()
+    {
+      super(ResourceType.D2_CLUSTER_OR_SERVICE_NAME);
+    }
+
+    /**
+     * Called when a D2ClusterOrServiceName resource is added or updated.
+     * @param resourceName the resource name of the D2ClusterOrServiceName that was added or updated.
+     * @param update       the new data for the D2ClusterOrServiceName resource
+     */
+    public abstract void onChanged(String resourceName, D2ClusterOrServiceNameUpdate update);
+
+    @Override
+    final void onChanged(String resourceName, ResourceUpdate update)
+    {
+      onChanged(resourceName, update);
+    }
+  }
+
   public interface ResourceUpdate
   {
     boolean isValid();
@@ -230,6 +251,54 @@ public abstract class XdsClient
     public String toString()
     {
       return MoreObjects.toStringHelper(this).add("_nodeData", _nodeData).toString();
+    }
+  }
+
+  public static final class D2ClusterOrServiceNameUpdate implements ResourceUpdate
+  {
+    XdsD2.D2ClusterOrServiceName _nameData;
+
+    D2ClusterOrServiceNameUpdate(XdsD2.D2ClusterOrServiceName nameData)
+    {
+      _nameData = nameData;
+    }
+
+    public XdsD2.D2ClusterOrServiceName getNameData()
+    {
+      return _nameData;
+    }
+
+    @Override
+    public boolean equals(Object object)
+    {
+      if (this == object)
+      {
+        return true;
+      }
+      if (object == null || getClass() != object.getClass())
+      {
+        return false;
+      }
+      D2ClusterOrServiceNameUpdate that = (D2ClusterOrServiceNameUpdate) object;
+      return Objects.equals(_nameData, that._nameData);
+    }
+
+    @Override
+    public int hashCode()
+    {
+      return Objects.hash(_nameData);
+    }
+
+    @Override
+    public boolean isValid()
+    {
+      return _nameData != null && (!_nameData.getClusterName().isEmpty() || !_nameData.getServiceName().isEmpty());
+    }
+
+    @Override
+    public String toString()
+    {
+      return MoreObjects.toStringHelper(this).add("_nameData", _nameData).toString();
     }
   }
 
@@ -302,12 +371,16 @@ public abstract class XdsClient
 
   public static final NodeUpdate EMPTY_NODE_UPDATE = new NodeUpdate(null);
   public static final D2URIMapUpdate EMPTY_D2_URI_MAP_UPDATE = new D2URIMapUpdate(null);
+  public static final D2ClusterOrServiceNameUpdate EMPTY_D2_CLUSTER_OR_SERVICE_NAME_UPDATE =
+      new D2ClusterOrServiceNameUpdate(null);
 
   enum ResourceType
   {
     NODE("type.googleapis.com/indis.Node", EMPTY_NODE_UPDATE),
     D2_URI_MAP("type.googleapis.com/indis.D2URIMap", EMPTY_D2_URI_MAP_UPDATE),
-    D2_URI("type.googleapis.com/indis.D2URI", EMPTY_D2_URI_MAP_UPDATE);
+    D2_URI("type.googleapis.com/indis.D2URI", EMPTY_D2_URI_MAP_UPDATE),
+    D2_CLUSTER_OR_SERVICE_NAME("type.googleapis.com/indis.D2ClusterOrServiceName",
+        EMPTY_D2_CLUSTER_OR_SERVICE_NAME_UPDATE);
 
     private static final Map<String, ResourceType> TYPE_URL_TO_ENUM = Arrays.stream(values())
         .filter(e -> e.typeUrl() != null)
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
@@ -135,6 +135,15 @@ public abstract class XdsClient
      * @param resourceName the name of the resource that was removed.
      */
     public abstract void onRemoval(String resourceName);
+
+    /**
+     * Just a signal to notify that all resources (including both changed and removed ones) have been processed.
+     * Default implementation does nothing.
+     */
+    public void onAllResourcesProcessed()
+    {
+      // do nothing
+    }
   }
 
   public static abstract class WildcardNodeResourceWatcher extends WildcardResourceWatcher
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
@@ -42,7 +42,6 @@ import io.grpc.stub.ClientResponseObserver;
 import io.grpc.stub.StreamObserver;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
@@ -72,7 +72,7 @@ public class XdsClientImpl extends XdsClient
 {
   private static final Logger _log = LoggerFactory.getLogger(XdsClientImpl.class);
   private static final RateLimitedLogger RATE_LIMITED_LOGGER =
-      new RateLimitedLogger(_log, TimeUnit.MINUTES.toMillis(10), SystemClock.instance());
+      new RateLimitedLogger(_log, TimeUnit.MINUTES.toMillis(10000), SystemClock.instance());
   public static final long DEFAULT_READY_TIMEOUT_MILLIS = 2000L;
 
   /**
@@ -625,7 +625,7 @@ public class XdsClientImpl extends XdsClient
     }
     catch (Exception e)
     {
-      _log.warn("Failed to decode nonce: {}", response.getNonce(), e);
+      RATE_LIMITED_LOGGER.warn("Failed to decode nonce: {}", response.getNonce(), e);
       remainingChunks = -1;
     }
 
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
@@ -141,25 +141,26 @@ public class XdsClientImpl extends XdsClient
   {
     _executorService.execute(() ->
     {
-      Map<String, ResourceSubscriber> resourceSubscriberMap = getResourceSubscriberMap(watcher.getType());
+      ResourceType originalType = watcher.getType();
+      Map<String, ResourceSubscriber> resourceSubscriberMap = getResourceSubscriberMap(originalType);
       ResourceSubscriber subscriber = resourceSubscriberMap.get(resourceName);
       if (subscriber == null)
       {
-        subscriber = new ResourceSubscriber(watcher.getType(), resourceName, _xdsClientJmx);
+        subscriber = new ResourceSubscriber(originalType, resourceName, _xdsClientJmx);
         resourceSubscriberMap.put(resourceName, subscriber);
-        ResourceType type;
+        ResourceType adjustedType;
         String adjustedResourceName;
-        if (watcher.getType() == ResourceType.D2_URI_MAP && _subscribeToUriGlobCollection)
+        if (originalType == ResourceType.D2_URI_MAP && _subscribeToUriGlobCollection)
         {
-          type = ResourceType.D2_URI;
+          adjustedType = ResourceType.D2_URI;
           adjustedResourceName = GlobCollectionUtils.globCollectionUrlForClusterResource(resourceName);
         }
         else
         {
-          type = watcher.getType();
+          adjustedType = originalType;
           adjustedResourceName = resourceName;
         }
-        _log.info("Subscribing to {} resource: {}", type, adjustedResourceName);
+        _log.info("Subscribing to {} resource: {}", adjustedType, adjustedResourceName);
 
         if (_adsStream == null && !isInBackoff())
         {
@@ -167,7 +168,7 @@ public class XdsClientImpl extends XdsClient
         }
         if (_adsStream != null)
         {
-          _adsStream.sendDiscoveryRequest(type, Collections.singletonList(adjustedResourceName));
+          _adsStream.sendDiscoveryRequest(adjustedType, Collections.singletonList(adjustedResourceName));
         }
       }
       subscriber.addWatcher(watcher);
@@ -179,16 +180,17 @@ public class XdsClientImpl extends XdsClient
   {
     _executorService.execute(() ->
     {
-      WildcardResourceSubscriber subscriber = getWildcardResourceSubscriber(watcher.getType());
+      ResourceType originalType = watcher.getType();
+      WildcardResourceSubscriber subscriber = getWildcardResourceSubscriber(originalType);
       if (subscriber == null)
       {
-        subscriber = new WildcardResourceSubscriber(watcher.getType());
-        _wildcardSubscribers.put(watcher.getType(), subscriber);
+        subscriber = new WildcardResourceSubscriber(originalType);
+        getWildcardResourceSubscribers().put(originalType, subscriber);
 
         ResourceType adjustedType =
-            (watcher.getType() == ResourceType.D2_URI_MAP && _subscribeToUriGlobCollection)
+            (originalType == ResourceType.D2_URI_MAP && _subscribeToUriGlobCollection)
                 ? ResourceType.D2_URI
-                : watcher.getType();
+                : originalType;
 
         _log.info("Subscribing to wildcard for resource type: {}", adjustedType);
 
@@ -230,7 +232,8 @@ public class XdsClientImpl extends XdsClient
   }
 
   // Start RPC stream. Must be called from the executor, and only if we're not backed off.
-  private void startRpcStreamLocal() {
+  @VisibleForTesting
+  void startRpcStreamLocal() {
     if (_shutdown) {
       _log.warn("RPC stream cannot be started after shutdown!");
       return;
@@ -330,6 +333,9 @@ public class XdsClientImpl extends XdsClient
       case NODE:
         handleD2NodeResponse(response);
         break;
+      case D2_CLUSTER_OR_SERVICE_NAME:
+        handleD2ClusterOrServiceNameResponse(response);
+        break;
       case D2_URI_MAP:
         handleD2URIMapResponse(response);
         break;
@@ -365,6 +371,36 @@ public class XdsClientImpl extends XdsClient
         // Assume that the resource doesn't exist if it cannot be deserialized instead of simply ignoring it. This way
         // any call waiting on the response can be satisfied instead of timing out.
         updates.put(resourceName, EMPTY_NODE_UPDATE);
+      }
+    }
+    sendAckOrNack(data.getResourceType(), data.getNonce(), errors);
+    handleResourceUpdate(updates, data.getResourceType());
+    handleResourceRemoval(data.getRemovedResources(), data.getResourceType());
+  }
+
+  private void handleD2ClusterOrServiceNameResponse(DiscoveryResponseData data)
+  {
+    Map<String, D2ClusterOrServiceNameUpdate> updates = new HashMap<>();
+    List<String> errors = new ArrayList<>();
+    String errMsg = "Failed to unpack D2ClusterOrServiceName response";
+
+    for (Resource resource : data.getResourcesList())
+    {
+      String resourceName = resource.getName();
+      try
+      {
+        XdsD2.D2ClusterOrServiceName clusterOrServiceName = resource.getResource()
+            .unpack(XdsD2.D2ClusterOrServiceName.class);
+        updates.put(resourceName, new D2ClusterOrServiceNameUpdate(clusterOrServiceName));
+      }
+      catch (InvalidProtocolBufferException e)
+      {
+
+        _log.warn(errMsg, e);
+        errors.add(errMsg);
+        // Assume that the resource doesn't exist if it cannot be deserialized instead of simply ignoring it. This way
+        // any call waiting on the response can be satisfied instead of timing out.
+        updates.put(resourceName, EMPTY_D2_CLUSTER_OR_SERVICE_NAME_UPDATE);
       }
     }
     sendAckOrNack(data.getResourceType(), data.getNonce(), errors);
@@ -554,14 +590,14 @@ public class XdsClientImpl extends XdsClient
 
   private void notifyStreamError(Status error)
   {
-    for (Map<String, ResourceSubscriber> subscriberMap : _resourceSubscribers.values())
+    for (Map<String, ResourceSubscriber> subscriberMap : getResourceSubscribers().values())
     {
       for (ResourceSubscriber subscriber : subscriberMap.values())
       {
         subscriber.onError(error);
       }
     }
-    for (WildcardResourceSubscriber wildcardResourceSubscriber : _wildcardSubscribers.values())
+    for (WildcardResourceSubscriber wildcardResourceSubscriber : getWildcardResourceSubscribers().values())
     {
       wildcardResourceSubscriber.onError(error);
     }
@@ -570,14 +606,14 @@ public class XdsClientImpl extends XdsClient
 
   private void notifyStreamReconnect()
   {
-    for (Map<String, ResourceSubscriber> subscriberMap : _resourceSubscribers.values())
+    for (Map<String, ResourceSubscriber> subscriberMap : getResourceSubscribers().values())
     {
       for (ResourceSubscriber subscriber : subscriberMap.values())
       {
         subscriber.onReconnect();
       }
     }
-    for (WildcardResourceSubscriber wildcardResourceSubscriber : _wildcardSubscribers.values())
+    for (WildcardResourceSubscriber wildcardResourceSubscriber : getWildcardResourceSubscribers().values())
     {
       wildcardResourceSubscriber.onReconnect();
     }
@@ -591,9 +627,21 @@ public class XdsClientImpl extends XdsClient
   }
 
   @VisibleForTesting
+  Map<ResourceType, Map<String, ResourceSubscriber>> getResourceSubscribers()
+  {
+    return _resourceSubscribers;
+  }
+
+  @VisibleForTesting
   WildcardResourceSubscriber getWildcardResourceSubscriber(ResourceType type)
   {
     return _wildcardSubscribers.get(type);
+  }
+
+  @VisibleForTesting
+  Map<ResourceType, WildcardResourceSubscriber> getWildcardResourceSubscribers()
+  {
+    return _wildcardSubscribers;
   }
 
   static class ResourceSubscriber
@@ -635,7 +683,8 @@ public class XdsClientImpl extends XdsClient
       }
     }
 
-    private void onData(ResourceUpdate data, XdsServerMetricsProvider metricsProvider)
+    @VisibleForTesting
+    void onData(ResourceUpdate data, XdsServerMetricsProvider metricsProvider)
     {
       if (Objects.equals(_data, data))
       {
@@ -659,13 +708,18 @@ public class XdsClientImpl extends XdsClient
           _log.warn("Received invalid data for {} {}, data: {}", _type, _resource, data);
         }
         _xdsClientJmx.incrementResourceInvalidCount();
+
+        if (_data == null)
+        {
+          _log.info("Initializing {} {} to empty data.", _type, _resource);
+          _data = _type.emptyData();
+        }
+        else {
+          // no update to the existing data, don't need to notify the watcher
+          return;
+        }
       }
 
-      if (_data == null)
-      {
-        _log.info("Initializing {} {} to empty data.", _type, _resource);
-        _data = _type.emptyData();
-      }
       for (ResourceWatcher watcher : _watchers)
       {
         watcher.onChanged(_data);
@@ -797,7 +851,8 @@ public class XdsClientImpl extends XdsClient
       }
     }
 
-    private void onData(String resourceName, ResourceUpdate update)
+    @VisibleForTesting
+    void onData(String resourceName, ResourceUpdate update)
     {
       if (Objects.equals(_data.get(resourceName), update))
       {
@@ -811,6 +866,7 @@ public class XdsClientImpl extends XdsClient
       }
       else
       {
+        // invalid data is received, log a warning and check if existing data is present.
         if (_type == ResourceType.D2_URI_MAP || _type == ResourceType.D2_URI)
         {
           RATE_LIMITED_LOGGER.warn("Received invalid data for {} {}, data: {}", _type, resourceName, update);
@@ -819,13 +875,18 @@ public class XdsClientImpl extends XdsClient
         {
           _log.warn("Received invalid data for {} {}, data: {}", _type, resourceName, update);
         }
+        // if no data has ever been set, init it to an empty data in case watchers are waiting for it
+        if (_data.get(resourceName) == null)
+        {
+          _log.info("Initializing {} {} to empty data.", _type, resourceName);
+          _data.put(resourceName, _type.emptyData());
+        }
+        else {
+          // no update to the existing data, don't need to notify the watcher
+          return;
+        }
       }
 
-      if (_data.get(resourceName) == null)
-      {
-        _log.info("Initializing {} {} to empty data.", _type, resourceName);
-        _data.put(resourceName, _type.emptyData());
-      }
       for (WildcardResourceWatcher watcher : _watchers)
       {
         watcher.onChanged(resourceName, update);
@@ -870,28 +931,33 @@ public class XdsClientImpl extends XdsClient
     public void run()
     {
       startRpcStreamLocal();
-      for (ResourceType type : _resourceSubscribers.keySet())
+      for (ResourceType originalType : getResourceSubscribers().keySet())
       {
-        Set<String> resources = new HashSet<>(getResourceSubscriberMap(type).keySet());
+        Set<String> resources = new HashSet<>(getResourceSubscriberMap(originalType).keySet());
         if (resources.isEmpty())
         {
           continue;
         }
 
-        ResourceType rewrittenType = type;
-        if (_subscribeToUriGlobCollection && type == ResourceType.D2_URI_MAP)
+        ResourceType adjustedType;
+        if (_subscribeToUriGlobCollection && originalType == ResourceType.D2_URI_MAP)
         {
           resources = resources.stream()
               .map(GlobCollectionUtils::globCollectionUrlForClusterResource)
               .collect(Collectors.toCollection(HashSet::new));
-          rewrittenType = ResourceType.D2_URI;
+          adjustedType = ResourceType.D2_URI;
         }
-        _adsStream.sendDiscoveryRequest(rewrittenType, resources);
+        else {
+          adjustedType = originalType;
+        }
+        _adsStream.sendDiscoveryRequest(adjustedType, resources);
       }
 
-      for (ResourceType type: _wildcardSubscribers.keySet())
+      for (ResourceType originalType: getWildcardResourceSubscribers().keySet())
       {
-        _adsStream.sendDiscoveryRequest(type, Collections.singletonList("*"));
+        ResourceType adjustedType = _subscribeToUriGlobCollection && originalType == ResourceType.D2_URI_MAP
+            ? ResourceType.D2_URI : originalType;
+        _adsStream.sendDiscoveryRequest(adjustedType, Collections.singletonList("*"));
       }
     }
   }
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
@@ -72,7 +72,7 @@ public class XdsClientImpl extends XdsClient
 {
   private static final Logger _log = LoggerFactory.getLogger(XdsClientImpl.class);
   private static final RateLimitedLogger RATE_LIMITED_LOGGER =
-      new RateLimitedLogger(_log, TimeUnit.SECONDS.toMillis(10), SystemClock.instance());
+      new RateLimitedLogger(_log, TimeUnit.MINUTES.toMillis(1), SystemClock.instance());
   public static final long DEFAULT_READY_TIMEOUT_MILLIS = 2000L;
 
   /**
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
@@ -383,7 +383,7 @@ public class XdsClientImpl extends XdsClient
   {
     Map<String, D2ClusterOrServiceNameUpdate> updates = new HashMap<>();
     List<String> errors = new ArrayList<>();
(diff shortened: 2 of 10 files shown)


d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -867,29 +873,26 @@ public void run()
...
-        }
-        // If there is a wildcard subscriber, we should always send a wildcard request to the server.
-        if (getWildcardResourceSubscriber(type) != null)
-        {
-          resources.add("*");
-        }
         _adsStream.sendDiscoveryRequest(rewrittenType, resources);
       }
+
+      for (ResourceType type: _wildcardSubscribers.keySet())
```
PapaCharlie, line 892: You have to use `getWildcardResourceSubscriber` here because it's how the tests are written. It overrides that function to make it return a specific map

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -867,29 +873,26 @@ public void run()
...
         {
           resources = resources.stream()
               .map(GlobCollectionUtils::globCollectionUrlForClusterResource)
               .collect(Collectors.toCollection(HashSet::new));
           rewrittenType = ResourceType.D2_URI;
         }
-        else
-        {
-          rewrittenType = type;
-        }
```
PapaCharlie, line 885: I actually prefer using the uninitialized variable declaration like `ResourceType rewrittenType;` because that way the compiler will complain if it's unset, so you are forced to look at both branches. For example, if a branch is added after, and it doesn't set `rewrittenType` the code will not compile. It's just a nice feature that makes it harder to mess up. If you don't mind keeping the original, I'd appreciate it

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -797,33 +797,39 @@ void addWatcher(WildcardResourceWatcher watcher)
...
-          _log.warn("Received invalid data for {} {}, data: {}", _type, resourceName, data);
+          _log.warn("Received invalid data for {} {}, data: {}", _type, resourceName, update);
         }
       }
+
+      if (_data.get(resourceName) == null)
+      {
+        _log.info("Initializing {} {} to empty data.", _type, resourceName);
+        _data.put(resourceName, _type.emptyData());
+      }
```
PapaCharlie, line 828: Why are you initializing it at all? Shouldn't it just be ignored if it's `null`, instead of notifying the watcher?

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -867,29 +873,26 @@ public void run()
...
-        if (getWildcardResourceSubscriber(type) != null)
-        {
-          resources.add("*");
-        }
         _adsStream.sendDiscoveryRequest(rewrittenType, resources);
       }
+
+      for (ResourceType type: _wildcardSubscribers.keySet())
+      {
+        _adsStream.sendDiscoveryRequest(type, Collections.singletonList("*"));
```
PapaCharlie, line 894: The key in the map is always `D2_URI_MAP` here even if glob collections are enabled, so this will actually break because it will resubscribe to the map instead of the glob collection. You have to check `_subscribeToUriGlobCollection && type == ResourceType.D2_URI_MAP` and do the translation accordingly.

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
```diff
@@ -0,0 +1,121 @@
...
+    if (names.isEmpty())
+    {
+      try {
+        wait(DEFAULT_WAIT_TIME);
+      } catch (InterruptedException e) {
+        // do nothing
+      }
+    }
+
+    callback.onSuccess(new ArrayList<>(names));
```
PapaCharlie, line 119: This feels clunky because if the data isn't all there you'd be returning a partial result, and if it's all there before 10s, you've wasted time. There's got to be a good middle ground here. The problem is that `onChanged` is called for each update, not for each xDS response, so you can't know when it's done iterating. Maybe we can add an extra API for the `WildcardResourceWatcher` which is `onResponseProcessed` or something? That would get called when `onChanged` has been called on all the elements of the response. That way, you can cancel this `wait` by calling `notify` (though I think it would be much more explicit if you had a `CountDownLatch` instead)

PapaCharlie (changes requested):

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: I noticed this today too. Will update it. thanks.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: good catch. Will do!

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: updated

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: sure, reverted.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: fixed.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
PapaCharlie: This actually plays into this comment: https://github.com/linkedin/rest.li/pull/1038#discussion_r1919413458
If we have a new API for the wildcard subscriber then we don't have to do this. I don't think it really makes sense to add this in the map, otherwise new subscribers, which call `addWatcher`, will be notified of a resource that doesn't exist:
```java
    void addWatcher(WildcardResourceWatcher watcher)
    {
      _watchers.add(watcher);
      for (Map.Entry<String, ResourceUpdate> entry : _data.entrySet())
      {
        watcher.onChanged(entry.getKey(), entry.getValue());
        _log.debug("Notifying watcher of current data for resource {} of type {}: {}",
            entry.getKey(), _type, entry.getValue());
      }
    }
```

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
bohhyang: that's true. One thing though is subscribing to the service/cluster names happen in the bootstrap when no real traffic come in yet, so the data should be ready when it's needed, so I hesitated a bit to make it robust. But you're right, the WildcardResourceWatcher API should be robust with onAllChangesProcessed, etc. (CountDownLatch wouldn't work since the caller wouldn't know beforehand how many resources will xds client fetches. I'll use a CyclicBarrier instead).

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: "new subscribers, which call addWatcher, will be notified of a resource that doesn't exist" ---- this is exactly what this is supposed to do, to fail the request fast when the resource has bad data responded rather than keeps them waiting.

bohhyang: And even having a new wildcard subscriber API which says "done with all updates" won't help, since the problematic resource has invalid data, if we don't put something in the map, even bus won't be updated and traffic request will hang until timing out. 
Image subscribing to d2 services with only wildcard, a call to d2://FooService won't get anything in the load balancer state if FooService has bad data returned by observer. It's not happening today because FooService is subscribed via individual subscriber, which add the empty holder when received bad data. A second call to d2://FooService will also fail immediately since event bus has a entry for it already (empty holder though).

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -659,13 +708,18 @@ private void onData(ResourceUpdate data, XdsServerMetricsProvider metricsProvide
...
-      {
-        _log.info("Initializing {} {} to empty data.", _type, _resource);
-        _data = _type.emptyData();
+        if (_data == null)
+        {
+          _log.info("Initializing {} {} to empty data.", _type, _resource);
+          _data = _type.emptyData();
+        }
+        else {
+          // no update to the existing data, don't need to notify the watcher
```
brycezhongqing, line 718: So this case is triggered when `data is null or data is invalid`. In this case, we think there is no update for the existing data (`_data`), right ?

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
```diff
@@ -0,0 +1,122 @@
...
+
+  private void addNameWatcher()
+  {
+    if (_watcher != null)
+    {
+      return;
+    }
+    _watcher = createNameWatcher();
+    _xdsClient.watchAllXdsResources(_watcher);
+  }
```
brycezhongqing, line 60: Should we consider about `race condition` for this segment ? Like the user client call `getServiceNames` and `getClusterNames` at the same time ?

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
```diff
@@ -0,0 +1,122 @@
...
+  private XdsClient.WildcardD2ClusterOrServiceNameResourceWatcher createNameWatcher()
+  {
+    return new XdsClient.WildcardD2ClusterOrServiceNameResourceWatcher()
+    {
+
+      @Override
+      public void onChanged(String resourceName, XdsClient.D2ClusterOrServiceNameUpdate update)
+      {
+        if (update == EMPTY_D2_CLUSTER_OR_SERVICE_NAME_UPDATE)
+        { // invalid data, ignore
```
brycezhongqing, line 71: Log there for debugging ?

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
```diff
@@ -0,0 +1,122 @@
...
+
+      @Override
+      public void onReconnect()
+      {
+        // do nothing
+      }
+    };
+  }
+
+  private void waitAndRespond(boolean isForService, Callback<List<String>> callback) {
```
brycezhongqing, line 108: How about make it more general ?
```suggestion
  private void waitAndRespond(Collection<String> names, Callback<List<String>> callback) {
```

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: only when current data (_data) is not null (checked at line 721), yes.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -587,13 +644,25 @@ private void notifyStreamReconnect()
   @VisibleForTesting
   Map<String, ResourceSubscriber> getResourceSubscriberMap(ResourceType type)
   {
-    return _resourceSubscribers.get(type);
+    return getResourceSubscribers().get(type);
+  }
```
PapaCharlie, line 648: Does this still need to be `@VisibleForTesting` anymore?

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -587,13 +644,25 @@ private void notifyStreamReconnect()
...
+  {
+    return _resourceSubscribers;
   }
 
   @VisibleForTesting
   WildcardResourceSubscriber getWildcardResourceSubscriber(ResourceType type)
   {
-    return _wildcardSubscribers.get(type);
+    return getWildcardResourceSubscribers().get(type);
+  }
```
PapaCharlie, line 660: Same

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
```diff
@@ -0,0 +1,150 @@
...
+  private XdsClient.WildcardD2ClusterOrServiceNameResourceWatcher createNameWatcher()
+  {
+    return new XdsClient.WildcardD2ClusterOrServiceNameResourceWatcher()
+    {
+
+      @Override
+      public void onChanged(String resourceName, XdsClient.D2ClusterOrServiceNameUpdate update)
+      {
+        _isUpdating.compareAndSet(false, true);
+        if (update == EMPTY_D2_CLUSTER_OR_SERVICE_NAME_UPDATE)
```
PapaCharlie, line 82: Just for consistency, you can't guarantee the XdsClientImpl will use `EMPTY_D2_CLUSTER_OR_SERVICE_NAME_UPDATE`
```suggestion
        if (EMPTY_D2_CLUSTER_OR_SERVICE_NAME_UPDATE.equals(update))
```

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
bohhyang: good point. fixed, thanks.

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
```diff
@@ -0,0 +1,150 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+
+import static com.linkedin.d2.xds.XdsClient.*;
+
+
+public class XdsDirectory implements Directory
+{
+  private final XdsClient _xdsClient;
+  private final ConcurrentMap<String, String> _serviceNames = new ConcurrentHashMap<>();
+  private final ConcurrentMap<String, String> _clusterNames = new ConcurrentHashMap<>();
```
PapaCharlie, line 23: Since this is just being used as a set, you can do this:
```suggestion
  private final Set<String> _serviceNames = ConcurrentHashMap.newKeySet();
  private final Set<String> _clusterNames = ConcurrentHashMap.newKeySet();
```

PapaCharlie: Actually, if you go with the `ReadWriteLock` approach, these don't need even need to be concurrent

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
```diff
@@ -0,0 +1,150 @@
...
+   * A flag that shows whether the service/cluster names data is being updated. Requests to the data should wait until
+   * the update is done.
+   */
+  private final AtomicBoolean _isUpdating = new AtomicBoolean(true);
+  /**
+   * This lock will be released when the service and cluster names data have been updated and is ready to serve.
+   * If the data is being updated, requests to access the data will wait indefinitely. Callers could set a shorter
+   * timeout on the callback passed in to getServiceNames or getClusterNames, as needed.
+   */
+  private final Object _dataReadyLock = new Object();
```
PapaCharlie, line 35: Using the `synchronized` block is making this more complicated than it needs to be I think. You could use a `ReadWriteLock` and do this in `onChanged` and `onRemoval`:
```java
if (_isUpdating.compareAndSet(false, true)) {
  _dataReadyLock.writeLock().lock();
}
```
And then the reverse in `onAllResourcesProcessed`:
```java
_isUpdating.set(false);
_dataReadyLock.writeLock().unlock();
```

Then in `waitAndRespond`:
```java
    List<String> names;
    try {
      _dataReadyLock.readLock().lock()
      names = new ArrayList<>(isForService ? _serviceNames.values() : _clusterNames.values());
    } finally {
      _dataReadyLock.readLock().unlock();
    }
    callback.onSuccess(names);
```

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
bohhyang: it's better to ease the caller with simple boolean param and do a bit more work in the helper function, compared to the caller otherwise having to call with serviceNames.values(), clusterNames.values, etc.  So making the caller easier would be less error prone then making the helper function code simpler.

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
PapaCharlie: Nevermind, they still do because `onChanged` doesn't always grab the lock

bohhyang: "Since this is just being used as a set" --- they're definitely used as a map when being updated. We need to keep track of which name is mapped for which d2 service/cluster, so that when the name changes for that d2 service/cluster, the old value can also be removed while new value is added (when calling `_serviceNames.put(resourceName, nameData.getServiceName());`). If we use a set, the old value will be left in the set.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -372,6 +378,36 @@ private void handleD2NodeResponse(DiscoveryResponseData data)
...
+      try
+      {
+        XdsD2.D2ClusterOrServiceName clusterOrServiceName = resource.getResource()
+            .unpack(XdsD2.D2ClusterOrServiceName.class);
+        updates.put(resourceName, new D2ClusterOrServiceNameUpdate(clusterOrServiceName));
+      }
+      catch (InvalidProtocolBufferException e)
+      {
+
+        _log.warn(errMsg, e);
```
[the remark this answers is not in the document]
bohhyang: makes sense. added.

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
bohhyang: it's logged in xds client impl already, added a comment for it.

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
```diff
@@ -180,6 +189,27 @@ final void onChanged(String resourceName, ResourceUpdate update)
...
+     * Called when a D2ClusterOrServiceName resource is added or updated.
+     * @param resourceName the resource name of the D2ClusterOrServiceName that was added or updated.
+     * @param update       the new data for the D2ClusterOrServiceName resource
+     */
+    public abstract void onChanged(String resourceName, D2ClusterOrServiceNameUpdate update);
+
+    @Override
+    final void onChanged(String resourceName, ResourceUpdate update)
+    {
+      onChanged(resourceName, update);
```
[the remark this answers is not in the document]
bohhyang: you're absolutely right.

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
bohhyang: yeah, updated.

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
bohhyang: smart. Updated.

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
```diff
@@ -135,6 +135,15 @@ final ResourceType getType()
...
     public abstract void onRemoval(String resourceName);
+
+    /**
+     * Just a signal to notify that all resources (including both changed and removed ones) have been processed.
+     * Default implementation does nothing.
+     */
+    public void onAllResourcesProcessed()
+    {
+      // do nothing
+    }
```
brycezhongqing, line 146: Why we don't use abstract method there ?
```java
 public abstract void onAllResourcesProcessed();
```

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -368,8 +376,36 @@ private void handleD2NodeResponse(DiscoveryResponseData data)
...
+    for (Resource resource : data.getResourcesList())
+    {
+      String resourceName = resource.getName();
+      try
+      {
+        XdsD2.D2ClusterOrServiceName clusterOrServiceName = resource.getResource()
+            .unpack(XdsD2.D2ClusterOrServiceName.class);
+        updates.put(resourceName, new D2ClusterOrServiceNameUpdate(clusterOrServiceName));
+      }
+      catch (InvalidProtocolBufferException e)
```
brycezhongqing, line 397: I'm wondering is there any other type of Exception ? Should we catch up all other exception either ?

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -141,33 +143,34 @@ public void watchXdsResource(String resourceName, ResourceWatcher watcher)
   {
     _executorService.execute(() ->
     {
-      Map<String, ResourceSubscriber> resourceSubscriberMap = getResourceSubscriberMap(watcher.getType());
+      ResourceType originalType = watcher.getType();
+      Map<String, ResourceSubscriber> resourceSubscriberMap = getResourceSubscriberMap(originalType);
       ResourceSubscriber subscriber = resourceSubscriberMap.get(resourceName);
```
PapaCharlie, line 148: This will throw an NPE, since this is how `_resourceSubscribers` is defined:
```java
  /**
   * The resource subscriber map stores the subscribers to specific resources of a given type. Note that it only has 2
   * keys: {@link ResourceType#D2_URI_MAP} and {@link ResourceType#NODE}. The {@link ResourceType#D2_URI} is absent from
   * this map because it should not be used as a key, as glob collection updates are translated to appear as normal map
   * updates to subscribers.
   */
  private final Map<ResourceType, Map<String, ResourceSubscriber>> _resourceSubscribers = Maps.immutableEnumMap(
      Stream.of(ResourceType.NODE, ResourceType.D2_URI_MAP)
          .collect(Collectors.toMap(Function.identity(), e -> new HashMap<>())));
```
I think you need to update it to just this, and remove that documentation altogether:
```java
  private final Map<ResourceType, Map<String, ResourceSubscriber>> _resourceSubscribers = Maps.immutableEnumMap(
      Stream.of(ResourceType.values())
          .collect(Collectors.toMap(Function.identity(), e -> new HashMap<>())));
```

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -510,17 +543,35 @@ void sendAckOrNack(ResourceType type, String nonce, List<String> errors)
...
   private void handleResourceUpdate(Map<String, ? extends ResourceUpdate> updates, ResourceType type)
   {
     Map<String, ResourceSubscriber> subscribers = getResourceSubscriberMap(type);
+    WildcardResourceSubscriber wildcardSubscriber = getWildcardResourceSubscriber(type);
+
     for (Map.Entry<String, ? extends ResourceUpdate> entry : updates.entrySet())
     {
-      ResourceSubscriber subscriber = subscribers.get(entry.getKey());
-      if (subscriber != null)
+      if (subscribers != null)
```
PapaCharlie, line 566: If you update the definition of `_resourceSubscribers` this is now impossible and isn't needed

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -534,34 +585,40 @@ private void handleResourceRemoval(Collection<String> removedResources, Resource
...
+    Map<String, ResourceSubscriber> subscribers = getResourceSubscriberMap(type);
+    WildcardResourceSubscriber wildcardSubscriber = getWildcardResourceSubscriber(type);
     for (String resourceName : removedResources)
     {
       _xdsClientJmx.incrementResourceNotFoundCount();
       _log.warn("Received response that {} {} was removed", type, resourceName);
-      ResourceSubscriber subscriber = getResourceSubscriberMap(type).get(resourceName);
-      if (subscriber != null)
+
+      if (subscribers != null)
```
PapaCharlie, line 596: Same as above, this isn't necessary anymnore

PapaCharlie (changes requested):

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -659,13 +727,18 @@ private void onData(ResourceUpdate data, XdsServerMetricsProvider metricsProvide
...
-        _data = _type.emptyData();
+        if (_data == null)
+        {
+          _log.info("Initializing {} {} to empty data.", _type, _resource);
+          _data = _type.emptyData();
+        }
+        else {
+          // no update to the existing data, don't need to notify the watcher
+          return;
+        }
```
PapaCharlie, line 739:
```suggestion
        else
        {
          // no update to the existing data, don't need to notify the watcher
          return;
        }
```

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -823,6 +894,21 @@ private void onData(String resourceName, ResourceUpdate data)
...
+        // if no data has ever been set, init it to an empty data in case watchers are waiting for it
+        if (_data.get(resourceName) == null)
+        {
+          _log.info("Initializing {} {} to empty data.", _type, resourceName);
+          _data.put(resourceName, _type.emptyData());
+        }
+        else {
+          // no update to the existing data, don't need to notify the watcher
+          return;
+        }
```
PapaCharlie, line 906:
```suggestion
        else
        {
          // no update to the existing data, don't need to notify the watcher
          return;
        }
```

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -864,31 +974,33 @@ final class RpcRetryTask implements Runnable
...
+        Set<String> resources = new HashSet<>(getResourceSubscriberMap(originalType).keySet());
+        if (resources.isEmpty())
         {
           continue;
         }
-        ResourceType rewrittenType;
-        if (_subscribeToUriGlobCollection && type == ResourceType.D2_URI_MAP)
+
+        ResourceType adjustedType;
+        if (_subscribeToUriGlobCollection && originalType == ResourceType.D2_URI_MAP)
```
PapaCharlie, line 986: The `_subscribeToUriGlobCollection && originalType == ResourceType.D2_URI_MAP` is checked in many places of the code. I think it'd be a little easier to read if it was separated into a standalone function with some docs, like:
```java
private boolean isAdjustedUriGlobCollectionType(ResourceType type) {
  return _subscribeToUriGlobCollection && type == ResourceType.D2_URI_MAP;
}
```

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -864,31 +974,33 @@ final class RpcRetryTask implements Runnable
...
-          resources.add("*");
-        }
-        _adsStream.sendDiscoveryRequest(rewrittenType, resources);
+        _adsStream.sendDiscoveryRequest(adjustedType, resources);
+      }
+
+      for (ResourceType originalType: getWildcardResourceSubscribers().keySet())
+      {
+        ResourceType adjustedType = _subscribeToUriGlobCollection && originalType == ResourceType.D2_URI_MAP
+            ? ResourceType.D2_URI : originalType;
```
PapaCharlie, line 1002: So for example, this would now look like this:
```suggestion
        ResourceType adjustedType = isAdjustedUriGlobCollectionType(originalType)
            ? ResourceType.D2_URI : originalType;
```

PapaCharlie (changes requested):

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
brycezhongqing: I see. Thanks for explanation.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -368,8 +376,36 @@ private void handleD2NodeResponse(DiscoveryResponseData data)
...
-    handleResourceUpdate(updates, data.getResourceType());
-    handleResourceRemoval(data.getRemovedResources(), data.getResourceType());
+    processResourceChanges(data.getResourceType(), updates, data.getRemovedResources());
+  }
+
+  private void handleD2ClusterOrServiceNameResponse(DiscoveryResponseData data)
+  {
+    Map<String, D2ClusterOrServiceNameUpdate> updates = new HashMap<>();
+    List<String> errors = new ArrayList<>();
+    String errMsg;
```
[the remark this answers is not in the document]
bohhyang: oh, just a tiny nit to keep just one instance of it here, instead of creating a new one in every iteration of the for loop at line 388.

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
```diff
@@ -0,0 +1,157 @@
...
+        {
+          _dataReadyLock.writeLock().lock();
+        }
+        if (EMPTY_D2_CLUSTER_OR_SERVICE_NAME_UPDATE.equals(update))
+        { // invalid data, ignore. Logged in xds client.
+          return;
+        }
+        XdsD2.D2ClusterOrServiceName nameData = update.getNameData();
+        // the data is guaranteed valid by the xds client. It has a non-empty name in either clusterName or serviceName.
+        if (!nameData.getClusterName().isEmpty())
```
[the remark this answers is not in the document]
bohhyang: I'm not sure. That's why I really want to QEI test this code before merge it after the observer side is available on EI.
I will also update D2ClusterOrServiceNameUpdate's isValid method to use Strings.isNullOrEmpty to check this.

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
```diff
@@ -0,0 +1,157 @@
...
+   * the update is done.
+   */
+  @VisibleForTesting
+  final AtomicBoolean _isUpdating = new AtomicBoolean(true);
+  /**
+   * The write lock will be released when the service and cluster names data have been updated and is ready to serve.
+   * If the data is being updated, requests to read the data will wait indefinitely. Callers could set a shorter
+   * timeout on the callback passed in to getServiceNames or getClusterNames, as needed.
+   */
+  private final ReadWriteLock _dataReadyLock = new ReentrantReadWriteLock();
```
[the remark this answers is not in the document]
bohhyang: anything else can be used here? I thought ReentrantReadWriteLock is the standard read write lock that one can instantiate.

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
bohhyang: as Paul explained here: https://github.com/linkedin/rest.li/pull/1038#discussion_r1927343223

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
bohhyang: it's updated now.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: oh, didn't notice this! Updated. Thanks for catching!

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: makes sense. Updated all occurrence of this.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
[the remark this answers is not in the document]
bohhyang: updated.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: updated all "{" in this file that were on a new line.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: updated.

brycezhongqing (approved):
LGTM. Thanks for the contribution.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -604,6 +602,34 @@ private void handleResourceRemoval(Collection<String> removedResources, Resource
...
+      }
+
+      if (remainingChunks == 0)
+      {
+        _log.info("Notifying wildcard subscriber of type {} for the end of response chunks.", type);
+        wildcardResourceSubscriber.onAllResourcesProcessed();
+      }
+    }
+  }
+
```
brycezhongqing:
```suggestion
private void notifyOnLastChunk(DiscoveryResponseData response) {
    ResourceType type = response.getResourceType();
    WildcardResourceSubscriber wildcardResourceSubscriber = getWildcardResourceSubscriber(type);

    if (wildcardResourceSubscriber == null) {
        return;
    }

    int remainingChunks = decodeRemainingChunks(response.getNonce());
    if (remainingChunks == 0) {
        _log.info("Notifying wildcard subscriber of type {} for the end of response chunks.", type);
        wildcardResourceSubscriber.onAllResourcesProcessed();
    }
}

private int decodeRemainingChunks(String nonce) {
    try {
        byte[] bytes = Hex.decodeHex(nonce.toCharArray());
        ByteBuffer bb = ByteBuffer.wrap(Arrays.copyOfRange(bytes, 8, 12));
        return bb.getInt();
    } catch (Exception e) {
        _log.warn("Failed to decode nonce: {}", nonce, e);
        return -1;
    }
```

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
```diff
@@ -0,0 +1,162 @@
...
+        // do nothing
+      }
+    };
+  }
+
+  private void waitAndRespond(boolean isForService, Callback<List<String>> callback)
+  {
+    synchronized (_dataReadyLock)
+    {
+      while (_isUpdating.get())
```
PapaCharlie, line 147: I think the `while` could just be replaced with an `if` here. After the first `_isUpdating` returns true, you know bootstrap is complete, since you are checking the chunks with the nonce. This means this only handles the case where something is waiting for the initial bootstrap, then in between them waking up from `_dataReadyLock.wait()` and checking `_isUpdating`, some new resource was added/removed. The response would still be 99% correct since responses after the bootstrap will realistically update a handful of resources. There is a (very unlikely) pathological case where updates happen just frequently enough where this thread keeps looping here forever and never returns. Just seems like something we can easily avoid, it'll be easier to debug :)

Maybe you can leave a comment here saying why we chose this

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -534,34 +581,67 @@ private void handleResourceRemoval(Collection<String> removedResources, Resource
...
-      removedResources.forEach(wildcardSubscriber::onRemoval);
+      return;
     }
-  }
 
+    int remainingChunks;
+    try
+    {
+      byte[] bytes = Hex.decodeHex(response.getNonce().toCharArray());
+      ByteBuffer bb = ByteBuffer.wrap(Arrays.copyOfRange(bytes, 8, 12));
```
PapaCharlie, line 619: You don't need to copy this actually, you can use the direct reference here, since you aren't modifying this. Just do:
```suggestion
      byte[] bytes = Hex.decodeHex(response.getNonce().toCharArray());
      ByteBuffer bb = ByteBuffer.wrap(bytes, 8, 4);
```

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -534,34 +581,67 @@ private void handleResourceRemoval(Collection<String> removedResources, Resource
...
+    {
+      _log.warn("Failed to decode nonce: {}", response.getNonce(), e);
+      remainingChunks = -1;
+    }
+
+    if (remainingChunks == 0)
+    {
+      _log.info("Notifying wildcard subscriber of type {} for the end of response chunks.", type);
+      wildcardResourceSubscriber.onAllResourcesProcessed();
+    }
```
PapaCharlie, line 632: Something about this feels wrong. For example, if you fail to deserialize the nonce for whatever reason, `onAllResourcesProcessed` may never be called and things will hang forever (for example if you're talking to a really old version of the observer or something weird like that). I think you should actually reverse this, where you *always* call `onAllResourcesProcessed` _unless_ you get explicit signal that `remainingChunks > 0`. Functionally it's the same, but it'll work better if the Observer doesn't send the chunk suffix. This way it's just "best effort" where we try to handle chunks if we can, but fire `onAllResourcesProcessed` early otherwise, which is what you would have to do if you weren't using `diderot` anyway

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -534,34 +581,67 @@ private void handleResourceRemoval(Collection<String> removedResources, Resource
...
+      {
+        wildcardSubscriber.onRemoval(resourceName);
+      }
     }
-    WildcardResourceSubscriber wildcardSubscriber = getWildcardResourceSubscriber(type);
-    if (wildcardSubscriber != null)
+  }
+
+  // If the nonce indicates that this is the last chunk of the response, notify the wildcard subscriber.
+  private void notifyOnLastChunk(DiscoveryResponseData response)
```
PapaCharlie, line 606: Can you link the [diderot PR](https://github.com/linkedin/diderot/pull/3) that introduced this behavior?

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
```diff
@@ -0,0 +1,161 @@
...
+      {
+        try
+        {
+          LOG.debug("waiting on lock for data to be ready");
+          _dataReadyLock.wait();
```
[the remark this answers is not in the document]
bohhyang: see the doc at the lock at the top of the class. The user callback should define their own timeout.

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
```diff
@@ -0,0 +1,162 @@
...
+        } catch (Exception e)
+        {
+          // do nothing
```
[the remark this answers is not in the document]
bohhyang: thread can be interrupted for unexpected reasons. While isUpdating flag is still true, the thread can be kept waiting.
"Since notify() and notifyAll() randomly wake up threads that are waiting on this object’s monitor, it’s not always important that the condition is met. Sometimes the thread is woken up, but the condition isn’t actually satisfied yet."
https://www.baeldung.com/java-wait-notify

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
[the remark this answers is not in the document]
bohhyang: ReadWriteLock is just an interface, can't be instantiated. And it can't be unlocked from another thread.

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
```diff
@@ -0,0 +1,162 @@
...
+
+      @Override
+      public void onAllResourcesProcessed()
+      {
+        synchronized (_dataReadyLock)
+        {
+          _isUpdating.compareAndSet(true, false);
+          _dataReadyLock.notifyAll();
+          LOG.debug("notified all threads waiting on lock");
+        }
```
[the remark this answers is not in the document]
bohhyang: nope. It needs the synchronized block so other threads waiting with synchronized block on that lock can be notified.
"when we’ve executed synchronized instance method for the given object"
https://www.baeldung.com/java-wait-notify

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
bohhyang: actually the thread can be interrupted for unexpected reasons. While isUpdating flag is still true, the thread should be kept waiting.
"Since notify() and notifyAll() randomly wake up threads that are waiting on this object’s monitor, it’s not always important that the condition is met. Sometimes the thread is woken up, but the condition isn’t actually satisfied yet."
https://www.baeldung.com/java-wait-notify

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
```diff
@@ -0,0 +1,162 @@
...
+    addNameWatcher();
+    waitAndRespond(false, callback);
+  }
+
+  private void addNameWatcher()
+  {
+    if (_watcher.get() != null)
+    {
+      return;
+    }
```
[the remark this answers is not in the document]
bohhyang: thought about this too. But createNameWatcher will be executed every time to compare with null.

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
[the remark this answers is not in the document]
brycezhongqing: Yeah. That's my question, if the current `thread` is interrupted for unexpected reasons or even killed deliberately , why we still keep listening here instead of telling the client : hi, there is a issue and you need to take care about now.

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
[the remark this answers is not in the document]
brycezhongqing: No, createNameWatcher() will not be executed every time. It will only be executed if _watcher is null at the time of compareAndSet(null, createNameWatcher()).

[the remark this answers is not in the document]
brycezhongqing: So how about this way ?
```java
    _watcher.updateAndGet(w -> {
      return w == null ? createNameWatcher() : w;
    });
```

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -534,34 +581,67 @@ private void handleResourceRemoval(Collection<String> removedResources, Resource
...
+    try
+    {
+      byte[] bytes = Hex.decodeHex(response.getNonce().toCharArray());
+      ByteBuffer bb = ByteBuffer.wrap(Arrays.copyOfRange(bytes, 8, 12));
+      remainingChunks = bb.getInt();
+    }
+    catch (Exception e)
+    {
+      _log.warn("Failed to decode nonce: {}", response.getNonce(), e);
+      remainingChunks = -1;
```
brycezhongqing, line 625: Could we need to add `wildcardResourceSubscriber.onError()` there? 
e.g
```java
wildcardResourceSubscriber.onError(Status.UNKNOWN.withDescription("Failed to decode nonce"));
```

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: added

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: good call! fixed.

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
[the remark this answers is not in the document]
bohhyang: that unexpected reasons could be notifyAll incorrectly woke the thread up without meeting the isUpdating condition. There is nothing wrong with the thread and it should keep waiting in this case. I added a comment about this here.

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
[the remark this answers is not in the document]
bohhyang: after a second thought, you're right, it's safer and won't hurt to set an internal timeout, just in case onAllResourcesProcessed is never or taking a very long time to be called. Returning partial result is better than nothing. Added, thanks.

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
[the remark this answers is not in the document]
bohhyang: It will throw IllegalMonitorStateException without the synchronized block

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
bohhyang: I added a timeout to avoid being waiting forever.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: No, onError is for notifying ads stream error when disconnected, such as notifyAvailabilityChanges is called in onError to switches the store to file system in XdsToD2PropertiesAdaptor.

brycezhongqing: I see. As I saw from the definition, the error should be called once the there is transient error for grpc communication.
```
   /**
     * Called when the resource discovery RPC encounters some transient error.
     */
```

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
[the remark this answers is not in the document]
PapaCharlie: OK if you need to be in the `synchronized` to call `notifyAll`, it makes sense. Some condition variables allow you to notify without holding the lock (for example in Go this is the case)

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
PapaCharlie: Yeah but `notify` in this case is only called when the condition is met, right? So it's still true that when the thread wakes up from notify, it's guaranteed that the bootstrap is complete

d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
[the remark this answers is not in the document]
bohhyang: Actually I misunderstood it. The thread will be woken only if notifyAll is called by my code or when the system interrupts it, which likely is when the JVM is shutting down. I updated to handle the exception explicitly now (also removed the while loop since there is no need to re-check the isUpdating condition now after caught an exception). Thanks.

PapaCharlie (approved):

PapaCharlie (approved):

PR #1039, opened 2025-01-31
Make xds directory lazy to subscribe names

bohhyang:
## Summary
We don't want all apps to subscribe to d2 cluster / service names at startup, which could be a bit waste. This change makes the subscription lazy to happen at when first real request comes in. INDIS Server response is very fast, so it should be ok to subscribe lazily.

## Test Done
UT.

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
@@ -27,8 +27,7 @@ public class XdsDirectory implements Directory
   final ConcurrentMap<String, String> _serviceNames = new ConcurrentHashMap<>();
   @VisibleForTesting
   final ConcurrentMap<String, String> _clusterNames = new ConcurrentHashMap<>();
-  @VisibleForTesting
-  final AtomicReference<WildcardD2ClusterOrServiceNameResourceWatcher> _watcher = new AtomicReference<>();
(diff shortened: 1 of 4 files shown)


d2/src/main/java/com/linkedin/d2/xds/balancer/XdsDirectory.java
```diff
@@ -51,8 +51,7 @@ public XdsDirectory(XdsClient xdsClient)
 
   public void start() {
     LOG.debug("Starting. Setting isUpdating to true");
-    _isUpdating.set(true); // initially set to true to block reads before the first update completes
-    addNameWatcher();
+    _isUpdating.set(true); // initially set to true to block reads before the first (lazy) update completes
```
brycezhongqing, line 54: Need we set the `_isUpdating.set(true);` obiviously ? As I saw the initial value of _isUpdating is `true`

bohhyang: hmm, I think it's safer to always set it to true at start in case someone accidentally changes the initial value in future.

brycezhongqing: Sure. I'm ok with this

brycezhongqing (approved):

PapaCharlie (approved):

PR #1040, opened 2025-02-15
Resiliency: mark a service as used even if the getClient call during warmUp fails

ZoabKapoor:
## Summary

We recently observed a bug with some users of D2 where:

1. During warmUp all their calls to `WarmUpLoadBalancer.getClient` would time out due to [the call in ZKDeterministicSubsettingMetadataProvider timing out](https://github.com/linkedin/rest.li/blob/master/d2/src/main/java/com/linkedin/d2/balancer/subsetting/ZKDeterministicSubsettingMetadataProvider.java#L128). 
2. Because the `service` was added to the set of `_usedServices` **after** the getClient call, it would never be added
3. Upon shutdown, that service was considered unused and its warm up backup store was deleted (even though it was eventually populated after warmUp).

This is wrong, as we should consider a service used even if the `getClient` call fails. I fixed this by moving the logic to add the service to the used services before making the getClient call instead of after

## Testing done

Updated the unit test `testNotDeletingFilesGetClient` to test both the case where getClient returns and errors.  

### Before

Tests fail with
```
After shutdown there should be just one service, the one that we 'get the client' on
Expected :0
Actual   :1
```

### After

Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.64.1] - 2025-02-15
+- Fix warmUp -- record service as used regardless of whether getClient succeeds
+
 ## [29.64.0] - 2025-01-31
 - Allow subscribing to a single D2URI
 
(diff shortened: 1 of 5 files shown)


bohhyang (approved):
Nice catch and test! LGTM.

PR #1045, opened 2025-03-11
Add d2 slow start configuration support

shivamgupta1:
Added `SlowStartProperties` class and support for ser/deser for the same. The values would be received from Indis/ZK via the LPS D2 flow.

Added UTs.

### DIFF
--- a/d2-schemas/src/main/pegasus/com/linkedin/d2/D2Cluster.pdl
+++ b/d2-schemas/src/main/pegasus/com/linkedin/d2/D2Cluster.pdl
@@ -39,4 +39,9 @@ record D2Cluster includes D2ChangeTimeStamps {
    * Holds the configuration for this cluster's dark canary clusters, if any. The map is keyed by the dark canary name.
    */
   darkClusters: map[string, DarkClusterConfig] = { }
+
+  /**
+   * D2 slow start properties for gRPC clusters.
+   */
+  slowStartProperties: optional SlowStartProperties
 }
--- a/d2-schemas/src/main/pegasus/com/linkedin/d2/D2Cluster.pdl
+++ b/d2-schemas/src/main/pegasus/com/linkedin/d2/D2Cluster.pdl
@@ -41,7 +41,7 @@ record D2Cluster includes D2ChangeTimeStamps {
   darkClusters: map[string, DarkClusterConfig] = { }
 
   /**
-   * D2 slow start properties for gRPC clusters.
+   * D2 slow start properties. Currently used by clients of gRPC clusters.
    */
   slowStartProperties: optional SlowStartProperties
 }
--- a/d2-schemas/src/main/pegasus/com/linkedin/d2/SlowStartProperties.pdl
+++ b/d2-schemas/src/main/pegasus/com/linkedin/d2/SlowStartProperties.pdl
@@ -2,6 +2,7 @@ namespace com.linkedin.d2
 
 /**
  * D2 Slow Start related configuration.
+ * See https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/slow_start for more info.
  */
 record SlowStartProperties {
   /**
--- a/d2-schemas/src/main/pegasus/com/linkedin/d2/SlowStartProperties.pdl
+++ b/d2-schemas/src/main/pegasus/com/linkedin/d2/SlowStartProperties.pdl
@@ -1,0 +1,26 @@
+namespace com.linkedin.d2
+
+/**
+ * D2 Slow Start related configuration.
+ */
+record SlowStartProperties {
+  /**
+   * Whether the feature is disabled
+   */
+  disabled: boolean = false
+
+  /**
+   * The duration within which the weight and traffic would be fully ramped
+   */
+  windowDurationSeconds: int
+
+  /**
+   * Non-linearly affects the ramp speed, generally should be within [1,10]
+   */
+  aggression: double
+
+  /**
+   * The percentage of weight to start from, (0,1)
+   */
+  minWeightPercent: double
+}
--- a/d2-schemas/src/main/pegasus/com/linkedin/d2/SlowStartProperties.pdl
+++ b/d2-schemas/src/main/pegasus/com/linkedin/d2/SlowStartProperties.pdl
@@ -15,7 +15,7 @@ record SlowStartProperties {
   windowDurationSeconds: int
 
   /**
-   * Non-linearly affects the ramp speed, generally should be within [1,10]
+   * Non-linearly affects the ramp speed, larger values lead to quicker ramping. Generally should be within [1,10]
    */
   aggression: double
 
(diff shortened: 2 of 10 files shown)


d2-schemas/src/main/pegasus/com/linkedin/d2/SlowStartProperties.pdl
```diff
@@ -0,0 +1,26 @@
...
+   */
+  disabled: boolean = false
+
+  /**
+   * The duration within which the weight and traffic would be fully ramped
+   */
+  windowDurationSeconds: int
+
+  /**
+   * Non-linearly affects the ramp speed, larger values lead to quicker ramping. Generally should be within [1,10]
```
bbarkley, line 18: It might be worth adding a link out to the envoy docs here or for the record as a whole so that people have a better idea of how these all work together.

d2-schemas/src/main/pegasus/com/linkedin/d2/D2Cluster.pdl
```diff
@@ -39,4 +39,9 @@ record D2Cluster includes D2ChangeTimeStamps {
    * Holds the configuration for this cluster's dark canary clusters, if any. The map is keyed by the dark canary name.
    */
   darkClusters: map[string, DarkClusterConfig] = { }
+
+  /**
+   * D2 slow start properties for gRPC clusters.
```
bbarkley, line 44: Any reason to limit this to grpc usage? Maybe just indicate that it currently isn't implemented for the restli client?

d2/src/main/java/com/linkedin/d2/balancer/properties/SlowStartProperties.java
```diff
@@ -0,0 +1,75 @@
+package com.linkedin.d2.balancer.properties;
+
+import com.google.common.base.MoreObjects;
+import java.util.Objects;
+
+
+/**
+ * Slow start configuration properties for gRPC P2C load balancer.
```
bbarkley, line 8: Same comment as above on pdl - link to envoy docs for better clarity on how these interact?

bbarkley (commented):
Minor stuff here from me - you should get a proper review from one of the owners, especially on whether they're okay with deprecating all of the old constructors.

d2-schemas/src/main/pegasus/com/linkedin/d2/SlowStartProperties.pdl
shivamgupta1: Done

d2/src/main/java/com/linkedin/d2/balancer/properties/SlowStartProperties.java
shivamgupta1: Done

d2-schemas/src/main/pegasus/com/linkedin/d2/D2Cluster.pdl
shivamgupta1: Done

mchen07:
I thought that you only need to add d2 slow start for gRPC path since rest.li already has d2 slow start supported, then why did you need to change rest.li here?

shivamgupta1:
>I thought that you only need to add d2 slow start for gRPC path since rest.li already has d2 slow start supported, then why did you need to change rest.li here?

@mchen07 this is needed because gRPC code still makes use of [XdsClient](https://github.com/linkedin/rest.li/blob/988c403b17edbe56a21fab1953b4dd2de4d89f02/d2/src/main/java/com/linkedin/d2/xds/XdsClient.java) and the resulting D2 classes (e.g. `ClusterProperties`) from Restli code.

d2/src/main/java/com/linkedin/d2/balancer/properties/ClusterProperties.java
```diff
@@ -143,6 +148,7 @@ public ClusterProperties(String clusterName,
     this(clusterName, prioritizedSchemes, properties, bannedUris, partitionProperties, sslSessionValidationStrings, (Map<String, Object>)null, false);
   }
 
+  @Deprecated
```
mchen07, line 151: why do you mark this as @Deprecated? I thought the reason that we deprecated previous two methods is because they are using `DarkClusterConfigMap` instead of `Map`.

d2/src/main/java/com/linkedin/d2/balancer/properties/ClusterProperties.java
```diff
@@ -156,6 +162,7 @@ public ClusterProperties(String clusterName,
         darkClusters, delegated, -1);
```
mchen07, line 162: use DEFAULT_VERSION for -1?

d2/src/main/java/com/linkedin/d2/balancer/properties/ClusterStoreProperties.java
```diff
@@ -43,23 +43,27 @@ public class ClusterStoreProperties extends ClusterProperties
   protected final CanaryDistributionStrategy _canaryDistributionStrategy;
   protected final FailoutProperties _failoutProperties;
 
+  @Deprecated
```
mchen07, line 46: why do you mark all such methods @Deprecated? Worth to add reason?

d2/src/main/java/com/linkedin/d2/balancer/properties/ClusterProperties.java
shivamgupta1: Done

d2/src/main/java/com/linkedin/d2/balancer/properties/ClusterProperties.java
```diff
@@ -166,6 +173,20 @@ public ClusterProperties(String clusterName,
       boolean delegated,
       long version)
   {
+    this(clusterName, prioritizedSchemes, properties, bannedUris, partitionProperties, sslSessionValidationStrings,
+        darkClusters, delegated, DEFAULT_VERSION, null);
```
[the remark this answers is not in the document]
shivamgupta1: Great catch, fixed!

d2/src/main/java/com/linkedin/d2/balancer/properties/ClusterStoreProperties.java
shivamgupta1: I had deprecated them since it seemed like the were old constructors that would've become obsolete when new params were added and these were left in just for backwards compatibility. However, I notice that making them deprecated results in unnecessary warnings during compilation, and may not be a great experience for OSS users. Hence, the deprecation has been reverted now.

d2/src/main/java/com/linkedin/d2/balancer/properties/ClusterProperties.java
shivamgupta1: I had deprecated them since it seemed like the were old constructors that would've become obsolete when new params were added and these were left in just for backwards compatibility. However, I notice that making them deprecated results in unnecessary warnings during compilation, and may not be a great experience for OSS users. Hence, the deprecation has been reverted now.

mchen07 (approved):

PR #1049, opened 2025-03-26
SI-37780, Change log level of ads stream closure from err to warn

bohhyang:
## Summary
As the title, to avoid excessive error exception tickets created to the app owner, where nothing needs their attention or action, since the stream will be re-established with another indis observer.

## Test Done
N/A


brycezhongqing (approved):
LGTM

PR #1052, opened 2025-04-09
fix d2ClientBuilder shutdown corner case

brycezhongqing:
Fix a corner case bug in d2ClientBuilder shutdown logic.

Corner case: when `_config._executorService == null`

```
  if (_config._executorService == null)
    {
      LOG.warn("No executor service passed as argument. Pass it for " +
        "enhanced monitoring and to have better control over the executor.");
      _config._executorService =
        Executors.newSingleThreadScheduledExecutor(new NamedThreadFactory("D2 PropertyEventExecutor"));
      executorsToShutDown.add(_config._executorService);
    }
```

Then `_executorService` will be added to `executorsToShutDown`. Finally, when d2Client call the shutdown method, the _executorService may be closed twice. Like

1. [first place](https://github.com/linkedin/rest.li/blob/38e48ecee0476ce49dae65c273f9aaff90e0ce16/d2/src/main/java/com/linkedin/d2/xds/balancer/XdsLoadBalancer.java#L192)
2. [second place](https://github.com/linkedin/rest.li/blob/38e48ecee0476ce49dae65c273f9aaff90e0ce16/d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java#L894)

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
@@ -886,17 +886,33 @@ public class D2ClientBuilder
         @Override
         public void onError(Throwable e)
         {
-          _executors.forEach(ExecutorService::shutdown);
+          shutdownExecutorServices();
           callback.onError(e);
         }
 
         @Override
         public void onSuccess(None result)
         {
-          _executors.forEach(ExecutorService::shutdown);
+          shutdownExecutorServices();
           callback.onSuccess(result);
         }
       });
+    }
+
+    private void shutdownExecutorServices()
+    {
+      for (ExecutorService executor : _executors)
+      {
+        if (executor.isShutdown())
+        {
+          LOG.warn("Executor is already shut down");
+        }
+        else
+        {
+          LOG.info("Shutting down executor");
+          executor.shutdown();
+        }
+      }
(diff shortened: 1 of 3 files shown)


d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
```diff
@@ -886,17 +886,33 @@ public void shutdown(Callback<None> callback)
...
       });
     }
+
+    private void shutdownExecutorServices()
+    {
+      for (ExecutorService executor : _executors)
+      {
+        if (executor.isShutdown())
+        {
+          LOG.warn("Executor is already shut down");
```
bohhyang, line 908: don't need warning, info level is fine.

bohhyang (commented):
should the first place be removed then? if the executors are shut down in the second place. But I guess it's ok as long as the second place has the isShutdown check guard.

brycezhongqing:
> should the first place be removed then? if the executors are shut down in the second place. But I guess it's ok as long as the second place has the isShutdown check guard.

Sadly. No. As `executorsToShutDown.add(_config._executorService);` will only execute when `if (_config._executorService == null)`

If `_config._executorService != null`, the first place should be executed

ChrisCarini:

PR #1054, opened 2025-04-09
Fix to default values issue

SrilakshmiBharadwaj:
DataTranslator adds null for default values of array and map fields, but record data schema and avro from proto add [] and {} for default values.
This PR fixes the issue by backfilling default values in data map when present in record data schema


ChrisCarini:

PR #1056, opened 2025-04-11
Remove invalid characters from multipart/mixed encoded requests

bbarkley:
The Java Mail code that is used for query tunneling creates an invalid Content-Type header for multipart/mixed encoded requests, containing a newline, return, and tab character (e.g. multipart/mixed; \r\n\tboundary="----=_Part_0_1587908352.1744332411531"). This isn't compliant with HTTP RFCs and the Netty client has become more strict about validating header values. This change replaces any whitespace of two or more characters with a single space.

Tests for encoding a request have been updated to validate the Content-Type doesn't contain invalid characters, and existing tests decoding these encoded requests continue to pass.


zackthehuman (approved):
Nice find.

PR #1058, opened 2025-04-18
Add options for connection jitter

TylerHorth:
Add options to d2 config for connection jittering. Will be used by the gRPC client to add random delay when establishing new connections to avoid a thundering herd problem.

### DIFF
--- a/d2-schemas/src/main/pegasus/com/linkedin/d2/ConnectionOptions.pdl
+++ b/d2-schemas/src/main/pegasus/com/linkedin/d2/ConnectionOptions.pdl
@@ -1,0 +1,24 @@
+namespace com.linkedin.d2
+
+/**
+ * Options for configuring the connection pool. Only used by gRPC clients.
+ */
+record ConnectionOptions {
+  /**
+   * Amount of jitter to apply when establishing a new connection. When a new host is added to the pool, the client will
+   * wait a random amount of time between 0 and this value before attempting to connect to the host. This is done to
+   * prevent a thundering herd problem when a large number of clients are trying to connect to the same host at the
+   * same time. A value of 0 disables connection jitter.
+   */
+  connectionJitterSeconds: int
+
+  /**
+   * Controls the maximum number of connections that can be delayed by connection jitter before the client will start
+   * immediately establishing connections. This value represents a ratio between the number of delayed connections and
+   * the total number of connections. For example, if this value is set to 0.2, the client will start immediately
+   * establishing connections when 20% of the connections are delayed by connection jitter. Connections are established
+   * by random selection from the delayed connections. Value must be between 0 and 1.0. Only applies if
+   * connectionJitterSeconds is enabled.
+   */
+  maxDelayedConnectionRatio: float
+}
--- a/d2/src/main/java/com/linkedin/d2/balancer/properties/ClusterPropertiesJsonSerializer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/properties/ClusterPropertiesJsonSerializer.java
@@ -18,12 +18,10 @@ package com.linkedin.d2.balancer.properties;
 
 import com.google.common.base.Preconditions;
 import com.google.protobuf.ByteString;
-import com.linkedin.d2.balancer.properties.util.PropertyUtil;
 import com.linkedin.d2.balancer.util.JacksonUtil;
 import com.linkedin.d2.discovery.PropertyBuilder;
 import com.linkedin.d2.discovery.PropertySerializationException;
 import com.linkedin.d2.discovery.PropertySerializer;
-import java.util.Optional;
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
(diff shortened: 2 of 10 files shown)


d2-schemas/src/main/pegasus/com/linkedin/d2/ConnectionOptions.pdl
```diff
@@ -0,0 +1,24 @@
...
+
+  /**
+   * Controls the maximum number of connections that can be delayed by connection jitter before the client will start
+   * immediately establishing connections. This value represents a ratio between the number of delayed connections and
+   * the total number of connections. For example, if this value is set to 0.2, the client will start immediately
+   * establishing connections when 20% of the connections are delayed by connection jitter. Connections are established
+   * by random selection from the delayed connections. Value must be between 0 and 1.0. Only applies if
+   * connectionJitterSeconds is enabled.
+   */
+  maxDelayedConnectionRatio: float
```
mchen07, line 23: Is this percentage to control in our design

```
The IDLE pool server count must not exceed a certain percentage threshold of the total cluster host count (e.g. 20%). This helps protect us when deployment proceeds too fast.

```

d2/src/main/java/com/linkedin/d2/balancer/properties/ConnectionOptions.java
```diff
@@ -0,0 +1,48 @@
+package com.linkedin.d2.balancer.properties;
+
+import java.util.Objects;
+import java.util.Optional;
+import javax.annotation.Nullable;
+
+
+public class ConnectionOptions {
```
mchen07, line 8: So even though we use `ConnectionOptions.pdl` to model, but we don't use PDL generated POJO as in-memory model, rather write our own POJO? In that case, I am curious why we even need to model d2 schema in PDL in the first place?

d2/src/main/java/com/linkedin/d2/balancer/properties/ClusterPropertiesJsonSerializer.java
```diff
@@ -240,11 +241,26 @@ private ClusterProperties buildClusterPropertiesFromMap(Map<String, Object> map)
...
+        darkClusterProperty, delegated, ClusterProperties.DEFAULT_VERSION, slowStartProperties, connectionOptions);
+  }
+
+  private ConnectionOptions getConnectionOptions(Map<String, Object> map)
+  {
+    Map<String, Object> connectionOptionsMap = mapGet(map, PropertyKeys.CONNECTION_OPTIONS);
+    if (connectionOptionsMap == null) {
+      return null;
+    }
+    int connectionJitterSeconds = PropertyUtil.checkAndGetValue(connectionOptionsMap, PropertyKeys.CONNECTION_JITTER_SECONDS,
```
shivamgupta1, line 260: Do these still need `PropertyUtil.` after the * import above?

d2/src/main/java/com/linkedin/d2/balancer/properties/ConnectionOptions.java
shivamgupta1: It's needed on the write side (d2-config-service, maestro etc.)

shivamgupta1 (approved):

gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=29.65.7
+version=29.65.8
```
nickibi, line 1: Please bump minor version if the change introduces new classes or new APIs. LinkedIn dependency check will only enforce bumping rest.li minor version not the patch version.

d2-schemas/src/main/pegasus/com/linkedin/d2/ConnectionOptions.pdl
TylerHorth: Yes, this is the percentage from the design. The design just says "a certain percentage". I decided to include this as a config rather than hard code a value since we don't know what the right percentage should be, and it's easier to update the config.

d2/src/main/java/com/linkedin/d2/balancer/properties/ConnectionOptions.java
mchen07: still not quite clear where PDL is used.

d2/src/main/java/com/linkedin/d2/balancer/properties/ClusterPropertiesJsonSerializer.java
TylerHorth: Updated

gradle.properties
TylerHorth: Thanks for the catch. Updated this.

d2/src/main/java/com/linkedin/d2/balancer/properties/ConnectionOptions.java
mchen07: synced offline on this.

mchen07 (approved):

PR #1060, opened 2025-04-24
detect LI raw d2 client builder usage

bohhyang:
## Summary
1. Detect LI raw d2 client builder usage, log a warning and system properties of the app (which will be saved to a znode later for revealing their identity). 
2. Mark non-indis load balancer factories as deprecated, and log a warning when they are used.
3. Downgrade backup file related errors to warn level, since backup files are non critical to SD behaviors.
(apps/offline jobs that don't use standard cfg2 will fail to resolve the default file path which contains the cfg2 properties of the app name, instance, fabric, etc.).

Note: this changes LoadBalancerWithFacilitiesFactory from interface to an abstract class. There is no direct usage of this interface in LI.  

Also cleans up a bunch of warnings introduced by adding LoadBalancerServer interface before.

## Testing Done
QEI test with indis-canary


d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
```diff
@@ -303,6 +302,18 @@ else if (_config.restRetryEnabled || _config.streamRetryEnabled)
     return d2Client;
   }
 
+  /**
+   * Check if the d2 client builder is to build a LI raw D2 client. When LI container D2ClientFactory is used, it sets
+   * hostName and d2JmxManagerPrefix with LI-specific values (app name, machine name, etc) at runtime. All LI raw D2
+   * client usages are known not setting these values according to code search.
+   * @return true if this is a LI raw D2 client, false otherwise.
+   */
+  boolean isLiRawD2Client()
```
brycezhongqing, line 311: tip:  make this method to be `private`

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
```diff
@@ -303,6 +302,18 @@ else if (_config.restRetryEnabled || _config.streamRetryEnabled)
...
+  {
+    return Objects.equals(_config.hostName, D2ClientConfig.HOST_NAME_DEFAULT)
+        || Objects.equals(_config.d2JmxManagerPrefix, D2ClientConfig.D2_JMX_MANAGER_PREFIX_DEFAULT);
```
brycezhongqing, line 314: The first thing come to my mind is :
```
return _config.hostName == null || _config.d2JmxManagerPrefix == null;
```
Could we learn more detials that create the default value there ?

d2/src/main/java/com/linkedin/d2/balancer/D2ClientConfig.java
```diff
@@ -172,6 +174,13 @@ public class D2ClientConfig
   public boolean loadBalanceStreamException = false;
   public boolean xdsInitialResourceVersionsEnabled = false;
 
+  /**
+   * When this is true, D2 client builder will detect if it's used to build a raw D2 client (as opposed to using standard
+   * D2 client factory in LI container library) and set the isLiRawD2Client flag accordingly in the load balancer factory.
+   * Open Source Users can set this to false to disable this behavior.
+   */
+  public boolean detectLiRawD2Client = true;
```
brycezhongqing, line 182: So just double confirm, are we going to set this to be `true` as default value ? Do we need to set this to be `false` as default, as this is a new feature for the d2 client builder. But as this feature introduce very little risk. I'm open to both

d2/src/main/java/com/linkedin/d2/balancer/LastSeenBalancerWithFacilitiesFactory.java
```diff
@@ -49,7 +49,8 @@
  * Implementation of {@link LoadBalancerWithFacilitiesFactory} interface, which creates an instance of
  * {@link LastSeenLoadBalancerWithFacilities}
  */
-public class LastSeenBalancerWithFacilitiesFactory implements LoadBalancerWithFacilitiesFactory
+@Deprecated
```
brycezhongqing, line 52: Are we going to add more comments to guide users which `LoadBalancerWithFacilities` they should use now ?

d2/src/main/java/com/linkedin/d2/discovery/util/D2Utils.java
```diff
@@ -17,6 +20,13 @@
 
 public class D2Utils
 {
+  // A set of system properties to be excluded as they are lengthy, not needed, etc.
+  private static final Set<String> SYSTEM_PROPS_TO_EXCLUDE = Stream.of(
+      "jdk.debug",
+      "line.separator",
+      "java.class.path",
+      "java.vm.inputarguments"
+  ).collect(Collectors.toSet());
```
brycezhongqing, line 29: How about directly create a set
```suggestion
private static final Set<String> systemPropsToExclude = new HashSet<>(Arrays.asList(
    "jdk.debug",
    "line.separator",
    "java.class.path",
    "java.vm.inputarguments"
));
```

d2/src/main/java/com/linkedin/d2/discovery/util/D2Utils.java
```diff
@@ -55,4 +65,20 @@ public static String getServicePathAsChildOfCluster(String clusterName, String s
   {
     return ZKFSUtil.clusterPath(basePath) + "/" + clusterName + "/" + serviceName;
   }
+
+  /**
+   * System properties could include properties set by LinkedIn, Java, Zookeepr, and more. It will be logged or saved
+   * on a znode to reveal identities of apps that are using hard-coded D2ClientBuilder.
+   * @return A string of system properties.
+   */
+  public static String getSystemProperties()
```
brycezhongqing, line 74: As this is not D2 properties only. Just wondering do we need to add it to more common util module ?

d2/src/main/java/com/linkedin/d2/balancer/LoadBalancerWithFacilitiesFactory.java
```diff
@@ -16,15 +16,45 @@
...
+import com.linkedin.d2.discovery.util.D2Utils;
+import javax.annotation.Nonnull;
+import org.slf4j.Logger;
+
+
 /**
  * Factory for creating instance of {@link LoadBalancerWithFacilities}
  */
-public interface LoadBalancerWithFacilitiesFactory
+public abstract class LoadBalancerWithFacilitiesFactory
```
brycezhongqing, line 27: https://jarvis.corp.linkedin.com/codesearch/result/?name=RtfLoadBalancerFactory.java&path=restli-test-fwk%2Frtf-d2%2Fsrc%2Fmain%2Fjava%2Fcom%2Flinkedin%2Frtf&reponame=linkedin-multiproduct%2Frestli-test-fwk#9

Have some concerns for this incompatible change. As some mps are still using `LoadBalancerWithFacilitiesFactory` as interface

d2/src/main/java/com/linkedin/d2/balancer/D2ClientConfig.java
```diff
@@ -61,7 +61,8 @@ public class D2ClientConfig
   @Deprecated
   String zkHosts = null;
   public String xdsServer = null;
-  public String hostName = null;
+  public String hostName = HOST_NAME_DEFAULT;
```
PapaCharlie, line 64: Is it possible for the factory to explicitly set this to `null`? In which case the check won't work, even though it may have been loaded through the factory. I'm just curious if there are any ways `isLiRawD2Client` will break

d2/src/main/java/com/linkedin/d2/balancer/D2ClientConfig.java
PapaCharlie: Yeah ideally this should be `disableDetectLiRawD2Client` or something. Booleans with default values of `true` almost always blow up in your face as most people don't expect it.

d2/src/main/java/com/linkedin/d2/balancer/LoadBalancerWithFacilitiesFactory.java
PapaCharlie: I was about to open the same comment! This is a backwards incompatible change. We need a major version bump for this because it breaks container

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
bohhyang: what do you mean? the default value is used in D2ClientConfig, they won't be null.

d2/src/main/java/com/linkedin/d2/balancer/D2ClientConfig.java
bohhyang: he default must be true so that detecting li raw d2 client can work. The raw d2 client bulider usages will only use default value.

d2/src/main/java/com/linkedin/d2/discovery/util/D2Utils.java
bohhyang: what's the difference? why to prefer one over the other?

d2/src/main/java/com/linkedin/d2/discovery/util/D2Utils.java
bohhyang: such as? since this is only used by d2, I think it's ok to keep it in d2 package.

d2/src/main/java/com/linkedin/d2/balancer/LoadBalancerWithFacilitiesFactory.java
bohhyang: ah, let me change it back to interface and see how to implement the logic there.

d2/src/main/java/com/linkedin/d2/balancer/D2ClientConfig.java
bohhyang: it's possible to set null, but users are not doing that, and they won't do it out of no where (no instruction tells them to it). Also, the d2JmxPrefix is a container-specific field that has even less chance to be set manually. So this should cover all cases well.

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
brycezhongqing: I meant `HOST_NAME_DEFAULT` is null now, then why we don't directly check `_config.hostName` is null or not ?

d2/src/main/java/com/linkedin/d2/discovery/util/D2Utils.java
brycezhongqing: It's just a tip, for the second one:
	•	Slightly more performant, especially for small sets.
	•	Less runtime overhead than streams.

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
bohhyang: we should use the var when available. If HOST_NAME_DEFAULT changes some day, all the places that supposed to check against will be still working. (just like if a = 1 and we want to check b == a, we won't check b == 1).

d2/src/main/java/com/linkedin/d2/discovery/util/D2Utils.java
bohhyang: the diff is very negligible.  Since introduced in java 8,  Stream has been super commonly used to handle Set, List, Map entries, etc. It also saves a bit typing. (If Java 9 is available, I would use Set.of then, but with 8, I think Stream is fine).

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
brycezhongqing: Yeah. Make sense. Thanks for the explanation

d2/src/main/java/com/linkedin/d2/discovery/util/D2Utils.java
bohhyang: that will require adding restli-tool as a [dependency](https://github.com/linkedin/rest.li/blob/f8b66e71860cce77cafb0bf81300be7612b930ab/d2/build.gradle#L6) to d2 package, which could affect build in some unexpected way due to legacy things, since we don't know much about any package out of d2.

d2/src/main/java/com/linkedin/d2/balancer/LastSeenBalancerWithFacilitiesFactory.java
bohhyang: sg, added.

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
bohhyang: yeah, it should be. thanks.

d2/src/main/java/com/linkedin/d2/balancer/D2ClientConfig.java
bohhyang: I renamed it to disableDetectLiRawD2Client and default to false.

d2/src/main/java/com/linkedin/d2/discovery/util/D2Utils.java
brycezhongqing: Yeah. I'm open to both method. It's just a tiny tip and yes, the diff is very negligible

d2/src/main/java/com/linkedin/d2/balancer/AbstractLoadBalancerWithFacilitiesFactory.java
```diff
@@ -0,0 +1,38 @@
...
+      + "stability. Non-INDIS D2 Client will CRASH in OCTOBER 2025.";
+
+  // As standard LinkedIn code set LI-specific properties in d2 clients, this is true when LI code uses raw d2 client
+  // builder to create a d2 client.
+  protected boolean _isLiRawD2Client = false;
+
+  @Override
+  public void setIsLiRawD2Client(boolean isRawD2Client)
+  {
+    _isLiRawD2Client = isRawD2Client;
```
brycezhongqing, line 26: Could we set `_isLiRawD2Client` in `D2ClientConfig` ? Then for `LastSeenBalancerWithFacilitiesFactory` and `ZKFSLoadBalancerWithFacilitiesFactory`, we could directly check `D2ClientConfig` instead of 

```
    if (_isLiRawD2Client)
    {
      //TODO: Set flag in ZooKeeperEphemeralStore to create a permanent znode about the app.
      logAppProps(LOG);
    }
```

bohhyang: that's true, but then after isLiRawD2Client in D2ClientBulider find out it's not a raw d2 client for OSS use cases, it has to set cfg.isLiRawD2Client = false (cfg is created at line 150 already), which modifies the config after it's created. In general config should be final and read-only once it's created.

brycezhongqing: >  cfg.isLiRawD2Client = false (cfg is created at line 150 already), which modifies the config after it's created. In general config should be final and read-only once it's created.

That's what I'm checking too.  Just wondering whether we could set `isLiRawD2Client`  here:
https://github.com/linkedin/rest.li/blob/f8b66e71860cce77cafb0bf81300be7612b930ab/d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java#L226


For `D2ClientConfig`, we could add a new field like `isLiRawD2Client` that default value is false

And the value could be set when we create `D2ClientConfig`.

```java
final D2ClientConfig cfg = new D2ClientConfig(_config.zkHosts,
```

and _config. isLiRawD2Client = isLiRawD2Client()

d2/src/main/java/com/linkedin/d2/balancer/D2ClientConfig.java
abhics1002, line 64: we should add some context why we are using two strings hostName & HOST_NAME_DEFAULT here?

d2/src/main/java/com/linkedin/d2/balancer/D2ClientConfig.java
```diff
@@ -151,7 +152,8 @@ public class D2ClientConfig
   public ScheduledExecutorService startUpExecutorService = null;
   public ScheduledExecutorService indisStartUpExecutorService = null;
   public JmxManager jmxManager = new NoOpJmxManager();
-  public String d2JmxManagerPrefix = "UnknownPrefix";
+  public String d2JmxManagerPrefix = D2_JMX_MANAGER_PREFIX_DEFAULT;
+  public static final String D2_JMX_MANAGER_PREFIX_DEFAULT = "UnknownPrefix";
```
abhics1002, line 156: we should add some context why we are using two strings d2JmxManagerPrefix & D2_JMX_MANAGER_PREFIX_DEFAULT here?

d2/src/main/java/com/linkedin/d2/discovery/util/D2Utils.java
```diff
@@ -55,4 +65,20 @@ public static String getServicePathAsChildOfCluster(String clusterName, String s
   {
     return ZKFSUtil.clusterPath(basePath) + "/" + clusterName + "/" + serviceName;
   }
+
+  /**
+   * System properties could include properties set by LinkedIn, Java, Zookeepr, and more. It will be logged or saved
```
abhics1002, line 70: nit: typo- `Zookeepr`

d2/src/main/java/com/linkedin/d2/discovery/util/D2Utils.java
```diff
@@ -17,6 +20,13 @@
 
 public class D2Utils
 {
+  // A set of system properties to be excluded as they are lengthy, not needed, etc.
+  private static final Set<String> SYSTEM_PROPS_TO_EXCLUDE = Stream.of(
```
abhics1002, line 24: should we read this from config, if in future, we need to exclude more config we need not to change the code. 
is it possible to keep these as a default property and add/remove more if needed.

d2/src/main/java/com/linkedin/d2/balancer/D2ClientConfig.java
bohhyang: Just to define the default value as a final var, so that we can use it here and in the checking method.

d2/src/main/java/com/linkedin/d2/balancer/D2ClientConfig.java
bohhyang: Just to define the default value as a final var, so that we can use it here and in the checking method.

d2/src/main/java/com/linkedin/d2/discovery/util/D2Utils.java
bohhyang: remember this is OSS code, there is no config in restli:)

d2/src/main/java/com/linkedin/d2/balancer/AbstractLoadBalancerWithFacilitiesFactory.java
bohhyang: hmm, that will be better (although still changing d2ClientBuilder's config when build() is called), let me see.

d2/src/main/java/com/linkedin/d2/balancer/D2ClientConfig.java
bohhyang: I moved all static final vars to the top so it's less weird.

d2/src/main/java/com/linkedin/d2/discovery/util/D2Utils.java
bohhyang: fixed.

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
```diff
@@ -220,21 +220,20 @@ public D2Client build()
...
       new ZKFSLoadBalancerWithFacilitiesFactory() :
       _config.lbWithFacilitiesFactory;
 
-    // Warn for non-INDIS load balancer factory usage
-    if (!(loadBalancerFactory instanceof XdsLoadBalancerWithFacilitiesFactory))
+    if (_config.detectLiRawD2Client && isLiRawD2Client())
     {
-      LOG.warn(String.format("ACTION REQUIRED: Zookeeper-based D2 Client is deprecated (except locally-deployed ZK) "
-          + "and must be migrated to INDIS. See instructions at " +
-          "https://iwww.corp.linkedin.com/wiki/cf/display/ENGS/INDIS+Rollout+Issue+Guidelines+for+Java+Apps" +
```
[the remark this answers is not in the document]
bohhyang: yeah, updated.

d2/src/main/java/com/linkedin/d2/balancer/AbstractLoadBalancerWithFacilitiesFactory.java
bohhyang: updated.

brycezhongqing (approved):
🚢

PR #1065, opened 2025-05-15
Don't call _executorService.execute on new tasks once the XdsClient has been shut down

ZoabKapoor:
We discovered an edge case where the `AdsStream`'s onError callback can be called **after** the `_executorService` has been shut down, which leads to a `RejectedExecutionException` being thrown, which the caller might not expect or be able to handle.

To avoid this, let's protect calls to the `_executorService` with a check on whether it's already been shut down.

## Testing done

1. Added a test `testExecutorServiceNotUsedAfterShutdown` which passes
2. Checked out the current version of `XdsClientImpl` from master and verified `testExecutorServiceNotUsedAfterShutdown` **fails** with `java.util.concurrent.RejectedExecutionException` (confirms that the test is in fact verifying the new behavior)

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
@@ -50,6 +50,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
@@ -97,7 +98,7 @@ public class XdsClientImpl extends XdsClient
   private BackoffPolicy _retryBackoffPolicy;
   @VisibleForTesting
   AdsStream _adsStream;
-  private boolean _shutdown;
+  private boolean _isXdsStreamShutdown;
   @VisibleForTesting
   ScheduledFuture<?> _retryRpcStreamFuture;
   private ScheduledFuture<?> _readyTimeoutFuture;
@@ -175,10 +176,6 @@ public class XdsClientImpl extends XdsClient
   @Override
   public void watchXdsResource(String resourceName, ResourceWatcher watcher)
   {
-    if (_shutdown) {
-      _log.warn("Attempting to start watching resource {} after shutdown, will do nothing", resourceName);
-      return;
-    }
     checkShutdownAndExecute(() ->
     {
       ResourceType originalType = watcher.getType();
@@ -273,7 +270,7 @@ public class XdsClientImpl extends XdsClient
   @VisibleForTesting
   void startRpcStreamLocal()
   {
-    if (_shutdown)
+    if (_isXdsStreamShutdown)
     {
       _log.warn("RPC stream cannot be started after shutdown!");
       return;
@@ -288,7 +285,7 @@ public class XdsClientImpl extends XdsClient
         AggregatedDiscoveryServiceGrpc.newStub(_managedChannel);
     AdsStream stream = new AdsStream(stub);
     _adsStream = stream;
-    _readyTimeoutFuture = _executorService.schedule(() ->
+    _readyTimeoutFuture = checkShutdownAndSchedule(() ->
     {
       // There is a race condition where the task can be executed right as it's being cancelled. This checks whether
       // the current state is still pointing to the right stream, and whether it is ready before notifying of an error.
@@ -309,15 +306,27 @@ public class XdsClientImpl extends XdsClient
   @Override
   public void shutdown()
   {
-    _executorService.execute(() ->
+    // create a new executor instead of using the _executorService because this shutdown should run
+    // regardless of whether _executorService is available.
+    Executors.newSingleThreadExecutor().execute(() ->
     {
-      _shutdown = true;
+      _isXdsStreamShutdown = true;
       _log.info("Shutting down");
       if (_adsStream != null)
       {
         _adsStream.close(Status.CANCELLED.withDescription("shutdown").asException());
       }
     });
+  }
+
+  private ScheduledFuture<?> checkShutdownAndSchedule(Runnable runnable, long delay, TimeUnit unit) {
+    if (_executorService.isShutdown())
+    {
+      _log.warn("Attempting to schedule a task after _executorService was shutdown, will do nothing");
+      return null;
+    }
+
+    return _executorService.schedule(runnable, delay, unit);
   }
 
   private void checkShutdownAndExecute(Runnable runnable)
@@ -1108,7 +1117,7 @@ public class XdsClientImpl extends XdsClient
       return;
     }
     _adsStream = testStream;
-    _retryRpcStreamFuture = _executorService.schedule(new RpcRetryTask(), 0, TimeUnit.NANOSECONDS);
+    _retryRpcStreamFuture = checkShutdownAndSchedule(new RpcRetryTask(), 0, TimeUnit.NANOSECONDS);
   }
 
   // Return true if the client should subscribe to URI glob collection for the given resource type.
@@ -1462,11 +1471,7 @@ public class XdsClientImpl extends XdsClient
         delayNanos = _retryBackoffPolicy.nextBackoffNanos();
       }
       _log.info("Retry ADS stream in {} ns", delayNanos);
-      if (_shutdown) {
-        _log.warn("Already shut down, will not retry ADS stream");
-        return;
-      }
-      _retryRpcStreamFuture = _executorService.schedule(new RpcRetryTask(), delayNanos, TimeUnit.NANOSECONDS);
+      _retryRpcStreamFuture = checkShutdownAndSchedule(new RpcRetryTask(), delayNanos, TimeUnit.NANOSECONDS);
     }
 
     private void close(Exception error)
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
@@ -322,9 +322,9 @@ public class XdsClientImpl extends XdsClient
 
   private void checkShutdownAndExecute(Runnable runnable)
   {
-    if (_shutdown)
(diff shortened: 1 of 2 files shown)


d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -175,6 +175,10 @@ public XdsClientImpl(Node node,
   @Override
   public void watchXdsResource(String resourceName, ResourceWatcher watcher)
   {
+    if (_shutdown) {
```
ZoabKapoor, line 178: Question for reviewers: would it be more idiomatic to use `_shutdown` or `_executorService.isShutdown` here? They should be the same based on the implementation of `shutdown()`

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -1365,12 +1381,20 @@ public void onNext(DeltaDiscoveryResponse response)
             @Override
             public void onError(Throwable t)
             {
+              if (_shutdown) {
```
brycezhongqing: TBH. I think for `_shutdown` means the `_xdsAdaptor` is shutdown, but it didn't shutdown `_executorService`

So what I'm suggesting here, is to check `_executorService` itself.

ZoabKapoor: Makes sense! That's what I had initially but i noticed the boolean flag `shutdown` in the class so was wondering if that might be more idiomatic to use (https://github.com/linkedin/rest.li/pull/1065#discussion_r2091934870).

But makes sense that `_shutdown` and `_executorService.shutdown()` don't check the same thing, so we should specifically check only `_executorService.shutdown()` so we are canceling tasks on only the most narrow surface area

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -1447,6 +1462,10 @@ private void handleRpcStreamClosed(Status error)
         delayNanos = _retryBackoffPolicy.nextBackoffNanos();
       }
       _log.info("Retry ADS stream in {} ns", delayNanos);
+      if (_shutdown) {
+        _log.warn("Already shut down, will not retry ADS stream");
+        return;
+      }
```
brycezhongqing, line 1468: note the format

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: They are different. _shutdown won't be set to true if _executorService is already shut down, will reject execution at line 312.
Can you rename _shutdown to _isXdsStreamShutdown?

At line 312, it also needs to call checkShutdownAndExecute instead of directly calling .execute. If the executor is already shutdown for some reason (not sure if it really happens in edge case), we should still run the function in a new thread, just not to be executed in the executor.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -1447,6 +1462,10 @@ private void handleRpcStreamClosed(Status error)
...
       _retryRpcStreamFuture = _executorService.schedule(new RpcRetryTask(), delayNanos, TimeUnit.NANOSECONDS);
```
bohhyang, line 1469: nit: we should also have a helper function checkAndSchedule (or checkShutdownAndSchedule), and call it here. There are other usage of .schedule in startRpcStreamLocal and testRetryTask. When it won't hurt to guard them, People unfamiliar with the code shouldn't need to look into the code and wonder why they don't need to be guarded, we should just guard all usages to be 100% safe.

bohhyang (changes requested):

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: it's not about xds adaptor but the ads stream.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
ZoabKapoor: Makes sense, it's just a little extra code to make sure we are protecting every surface area

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
ZoabKapoor: Makes sense! For line 312, would it be better to just always run it outside the executor? I guess we would want a new thread to make it async but we could just always spawn a new thread for the shutdown 🤔

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
ZoabKapoor: Thanks! I'm going to get rid of this actually and make a helper `checkShutdownAndSchedule` as @bohhyang suggested 🎉

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: it's the executor is not shutdown, no need to spawn a new thread right? But in general, yeah, we need to make sure shutdown code is always executed and need to be async (don't want to be blocking in case ads stream hangs somehow).

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
ZoabKapoor: One challenge with `checkShutdownAndSchedule` is that `_executorService.schedule` returns a `ScheduledFuture<V>` and it's not clear what we should return when the `_executorService` is shutdown.

I have 
```
  private ScheduledFuture<?> checkShutdownAndSchedule(Runnable runnable, int delay, TimeUnit unit) {
    if (_executorService.isShutdown())
    {
      _log.warn("Attempting to schedule a task after _executorService was shutdown, will do nothing");
      return null;
    }

    return _executorService.schedule(runnable, delay, unit);
  }
```

but I worry that passing null to the caller may lead to a `NullPointerException` somewhere

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -316,6 +319,27 @@ public void shutdown()
     });
   }
 
+  private ScheduledFuture<?> checkShutdownAndSchedule(Runnable runnable, long delay, TimeUnit unit) {
+    if (_executorService.isShutdown())
+    {
+      _log.warn("Attempting to schedule a task after _executorService was shutdown, will do nothing");
+      return null;
```
ZoabKapoor, line 326: Reviewers: is this safe? Unlike `execute()`, `executorservice.schedule` returns a `ScheduledFuture<?>` and so we need to return **something** here.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
ZoabKapoor: Makes sense! I've renamed the variable to be clearer, too

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
ZoabKapoor: Update: I checked all the places `checkShutdownAndSchedule` is called and the variable that the result is assigned to is always checked for null, so we should actually be good here

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
ZoabKapoor: https://github.com/linkedin/rest.li/pull/1065#discussion_r2093823724

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
ZoabKapoor: Sounds good; I've created a new executor for this just for simplicity, but let me know if it doesn't make sense

d2/src/test/java/com/linkedin/d2/xds/TestXdsClientImpl.java
```diff
@@ -982,8 +1002,10 @@ private static final class XdsClientImplFixture
         setResourceVersions(useGlobCollections);
       }
 
+      _executorService = spy(Executors.newScheduledThreadPool(1));
```
ZoabKapoor, line 1005: FYI for reviewers: I changed the executor from `Executors.newSingleThreadScheduledExecutor(new NamedThreadFactory("test executor")` to `Executors.newScheduledThreadPool(1)` because `spy` didn't work on the former executor

```
Mockito cannot mock this class: class java.util.concurrent.Executors$DelegatedScheduledExecutorService
Mockito can only mock visible & non-final classes.
```

As far as I'm aware, these should be equivalent

bohhyang (approved):
great. Thank you for fixing this issue holistically.

brycezhongqing:
Thanks for this issue fixed!

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -305,9 +306,11 @@ void startRpcStreamLocal()
   @Override
   public void shutdown()
   {
-    _executorService.execute(() ->
+    // create a new executor instead of using the _executorService because this shutdown should run
+    // regardless of whether _executorService is available.
+    Executors.newSingleThreadExecutor().execute(() ->
```
bohhyang, line 311: actually, this new executor needs to be shutdown too. @ZoabKapoor

ZoabKapoor: Oh no! When should it be shutdown, maybe immediately after the executed command returns? 

Though from the docs it looks like `execute` is [async](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Executor.html#execute-java.lang.Runnable-) and so we'd have to explicitly wait for it to return

PR #1066, opened 2025-05-19
Release rest.li version 29.69.1

ZoabKapoor:
This releases the changes from #1065

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-version=29.69.0
+version=29.69.1
 group=com.linkedin.pegasus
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=29.69.0
+version=29.69.1
```
ZoabKapoor, line 1: We could bump the minor version instead of the patch, but since this is just a bug fix and not adding any new APIs I thought it could be a patch bump. @reviewers Let me know if you disagree

bohhyang (approved):

PR #1067, opened 2025-05-20
Skip node creation for post commit

abhics1002:
Skip node creation for post commit

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.69.2] - 2025-05-20
+
 ## [29.69.1] - 2025-05-19
 - Don't execute or schedule new tasks in XdsClientImpl's ScheduledExecutorService once the XdsClient has been shut down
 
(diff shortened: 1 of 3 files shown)


brycezhongqing (approved):

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.69.2] - 2025-05-20
+- Skipping ZK node creation for Raw D2 Client for post commit
```
bohhyang, line 18: hmm, there should be an empty line between here. Did you use the /scripts/update-changelog to update this file?

d2/src/main/java/com/linkedin/d2/discovery/stores/zk/ZooKeeperEphemeralStore.java
```diff
@@ -476,6 +476,12 @@ private void setRawClientTrackingNode()
     String rawD2ClientTrackingPath = D2Utils.getRawClientTrackingPath();
     try
     {
+      if (rawD2ClientTrackingPath.contains("post-commit"))
```
bohhyang, line 479: adding delimiters around it "**-**post-commit**-**" feels better just in case some app names accidentally match it, for example "blogpost-committee"

bohhyang, line 479: also would be good to add a doc about why this check is needed with a sample path like "export-content-data-multiproduct-post-commit-mpdep-i001-workspace-credential-store_7c08d85e507e41cab5598191ca6d459ccbce3f89-credential-store-credential-store-integ-test"

PR #1068, opened 2025-05-21
Fix the shutdown issue for XdsClientImpl and add more comments for Raw D2 Client usages tracking

brycezhongqing:
As title


bohhyang (approved):
thanks for fixing it. please see if @zkapoor is on the same page

PR #1072, opened 2025-05-28
Handle exceptions thrown from CustomAffinityRoutingURIProvider

shivamgupta1:
Summary
---------
Add `SafeCustomAffinityRoutingURIProviderDecorator` to safely call user-supplied `CustomAffinityRoutingURIProvider` implementations. This is needed to avoid deadlocked requests since any exceptions thrown from here aren't caught and they result in the executor thread being killed.

Testing Done
---------
Added UTs

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/simple/SafeCustomAffinityRoutingURIProviderDecorator.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/simple/SafeCustomAffinityRoutingURIProviderDecorator.java
@@ -1,0 +1,84 @@
+/*
+   Copyright (c) 2022 LinkedIn Corp.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


d2/src/main/java/com/linkedin/d2/balancer/simple/SafeCustomAffinityRoutingURIProviderDecorator.java
```diff
@@ -0,0 +1,84 @@
...
+  public Optional<URI> getTargetHostURI(String clusterName) {
+    if (_delegate == null) {
+      return Optional.empty();
+    }
+    try {
+      // Attempt to get the target host URI from the delegate
+      return _delegate.getTargetHostURI(clusterName);
+    } catch (RuntimeException ex) {
+      RATE_LIMITED_LOGGER.error("Error getting target host URI for cluster: " + clusterName, ex);
+      return Optional.empty();
```
golthitarun, line 68: Would returning empty cause any issues where this method is being called ? or its already handled there ?

shivamgupta1: It'll fallback to the usual d2 routing in that case: https://github.com/linkedin/rest.li/blob/fc2d8eedc1564165da8a4d1ab9da897e743dd305/d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancer.java#L275

PR #1074, opened 2025-05-31
fixing child count logic

abhics1002:
changing the way child count is being determined.


brycezhongqing (approved):
LGTM. Could you please elaborate more on the description? Like the context, why we do such change

gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=29.69.6
+version=29.69.7-rc.1
```
brycezhongqing: Please correct this to right version

bohhyang:
LGTM, just the version number as Zhongqing mentioned above

d2/src/main/java/com/linkedin/d2/discovery/util/D2Utils.java
```diff
@@ -36,7 +36,7 @@ public class D2Utils
   ).collect(Collectors.toSet());
 
   // Keeping the max threshold to 10K, this would ensure that we accidentally won't create more than max ZK tracking nodes.
-  public static final int RAW_D2_CLIENT_MAX_TRACKING_NODE = 10000;
+  public static final int RAW_D2_CLIENT_MAX_TRACKING_NODE = 1000;
```
bohhyang, line 39: actually the RAW_D2_CLIENT_BASE_PATH could be either /d2/..., or /shared/..., or /kafka/..., varies with the fabric.

d2/src/main/java/com/linkedin/d2/discovery/stores/zk/ZooKeeperEphemeralStore.java
```diff
@@ -491,13 +491,14 @@ private void setRawClientTrackingNode()
         return;
       }
 
-      List<String> childNodes = _zk.getChildren(D2Utils.getRawClientTrackingBasePath(), false);
+      Stat nodeStat = _zk.exists(D2Utils.getRawClientTrackingBasePath(), false);
```
bohhyang, line 494: actually the base path could be either /d2/..., or /shared/..., or /kafka/..., varies with the fabric. We need to check the path existence one by one.

btw, what value will nodeStat be if the path not exists? also what value will nodeStat.getNumChildren() be in that case?

bohhyang (changes requested):

d2/src/main/java/com/linkedin/d2/discovery/stores/zk/ZooKeeperEphemeralStore.java
abhics1002: > actually the base path could be either /d2/..., or /shared/..., or /kafka/..., varies with the fabric. We need to the path existence one by one.
i can see that tracking node is getting created correctly at /kafka/d2/rawD2ClientBuilders in corp-lva1, so we needn't check each path.

abhics1002: > btw, what value will nodeStat be if the path not exists? also what value will nodeStat.getNumChildren() be in that case?
nodeStat will be null in this case.

bohhyang: ok, false alarm. Although the zk ensemble name is different, the base path is the same in each ensemble.

PR #1075, opened 2025-06-06
Add fields to D2Uri which support IP-for-hostname substitution in nodeUri

ZoabKapoor:
This adds fields to the proto definition of D2Uri which specify the hostnames, IP addresses, and whether these IPs are in an app's cert. 

We are doing this for a couple of reasons:

1. These fields can be used by either the service discovery server (which we are implementing presently) or the service discovery client (which is not currently planned or implemented, but could be in the future) to decouple service discovery from DNS by making routing decisions directly on the IP address instead of the hostname or nodeUri. 
2. This D2URI schema is duplicated internally and we are adding these new fields to the internal version, so we must keep it in sync here or risk proto deserialization issues

### DIFF
--- a/d2/src/main/proto/XdsD2.proto
+++ b/d2/src/main/proto/XdsD2.proto
@@ -133,6 +133,21 @@ message D2URI {
   // comes from the announcing server (which sets a UUID for each announcement) and for Zookeeper,
   // this will be the full path to the d2URI. eg: /d2/uris/FooCluster/lor1-app000-xxxxx
   string tracing_id = 7;
+
+  // fields 8-12 are added to support decoupling Service Discovery from the DNS stack -- if provided, the IPv4 or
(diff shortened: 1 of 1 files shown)


bohhyang (approved):

ZoabKapoor:
> > "service discovery client (which is not implemented, but could be in the future)  to decouple service discovery from DNS by substituting the IP for the hostname in the nodeUri. "
> 
> what needs to be implemented on the client side? Once the observer replaced the hostname in the nodeUri by ip, clients current code (calling dns method with an address that's already an IP) just works as no op, doesn't it? You mean they may be used by future projects if needed? It reads like something is still missing in order to decouple SD from DNS. Maybe clarify in the PR title and the description?

Yes, I meant to say that once the client has these fields it could be modified to directly use them in routing decisions, but I see your point about how the current working makes it seem like we're going to implement this right now! Let me remove it from the title and reword the description

PR #1077, opened 2025-06-23
Configure xds stream retry max backoff time

bohhyang:
## Summary
Current default max retry backoff time in [grpc](https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/ExponentialBackoffPolicy.java#L41) is 2min, which is way too long for INDIS. This change makes it configurable so that we can set it for our need (like 10s, 30s, etc.).

## Testing Done
TODO

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
@@ -95,6 +95,7 @@ public class XdsClientImpl extends XdsClient
   private final boolean _subscribeToUriGlobCollection;
   private final BackoffPolicy.Provider _backoffPolicyProvider = new ExponentialBackoffPolicy.Provider();
   private BackoffPolicy _retryBackoffPolicy;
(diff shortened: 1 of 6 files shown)


d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -73,6 +73,7 @@ public class XdsClientImpl extends XdsClient
   private static final RateLimitedLogger RATE_LIMITED_LOGGER =
       new RateLimitedLogger(_log, TimeUnit.MINUTES.toMillis(1), SystemClock.instance());
   public static final long DEFAULT_READY_TIMEOUT_MILLIS = 2000L;
+  public static final long DEFAULT_MAX_RETRY_BACKOFF_SECS = 30L; // default value for max retry backoff seconds
```
brycezhongqing: seems we are assume this value is always `integer` ?
refer to
```java
Integer maxRetryBackoffSeconds
```

bohhyang: yeah, updated.

brycezhongqing (approved):

PR #1078, opened 2025-07-07
Update FutureCallback.java

inonwir:
We are working on a research project for improving open-source projects, by using the latest accepted answer from Stack Overflow that matched with your code snippet. We found this recommendation for improving your code from [https://stackoverflow.com/questions/2180534](https://stackoverflow.com/questions/2180534). 

Note: Our study is approved by the Institutional Review Board of Mahidol University. You can find the participant information sheet explaining this study [HERE](https://drive.google.com/file/d/1ml5AqrtWQ9pnifTQyTFTcWQmwp6RuPA7/view?usp=sharing).

Type of Recommendation: Add Null Check
Reason:  Prevent NullPointerException
Type: Bug
Influences: Fix method bytesToHexString (edited)
[10:42]

Type of Recommendation: Add safe cancellation support to FutureCallback
Reason: Prevent hanging get() calls and ensure compliance with Future API
Type: Bug Fix
Influences: Implements cancel(), updates get() to throw CancellationException, and adds proper isCancelled() state handling

### DIFF
--- a/pegasus-common/src/main/java/com/linkedin/common/callback/FutureCallback.java
+++ b/pegasus-common/src/main/java/com/linkedin/common/callback/FutureCallback.java
@@ -17,6 +17,7 @@
 /* $Id$ */
 package com.linkedin.common.callback;
 
+import java.util.concurrent.CancellationException;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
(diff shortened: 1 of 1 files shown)


zackthehuman:
@inonwir Thanks for the contribution. I have some questions about why this change is necessary. If you look at the implementation of `FutureCallback`, it was intentionally designed to not support cancellation. Since it is an adapter that makes a `Callback` work as a `Future`, and `Callback` doesn't have any kind of cancellation semantics, neither would a `FutureCallback`.

What does it mean to "cancel" a callback? It doesn't make sense.

inonwir:
@zackthehuman 
Thanks for pointing this out! You’re right — this doesn’t cancel the underlying Callback work. The change only lets callers cancel their interest in the Future: cancel() marks the Future as cancelled, unblocks any waiting threads, and get() throws CancellationException. This avoids hanging get() calls and keeps FutureCallback consistent with the standard Future API, without breaking existing behavior.

ChrisCarini:

PR #1088, opened 2025-08-06
Add xds client metrics to count requests, IRV, and responses

bohhyang:
## Summary
See https://github.com/linkedin-multiproduct/container/pull/1855

Also added a time check so that after Oct 1st 2025 9AM PST, zk-reading raw d2 clients will start crashing. 

## Test Done
See https://github.com/linkedin-multiproduct/container/pull/1855


d2/src/main/java/com/linkedin/d2/jmx/XdsClientJmx.java
```diff
@@ -128,6 +150,26 @@ public void incrementReconnectionCount()
     _reconnectionCount.incrementAndGet();
   }
 
+  public void incrementRequestSentCount()
+  {
+    _resquestSentCount.incrementAndGet();
+  }
+
+  public void addToIrvSentCount(int addend)
```
PapaCharlie: `addend` isn't a word, maybe `delta` is better? It's the name of the param of `AtomicLong#addAndGet`

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -1162,6 +1162,7 @@ final class RpcRetryTask implements Runnable
     @Override
     public void run()
     {
+      _xdsClientJmx.resetIrvSentCount();
```
PapaCharlie: Why reset the send count? When showing a counter, inGraphs/Observe will show the diff between the previous minute and the current minute, so you never need to reset counters

bohhyang: just worried that IRVs in apps like RRE could be huge, like 800k, etc.. Over time if the app is long running for months or even longer, the count may have a tiny risk to overflow theoretically if it's never reset. Maybe it's overthinking? (I know the client needs to reconnect probably millions of times to overflow, but who knows)

d2/src/main/java/com/linkedin/d2/jmx/XdsClientJmx.java
bohhyang: hmm, looks like it's [a word](https://www.google.com/search?q=addend+meaning&rlz=1C5GCCM_en&oq=addend+meaning&gs_lcrp=EgZjaHJvbWUyDggAEEUYORhGGPkBGIAEMgcIARAAGIAEMgcIAhAAGIAEMgcIAxAAGIAEMgcIBBAAGIAEMgcIBRAAGIAEMgcIBhAAGIAEMgcIBxAAGIAEMgcICBAAGIAEMgcICRAAGIAE0gEIMjQ2OWowajeoAgiwAgHxBcEwnYguXzDF8QXBMJ2ILl8wxQ&sourceid=chrome&ie=UTF-8). But delta is fine too.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
PapaCharlie: This is a long, so it would take at least thousands of years to overflow! And even so, I'm almost certain inGraphs will render it properly. Plus, remember that metrics are only sampled once per minute. If the reconnect happens in less than a minute, you will lose some data. I think you can skip resetting

d2/src/main/java/com/linkedin/d2/jmx/XdsClientJmx.java
PapaCharlie: TIL! Well then it's up to you haha, however I think this might raise some eyebrows 🤷

bohhyang: yeah, updated to delta.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: makes sense. Removed it, thanks!

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
```diff
@@ -250,10 +252,21 @@ public D2Client build()
...
-          + "See instructions at go/onboardindis.\n"
-          + "Using in stack: {}", stackTrace);
+      //After Oct 1st 9AM PST, throw exception to hard fail non INDIS raw d2 client
+      ZonedDateTime zdt = ZonedDateTime.of(2025, 10, 1, 9, 0, 0, 0,
+          ZoneId.of("America/Los_Angeles"));
+      if (Instant.now().isAfter(zdt.toInstant()))
+      {
+        throw new IllegalStateException("Creating Zookeeper-reading raw D2 Client in app-custom code is prohibited. "
+            + "See instructions at go/onboardindis to find the code owner and migrate to INDIS.\n"
+            + "Creating in stack: " + stackTrace);
```
PapaCharlie: Strictly speaking, the stacktrace will be included by the exception itself. You don't actually need to include it here.
```suggestion
        throw new IllegalStateException("Creating Zookeeper-reading raw D2 Client in app-custom code is prohibited. "
            + "See instructions at go/onboardindis to find the code owner and migrate to INDIS.")
```

d2/src/main/java/com/linkedin/d2/jmx/XdsClientJmx.java
```diff
@@ -155,9 +155,9 @@ public void incrementRequestSentCount()
...
   }
 
-  public void addToIrvSentCount(int addend)
+  public void addToIrvSentCount(int delta)
   {
-    _irvSentCount.addAndGet(addend);
+    _irvSentCount.addAndGet(delta);
   }
 
   public void resetIrvSentCount()
```
PapaCharlie: This should be deleted as well right?

d2/src/main/java/com/linkedin/d2/jmx/XdsClientJmx.java
```diff
@@ -128,6 +150,26 @@ public void incrementReconnectionCount()
     _reconnectionCount.incrementAndGet();
   }
 
+  public void incrementRequestSentCount()
+  {
+    _resquestSentCount.incrementAndGet();
+  }
+
+  public void addToIrvSentCount(int delta)
```
PapaCharlie, line 158: Hude nit, but this should match the other function names and be called `incrementIrvSentCount`

PapaCharlie (changes requested):

d2/src/main/java/com/linkedin/d2/jmx/XdsClientJmx.java
bohhyang: not really, incrementXXX methods only increment the counter by 1.

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
bohhyang: you're totally right. Removed this after all as we discussed.

d2/src/main/java/com/linkedin/d2/jmx/XdsClientJmx.java
bohhyang: whoops, removed.

d2/src/main/java/com/linkedin/d2/jmx/XdsClientJmx.java
PapaCharlie: You can increment by any delta, the default happens to be one. It's a weird signature, but it's not a deal breaker

PR #1089, opened 2025-08-06
Introduce new IPv6 aware Pick First policy

PapaCharlie:
This new policy, registered under `ipv6_aware_random_pick_first`, is smart enough to only shuffle the addresses within their group. In other words, instead of shuffling IPv4 addresses in between IPv6 addresses, it shuffles the IPv6 and IPv4 addresses independently, and an additional property determines whether IPv4 addresses should appear before IPv6 addresses (defaults to IPv6 first).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.71.1] - 2025-08-07
+- Introduce new IPv6 aware Pick First policy
+
 ## [29.71.0] - 2025-08-04
 - Add isIndisOnly to LoadBalancerWithFacilitiesFactory
 
@@ -5858,7 +5861,8 @@ patch operations can re-use these classes for generating patch messages.
 
 ## [0.14.1]
 
-[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.71.0...master
+[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.71.1...master
+[29.71.1]: https://github.com/linkedin/rest.li/compare/v29.71.0...v29.71.1
 [29.71.0]: https://github.com/linkedin/rest.li/compare/v29.70.2...v29.71.0
 [29.70.2]: https://github.com/linkedin/rest.li/compare/v29.70.1...v29.70.2
 [29.70.1]: https://github.com/linkedin/rest.li/compare/v29.70.0...v29.70.1
--- a/d2/src/main/java/com/linkedin/d2/xds/IPv6AwarePickFirstLoadBalancer.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/IPv6AwarePickFirstLoadBalancer.java
@@ -1,0 +1,158 @@
+package com.linkedin.d2.xds;
+
+import com.google.common.collect.ImmutableList;
+import io.grpc.EquivalentAddressGroup;
+import io.grpc.LoadBalancer;
+import io.grpc.LoadBalancerProvider;
+import io.grpc.LoadBalancerRegistry;
+import io.grpc.NameResolver.ConfigOrError;
+import io.grpc.Status;
+import io.grpc.internal.JsonUtil;
+import io.grpc.internal.PickFirstLoadBalancerProvider;
+import java.net.Inet6Address;
+import java.net.InetSocketAddress;
+import java.net.SocketAddress;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+public class IPv6AwarePickFirstLoadBalancer extends LoadBalancer
+{
+  public static final String POLICY_NAME = "ipv6_aware_random_pick_first";
+  public static final String IPV4_FIRST = "ipv4_first";
+
+  private final LoadBalancer _pickFirst;
+
+  IPv6AwarePickFirstLoadBalancer(Helper helper)
+  {
+    _pickFirst = new PickFirstLoadBalancerProvider().newLoadBalancer(helper);
+  }
+
+  @Override
+  public boolean acceptResolvedAddresses(ResolvedAddresses addresses)
+  {
+    List<EquivalentAddressGroup> ipv6EAGs = new ArrayList<>();
+    List<EquivalentAddressGroup> ipv4EAGs = new ArrayList<>();
+    for (EquivalentAddressGroup eag : addresses.getAddresses())
+    {
+      (hasIPv6Address(eag) ? ipv6EAGs : ipv4EAGs).add(eag);
+    }
+
+    Collections.shuffle(ipv6EAGs);
+    Collections.shuffle(ipv4EAGs);
+
+    List<EquivalentAddressGroup> shuffledEAGs = new ArrayList<>();
+    if (addresses.getLoadBalancingPolicyConfig() instanceof Config
+        && ((Config) addresses.getLoadBalancingPolicyConfig())._ipv4First)
+    {
+      shuffledEAGs.addAll(ipv4EAGs);
+      shuffledEAGs.addAll(ipv6EAGs);
+    }
+    else
+    {
+      shuffledEAGs.addAll(ipv6EAGs);
+      shuffledEAGs.addAll(ipv4EAGs);
+    }
+
+    return _pickFirst.acceptResolvedAddresses(
+        ResolvedAddresses.newBuilder()
+            .setAddresses(shuffledEAGs)
+            .setAttributes(addresses.getAttributes())
+            .setLoadBalancingPolicyConfig(addresses.getLoadBalancingPolicyConfig())
+            .build()
+    );
+  }
+
+  private static boolean hasIPv6Address(EquivalentAddressGroup eag)
+  {
+    for (SocketAddress address : eag.getAddresses())
+    {
+      if (!(address instanceof InetSocketAddress))
+      {
+        continue;
+      }
+      if (((InetSocketAddress) address).getAddress() instanceof Inet6Address)
+      {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  @Override
+  public void handleNameResolutionError(Status error)
+  {
+    _pickFirst.handleNameResolutionError(error);
+  }
+
+  @Override
+  public void shutdown()
+  {
+    _pickFirst.shutdown();
+  }
+
+  @Override
+  public void requestConnection()
+  {
+    _pickFirst.requestConnection();
+  }
+
+  private static class Config
+  {
+    private final boolean _ipv4First;
+
+    private Config(@Nullable Boolean ipv4First)
+    {
+      _ipv4First = (ipv4First != null) ? ipv4First : false;
+    }
+  }
+
+  private static final class Provider extends LoadBalancerProvider
+  {
+    static
+    {
+      LoadBalancerRegistry.getDefaultRegistry().register(new Provider());
+    }
+
+    @Override
+    public boolean isAvailable()
+    {
+      return true;
+    }
+
+    @Override
+    public int getPriority()
+    {
+      return 5;
+    }
+
+    @Override
+    public String getPolicyName()
+    {
+      return POLICY_NAME;
+    }
+
+    @Override
+    public LoadBalancer newLoadBalancer(LoadBalancer.Helper helper)
+    {
+      return new IPv6AwarePickFirstLoadBalancer(helper);
+    }
+
+    @Override
+    public ConfigOrError parseLoadBalancingPolicyConfig(Map<String, ?> rawLoadBalancingPolicyConfig)
+    {
+      try
+      {
+        return ConfigOrError.fromConfig(new Config(JsonUtil.getBoolean(rawLoadBalancingPolicyConfig, IPV4_FIRST)));
+      }
+      catch (RuntimeException e)
+      {
+        return ConfigOrError.fromError(
+            Status.UNAVAILABLE.withCause(e).withDescription(
+                "Failed parsing configuration for " + getPolicyName()));
+      }
+    }
+  }
+}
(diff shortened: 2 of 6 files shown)


d2/src/main/java/com/linkedin/d2/xds/IPv6AwarePickFirstLoadBalancer.java
```diff
@@ -0,0 +1,158 @@
...
+    {
+      (hasIPv6Address(eag) ? ipv6EAGs : ipv4EAGs).add(eag);
+    }
+
+    Collections.shuffle(ipv6EAGs);
+    Collections.shuffle(ipv4EAGs);
+
+    List<EquivalentAddressGroup> shuffledEAGs = new ArrayList<>();
+    if (addresses.getLoadBalancingPolicyConfig() instanceof Config
+        && ((Config) addresses.getLoadBalancingPolicyConfig())._ipv4First)
```
TylerHorth: I don't think we should be changing the relative ordering of IPv4 and IPv6 addresses, since these will already be ordered based on the users preference (i.e. preferIPv6Addresses / preferIPv4Stack) and I don't think we should be trying to reproduce the same logic. 

I think we can keep the previous ordering by iterating over the existing list and replacing elements based on IP version. i.e.:

```java
int ipv4Index = 0;
int ipv6Index = 0;

List<EquivalentAddressGroup> shuffledEAGs = new ArrayList<>(addresses.getAddresses());

for (int i = 0; i < shuffledEAGs.size(); i++) {
    EquivalentAddressGroup prevEag = shuffledEAGs.get(i);
    if (hasIPv6Address(prevEag)) {
        shuffledEAGs.set(i, ipv6EAGs.get(ipv6Index++));
    } else {
        shuffledEAGs.set(i, ipv4EAGs.get(ipv4Index++));
    }
}
```

PapaCharlie: Not bad! That's pretty clever, I like it

d2/src/main/java/com/linkedin/d2/xds/XdsChannelFactory.java
```diff
@@ -16,11 +16,28 @@
...
+import io.grpc.internal.PickFirstLoadBalancerProvider;
 import io.grpc.netty.shaded.io.grpc.netty.NettyChannelBuilder;
 import io.grpc.netty.shaded.io.netty.handler.ssl.SslContext;
+import java.net.Inet6Address;
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.net.SocketAddress;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
```
TylerHorth: What's with all the new imports?

TylerHorth (approved):
Looks good to me besides the random imports.

d2/src/main/java/com/linkedin/d2/xds/XdsChannelFactory.java
ZoabKapoor: I'd guess @PapaCharlie maybe directly implemented this in the `XdsChannelFactory` before creating a new class -- doing an `Optimize Imports` should get rid of these

d2/src/main/java/com/linkedin/d2/xds/IPv6AwarePickFirstLoadBalancer.java
```diff
@@ -0,0 +1,162 @@
...
+ * desired effect of random pick_first load balancing.
+ */
+public class IPv6AwarePickFirstLoadBalancer extends LoadBalancer
+{
+  static
+  {
+    LoadBalancerRegistry.getDefaultRegistry().register(new Provider());
+  }
+
+  public static final String POLICY_NAME = "ipv6_aware_random_pick_first";
```
ZoabKapoor: Nit: I think the policy names `PICK_FIRST` and `RANDOM_PICK_FIRST` we use today are all caps, do we want to match that here?

d2/src/main/java/com/linkedin/d2/xds/IPv6AwarePickFirstLoadBalancer.java
```diff
@@ -0,0 +1,162 @@
...
+    return _delegate.acceptResolvedAddresses(resolvedAddresses.toBuilder()
+        .setAddresses(ipAwareShuffle(resolvedAddresses.getAddresses()))
+        .build());
+  }
+
+  /**
+   * Shuffles the given addresses such that the original interleaving of IPv4 and IPv6 addresses is respected. For
+   * example, the following input [IPv6a, IPv4a, IPv4b, IPv6b, IPv4c] could be shuffled into the following output
+   * [IPv6b, IPv4c, IPv4b, IPv6a, IPv4a]. The IPs were shuffled relative to other IPs of the same version, but the
+   * original interleaving is the same.
```
ZoabKapoor, line 71: This is super clean imo, thanks for the suggestion @TylerHorth !

d2/src/main/java/com/linkedin/d2/xds/IPv6AwarePickFirstLoadBalancer.java
```diff
@@ -0,0 +1,162 @@
...
+ * IPv6 addresses, only IPv4 addresses, IPv6 addresses followed by IPv4 addresses, IPv4 addresses followed by IPv6
+ * addresses, and everything in between. This load balancer shuffles the given addresses in a way that respects the
+ * original interleaving of address types, such that the client's preference is respected, while still achieving the
+ * desired effect of random pick_first load balancing.
+ */
+public class IPv6AwarePickFirstLoadBalancer extends LoadBalancer
+{
+  static
+  {
+    LoadBalancerRegistry.getDefaultRegistry().register(new Provider());
```
ZoabKapoor, line 44: The only thing I'm worried about on this implementation is whether we can correctly select and apply config from the D2ClientFactory in container. Should we try to snapshot this and test it with a real app before merging?

d2/src/main/java/com/linkedin/d2/xds/IPv6AwarePickFirstLoadBalancer.java
```diff
@@ -0,0 +1,162 @@
...
+
+    List<EquivalentAddressGroup> shuffledEAGs = new ArrayList<>(addresses);
+
+    int ipv4Index = 0;
+    int ipv6Index = 0;
+    for (int i = 0; i < shuffledEAGs.size(); i++)
+    {
+      if (hasIPv6Address(shuffledEAGs.get(i)))
+      {
+        shuffledEAGs.set(i, ipv6EAGs.get(ipv6Index++));
```
ZoabKapoor: Nit: I suspect it's more idiomatic to do this with `ListIterator::next` instead of `ipv6EAGs.get(ipv6Index++)`, so you don't have to do the index tracking in ipv6EAGs etc yourself

d2/src/main/java/com/linkedin/d2/xds/IPv6AwarePickFirstLoadBalancer.java
```diff
@@ -0,0 +1,172 @@
...
+    @Override
+    public boolean isAvailable()
+    {
+      return true;
+    }
+
+    @Override
+    public int getPriority()
+    {
+      return 5;
```
ZoabKapoor, line 153: Curious: what does the priority 5 represent here?

d2/src/test/java/com/linkedin/d2/xds/IPv6AwarePickFirstLoadBalancerTest.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.concurrent.ThreadLocalRandom;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Mockito;
+import org.testng.annotations.Test;
+
+import static com.linkedin.d2.xds.IPv6AwarePickFirstLoadBalancer.*;
+import static org.mockito.Mockito.verify;
+import static org.testng.Assert.*;
+
+public class IPv6AwarePickFirstLoadBalancerTest
```
ZoabKapoor, line 25: Should we explicitly test a case of `contiguous block of ipv4s followed by contiguous block of ipv6s` to ensure the LB behaves right in that scenario?

ZoabKapoor (changes requested):
Nice, I think this will help a lot! Added a couple of questions (the only blocker is if we can test the config with a real app) but the approach overall lgtm

d2/src/main/java/com/linkedin/d2/xds/XdsChannelFactory.java
PapaCharlie: Yep those are fixed

d2/src/main/java/com/linkedin/d2/xds/IPv6AwarePickFirstLoadBalancer.java
PapaCharlie: Those are different, they're members of an enum in `container`. This one is just a standalone variable name

d2/src/main/java/com/linkedin/d2/xds/IPv6AwarePickFirstLoadBalancer.java
PapaCharlie: If we do it like this:
```
  /**
   * Specifies a set of policy parameters to pass into {@link com.linkedin.d2.xds.XdsChannelFactory}.
   */
  private enum XdsChannelLoadBalancingPolicy {
    /**
     * Uses the "pick_first" policy.
     */
    PICK_FIRST("pick_first", null),
    /**
     * When passed in with the "pick_first" load balancing policy, setting the shuffleAddressList property to
     * {@code true} shuffles the addresses it resolves from DNS instead of using the default order. Only applicable
     * to gRPC-java versions after 1.59. Noops to standard "pick_first" otherwise.
     */
    RANDOM_PICK_FIRST("pick_first", Collections.singletonMap("shuffleAddressList", true)),
    
    IPV6_AWARE_RANDOM_PICK_FIRST(IPv6AwarePickFirstLoadBalancer.POLICY_NAME, null);

    private final String _policyName;
    private final Map<String, ?> _config;

    XdsChannelLoadBalancingPolicy(String policyName, Map<String, ?> config) {
      _policyName = policyName;
      _config = config;
    }
  }
```
Then it's guaranteed to be loaded if the config is `IPV6_AWARE_RANDOM_PICK_FIRST`. That's *specifically* why I defined it like this

d2/src/main/java/com/linkedin/d2/xds/IPv6AwarePickFirstLoadBalancer.java
PapaCharlie: I think it doesn't even need to be a `ListIterator`, a normal `Iterator` will work here fine as well. Udpated

d2/src/main/java/com/linkedin/d2/xds/IPv6AwarePickFirstLoadBalancer.java
PapaCharlie: 🤷 
It's the priority `PickFirstLoadBalancer` uses. It's used if there are two policies with the same name, where the policy with the higher priority will be used. I guess if you wanted to replace some built-in policy like `pick_first` across the board or something.

PapaCharlie: Added a comment

d2/src/test/java/com/linkedin/d2/xds/IPv6AwarePickFirstLoadBalancerTest.java
PapaCharlie: I think this is functionally the same, in fact it's more thorough. This test that it handles *any* interleaving.

d2/src/main/java/com/linkedin/d2/xds/IPv6AwarePickFirstLoadBalancer.java
ZoabKapoor: Gotcha, this should be fine then

d2/src/test/java/com/linkedin/d2/xds/IPv6AwarePickFirstLoadBalancerTest.java
ZoabKapoor: It definitely is, but from a readability perspective for future maintainers I'm wondering if it would be helpful to have a test that exactly tests the scenario we saw of `N ipv6s + N ipv4s` and making sure they're not interleaved.

It's not a strict necessity I guess but it might make it more clear to someone reading the tests to understand **why** this policy was created

d2/src/main/java/com/linkedin/d2/xds/IPv6AwarePickFirstLoadBalancer.java
ZoabKapoor: Gotcha, so the thought is that we don't need to snapshot test it and can instead just pick up the new version when it's published? 👌

gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=29.72.0
+version=29.72.1
```
ZoabKapoor, line 1: Is the change big enough to merit a new minor version?

PapaCharlie: It doesn't technically introduce any behavior changes since it's not on by default. Not *everything* needs a minor version!

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.72.1] - 2025-08-07
```
bohhyang, line 17: The implementation is good, but the version is wrong. It has to bump the minor version so that the restli version will be required to bump when the corresponding container version that uses this code is bumped in an app (otherwise it will be a build error).

The container code change PR should be created at the same time and be tested along with restli snapshot in an app, like indis-canary to verify the real behavior (the resolved shuffled addresses is as expected, etc.). Since releasing bad version of restli/container is painful and expensive to clean up in adopted apps, we need to be careful and verify thoroughly before release.

PR #1091, opened 2025-08-08
Add fields from D2URI proto to the pdl to return for debugging with restli-resource-explorer

ZoabKapoor:
the D2URI proto defined in XdsD2.proto has a lot of fields that might be relevant for debuggging D2-related issues, but none of these are surfaced in the legacy PDL. This PR backports these fields into the D2Uri.pdl for ease of debugging when all you have is the PDL

### DIFF
--- a/d2-schemas/src/main/pegasus/com/linkedin/d2/D2Uri.pdl
+++ b/d2-schemas/src/main/pegasus/com/linkedin/d2/D2Uri.pdl
@@ -11,6 +11,23 @@ record D2Uri {
   URI: string
 
   /**
+   * [optional] URI for this machine with the hostname replaced with an ipv4.
+   */
+  ipv4VariantURI: string
+
+  /**
+   * [optional] URI for this machine with the hostname replaced with an ipv4.
+   */
+  ipv6VariantURI: string
+
+  // TODO: add modified_time
+  hostname: string
+  ipv4Address: string
+  ipv6Address: string
+  isIpv4InSan: boolean
+  isIpv6InSan: boolean
+
+  /**
    * The cluster where this URI belongs.
    */
   clusterName: string


d2-schemas/src/main/pegasus/com/linkedin/d2/D2Uri.pdl
```diff
@@ -19,4 +19,44 @@ record D2Uri {
...
+
+  /**
+   * Whether the ipv4 provided is also in the app's TLS cert
+   */
+  isIpv4InSan: optional boolean
+
+  /**
+   * Whether the ipv6 provided is also in the app's TLS cert
+   */
+  isIpv6InSan: optional boolean
```
bohhyang, line 61: should have app specific properties too, some apps set custom values in it. It's also good for everyone to see what's set.

ZoabKapoor: You mean the `UriSpecificProperties`? We can look into it, though it may be tough to translate because the type is `Struct` which is way more flexible than any of the types we're including so far

ZoabKapoor: Looking closer into this, I see that a `Struct` can be arbitrarily nested... for simplicity and also considering nobody has yet requested this data, I would suggest we defer trying to add the `UriSpecificProperties` future work

d2-schemas/src/main/pegasus/com/linkedin/d2/D2Uri.pdl
```diff
@@ -19,4 +19,44 @@ record D2Uri {
...
+
+  /**
+   * URI for this machine with the hostname replaced with an ipv6. Only set if an ipv6 is provided and isIpv6InSan is true
+   */
+  ipv6VariantURI: optional string
+
+  /**
+   * The time this URI was last modified in milliseconds since the epoch
+   */
+  modifiedTime: optional long
```
bohhyang: this needs to be a formatted date time hour sec.

ZoabKapoor: We can do that! The reason I made this a long is because when I looked elsewhere in restli that seems how it handles times: https://github.com/search?q=repo%3Alinkedin%2Frest.li%20pdl%20time&type=code

However, I can also make this an `optional string` and then use https://protobuf.dev/reference/java/api-docs/com/google/protobuf/util/Timestamps.html#toString-com.google.protobuf.Timestamp- to `Convert Timestamp to RFC 3339 date string format`

d2-schemas/src/main/pegasus/com/linkedin/d2/D2Uri.pdl
```diff
@@ -19,4 +19,44 @@ record D2Uri {
...
+
+  /**
+   * URI for this machine with the hostname replaced with an ipv4. Only set if an ipv4 is provided and isIpv4InSan is true.
+   */
+  ipv4VariantURI: optional string
+
+  /**
+   * URI for this machine with the hostname replaced with an ipv6. Only set if an ipv6 is provided and isIpv6InSan is true
+   */
+  ipv6VariantURI: optional string
```
bohhyang, line 31: why we need these? There should be only one uri (ip-sub completed by the observer) which is the only one received and used by the client to send RPCs. Multiple variants will just be confusing.

ZoabKapoor: with the IP substitution change, what URI the client will get depend on whether it connects to the Observer with IPv6 or IPv4, so if it connected with IPv4 it would get the IPv4 variant and if it connected with IPv6 it will get the IPv6 variant

ZoabKapoor: So I don't think there's a single URI that makes sense to provide here

d2-schemas/src/main/pegasus/com/linkedin/d2/D2Uri.pdl
ZoabKapoor: Updated to a `String` and specified in the docs that it's `RFC 3339 date string format` , let me know if this is not what you had in mind!

d2-schemas/src/main/pegasus/com/linkedin/d2/D2Uri.pdl
bohhyang: Oh, yeah, I forgot that RRE getting the result is not equivalent to what real clients will receive. Yes, we need the variants. thanks. 
btw, will one client app be distributed on a mixed ipv6 and v4 machines?

CHANGELOG.md
```diff
@@ -5870,8 +5873,14 @@ patch operations can re-use these classes for generating patch messages.
...
 ## [0.14.1]
 
+<<<<<<< HEAD
 [Unreleased]: https://github.com/linkedin/rest.li/compare/v29.74.0...master
 [29.74.0]: https://github.com/linkedin/rest.li/compare/v29.73.0...v29.74.0
+=======
+[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.74.1...master
+[29.74.1]: https://github.com/linkedin/rest.li/compare/v29.73.1...v29.74.1
+[29.73.1]: https://github.com/linkedin/rest.li/compare/v29.73.0...v29.73.1
+>>>>>>> 7c9c7adab (Update changelog)
```
bohhyang: !! :)

ZoabKapoor: Oh noooo

d2-schemas/src/main/pegasus/com/linkedin/d2/D2Uri.pdl
ZoabKapoor: Like, some hosts for some client app use ipv6 and others use ipv4? I think it should be unlikely (once https://github.com/linkedin/rest.li/pull/1092 is adopted), but I don't know for sure if I can rule it out

PR #1092, opened 2025-08-12
Bump gRPC version to fix IPv6AwarePickFirstLoadBalancer

PapaCharlie:
The signature of `LoadBalancer#acceptResolvedAddresses` was changed from returning a boolean to a `io.grpc.Status`. This meant invoking `IPv6AwarePickFirstLoadBalancer#acceptResolvedAddresses` was invoking the default definition in the parent abstract class instead of the actual overridden definition in `IPv6AwarePickFirstLoadBalancer`. Strangely, it did not cause any runtime exceptions such as NoSuchMethodError.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,6 +14,8 @@
    limitations under the License.
 */
 
+//import org.gradle.api.attributes.java.TargetJvmEnvironment;
+
 buildscript {
   repositories {
     mavenCentral()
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -14,6 +14,8 @@
    limitations under the License.
 */
 
+//import org.gradle.api.attributes.java.TargetJvmEnvironment;
```
ZoabKapoor: Nit: remove this?

build.gradle
```diff
@@ -434,6 +436,25 @@ subprojects {
     }
     logger.lifecycle "Tests for subproject ${project.name} will be skipped."
   }
+
+  dependencies.constraints {
+    add("compile", "com.google.guava:guava") {
+      attributes {
+        attribute(Attribute.of("org.gradle.jvm.environment", String),"standard-jvm")
+//        attribute(TargetJvmEnvironment.TARGET_JVM_ENVIRONMENT_ATTRIBUTE,
```
ZoabKapoor: Nit: remove these commented out lines? Is the attribute above needed too?

build.gradle
```diff
@@ -434,6 +436,25 @@ subprojects {
...
+//                objects.named(TargetJvmEnvironment.STANDARD_JVM))
+      }
+    }
+  }
+
+  configurations.configureEach {
```
ZoabKapoor: If we need to keep this in the build.gradle we might want to add comments here and above about what this does and/or why it's needed

ZoabKapoor (changes requested):
Overall LGTM, the only thing would be adding some comments explaining the new gradle fu

build.gradle
PapaCharlie: Done

build.gradle
PapaCharlie: Done

build.gradle
PapaCharlie: Done

PR #1093, opened 2025-08-14
Correct xds client latency metric and SD update receipt event

bohhyang:
## Summary
This PR corrects a few things for calculation xds client latency and the emission of SD update receipt event for e2e propagation latency in Kusto dashboard. It will remove false spikes for these two latencies caused by server updates happened before the client starts its subscription. 
1. When resource modified time is before the client starts subscribing, use the subscribe time.
2. When resource modified time is before the watcher starts watching, skip emitting SD update receipt event (Kusto will blindly associate the server update sent event with the client receipt event based on tracing ID, causing incorrect huge e2e latency on the dashboard. So just skip client receipt event for this case).
3. Reset subscribe/watch time when reconnected to INDIS.
4. Also correct wildcard subscribes to also calculate xds client latency for all resources it receives.

#Test Done
Verified in UTs for various cases:
1) resource modified time < subscribe/watch time
2) resource modified time > subscribe/watch time
3) reset subscribe/watch time

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
@@ -65,7 +65,7 @@ public abstract class XdsClient
      */
     public void onReconnect()
     {
-      resetWatchedAt();
+      resetWatchedAt(); // Reconnected needs to reset the watchedAt time to the current time for tracking purposes.
     }
 
     abstract void onChanged(ResourceUpdate update);
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
@@ -19,6 +19,7 @@ package com.linkedin.d2.xds;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Strings;
 import com.linkedin.d2.jmx.XdsClientJmx;
+import com.linkedin.util.clock.SystemClock;
 import indis.XdsD2;
 import io.grpc.Status;
 import java.util.Arrays;
@@ -37,6 +38,7 @@ public abstract class XdsClient
   public static abstract class ResourceWatcher
   {
     private final ResourceType _type;
+    protected long _watchedAt = SystemClock.instance().currentTimeMillis();
 
     /**
      * Defining a private constructor means only classes that are defined in this file can extend this class. This way,
@@ -61,9 +63,17 @@ public abstract class XdsClient
     /**
      * Called when the resource discovery RPC reestablishes connection.
      */
-    public abstract void onReconnect();
+    public void onReconnect()
+    {
+      resetWatchedAt();
+    }
 
     abstract void onChanged(ResourceUpdate update);
+
+    private void resetWatchedAt()
+    {
+      _watchedAt = SystemClock.instance().currentTimeMillis();
+    }
   }
 
   public static abstract class NodeResourceWatcher extends ResourceWatcher
@@ -117,6 +127,7 @@ public abstract class XdsClient
   public static abstract class WildcardResourceWatcher
   {
     private final ResourceType _type;
+    protected long _watchedAt = SystemClock.instance().currentTimeMillis();
 
     /**
      * Defining a private constructor means only classes that are defined in this file can extend this class (see
@@ -140,7 +151,10 @@ public abstract class XdsClient
     /**
      * Called when the resource discovery RPC reestablishes connection.
      */
-    public abstract void onReconnect();
+    public void onReconnect()
+    {
+      resetWatchedAt();
+    }
 
     /**
      * Called when a resource is added or updated.
@@ -163,6 +177,11 @@ public abstract class XdsClient
     public void onAllResourcesProcessed()
     {
       // do nothing
+    }
+
+    private void resetWatchedAt()
+    {
+      _watchedAt = SystemClock.instance().currentTimeMillis();
     }
   }
 
--- a/d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
+++ b/d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
@@ -21,6 +21,7 @@ import com.google.common.base.Joiner;
 import com.google.common.base.Strings;
 import com.google.common.collect.MapDifference;
 import com.google.common.collect.Maps;
+import com.google.protobuf.util.Timestamps;
 import com.google.rpc.Code;
 import com.linkedin.d2.jmx.NoOpXdsServerMetricsProvider;
 import com.linkedin.d2.jmx.XdsClientJmx;
@@ -709,7 +710,7 @@ public class XdsClientImpl extends XdsClient
 
       if (wildcardSubscriber != null)
       {
-        wildcardSubscriber.onData(entry.getKey(), entry.getValue());
+        wildcardSubscriber.onData(entry.getKey(), entry.getValue(), _serverMetricsProvider);
       }
     }
   }
@@ -843,6 +844,7 @@ public class XdsClientImpl extends XdsClient
     private final XdsClientJmx _xdsClientJmx;
     @Nullable
     private ResourceUpdate _data;
+    private long _subscribedAt = SystemClock.instance().currentTimeMillis();
 
     @VisibleForTesting
     @Nullable
@@ -885,7 +887,7 @@ public class XdsClientImpl extends XdsClient
       // null value guard to avoid overwriting the property with null
       if (data != null && data.isValid())
       {
-        trackServerLatency(data, metricsProvider); // data updated, track xds server latency
+        trackServerLatency(data, _data, metricsProvider, _subscribedAt); // data updated, track xds server latency
         _data = data;
       }
       else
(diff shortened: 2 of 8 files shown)


d2/src/main/java/com/linkedin/d2/jmx/XdsServerMetricsProvider.java
```diff
@@ -2,42 +2,45 @@
 
 
 /**
- * Interface for providing metrics for Xds Server
+ * Interface for providing metrics for Xds Server.
+ * Xds server latency is calculated:
+ * -- When IRV is not enabled, the client will receive all its interested resources every time it (re-)connects,
+ * so the latency should be tracked based on the max of (resource modified time, subscribed time). Caveat in
+ * this is that if some resource is modified and the update is not received for network issues, then the
+ * client reconnects, the latency will be tracked based on the new subscribed time, and the real latency of
```
brycezhongqing, line 10: > the latency will be tracked based on the new subscribed time,

Thanks for catching up this corner case

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
```diff
@@ -338,7 +339,7 @@ public static final class D2URIMapUpdate implements ResourceUpdate
     private final boolean _globCollectionEnabled;
     private final Set<String> _updatedUrisName = new HashSet<>();
     private final Set<String> _removedUrisName = new HashSet<>();
-
+    private Map<String, Boolean> _isStaleModifiedTimes = new HashMap<>();
```
brycezhongqing: Add `final` here

PapaCharlie: Should be `final`

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -1015,13 +967,26 @@ void onRemoval()
         watcher.onChanged(_data);
       }
     }
+
+    void resetSubscribedAt()
```
PapaCharlie: I think this can be upgraded to `private`?

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -843,6 +844,7 @@ static class ResourceSubscriber
     private final XdsClientJmx _xdsClientJmx;
     @Nullable
     private ResourceUpdate _data;
+    private long _subscribedAt = SystemClock.instance().currentTimeMillis();
```
PapaCharlie: Is there a reason you're using `SystemClock` instead of just `System.currentTimeMillis()`? It makes sense to use `SystemClock` if you're actually injecting a mocked `com.linkedin.util.clock.Clock` for testing but otherwise you can just use the built-in one.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -1157,6 +1143,90 @@ private boolean shouldSubscribeUriGlobCollection(ResourceType type)
...
+          isIrvEnabled, isFirstFetch);
+    }
+    else if (resourceUpdate instanceof D2URIMapUpdate)
+    {
+      D2URIMapUpdate update = (D2URIMapUpdate) resourceUpdate;
+      // only track server latency for the updated/new uris in the update
+      Map<String, XdsD2.D2URI> currentUriMap = currentData == null ? Collections.emptyMap()
+          : ((D2URIMapUpdate) currentData).getURIMap();
+      MapDifference<String, XdsD2.D2URI> rawDiff = Maps.difference(update.getURIMap(),
+          currentUriMap == null ? Collections.emptyMap() : currentUriMap);
```
brycezhongqing, line 1235: Is it possible that `update.getURIMap()` is `null` ?

As from the above code , the `getURIMap` sometimes will return `null`

d2/src/test/java/com/linkedin/d2/xds/TestXdsClientImpl.java
```diff
@@ -928,7 +907,177 @@ public void testWatchD2Uri()
...
+
+  private static void verifyLagGreaterThanTenMins(int times, XdsServerMetricsProvider serverMetricsProvider)
+  {
+    verify(serverMetricsProvider, times(times))
+        .trackLatency(longThat(new GreaterThan<>(TEN_MINS)));
+  }
+
+  private static final class PositiveLagLessThan extends ArgumentMatcher<Long>
+  {
+    private static final long deviation = 50; // 50ms deviation to account for clock skew
```
PapaCharlie: This seems suspicious, and usually leads to flakiness in tests. See my other comment on using the `Clock`. Why not inject a `Clock` here to prevent this from being a problem altogether?

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
bohhyang: fixed

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: fixed

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
bohhyang: fixed

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: no, "data != null && data.isValid()" is called before trackServerLatency is called, which makes sure the new update's uriMap is not null.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: Thanks for suggestion. Updated.

d2/src/test/java/com/linkedin/d2/xds/TestXdsClientImpl.java
bohhyang: makes sense. Updated.

d2/src/test/java/com/linkedin/d2/xds/TestXdsToD2PropertiesAdaptor.java
```diff
@@ -380,6 +395,10 @@ private XdsD2.D2URI getD2URI(String clusterName, String uri, long version)
         .putPartitionDesc(0, 42)
         .putPartitionDesc(1, 27)
         .setTracingId(TRACING_ID)
+        .setModifiedTime(Timestamp.newBuilder()
+            .setSeconds(modifiedAt/1000)
+            .setNanos((int) ((modifiedAt % 1000) * 1_000_000))
+            .build())
```
PapaCharlie: There is a better way to create timestamps: `com.google.protobuf.util.Timestamps.fromMillis`

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
```diff
@@ -55,7 +56,7 @@ public class XdsToD2PropertiesAdaptor
   private static final String D2_SERVICE_NODE_PREFIX = "/d2/services/";
   private static final String D2_URI_NODE_PREFIX = "/d2/uris/";
   private static final char PATH_SEPARATOR = '/';
-  private static final String NON_EXISTENT_CLUSTER = "NonExistentCluster";
+  public static final String NON_EXISTENT_CLUSTER = "NonExistentCluster";
```
PapaCharlie: Does this actually get used?

d2/src/test/java/com/linkedin/d2/xds/TestXdsClientImpl.java
```diff
@@ -928,7 +910,225 @@ public void testWatchD2Uri()
...
+      Map<Integer, Double> partitionDesc, long modifiedTimeMillis)
+  {
+    XdsD2.D2URI.Builder builder = XdsD2.D2URI.newBuilder()
+        .setVersion(Long.parseLong(version))
+        .setClusterName(clusterName)
+        .setUri(uri)
+        .setModifiedTime(Timestamp.newBuilder()
+            .setSeconds(modifiedTimeMillis / 1000)
+            .setNanos((int) (modifiedTimeMillis % 1000 * 1_000_000))
+            .build());
```
PapaCharlie: Same, you can use `com.google.protobuf.util.Timestamps.fromMillis`

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
```diff
@@ -547,6 +568,8 @@ static ResourceType fromTypeUrl(String typeUrl)
     }
   }
 
+  public abstract void start();
```
PapaCharlie, line 570: Having both `start` and `startRpcStream` feels dangerous. What happens if you call one but not the other? Are you supposed to call both? I know that `startRpcStream` is a bad name, and it should likely have been just `start` from the beginning, but I think it is too late to change this now.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -1125,13 +1156,39 @@ void onRemoval(String resourceName)
...
       for (WildcardResourceWatcher watcher : _watchers)
       {
         watcher.onAllResourcesProcessed();
       }
     }
+
+    private void resetSubscribedAt()
+    {
+      _subscribedAt = _clock.currentTimeMillis();
+    }
```
PapaCharlie: Should this function also reset `_isFirstFetch`? Otherwise on a reconnect, the `getActiveInitialWaitTimeMillis` won't reflect it

PapaCharlie (changes requested):

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
bohhyang: since "_xdsClientJmx.setXdsClient(this);" has to be done in XdsClientImpl. If we put it into another public method, and if there are more things to be done at startup for xdsClientimpl we put them into different public methods, then external class like XdsToD2PropertiesAdaptor need to call them all. It feels dirty and odd.

Currently only XdsToD2PropertiesAdaptor is calling startRpcStream, I wonder if we could start using start() from now on, and just avoid calling startRpcStream from external class. I can put a comment on startRpcStream to note that.

Or do you have a better way to solve this?

d2/src/test/java/com/linkedin/d2/xds/TestXdsToD2PropertiesAdaptor.java
bohhyang: thanks, updated

d2/src/main/java/com/linkedin/d2/xds/XdsToD2PropertiesAdaptor.java
bohhyang: whoops, was about to use it but ended up not, restored to private.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: ah, good catch, updated, thanks.

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
bohhyang: made startRpcStream package private as we discussed in DM, thanks.

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
```diff
@@ -563,9 +585,10 @@ static ResourceType fromTypeUrl(String typeUrl)
   public abstract void watchAllXdsResources(WildcardResourceWatcher watcher);
 
   /**
+   * NOTE: Should NOT be called directly; instead, call {@link #start()}.
    * Initiates the RPC stream to the xDS server.
    */
-  public abstract void startRpcStream();
+  abstract void startRpcStream();
```
PapaCharlie: Do you even need this in the abstract class anymore at this point?

PapaCharlie (approved):

d2/src/main/java/com/linkedin/d2/xds/XdsClient.java
bohhyang: removed

PapaCharlie (approved):

PR #1097, opened 2025-08-26
Add outlier detection config into D2Cluster

whutwhu:
Outlier detection is a load-balancing feature provided by Envoy clusters. In this PR, we add a PDL for outlier detection based on Envoy’s protocol and integrate it into D2Cluster. For more details, please see [config.cluster.v3.OutlierDetection](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/outlier_detection.proto).

### DIFF
--- a/d2-schemas/src/main/pegasus/com/linkedin/d2/OutlierDetectionConfig.pdl
+++ b/d2-schemas/src/main/pegasus/com/linkedin/d2/OutlierDetectionConfig.pdl
@@ -1,0 +1,173 @@
+namespace com.linkedin.d2
+
+/**
+ * Options for configuring the outlier detection. Only used by gRPC clients.
+ */
+record OutlierDetectionConfig {
+  /** 
+   * The number of consecutive server-side error responses (for HTTP traffic,
+   * 5xx responses; for TCP traffic, connection failures; for Redis, failure to
+   * respond PONG; etc.) before a consecutive 5xx ejection occurs. Defaults to 5.
+   */
+  consecutive_5xx: int
+
+  /** The time interval between ejection analysis sweeps. This can result in
+   * both new ejections as well as hosts being returned to service. Defaults
+   * to 10000ms or 10s.
+   */
+  interval: int
+
+  /** The base time that a host is ejected for. The real time is equal to the
+   * base time multiplied by the number of times the host has been ejected and is
+   * capped by :ref:`max_ejection_time<envoy_v3_api_field_config.cluster.v3.OutlierDetection.max_ejection_time>`.
+   * Defaults to 30000ms or 30s.
+   */
+  base_ejection_time: int
+
+  /** The maximum % of an upstream cluster that can be ejected due to outlier detection. Defaults to 10% .
+   * Will eject at least one host regardless of the value if :ref:`always_eject_one_host<envoy_v3_api_field_config.cluster.v3.OutlierDetection.always_eject_one_host>` is enabled.
+   */
+  max_ejection_percent: int
+
+  /** The % chance that a host will be actually ejected when an outlier status
+   * is detected through consecutive 5xx. This setting can be used to disable
+   * ejection or to ramp it up slowly. Defaults to 100.
+   */
+  enforcing_consecutive_5xx: int
+
+  /** The % chance that a host will be actually ejected when an outlier status
+   * is detected through success rate statistics. This setting can be used to
+   * disable ejection or to ramp it up slowly. Defaults to 100.
+   */
+  enforcing_success_rate: int
+
+  /** The number of hosts in a cluster that must have enough request volume to
+   * detect success rate outliers. If the number of hosts is less than this
+   * setting, outlier detection via success rate statistics is not performed
+   * for any host in the cluster. Defaults to 5.
+   */
+  success_rate_minimum_hosts: int
+
+  /** The minimum number of total requests that must be collected in one
+   * interval (as defined by the interval duration above) to include this host
+   * in success rate based outlier detection. If the volume is lower than this
+   * setting, outlier detection via success rate statistics is not performed
+   * for that host. Defaults to 100.
+   */
+  success_rate_request_volume: int
+
+  /** This factor is used to determine the ejection threshold for success rate
+   * outlier ejection. The ejection threshold is the difference between the
+   * mean success rate, and the product of this factor and the standard
+   * deviation of the mean success rate: mean - (stdev *
+   * success_rate_stdev_factor). This factor is divided by a thousand to get a
+   * double. That is, if the desired factor is 1.9, the runtime value should
+   * be 1900. Defaults to 1900.
+   */
+  success_rate_stdev_factor: int
+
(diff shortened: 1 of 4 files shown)


d2-schemas/src/main/pegasus/com/linkedin/d2/OutlierDetectionConfig.pdl
```diff
@@ -0,0 +1,173 @@
+namespace com.linkedin.d2
+
+/**
+ * Options for configuring the outlier detection. Only used by gRPC clients.
+ */
```
gu0keno0, line 7: Add a link to Envoy proto , and we should mention that this is a port from Envoy proto. Then the following comments on the fields would make better sense because we just copied them from Envoy

d2-schemas/src/main/pegasus/com/linkedin/d2/OutlierDetectionConfig.pdl
```diff
@@ -0,0 +1,173 @@
...
+record OutlierDetectionConfig {
+  /** 
+   * The number of consecutive server-side error responses (for HTTP traffic,
+   * 5xx responses; for TCP traffic, connection failures; for Redis, failure to
+   * respond PONG; etc.) before a consecutive 5xx ejection occurs. Defaults to 5.
```
gu0keno0, line 12: what's the behavior for gRPC? gRPC does not have 5xx

d2-schemas/src/main/pegasus/com/linkedin/d2/OutlierDetectionConfig.pdl
```diff
@@ -0,0 +1,173 @@
...
+   */
+  consecutive_5xx: int
+
+  /** The time interval between ejection analysis sweeps. This can result in
+   * both new ejections as well as hosts being returned to service. Defaults
+   * to 10000ms or 10s.
```
gu0keno0, line 21: I thought we will have different defaults for LinkedIn use cases?

d2-schemas/src/main/pegasus/com/linkedin/d2/OutlierDetectionConfig.pdl
```diff
@@ -0,0 +1,173 @@
...
+   * is detected through consecutive 5xx. This setting can be used to disable
+   * ejection or to ramp it up slowly. Defaults to 100.
+   */
+  enforcing_consecutive_5xx: int
+
+  /** The % chance that a host will be actually ejected when an outlier status
+   * is detected through success rate statistics. This setting can be used to
+   * disable ejection or to ramp it up slowly. Defaults to 100.
+   */
+  enforcing_success_rate: int
```
gu0keno0, line 47: Let's double check if we enable success rate based detection already for gRPC.

gu0keno0 (commented):
Let's improve the documentation and the defaults, as the use case is different from Envoy.

d2-schemas/src/main/pegasus/com/linkedin/d2/OutlierDetectionConfig.pdl
whutwhu: I've double checked and the gRPC also sets it to 100%.

d2-schemas/src/main/pegasus/com/linkedin/d2/OutlierDetectionConfig.pdl
whutwhu: Exactly! I’ve updated the comments for all configurations that have different defaults for LinkedIn use cases.

d2-schemas/src/main/pegasus/com/linkedin/d2/OutlierDetectionConfig.pdl
whutwhu: Envoy’s outlier detection supports gRPC by relying on the HTTP status codes derived from the [grpc-status](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#responses) response header ([link](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/outlier#grpc)).

d2-schemas/src/main/pegasus/com/linkedin/d2/OutlierDetectionConfig.pdl
whutwhu: I’ve updated the comment to include this detail.

whutwhu:
> Let's improve the documentation and the defaults, as the use case is different from Envoy.

Makes sense. Thanks for reviewing!

d2-schemas/src/main/pegasus/com/linkedin/d2/OutlierDetectionConfig.pdl
gu0keno0: The mapping rule is not very apparent from the doc.

gu0keno0 (approved):

d2-schemas/src/main/pegasus/com/linkedin/d2/OutlierDetectionConfig.pdl
whutwhu: The code is shown below, and I’ve included the instructions in the comments:
https://github.com/envoyproxy/envoy/blob/main/source/common/grpc/status.cc#L28

TylerHorth: Seems to be documented in the proto definition also: https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto
Maybe we can link that instead of the envoy source code?

whutwhu: Hi @TylerHorth, sounds good! I’ve updated the link in the comment.

d2-schemas/src/main/pegasus/com/linkedin/d2/OutlierDetectionConfig.pdl
```diff
@@ -0,0 +1,178 @@
...
+
+  /** The number of hosts in a cluster that must have enough request volume to
+   * detect success rate outliers. If the number of hosts is less than this
+   * setting, outlier detection via success rate statistics is not performed
+   * for any host in the cluster. The default has been changed from 5 in Envoy to 3 in LinkedIn.
+   */
+  success_rate_minimum_hosts: int
```
shivamgupta1, line 54: Consider if we wish to create separate schemas for each kind of outlier detection mechanism (success rate, consecutive error rate, etc.), rather than having several fields with the type prefixed to the names (i.e. `sucess_rate_`).

whutwhu: Currently, we align with OSS Envoy to use the exact same schemas, which are also adopted by OSS gRPC, allowing us to stay consistent with industry.

PR #1100, opened 2025-09-03
Add a pre-check step for creating INDIS connections

brycezhongqing:
# PR description

During the INDIS migration, we observed that many clients encountered various issues when connecting to INDIS. Troubleshooting these problems often consumed significant time from both customers and the INDIS team.

This PR addresses that pain point by **introducing a pre-check function that runs before attempting to establish an INDIS connection**. The pre-check **introduces no side effects**—it will not alter or break the normal workflow. Instead, it provides clear log messages that indicate which errors might prevent a successful INDIS connection.

With these logs, customers and the INDIS team can more easily identify root causes, accelerating troubleshooting and enabling customers to resolve issues on their own.

🚨 Pre-Check Failure Scenarios
- `Java version too low`
- `Network connectivity issues`
- `gRPC health check failures`
- `Missing dependencies`

# Test Done
[✅]manully test
case1: Everything goes well:
```
2025/09/03 10:36:31.236 INFO [XdsClientImpl] [Indis xDS client executor-4-1] [indis-canary] [AAY96QuUvrF6rDL6zAFY4Q==] [xds pre-check] INDIS xDS server authority: main.indis-registry-observer.ei-ltx1.atd.disco.linkedin.com:32123
2025/09/03 10:36:31.236 INFO [XdsClientImpl] [Indis xDS client executor-4-1] [indis-canary] [AAY96QuUvrF6rDL6zAFY4Q==] [xds pre-check] Testing socket connection to INDIS xDS server: main.indis-registry-observer.ei-ltx1.atd.disco.linkedin.com:32123
2025/09/03 10:36:31.480 INFO [XdsClientImpl] [Indis xDS client executor-4-1] [indis-canary] [AAY96QuUvrF6rDL6zAFY4Q==] [xds pre-check] Successfully connected to INDIS xDS server at main.indis-registry-observer.ei-ltx1.atd.disco.linkedin.com:32123
2025/09/03 10:36:32.000 INFO [XdsClientImpl] [Indis xDS client executor-4-1] [indis-canary] [AAY96QuUvrF6rDL6zAFY4Q==] [xds pre-check] Health check for managed channel passed - channel is SERVING
2025/09/03 10:36:32.001 INFO [XdsClientImpl] [Indis xDS client executor-4-1] [indis-canary] [AAY96QuUvrF6rDL6zAFY4Q==] [xds pre-check] Protobuf Descriptor classes are available
2025/09/03 10:36:32.002 INFO [XdsClientImpl] [Indis xDS client executor-4-1] [indis-canary] [AAY96QuUvrF6rDL6zAFY4Q==] [xds pre-check] Envoy API classes are available
2025/09/03 10:36:32.002 INFO [XdsClientImpl] [Indis xDS client executor-4-1] [indis-canary] [AAY96QuUvrF6rDL6zAFY4Q==] [xds pre-check] Protobuf Descriptor and module dependency checks passed
2025/09/03 10:36:32.002 INFO [XdsClientImpl] [Indis xDS client executor-4-1] [indis-canary] [AAY96QuUvrF6rDL6zAFY4Q==] [xds pre-check] Managed channel is healthy and ready to use
2025/09/03 10:36:33.253 INFO [XdsClientImpl] [Grpc xDS Client Executor-13-1] [indis-canary] [AAY96QuziYpYykR0449H6w==] [xds pre-check] INDIS xDS server authority: main.indis-registry-observer.ei-ltx1.atd.disco.linkedin.com:32123
2025/09/03 10:36:33.253 INFO [XdsClientImpl] [Grpc xDS Client Executor-13-1] [indis-canary] [AAY96QuziYpYykR0449H6w==] [xds pre-check] Testing socket connection to INDIS xDS server: main.indis-registry-observer.ei-ltx1.atd.disco.linkedin.com:32123
2025/09/03 10:36:33.338 INFO [XdsClientImpl] [Grpc xDS Client Executor-13-1] [indis-canary] [AAY96QuziYpYykR0449H6w==] [xds pre-check] Successfully connected to INDIS xDS server at main.indis-registry-observer.ei-ltx1.atd.disco.linkedin.com:32123
2025/09/03 10:36:33.756 INFO [XdsClientImpl] [Grpc xDS Client Executor-13-1] [indis-canary] [AAY96QuziYpYykR0449H6w==] [xds pre-check] Health check for managed channel passed - channel is SERVING
2025/09/03 10:36:33.756 INFO [XdsClientImpl] [Grpc xDS Client Executor-13-1] [indis-canary] [AAY96QuziYpYykR0449H6w==] [xds pre-check] Protobuf Descriptor classes are available
2025/09/03 10:36:33.756 INFO [XdsClientImpl] [Grpc xDS Client Executor-13-1] [indis-canary] [AAY96QuziYpYykR0449H6w==] [xds pre-check] Envoy API classes are available
2025/09/03 10:36:33.756 INFO [XdsClientImpl] [Grpc xDS Client Executor-13-1] [indis-canary] [AAY96QuziYpYykR0449H6w==] [xds pre-check] Protobuf Descriptor and module dependency checks passed
2025/09/03 10:36:33.756 INFO [XdsClientImpl] [Grpc xDS Client Executor-13-1] [indis-canary] [AAY96QuziYpYykR0449H6w==] [xds pre-check] Managed channel is healthy and ready to use
2025/09/03 10:36:40.553 INFO [XdsClientImpl] [Indis xDS client executor-15-1] [indis-canary] [AAY96Qwi6qRJyxkHLVQbKQ==] [xds pre-check] INDIS xDS server authority: main.indis-registry-observer.ei-ltx1.atd.disco.linkedin.com:32123
2025/09/03 10:36:40.553 INFO [XdsClientImpl] [Indis xDS client executor-15-1] [indis-canary] [AAY96Qwi6qRJyxkHLVQbKQ==] [xds pre-check] Testing socket connection to INDIS xDS server: main.indis-registry-observer.ei-ltx1.atd.disco.linkedin.com:32123
2025/09/03 10:36:40.640 INFO [XdsClientImpl] [Indis xDS client executor-15-1] [indis-canary] [AAY96Qwi6qRJyxkHLVQbKQ==] [xds pre-check] Successfully connected to INDIS xDS server at main.indis-registry-observer.ei-ltx1.atd.disco.linkedin.com:32123
2025/09/03 10:36:41.115 INFO [XdsClientImpl] [Indis xDS client executor-15-1] [indis-canary] [AAY96Qwi6qRJyxkHLVQbKQ==] [xds pre-check] Health check for managed channel passed - channel is SERVING
2025/09/03 10:36:41.115 INFO [XdsClientImpl] [Indis xDS client executor-15-1] [indis-canary] [AAY96Qwi6qRJyxkHLVQbKQ==] [xds pre-check] Protobuf Descriptor classes are available
2025/09/03 10:36:41.115 INFO [XdsClientImpl] [Indis xDS client executor-15-1] [indis-canary] [AAY96Qwi6qRJyxkHLVQbKQ==] [xds pre-check] Envoy API classes are available
2025/09/03 10:36:41.115 INFO [XdsClientImpl] [Indis xDS client executor-15-1] [indis-canary] [AAY96Qwi6qRJyxkHLVQbKQ==] [xds pre-check] Protobuf Descriptor and module dependency checks passed
2025/09/03 10:36:41.115 INFO [XdsClientImpl] [Indis xDS client executor-15-1] [indis-canary] [AAY96Qwi6qRJyxkHLVQbKQ==] [xds pre-check] Managed channel is healthy and ready to use
```

case2: NACL issue:
```
2025/09/03 11:35:37.250 INFO [XdsClientImpl] [Indis xDS client executor-4-1] [indis-canary] [AAY96d7woRC2h5PagBm8/Q==] [xds pre-check] INDIS xDS server authority: main.indis-registry-observer.prod-ltx1.atd.disco.linkedin.com:32123
2025/09/03 11:35:37.250 INFO [XdsClientImpl] [Indis xDS client executor-4-1] [indis-canary] [AAY96d7woRC2h5PagBm8/Q==] [xds pre-check] Testing socket connection to INDIS xDS server: main.indis-registry-observer.prod-ltx1.atd.disco.linkedin.com:32123
2025/09/03 11:35:39.765 ERROR [XdsClientImpl] [Indis xDS client executor-4-1] [indis-canary] [AAY96d7woRC2h5PagBm8/Q==] [xds pre-check] Failed to connect to INDIS xDS server at authority main.indis-registry-observer.prod-ltx1.atd.disco.linkedin.com:32123: Connect timed out, check go/onboardindis Guidelines #5 and #7 for more details
2025/09/03 11:35:41.867 INFO [XdsClientImpl] [Grpc xDS Client Executor-13-1] [indis-canary] [AAY96d83E8/uDEJMGwyhQg==] [xds pre-check] INDIS xDS server authority: main.indis-registry-observer.prod-ltx1.atd.disco.linkedin.com:32123
2025/09/03 11:35:41.867 INFO [XdsClientImpl] [Grpc xDS Client Executor-13-1] [indis-canary] [AAY96d83E8/uDEJMGwyhQg==] [xds pre-check] Testing socket connection to INDIS xDS server: main.indis-registry-observer.prod-ltx1.atd.disco.linkedin.com:32123
2025/09/03 11:35:43.868 ERROR [XdsClientImpl] [Grpc xDS Client Executor-13-1] [indis-canary] [AAY96d83E8/uDEJMGwyhQg==] [xds pre-check] Failed to connect to INDIS xDS server at authority main.indis-registry-observer.prod-ltx1.atd.disco.linkedin.com:32123: Connect timed out, check go/onboardindis Guidelines #5 and #7 for more details
2025/09/03 11:35:49.096 INFO [XdsClientImpl] [Indis xDS client executor-15-1] [indis-canary] [AAY96d+lYr3FzkfvnYxWnQ==] [xds pre-check] INDIS xDS server authority: main.indis-registry-observer.prod-ltx1.atd.disco.linkedin.com:32123
2025/09/03 11:35:49.096 INFO [XdsClientImpl] [Indis xDS client executor-15-1] [indis-canary] [AAY96d+lYr3FzkfvnYxWnQ==] [xds pre-check] Testing socket connection to INDIS xDS server: main.indis-registry-observer.prod-ltx1.atd.disco.linkedin.com:32123
2025/09/03 11:35:51.097 ERROR [XdsClientImpl] [Indis xDS client executor-15-1] [indis-canary] [AAY96d+lYr3FzkfvnYxWnQ==] [xds pre-check] Failed to connect to INDIS xDS server at authority main.indis-registry-observer.prod-ltx1.atd.disco.linkedin.com:32123: Connect timed out, check go/onboardindis Guidelines #5 and #7 for more details
```


d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -74,6 +75,7 @@ public class XdsClientImpl extends XdsClient
       new RateLimitedLogger(_log, TimeUnit.MINUTES.toMillis(1), SystemClock.instance());
   public static final long DEFAULT_READY_TIMEOUT_MILLIS = 2000L;
   public static final Integer DEFAULT_MAX_RETRY_BACKOFF_SECS = 30; // default value for max retry backoff seconds
+  private static final String MINIMUM_VALID_JAVA_VERSION = "1.8.0_282-msft";
```
bohhyang: this better to be configurable, so that we can control later if higher java version is needed.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -807,6 +818,123 @@ private void notifyStreamReconnect()
     _xdsClientJmx.setIsConnected(true);
   }
 
+
+
+
```
bohhyang: nit remove empty lines

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -307,6 +309,15 @@ void startRpcStreamLocal()
       _log.warn("Tried to create duplicate RPC stream, ignoring!");
       return;
     }
+
+    if (!preCheckForIndisConnection())
```
bohhyang: The pre-check should be run only ONCE at startup, instead of in every startRpcStreamLocal (every reconnect). After https://github.com/linkedin/rest.li/pull/1093 is merged, there will be a start() method in XdsClientImpl, so you can add it there (cleaner).
Or add it to startRpcStream().

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -807,6 +818,123 @@ private void notifyStreamReconnect()
...
+      if (!D2Utils.isJavaVersionAtLeast(javaVersion, MINIMUM_VALID_JAVA_VERSION))
+      {
+        _log.error("The current Java version {} is too low, please upgrade to at least {}, " +
+                       "otherwise the service couldn't create grpc connection between service and INDIS," +
+                       "check go/onboardindis Guidelines #2 for more details",
+                   javaVersion,
+                   MINIMUM_VALID_JAVA_VERSION);
+        return false;
+      }
+      // Check if we can actually reach the server at the given authority to rule out NACL issue
```
bohhyang: the different checks should be in their individual helper functions to make code cleaner and manageable (serverAuthority, socket connection, grpc health check, dependencies)

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -807,6 +818,123 @@ private void notifyStreamReconnect()
...
+        _log.error("[xds pre-check] Required classes not found: {},  check go/onboardindis Guidelines #3 and #4",
+                   e.getMessage());
+        return false;
+      }
+      catch (NoClassDefFoundError err)
+      {
+        _log.error("[xds pre-check] Class definition missing: {},  check go/onboardindis Guidelines #3 and #4",
+                   err.getMessage(), err);
+        return false;
+      }
```
bohhyang: Seems these two are very similar with the same action item anyway. 
Can they be merged?
```
catch (ClassNotFoundException | NoClassDefFoundError e)
```

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -807,6 +818,123 @@ private void notifyStreamReconnect()
...
+      catch (Exception e)
+      {
+        Throwable c = e;
+        while (c.getCause() != null) c = c.getCause();
+        String errorMessage = c.getClass().getName() + " | " + c.getMessage();
+        _log.error("[xds pre-check] Health check failed for managed channel: {}， " +
+                      "check go/onboardindis Guidelines #2 and #9 for more details, " +
+                      "if there is any other sslContext issue, please check with #pki team", errorMessage);
+        return false;
+      }
```
bohhyang: Sometimes connections can't be made within readyTimeout but it's temporary. We should catch timeout exception separately and log that it may be transient and users should check the wiki if the issue persists.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -807,6 +818,123 @@ private void notifyStreamReconnect()
...
+      {
+        _log.error("The current Java version {} is too low, please upgrade to at least {}, " +
+                       "otherwise the service couldn't create grpc connection between service and INDIS," +
+                       "check go/onboardindis Guidelines #2 for more details",
+                   javaVersion,
+                   MINIMUM_VALID_JAVA_VERSION);
+        return false;
+      }
+      // Check if we can actually reach the server at the given authority to rule out NACL issue
+      String serverAuthority = getXdsServerAuthority();
```
bohhyang: the _manageChannel can be mocked to test the serverAuthority and health check related checks.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: btw, this precheck shouldn't be run on the executor but should be blocking xds client startup, so that later we can control whether to fail the app startup if the precheck fails.

d2/src/main/java/com/linkedin/d2/discovery/util/D2Utils.java
```diff
@@ -199,6 +199,77 @@ private static String getProp(String name, Properties sysProps, Map<String, Stri
...
+    try
+    {
+      // Simple approach: Java 9+ is always acceptable
+      if (isJava9OrHigher(currentVersion))
+      {
+        return true;
+      }
+      
+      // For Java 8, check build number
+      if (currentVersion.startsWith("1.8.0_") && minVersion.startsWith("1.8.0_"))
```
PapaCharlie: Can you make `"1.8.0_"` a constant? It's used in multiple places

d2/src/main/java/com/linkedin/d2/discovery/util/D2Utils.java
```diff
@@ -199,6 +199,77 @@ private static String getProp(String name, Properties sysProps, Map<String, Stri
...
+  }
+  
+  /**
+   * Checks if the Java version is 9 or higher.
+   */
+  private static boolean isJava9OrHigher(String version)
+  {
+    // Java 9+ versions start with "9.", "10.", "11.", etc.
+    return version.matches("^([9-9]|[1-9][0-9]+)\\..*");
+  }
```
PapaCharlie: Feels like it would be easier to check this explicitly rather than using a regex:
```
Integer.parseInt(version.substring(version.indexOf('.'))) >= 9
```
Regexes are hard to understand even for those who know how to write them, and they perform very poorly.

d2/src/main/java/com/linkedin/d2/discovery/util/D2Utils.java
```diff
@@ -199,6 +199,77 @@ private static String getProp(String name, Properties sysProps, Map<String, Stri
...
+   * Example: "1.8.0_282-msft" -> 282
+   */
+  private static int extractBuildNumber(String version)
+  {
+    String buildPart = version.substring("1.8.0_".length());
+    if (buildPart.contains("-"))
+    {
+      buildPart = buildPart.split("-")[0];
+    }
+    return Integer.parseInt(buildPart);
```
PapaCharlie: Nit: splitting creates a bunch of garbage. This is equivalent to:
```
int dash = buildPart.indexOf('-');
if (dash != -1)
{
  buildPart = buildPart.substring(0, dash);
}
```

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -807,6 +818,123 @@ private void notifyStreamReconnect()
...
+      {
+        String[] parts = serverAuthority.split(":");
+        if (parts.length != 2)
+        {
+          _log.error("[xds pre-check] Invalid server authority format: {}, expected format: host:port", serverAuthority);
+          return false;
+        }
+
+        String host = parts[0];
+        int port = Integer.parseInt(parts[1]);
```
PapaCharlie: This technically breaks if the string is `[::1]:32123` or any other IPv6 address. I'm going to try and see if there is a good equivalent of https://pkg.go.dev/net#SplitHostPort for Java

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
PapaCharlie: +1

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -807,6 +818,123 @@ private void notifyStreamReconnect()
...
+
+        _log.info("[xds pre-check] Testing socket connection to INDIS xDS server: {}:{}", host, port);
+        // Check if we can connect to the server using a socket
+        try (java.net.Socket socket = new java.net.Socket())
```
PapaCharlie: Can you not import `java.net.Socket` here?

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
PapaCharlie: IMO all of this should be in its own file within this package, like `XdsClientValidation` or something. This includes `isJavaVersionAtLeast` in `D2Utils`

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -807,6 +818,123 @@ private void notifyStreamReconnect()
...
+        {
+          socket.connect(new java.net.InetSocketAddress(host, port), (int) _readyTimeoutMillis);
```
PapaCharlie: Same, can you import `java.net.InetSocketAddress`?

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -807,6 +818,123 @@ private void notifyStreamReconnect()
...
+          _log.info("[xds pre-check] Successfully pinged to INDIS xDS server at {}:{}", host, port);
+        }
+      }
+      catch (Exception e)
+      {
+        _log.error("[xds pre-check] Failed to connect to INDIS xDS server at authority {}: {}, " +
+                      "check go/onboardindis Guidelines #5 and #7 for more details",
+                  serverAuthority, e.getMessage(), e);
```
PapaCharlie: Why are you calling `e.getMessage`? This will print the message twice. Better to let the logger handle it:
```suggestion
        _log.error("[xds pre-check] Failed to connect to INDIS xDS server at authority {}, " +
                      "check go/onboardindis Guidelines #5 and #7 for more details",
                  serverAuthority, e);
```

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
PapaCharlie: I would probably avoid this specific check. There is a good chance that between the DNS lookup and actually trying the healthcheck, the observer went into AT_CAPACITY. I think this is somewhat likely to fail

PapaCharlie: The alternative here would be to explicitly check the channel's state, with something like this:
```
  private boolean waitForReady(long timeoutMs)
  {
    // getState(true) forces the underlying connection to start.
    ConnectivityState state = _managedChannel.getState(true);
    long end = System.currentTimeMillis() + timeoutMs;
    while (!ConnectivityState.READY.equals(state))
    {
      CountDownLatch latch = new CountDownLatch(1);
      _managedChannel.notifyWhenStateChanged(state, latch::countDown);
      try
      {
        // If waitTime is 0 (or negative), latch.await immediately returns false, exiting the loop.
        long waitTime = end - System.currentTimeMillis();
        if (!latch.await(waitTime, TimeUnit.MILLISECONDS))
        {
          return false;
        }
      }
      catch (InterruptedException e)
      {
        Thread.currentThread().interrupt();
        throw new RuntimeException(e);
      }
      state = _managedChannel.getState(false);
    }
    return true;
  }
```

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
PapaCharlie: You should check this before making a network call, so definitely move it above the ping

PapaCharlie (changes requested):

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
brycezhongqing: Sure. Good idea. Updated

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
brycezhongqing: Yeah. I add some logic to make sure the precheck will only run one-time

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
brycezhongqing: Yeah. Good idea. Updated

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
brycezhongqing: Updated. Thanks

d2/src/main/java/com/linkedin/d2/discovery/util/D2Utils.java
brycezhongqing: Sure. I have update the whole java version check logic.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
brycezhongqing: Yes. Thanks for catching this up!. Then [HostAndPort](https://guava.dev/releases/19.0/api/docs/com/google/common/net/HostAndPort.html) is watch we need

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
brycezhongqing: Sure. Removed

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
brycezhongqing: Yeah. We could remove it. And I have refactor the error message log part now

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
brycezhongqing: Yes! That was my initial idea as well. I’ve also tried using **_managedChannel.getState(true)**; to debug and trace the root cause. **Unfortunately, that doesn’t help with SSL context issues**—it only reports **TRANSIENT_FAILURE**. That’s why I switched to using io.grpc.health.v1.HealthGrpc.newBlockingStub(_managedChannel) to dig deeper and uncover more details about the issue. You can find additional context here:

https://linkedin-randd.slack.com/archives/CDAKCBW92/p1756248008710549?thread_ts=1755215815.181029&cid=CDAKCBW92

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
brycezhongqing: Make senses

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
brycezhongqing: Yeah. Good point! Updated

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
brycezhongqing: Updated!

CHANGELOG.md
```diff
@@ -14,8 +14,20 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
-## [29.74.4] - 2025-09-03
-- Add a pre-check step for initializing xDS stream connections
+## [29.75.3] - 2025-09-05
```
bohhyang: this needs minor version bump since container needs to pass config to it, to do general check for all standard d2 clients too.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -78,7 +78,7 @@ public class XdsClientImpl extends XdsClient
       new RateLimitedLogger(_log, TimeUnit.MINUTES.toMillis(1), SystemClock.instance());
   public static final long DEFAULT_READY_TIMEOUT_MILLIS = 2000L;
   public static final Integer DEFAULT_MAX_RETRY_BACKOFF_SECS = 30; // default value for max retry backoff seconds
-  private static final String MINIMUM_VALID_JAVA_VERSION = "1.8.0_282-msft";
+  public static final String DEFAULT_MINIMUM_JAVA_VERSION = "1.8.0_282"; // default value for minimum Java version required for XDS client
```
bohhyang, line 49: should use XdsClientValidator one, don't redefine this.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -178,6 +182,21 @@ public XdsClientImpl(Node node,
...
+
+  public XdsClientImpl(Node node,
+      ManagedChannel managedChannel,
+      ScheduledExecutorService executorService,
+      long readyTimeoutMillis,
+      boolean subscribeToUriGlobCollection,
+      XdsServerMetricsProvider serverMetricsProvider,
+      boolean irvSupport,
+      Integer maxRetryBackoffSeconds,
+      String minimumJavaVersion,
```
bohhyang, line 181: could set this `@Nonnull` and don't need the null check then

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -178,6 +182,21 @@ public XdsClientImpl(Node node,
...
+      XdsClientValidator.ActionOnPrecheckFailure actionOnPrecheckFailure)
```
bohhyang, line 182: same here for Nonnull

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -208,6 +233,14 @@ public XdsClientImpl(Node node,
   public void start()
   {
     _xdsClientJmx.setXdsClient(this);
+
+    // Run pre-check only once at startup
+    if (!_preCheckCompleted.get())
+    {
+      XdsClientValidator.preCheckForIndisConnection(_managedChannel, _readyTimeoutMillis, _minimumJavaVersion, _actionOnPrecheckFailure);
+      _preCheckCompleted.set(true);
```
bohhyang, line 220: should set the flag first before running the precheck, in case in edge case, two threads call start() and both see flag being false.

d2/src/main/java/com/linkedin/d2/xds/XdsClientValidator.java
```diff
@@ -0,0 +1,490 @@
...
+
+/**
+ * Validation utilities for XDS client pre-checks.
+ * This class contains all the validation logic that should be performed before starting an XDS client.
+ */
+public class XdsClientValidator
+{
+  private static final Logger LOG = LoggerFactory.getLogger(XdsClientValidator.class);
+
+  public static final String DEFAULT_MINIMUM_JAVA_VERSION = "1.8.0_282";
```
bohhyang, line 37: should also define a static final var for default action, like "DEFAULT_ACTION_ON_PRECHECK_FAILURE", and let XdsClientImpl uses it.

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
```diff
@@ -853,6 +856,18 @@ public D2ClientBuilder setXdsStreamMaxRetryBackoffSeconds(int xdsStreamMaxRetryB
     return this;
   }
 
+  public D2ClientBuilder setXdsMinimumJavaVersion(String xdsMinimumJavaVersion)
```
bohhyang, line 860: could add nonnull

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
```diff
@@ -853,6 +856,18 @@ public D2ClientBuilder setXdsStreamMaxRetryBackoffSeconds(int xdsStreamMaxRetryB
...
+  {
+    _config.xdsMinimumJavaVersion = xdsMinimumJavaVersion;
+    return this;
+  }
+
+  public D2ClientBuilder setActionOnPrecheckFailure(XdsClientValidator.ActionOnPrecheckFailure actionOnPrecheckFailure)
```
bohhyang, line 866: could add nonnull

d2/src/main/java/com/linkedin/d2/xds/XdsClientValidator.java
```diff
@@ -0,0 +1,490 @@
...
+    switch (actionOnFailure)
+    {
+      case ERROR:
+        LOG.error(errorMessage);
+        break;
+      case THROW:
+        throw new IllegalStateException(errorMessage);
+      default:
+        LOG.warn(errorMessage);
+        break;
```
bohhyang:
```suggestion
      case WARN:
        LOG.warn(errorMessage);
      default:
        throw new IllegalArgumentException("Unknown action on precheck failure");
        break;
```

Should always explicitly handle enum cases, so that if new enum is added but not implemented, it won't be swallowed and will throw an error.

brycezhongqing: Make sense. Updated

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
brycezhongqing: Sure. Updated

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
brycezhongqing: Good point. Updated

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
brycezhongqing: Sure. Updated

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
brycezhongqing: Make sense

CHANGELOG.md
brycezhongqing: Sure!

bohhyang (approved):

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
```diff
@@ -853,6 +857,18 @@ public D2ClientBuilder setXdsStreamMaxRetryBackoffSeconds(int xdsStreamMaxRetryB
     return this;
   }
 
+  public D2ClientBuilder setXdsMinimumJavaVersion(@Nonnull String xdsMinimumJavaVersion)
```
PapaCharlie: IMO `@Nonnull` is implied unless a field is explicitly listed as `@Nullable`. It doesn't really add much to have this annotation

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -176,6 +181,21 @@ public XdsClientImpl(Node node,
...
+  public XdsClientImpl(Node node,
+      ManagedChannel managedChannel,
+      ScheduledExecutorService executorService,
+      long readyTimeoutMillis,
+      boolean subscribeToUriGlobCollection,
+      XdsServerMetricsProvider serverMetricsProvider,
+      boolean irvSupport,
+      Integer maxRetryBackoffSeconds,
+      @Nonnull String minimumJavaVersion,
+      @Nonnull XdsClientValidator.ActionOnPrecheckFailure actionOnPrecheckFailure)
```
PapaCharlie: See my other comment, IMO `@Nonnull` isn't really necessary. However, having the `@Nullable` annotation on `serverMetricsProvider` and `maxRetryBackoffSeconds` *does* seem necessary, since they *are* nullable.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -206,6 +232,13 @@ public XdsClientImpl(Node node,
   public void start()
   {
     _xdsClientJmx.setXdsClient(this);
+
+    // Run pre-check only once at startup
+    if (_preCheckCompleted.compareAndSet(false, true))
+    {
+      XdsClientValidator.preCheckForIndisConnection(_managedChannel, _readyTimeoutMillis, _minimumJavaVersion, _actionOnPrecheckFailure);
+    }
```
PapaCharlie: I don't think you really need the `_preCheckCompleted` flag. I don't think it's valid for `start` to be called more than once. If anything, this needs something like:
```java
private final AtomicBoolean _started = new AtomicBoolean();

....


@Override
public void start()
{
  if (!_started.compareAndSet(false, true))
  {
    throw new IllegalStateException("Cannot start XdsClient more than once");
  }
  ...
```

d2/src/main/java/com/linkedin/d2/xds/XdsClientValidator.java
```diff
@@ -0,0 +1,493 @@
...
+  private static final String PROTOBUF_FILE_DESCRIPTOR_CLASS = "com.google.protobuf.Descriptors$FileDescriptor";
+  private static final String ENVOY_AGGREGATED_DISCOVERY_SERVICE_CLASS = "io.envoyproxy.envoy.service.discovery.v3.AggregatedDiscoveryServiceGrpc";
+
+  /**
+   * Action to take when pre-check validation fails.
+   */
+  public enum ActionOnPrecheckFailure
+  {
+    // Log warning and continue (default behavior for stability)
+    WARN,
```
PapaCharlie: This doesn't seem necessary, there isn't much of a difference between WARN and ERROR. You can just remove this one and have just ERROR and THROW

d2/src/main/java/com/linkedin/d2/xds/XdsClientValidator.java
```diff
@@ -0,0 +1,493 @@
...
+   * Performs all pre-checks required for INDIS connection.
+   * This includes Java version validation, class availability checks, network connectivity tests, etc.
+   *
+   * @param managedChannel     the gRPC managed channel
+   * @param readyTimeoutMillis timeout for connection checks in milliseconds
+   * @param minimumJavaVersion the minimum required Java version
+   * @param actionOnFailure    action to take when validation fails
+   */
+  public static void preCheckForIndisConnection(ManagedChannel managedChannel, long readyTimeoutMillis,
+                                                String minimumJavaVersion, ActionOnPrecheckFailure actionOnFailure)
```
PapaCharlie: Don't format things this way, it's against the LinkedIn Java code convention: https://linkedin.atlassian.net/wiki/spaces/ENGS/pages/525084785/Java+Coding+Conventions
> Indentation for continuation lines is now always four spaces rather than attempting to align with the previous line. This protects code against bad formatting due to name changes during refactoring.

Basically, don't indent line breaks to the function name, just leave it at 4 spaces

d2/src/main/java/com/linkedin/d2/xds/XdsClientValidator.java
```diff
@@ -0,0 +1,493 @@
...
+
+  /**
+   * This method focuses purely on validation and returns error messages on failure.
+   *
+   * @param managedChannel     the gRPC managed channel
+   * @param readyTimeoutMillis timeout for connection checks in milliseconds
+   * @param minimumJavaVersion the minimum required Java version
+   * @return null if all pre-checks pass, error message if validation fails
+   */
+  static String processValidation(ManagedChannel managedChannel, long readyTimeoutMillis,
```
PapaCharlie: `@Nullable`?

d2/src/main/java/com/linkedin/d2/xds/XdsClientValidator.java
```diff
@@ -0,0 +1,493 @@
...
+                                  String minimumJavaVersion)
```
PapaCharlie: See comment above, don't format line breaks this way

d2/src/main/java/com/linkedin/d2/xds/XdsClientValidator.java
PapaCharlie: Seems you have a bunch of these methods that return null or an error message. They should probably all be annotated with @nullable

d2/src/main/java/com/linkedin/d2/xds/XdsClientValidator.java
```diff
@@ -0,0 +1,493 @@
...
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package com.linkedin.d2.xds;
+
+import io.grpc.ManagedChannel;
+
```
PapaCharlie: Why have this line break in the imports?

d2/src/main/java/com/linkedin/d2/xds/XdsClientValidator.java
```diff
@@ -0,0 +1,493 @@
...
+import java.net.InetSocketAddress;
+import java.net.Socket;
+import java.net.SocketTimeoutException;
+
```
PapaCharlie: Same

PapaCharlie (changes requested):

d2/src/main/java/com/linkedin/d2/xds/XdsClientValidator.java
brycezhongqing: IMO. We could keep the same action like [_actionOnWeightBreach](https://github.com/linkedin/rest.li/blob/ecfa350d0ceab90b3265addd03e83673669e7d51/d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java#L942)

d2/src/main/java/com/linkedin/d2/xds/XdsClientValidator.java
brycezhongqing: Updated

d2/src/main/java/com/linkedin/d2/xds/XdsClientValidator.java
brycezhongqing: Sure. Updated

d2/src/main/java/com/linkedin/d2/xds/XdsClientValidator.java
brycezhongqing: Updated

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
brycezhongqing: Sure. Remove the `_preCheckCompleted` flag

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
brycezhongqing: Updated

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
brycezhongqing: Updated

d2/src/main/java/com/linkedin/d2/xds/XdsClientValidator.java
PapaCharlie: But they are different? The possible values for weight breach are `WARN`, `THROW`, `RECTIFY` and `IGNORE`, which each do very different things. Here, `WARN` and `ERROR` are almost the same... When would you ever use `WARN` instead of `ERROR`? Remember that if the precheck finds an issue, the ADS stream will never connect so this is definitely an ERROR, not a WARN that the app can recover from.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
PapaCharlie: Can you add the `_started` flag? This is technically a bug that needs to be fixed in `start`

brycezhongqing: Sure. Updated

d2/src/main/java/com/linkedin/d2/xds/XdsClientValidator.java
brycezhongqing: Sure. Updated

PR #1101, opened 2025-09-04
mark fields as optional

whutwhu:
We should mark fields as optional since they have default values.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,11 +14,14 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
-## [29.74.5] - 2025-09-03
+## [29.75.1] - 2025-09-04
 - mark fields as optional
 
+## [29.75.0] - 2025-09-03
+- correct xds client latency metric, add active initial wait time metric, and SD update receipt event
+
 ## [29.74.4] - 2025-09-02
-- DataTranslator: fall back to matching simple 
+- DataTranslator: fall back to matching simple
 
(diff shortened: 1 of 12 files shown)


d2-schemas/src/main/pegasus/com/linkedin/d2/OutlierDetectionConfig.pdl
```diff
@@ -14,52 +14,52 @@ record OutlierDetectionConfig {
...
    */
-  enforcing_consecutive_5xx: int
+  enforcing_consecutive_5xx: optional int
 
   /** The % chance that a host will be actually ejected when an outlier status
    * is detected through success rate statistics. This setting can be used to
    * disable ejection or to ramp it up slowly. Defaults to 100.
    */
-  enforcing_success_rate: int
+  enforcing_success_rate: optional int
```
mchen07, line 45: Based on comment, there is default value for them, then you should define default value in their PDL?

mchen07: See example here: https://linkedin.github.io/rest.li/pdl_schema#primitive-types

whutwhu: Thanks for the info! We’ve removed all default values from the comments to avoid confusion; defaults is handled directly in the code.

mchen07 (approved):

PR #1102, opened 2025-09-05
Rename the outlier detection field in the D2 cluster

whutwhu:
As title. Make the PDL and proto naming consistent.


mchen07 (approved):
Let's also deprecate previously released version to avoid any potential mismatch after releasing this one.

PR #1103, opened 2025-09-05
Add D2 service MethodLevelProperties configuration support

golthitarun:
Description:

Add D2 service MethodLevelProperties configuration support

Testing Done:

Added unit tests

### DIFF
--- a/d2-schemas/src/main/pegasus/com/linkedin/d2/MethodLevelProperties.pdl
+++ b/d2-schemas/src/main/pegasus/com/linkedin/d2/MethodLevelProperties.pdl
@@ -1,0 +1,28 @@
+namespace com.linkedin.d2
+
+/**
+ * D2 Service Method Level properties related configuration.
+ * This property is only used for gRPC method level configuration and
+ * with this service owners can set configurations with the following levels of granularity.
+ * - Exact method match (service + method)
+ * - Service match (service only)
+ */
+record MethodLevelProperties {
+  /**
+   * name is a list of NameProperties object, on which these method level properties will be configured.
+   * NameProperties consists of service (gRPC service name) and method (gRPC method name)
+   */
+  name: array[NameProperties]
+
+  /**
+   * Metadata properties about the service e.g. retry, hedging etc
+   * configuring any properties not supported by gRPC, will be ignored
+   */
+  serviceMetadataProperties: map[string, string]
+
+  /**
+   * The transport client properties for this service
+   * configuring any properties not supported by gRPC, will be ignored
+   */
+  transportClientProperties: D2TransportClientProperties
+}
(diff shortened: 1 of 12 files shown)


d2-schemas/src/main/pegasus/com/linkedin/d2/MethodLevelProperties.pdl
```diff
@@ -0,0 +1,28 @@
...
+ * with this service owners can set configurations with the following levels of granularity.
+ * - Exact method match (service + method)
+ * - Service match (service only)
+ */
+record MethodLevelProperties {
+  /**
+   * name is a list of Name objects, on which these method level properties will be configured.
+   * Name consists of service (gRPC service name) and method (gRPC method name)
+   */
+  name: array[Name]
```
TylerHorth: I think copying the gRPC naming for this property doesn't make sense. The context "MethodLevelProperties" is different as is the content of this record. I think we should call this "methods" or at least make it plural "names" instead.

TylerHorth (commented):
Minor comment on naming (pun not intended 😛). Otherwise lgtm

d2-schemas/src/main/pegasus/com/linkedin/d2/Name.pdl
```diff
@@ -0,0 +1,16 @@
+namespace com.linkedin.d2
+
+/**
+ * Name configuration, used by D2 service method level properties
+ */
+record Name {
+  /**
+   * service is gRPC service name
```
shivamgupta1, line 8: nit: add an example? `com.linkedin.protobuf.Example`

d2-schemas/src/main/pegasus/com/linkedin/d2/MethodLevelProperties.pdl
golthitarun: Maybe I will update it to "names" as they can also place service as a whole in this

d2/src/main/java/com/linkedin/d2/balancer/properties/Name.java
```diff
@@ -0,0 +1,55 @@
...
+
+import com.google.common.base.MoreObjects;
+import java.util.Objects;
+import javax.annotation.Nullable;
+
+
+/**
+ * Name properties
+ */
+public class Name {
```
shivamgupta1, line 11: Is `Endpoint` a better name for this?

d2-schemas/src/main/pegasus/com/linkedin/d2/Name.pdl
golthitarun: sure will do

d2/src/main/java/com/linkedin/d2/balancer/properties/Name.java
golthitarun: I feel as this is specific to gRPC maybe better to keep it as names ?

shivamgupta1: Okay

PR #1106, opened 2025-09-11
Fix notifyOnLastChunk handling for D2URIMap WildcardResourceSubscriber

brycezhongqing:
Trigger condition: D2URIMap with glob collection.

When creating the `WildcardResourceSubscriber` for D2URIMap, the code remaps the resource type to **D2URI** and performs the wildcard subscription ([code link](https://github.com/linkedin/rest.li/blob/a5f43469556d480c6af52e905b07d19f1f613319/d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java#L263)).

However, in `notifyOnLastChunk` ([code is here](https://github.com/linkedin/rest.li/blob/a5f43469556d480c6af52e905b07d19f1f613319/d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java#L772)), we look up the **WildcardResourceSubscriber** using D2URI . Since the subscriber was only created for D2URIMap, onAllResourcesProcessed is never called in this scenario.

## Test Done
Override  `onAllResourcesProcessed` in RRE.
Behavior: Each time an update is received from the Observer, the method logs the size of the URIMap.
Note: The callback is invoked whenever the RRE receives a URI update from INDIS.
```
2025/09/11 13:48:13.866 INFO [D2IndisStoreAccessorImpl] [D2IndisStoreAccessor Executor Thread-4-1] [restli-resource-explorer] [] Total URI Maps: 5261
2025/09/11 13:48:14.595 INFO [D2IndisStoreAccessorImpl] [D2IndisStoreAccessor Executor Thread-4-1] [restli-resource-explorer] [] Total URI Maps: 5261
2025/09/11 13:48:16.028 INFO [D2IndisStoreAccessorImpl] [D2IndisStoreAccessor Executor Thread-4-1] [restli-resource-explorer] [] Total URI Maps: 5261
2025/09/11 13:48:17.079 INFO [D2IndisStoreAccessorImpl] [D2IndisStoreAccessor Executor Thread-4-1] [restli-resource-explorer] [] Total URI Maps: 5261
```

### DIFF
--- a/d2/src/test/java/com/linkedin/d2/xds/TestXdsClientImpl.java
+++ b/d2/src/test/java/com/linkedin/d2/xds/TestXdsClientImpl.java
@@ -65,7 +65,8 @@ public class TestXdsClientImpl
   private static final String VERSION1 = "1";
   private static final String VERSION2 = "2";
   private static final String VERSION3 = "3";
-  private static final String NONCE = "nonce";
+  private static final String NONCE = "000000000000000000000000"; // this is used as LAST CHUNK NONCE
+  private static final String NON_LAST_NONCE = "000000000000000000000001";
 
   private static final long TEN_MINS = Time.minutes(10);
   private static final XdsD2.Node NODE_WITH_DATA = createNodeWithData(DATA,
@@ -164,6 +165,12 @@ public class TestXdsClientImpl
           .setResource(Any.pack(D2URI_1))
           .build()
   ), null, NONCE, null);
+  private static final DiscoveryResponseData RESPONSE_D2URI1_WITH_NON_LAST_NONCE = new DiscoveryResponseData(D2_URI, Collections.singletonList(
+      Resource.newBuilder()
+          .setVersion(VERSION1)
+          .setName(URI_URN1)
+          .setResource(Any.pack(D2URI_1))
+          .build()), null, NON_LAST_NONCE, null);
   private static final DiscoveryResponseData RESPONSE_D2URI2_DELETE_URI1 = new DiscoveryResponseData(D2_URI, Collections.singletonList(
       Resource.newBuilder()
           .setVersion(VERSION1)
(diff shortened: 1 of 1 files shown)


d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -772,12 +772,16 @@ private void handleResourceRemoval(Collection<String> removedResources, Resource
   private void notifyOnLastChunk(DiscoveryResponseData response)
   {
     ResourceType type = response.getResourceType();
-    WildcardResourceSubscriber wildcardResourceSubscriber = getWildcardResourceSubscriber(type);
+    // For the D2URI with glob collection, we need to get D2URIMap WildcardSubscriber
+    ResourceType originalType =
+        (type == ResourceType.D2_URI && shouldSubscribeUriGlobCollection(ResourceType.D2_URI_MAP))
+            ? ResourceType.D2_URI_MAP
+            : type;
+    WildcardResourceSubscriber wildcardResourceSubscriber = getWildcardResourceSubscriber(originalType);
```
bohhyang:
```suggestion
    if (type == ResourceType.D2_URI)
    {
       type = ResourceType.D2_URI_MAP;
    }
    WildcardResourceSubscriber wildcardResourceSubscriber = getWildcardResourceSubscriber(type);
```

1. Should be directly updating "type" since it's used in logging on line 800 too. (It should be logging "Notifying wildcard subscriber of type D2_URI_MAP"
2. Doesn't need to call shouldSubscribeUriGlobCollection, when response type is D2_URI, wildcard D2_URI_MAP subscriber, if any, always needs to be notified (just like handleD2URICollectionResponse method directly call getWildcardResourceSubscriber(ResourceType.D2_URI_MAP)).  shouldSubscribeUriGlobCollection should be called only when subscribing to resources (to decide if it needs to convert the type before sending the request), not when processing the response.

brycezhongqing: Yeah. Make senses. Updated

PR #1107, opened 2025-09-15
Preliminary readiness management

bohhyang:
## Summary
As designed in [the doc](https://docs.google.com/document/d/1oa-gCLbNlWuEIQxIqscIEC24h0slI-7S_aItpjJZZLU/edit?usp=sharing).
Register each announcer's announcement status to readiness manager in container, see container PR.

## Test Done
UT. 
Will do QEI deploy indis-canary

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.76.0] - 2025-09-15
+- Preliminary readiness management
+
 ## [29.75.4] - 2025-09-10
 - Fix notifyOnLastChunk handling for D2URIMap WildcardResourceSubscriber 
 
@@ -5897,7 +5900,8 @@ patch operations can re-use these classes for generating patch messages.
 
 ## [0.14.1]
 
-[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.75.4...master
+[Unreleased]: https://github.com/linkedin/rest.li/compare/v29.76.0...master
+[29.76.0]: https://github.com/linkedin/rest.li/compare/v29.75.4...v29.76.0
 [29.75.4]: https://github.com/linkedin/rest.li/compare/v29.75.3...v29.75.4
 [29.75.3]: https://github.com/linkedin/rest.li/compare/v29.75.2...v29.75.3
 [29.75.2]: https://github.com/linkedin/rest.li/compare/v29.75.1...v29.75.2
--- a/d2/src/main/java/com/linkedin/d2/balancer/servers/NoOpReadinessStatusManager.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/servers/NoOpReadinessStatusManager.java
@@ -1,0 +1,22 @@
+package com.linkedin.d2.balancer.servers;
+
+public class NoOpReadinessStatusManager implements ReadinessStatusManager
+{
+  @Override
+  public void registerAnnouncerStatus(AnnouncerStatus status)
+  {
+    // no-op
+  }
+
+  @Override
+  public void onAnnouncerStatusUpdated()
+  {
+    // no-op
+  }
+
+  @Override
+  public void addWatcher(ReadinessStatusWatcher watcher)
+  {
+    // no-op
+  }
+}
--- a/d2/src/main/java/com/linkedin/d2/balancer/servers/ReadinessStatusManager.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/servers/ReadinessStatusManager.java
@@ -1,0 +1,113 @@
+package com.linkedin.d2.balancer.servers;
+
+import javax.annotation.Nonnull;
+
+
+/**
+ * This class manages the readiness status of the server.
+ * Currently, it's based on D2 announcements' statuses as a preliminary design, can be extended in future for internal
+ * components' readiness status.
+ * See the <a href="preliminary design doc">https://shorturl.at/hmjz8</a> for details of determining the readiness status based on
+ * D2 announcements. Treating D2 announcements as the app servers’ intent to serve traffic, the readiness status is
+ * determined by the combination of D2 announcements' sent status.
+ */
+public interface ReadinessStatusManager
+{
+  void registerAnnouncerStatus(AnnouncerStatus status);
+
+  void onAnnouncerStatusUpdated();
+
+  void addWatcher(ReadinessStatusWatcher watcher);
+
+  interface ReadinessStatusWatcher
+  {
+    void onReadinessStatusChanged(ReadinessStatus newStatus);
+  }
+
+  enum ReadinessStatus
+  {
+    /**
+     * When the D2 clusters that are intended to be announced have sent announcements successfully.
+     * (And if any D2 cluster is intended to be de-announced, the corresponding de-announcement is also sent successfully)
+     */
+    SERVING,
+    /**
+     * When all D2 clusters are intended to be de-announced and the de-announcements are sent successfully, or never
+     * announced before.
+     * OR: if some D2 clusters have isRequiredToServe set to true, any of them is de-announced.
+     */
+    NOT_SERVING,
+    /**
+     * There is a time gap between when a serving intent changes (either static config loaded at startup, or dynamic
+     * markup/markdown API is called at runtime) and when the intent is fulfilled ---- (de-)announcement sent successfully.
+     * During this gap, the intent is not fulfilled, and the readiness status is INCONSISTENT. Or, the (de-)announcements
+     * failed to be sent, the readiness status is also INCONSISTENT.
+     * This status can be deprecated in future when the readiness status is based on internal components' readiness
+     * rather than D2 announcements' statuses.
+     */
+    INCONSISTENT
+  }
+
+  class AnnouncerStatus
+  {
+    // whether this announcer is to be announced and defined in static config
+    private final boolean isToBeAnnouncedFromStaticConfig;
+
+    private AnnouncementStatus announcementStatus;
+
+    public AnnouncerStatus(boolean isToBeAnnouncedFromStaticConfig, @Nonnull AnnouncementStatus announcementStatus)
+    {
+      this.isToBeAnnouncedFromStaticConfig = isToBeAnnouncedFromStaticConfig;
+      this.announcementStatus = announcementStatus;
+    }
+
+    public boolean isToBeAnnouncedFromStaticConfig()
+    {
+      return isToBeAnnouncedFromStaticConfig;
+    }
+
+    public boolean isAnnounced()
+    {
+      return announcementStatus == AnnouncementStatus.ANNOUNCED;
+    }
+
+    public boolean isDeAnnounced()
+    {
+      return announcementStatus == AnnouncementStatus.DE_ANNOUNCED;
+    }
+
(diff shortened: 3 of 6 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.76.0] - 2025-09-15
+- Preliminary readiness management
```
ZoabKapoor, line 18: Nit: what's `preliminary` about it? Would it make more sense to say `initial implementation of readiness management system` or similar?

d2/src/main/java/com/linkedin/d2/balancer/servers/NoOpReadinessStatusManager.java
```diff
@@ -0,0 +1,22 @@
+package com.linkedin.d2.balancer.servers;
+
+public class NoOpReadinessStatusManager implements ReadinessStatusManager
```
ZoabKapoor, line 3: Thinking out loud -- it'd be nice if Java had a way for us to concisely define a no-op implementation of an interface rather than having to add the boilerplate of all the empty`@Override` methods 

As far as I'm aware this isn't a thing though 😔

d2/src/main/java/com/linkedin/d2/balancer/servers/ReadinessStatusManager.java
```diff
@@ -0,0 +1,122 @@
+package com.linkedin.d2.balancer.servers;
+
+import javax.annotation.Nonnull;
+
+
+/**
+ * This class manages the readiness status of the server.
```
ZoabKapoor: Super nit: `This interface`

d2/src/main/java/com/linkedin/d2/balancer/servers/ReadinessStatusManager.java
```diff
@@ -0,0 +1,122 @@
...
+     * rather than D2 announcements' statuses.
+     */
+    INCONSISTENT,
+    /**
+     * When all D2 clusters are intended to be de-announced and the de-announcements are sent successfully, or never
+     * announced before.
+     * Note: if readinessStatusManager.requireStaticD2ServersAnnounced is set to true, and some D2 clusters have
+     * isToBeAnnouncedFromStaticConfig set to true, then any of them is de-announced will become NOT_SERVING.
+     */
+    NOT_SERVING,
```
ZoabKapoor, line 45: Nit: should we mention that `NOT_SERVING` is/should be the default `ReadinessStatus` if there's no other information? It looks like in our internal implementation we use it as the default, not sure if we want to document that here

(just considering I could also imagine an argument for `INCONSISTENT` as the default)

d2/src/main/java/com/linkedin/d2/balancer/servers/ReadinessStatusManager.java
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  class AnnouncerStatus
+  {
+    // whether this announcer is to be announced and defined in static config
+    private final boolean isToBeAnnouncedFromStaticConfig;
+
+    private AnnouncementStatus announcementStatus;
+
+    public AnnouncerStatus(boolean isToBeAnnouncedFromStaticConfig, @Nonnull AnnouncementStatus announcementStatus)
```
PapaCharlie: Nit: IMO `Nonnull` should be considered the default. You can assume everything is Nonnull unless it's explicitly marked `@Nullable`. Otherwise you end up having to check for null everywhere all the time and it gets tedious and error prone

d2/src/main/java/com/linkedin/d2/balancer/servers/ReadinessStatusManager.java
```diff
@@ -0,0 +1,122 @@
...
+
+    public boolean isAnnouncing()
+    {
+      return announcementStatus == AnnouncementStatus.ANNOUNCING;
+    }
+
+    public boolean isDeAnnouncing()
+    {
+      return announcementStatus == AnnouncementStatus.DE_ANNOUNCING;
+    }
```
PapaCharlie, line 91: How much does this really give you on top of just checking the value of `getAnnouncementStatus`?

d2/src/main/java/com/linkedin/d2/balancer/servers/ReadinessStatusManager.java
```diff
@@ -0,0 +1,122 @@
...
+     */
+    SERVING
+  }
+
+  class AnnouncerStatus
+  {
+    // whether this announcer is to be announced and defined in static config
+    private final boolean isToBeAnnouncedFromStaticConfig;
+
+    private AnnouncementStatus announcementStatus;
```
PapaCharlie: Cursory glance at the code suggests this may be accessed from different threads? Needs to be `volatile` if so (or an `AtomicReference`)

d2/src/main/java/com/linkedin/d2/balancer/servers/ReadinessStatusManager.java
bohhyang: It makes the logic check in [onAnnouncerStatusUpdated](https://github.com/linkedin-multiproduct/container/pull/1948/files#diff-f57b6c847456cbffbbb7bb19d8b2f039157d4f47a5f290d00447f09517e364b2R36) method in container shorter and handy. And it unifies the way of checking the enum equality (I've updated it to use AnnouncementStatus.XXX.equals(announcementStatus)).

d2/src/main/java/com/linkedin/d2/balancer/servers/ReadinessStatusManager.java
bohhyang: yeah, added volatile

d2/src/main/java/com/linkedin/d2/balancer/servers/ReadinessStatusManager.java
bohhyang: makes sense. Removed

d2/src/main/java/com/linkedin/d2/balancer/servers/ReadinessStatusManager.java
```diff
@@ -0,0 +1,123 @@
...
+    {
+      return AnnouncementStatus.DE_ANNOUNCING.equals(announcementStatus);
+    }
+
+    public AnnouncementStatus getAnnouncementStatus()
+    {
+      return announcementStatus;
+    }
+
+    public synchronized void setAnnouncementStatus(AnnouncementStatus announcementStatus)
```
PapaCharlie: Why `synchronized` this? `announcementStatus` is volatile, so you don't technically need to

PapaCharlie (approved):

d2/src/main/java/com/linkedin/d2/balancer/servers/ReadinessStatusManager.java
bohhyang: makes sense. Removed, thanks.

d2/src/main/java/com/linkedin/d2/jmx/ZooKeeperAnnouncerJmx.java
```diff
@@ -211,4 +211,8 @@ public int getWeightDecimalPlacesBreachedCount()
   public int getServerAnnounceMode() {
     return _announcer.getServerAnnounceMode().ordinal();
   }
+
+  public int getAnnouncementStatus() {
```
brycezhongqing, line 216: Should we need to add  `@Override` ?

d2/src/test/java/com/linkedin/d2/balancer/servers/TestZooKeeperAnnouncer.java
```diff
@@ -145,4 +145,271 @@ null, new LogOnlyServiceDiscoveryEventEmitter(),
...
+            // dark warmup disabled, real cluster markup successful
+            {ANNOUNCING, false, true, true, true, Collections.singletonList(ANNOUNCED)},
+            // dark warmup disabled, real cluster markup failed
+            {ANNOUNCING, false, true, true, false, Collections.emptyList()},
+
+            // --- other init statuses have the same behavior ---
+            {ANNOUNCED, false, true, true, true, Arrays.asList(ANNOUNCING, ANNOUNCED)},
+            {ANNOUNCED, false, true, true, false, Collections.singletonList(ANNOUNCING)},
+            {DE_ANNOUNCING, false, true, true, true, Arrays.asList(ANNOUNCING, ANNOUNCED)},
+            {DE_ANNOUNCING, false, true, true, false, Collections.singletonList(ANNOUNCING)}
```
ZoabKapoor, line 195: Nice, this is super thorough!

CHANGELOG.md
bohhyang: No, please see the doc in the PR description, also explained in this class's doc. It's preliminary as a short-term solution since it's based on d2 announcements, which shouldn't be in long run with topology aware SD will change it to base on internal components' readiness.

d2/src/main/java/com/linkedin/d2/balancer/servers/ReadinessStatusManager.java
bohhyang: INCONSISTENT is when some announcement/de-announcement happened and still ongoing or failed. At start up, no announcement is made yet, so not serving is default. I will comment it in the impl class in container. The interface here should be general.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
```diff
@@ -273,6 +296,9 @@ public ZooKeeperAnnouncer(LoadBalancerServer server, boolean initialIsUp,
     {
       ((ZooKeeperServer) server).setServiceDiscoveryEventHelper(this);
     }
+
+    _status = status;
+    _readinessManager = readinessManager;
```
brycezhongqing, line 301: why we don't register the _status here ?
```
_readinessManager.registerAnnouncerStatus(_status);
```

I saw the code is in container side.

bohhyang: I think in this case it's better for the "managing" code (d2serverfactory in container) to register the status explicitly which creates the announcer, instead of having the announcer registers itself in the constructor (to avoid doing work in constructor as much as possible. For example, If the manager needs to call back to the announcer when registering it, the announcer's construction may not be done yet), so the announcer just needs to report its subsequent updates.

d2/src/main/java/com/linkedin/d2/jmx/ZooKeeperAnnouncerJmx.java
bohhyang: ah, right, added.

d2/src/main/java/com/linkedin/d2/balancer/servers/ReadinessStatusManager.java
bohhyang: updated.

d2/src/main/java/com/linkedin/d2/balancer/servers/ZooKeeperAnnouncer.java
brycezhongqing: Make senses.

PR #1108, opened 2025-09-17
Dark Cluster Partition Support

jcleezer:
## Summary
This PR adds support for routing dark cluster traffic to partitioned hosts. For the most part the logic needed for this was present and working however the ClusterInfoProvider class was only using the Default Partition (0) to calculate the different host sizes. This has two problems, first the rate wont match what the user sets. And more importantly, if the cluster has no hosts that announce to 0, then the calculated QPS rate will be 0 and no traffic will be sent.

## Testing
This issue was first reported with a cluster that did not announce to the default partition for both source and dark cluster. With the changes here, dark cluster traffic was routed to their dark host announced to partition 10.

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancer.java
@@ -1294,7 +1294,6 @@ public class SimpleLoadBalancer implements LoadBalancer, HashRingProvider, Clien
     return -1;
   }
 
-
   @Override
   public DarkClusterConfigMap getDarkClusterConfigMap(String clusterName) throws ServiceUnavailableException
   {
--- a/d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancer.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancer.java
@@ -1295,6 +1295,58 @@ public class SimpleLoadBalancer implements LoadBalancer, HashRingProvider, Clien
   }
 
   @Override
+  public int getClusterCountAcrossPartitions(String clusterName, String scheme) throws ServiceUnavailableException
+  {
+    FutureCallback<Integer> clusterCountFutureCallback = new FutureCallback<>();
+
+    _state.listenToCluster(clusterName, (type, name) ->
+    {
+      if (_state.getUriProperties(clusterName) != null && _state.getUriProperties(clusterName).getProperty() != null)
+      {
+        Set<URI> allUris = _state.getUriProperties(clusterName).getProperty().Uris();
+        int count = (int) allUris.stream().filter(uri -> scheme.equalsIgnoreCase(uri.getScheme())).count();
+        clusterCountFutureCallback.onSuccess(count);
+      }
+      else
+      {
+        // there won't be a UriProperties if there are no Uris announced. Return zero in this case.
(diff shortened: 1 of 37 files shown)


d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancer.java
```diff
@@ -1294,6 +1294,58 @@ int getClusterCountFromCache(String clusterName, String scheme, int partitionId)
     return -1;
   }
 
+  @Override
+  public int getClusterCountAcrossPartitions(String clusterName, String scheme) throws ServiceUnavailableException
```
shivamgupta1: It seems to me that all partitions should be handled independently. As in, this method should return `Map<Integer, Integer>`. The overall host count might not be a good indicator of the traffic multipliers. The same server host can register with multiple partitions at the service owners' discretion. It's possible that some partitions have 10x the number of hosts in the dark cluster while others have only 1x.

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterManagerImpl.java
```diff
@@ -164,6 +165,20 @@ public boolean handleDarkRequest(RestRequest originalRequest, RequestContext ori
...
 
+  private int getPartitionId(RestRequest request)
+  {
+    try
+    {
+      String serviceName = com.linkedin.d2.balancer.util.LoadBalancerUtil.getServiceNameFromUri(request.getURI());
+      com.linkedin.d2.balancer.util.partitions.PartitionAccessor accessor = _facilities.getPartitionInfoProvider().getPartitionAccessor(serviceName);
+      return accessor.getPartitionId(request.getURI());
+    }
+    catch (Throwable t)
```
shivamgupta1: `Throwable` may be too much, we don't wanna catch JVM errors and ignore them. Perhaps change this to `RuntimeException`? Same elsewhere.

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
```diff
@@ -138,16 +142,44 @@ public void shutdown()
...
+
+    try
+    {
+      // Lazily create the strategy if it doesn't exist.
+      DarkClusterConfigMap darkClusterConfigMap = _facilities.getClusterInfoProvider().getDarkClusterConfigMap(_sourceClusterName);
+      if (darkClusterConfigMap != null && darkClusterConfigMap.containsKey(darkClusterName))
+      {
+        DarkClusterConfig config = darkClusterConfigMap.get(darkClusterName);
+        DarkClusterStrategy strategy = createStrategy(darkClusterName, config, partitionId);
+        darkMap.put(darkClusterName, strategy);
```
shivamgupta1: It seems better to put this logic in a `computeIfAbsent` to make it thread safe?

jcleezer: Yeah much nicer. Thanks.

darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterManagerImpl.java
jcleezer: Makes sense, fixed. Thanks.

PR #1110, opened 2025-10-02
null guards for xds server and channel

bohhyang:
As the title, add null guards for critical params. Also add a setter to disable detecting raw d2 client, so test apps like indis-canary can use it to disable the detection as needed.

## Test Done
N/A

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
@@ -78,6 +78,8 @@ import javax.net.ssl.SSLParameters;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static com.google.common.base.Preconditions.*;
+
 
 /**
  * ATTENTION: Using this class MUST be reading from INDIS instead of Zookeeper. ZK read will crash in October 2025.
@@ -244,8 +246,7 @@ public class D2ClientBuilder
     );
 
     final LoadBalancerWithFacilitiesFactory loadBalancerFactory = (_config.lbWithFacilitiesFactory == null) ?
-      new ZKFSLoadBalancerWithFacilitiesFactory() :
-      _config.lbWithFacilitiesFactory;
+      new ZKFSLoadBalancerWithFacilitiesFactory() : _config.lbWithFacilitiesFactory;
 
     // log error for not using INDIS in raw d2 client
     if (_config.isLiRawD2Client && !loadBalancerFactory.isIndisOnly())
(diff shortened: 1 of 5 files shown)


d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -199,6 +200,7 @@ public XdsClientImpl(Node node,
   {
     _readyTimeoutMillis = readyTimeoutMillis;
     _node = node;
+    checkNotNull(managedChannel, "managedChannel");
```
brycezhongqing, line 203: We already has the [check here](https://github.com/linkedin/rest.li/blob/fcebe9261ad14fea6262ead50f28849aa38e23ae/d2/src/main/java/com/linkedin/d2/xds/XdsChannelFactory.java#L117). What do you think we directly to throwException ?

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
```diff
@@ -352,6 +359,7 @@ public D2ClientBuilder setZkHosts(String zkHosts)
 
   public D2ClientBuilder setXdsServer(String xdsServer)
   {
+    checkNotNull(xdsServer, "xdsServer");
```
brycezhongqing, line 362: We don't need this right ? See my other comment we could directly to throw Exception [there](https://github.com/linkedin/rest.li/blob/fcebe9261ad14fea6262ead50f28849aa38e23ae/d2/src/main/java/com/linkedin/d2/xds/XdsChannelFactory.java#L117).

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: I think it's not bad to guard here too, so that for whatever reason the channel is null (maybe due to other failures too), the error will be thrown

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
bohhyang: this is called earlier than when createChannel is called. This can throw as soon as anyone is setting null, instead of wait until XdsChannelFactory actually try to use the null.

brycezhongqing: IIRC, without the nullable annotation, the parameter should already be treated as non-null.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
brycezhongqing: Yeah. For sure.

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
bohhyang: For coding wise, yes, but it won't prevent any user from actually calling it with a null. That's what happened to some espresso-pub users. So we need this explicit null check and throw.

brycezhongqing (approved):

PR #1113, opened 2025-10-15
Update pre-check to prevent duplicate xDS client start.

brycezhongqing:
While the pre-check checks for various issues for indis connection. **The start()** method indeed can be called multiple times for valid reasons, since app owners are allowed to explicitly call d2Client.start and wait for the callback to ensure a d2 client is started before using it. We should change the throw to an info log and just return (no-op) in that case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.80.1] - 2025-10-15
+- Update pre-check to prevent duplicate xDS client start
+
 ## [29.80.0] - 2025-10-10
 - Update methodLevelProperties in ServiceProperties constructor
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.80.1] - 2025-10-15
+- Update pre-check to prevent duplicate xDS client start
```
bohhyang, line 18: "prevent" should be "log", right? since it's ok to call start again.

brycezhongqing: My thought is that returning early on duplicate calls can act as a preventive measure., then it could be a prevent scenario. But, yes, log is more accurate.

ChrisCarini:

PR #1114, opened 2025-10-15
Not to throw for duplicate xDS client start. Throw when setting xds server address to null only for indis-only.

bohhyang:
## Summary
1. Change throw to log for duplicate xds client start. The start() method can be called multiple times for valid reasons, since app owners are allowed to explicitly call d2Client.start and wait for the callback to ensure a d2 client is started before using it. We should change the throw to an info log and just return (no-op) in that case.
2. Only in indis-only read, we should throw when setting xds server address to null. Currently the check is mis-applied to zk read too, zk read is still used in DEV testings with local zk.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.80.1] - 2025-10-15
+- Not to throw for duplicate xDS client start. Throw when setting xds server address to null only for indis-only.  
+
 ## [29.80.0] - 2025-10-10
 - Update methodLevelProperties in ServiceProperties constructor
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ When updating the changelog, remember to be very clear about what behavior has c
 and what APIs have changed, if applicable.
 
 ## [Unreleased]
+- Not to throw for duplicate xDS client start. Throw when setting xds server address to null only for indis-only.
```
brycezhongqing, line 18: Do we missed the version here ? Seems a little weird

bohhyang: yeah, just updated

PR #1115, opened 2025-10-15
Rate limit dark cluster exceptoin logging

jcleezer:
Exceptions can lead to a large amount of logs so use a rate limited logger.

### DIFF
--- a/darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterManagerImpl.java
+++ b/darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterManagerImpl.java
@@ -56,7 +56,7 @@ import org.slf4j.LoggerFactory;
  */
 public class DarkClusterManagerImpl implements DarkClusterManager
 {
-  private static final Logger _rateLimitedPartitionLogger =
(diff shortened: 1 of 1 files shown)


darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterManagerImpl.java
```diff
@@ -177,7 +181,7 @@ private int getPartitionId(RestRequest request)
     }
     catch (RuntimeException | PartitionAccessException | ServiceUnavailableException e)
     {
-      _log.error("Cannot find partition id for request: {}, defaulting to 0", request.getURI(), e);
+      _rateLimitedLogger.error("Cannot find partition id for request: {}, defaulting to 0", request.getURI(), e);
```
shivamgupta1: It's possible that if there are many logs of one of these two types, only one of them will be ever seen. To avoid this, two rate limited logger instances need to be created.

jcleezer: Updated to use two loggers.

PR #1116, opened 2025-10-15
Throw when setting xds server address to null while lbWithFacilitiesFactory is null

bohhyang:
A quick add-on to the [previous change](https://github.com/linkedin/rest.li/pull/1114) to catch a corner case that when setXdsServer is called with null, lbWithFacilitiesFactory may be null (user code will either use the default ZK LB in Dev env, or will call setLoadBalancerWithFacilitiesFactory to use indis LB later)

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
@@ -359,7 +359,7 @@ public class D2ClientBuilder
 
   public D2ClientBuilder setXdsServer(String xdsServer)
   {
-    if (_config.lbWithFacilitiesFactory.isIndisOnly())
+    if (_config.lbWithFacilitiesFactory == null || _config.lbWithFacilitiesFactory.isIndisOnly())
     {
       checkNotNull(xdsServer, "xdsServer");
     }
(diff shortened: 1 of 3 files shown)


d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
```diff
@@ -359,7 +359,7 @@ public D2ClientBuilder setZkHosts(String zkHosts)
 
   public D2ClientBuilder setXdsServer(String xdsServer)
   {
-    if (_config.lbWithFacilitiesFactory.isIndisOnly())
+    if (_config.lbWithFacilitiesFactory == null || _config.lbWithFacilitiesFactory.isIndisOnly())
```
brycezhongqing, line 362: A little confused for this check, if _config.lbWithFacilitiesFactory == null why we still continue to checknull ?
```java
_config.lbWithFacilitiesFactory == null
```

brycezhongqing (approved):

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
bohhyang: It's about the order that the setters are called. Users can do:
```
.setXdsServer(null)
.setLoadBalancerWithFacilitiesFactory(new XdsLoadBalancerWithFacilitiesFactory())
```
Then when setXdsServer is called, the lbWithFacilitiesFactory is still null, but we shall still catch that they are setting null to the xds server.

PR #1120, opened 2025-10-21
Cache response from getPotentialClientsFromCache() method

goktuggurler:
This change introduces the feature of caching the response from getPotentialClients() method. The cache is valid for the given duration. In case the cache is empty or the data is stale, a new map is fetched using the method. 

The reason for caching is to avoid the iterating operation, which is heavy both in terms of computation and memory.


ChrisCarini:

PR #1122, opened 2025-10-23
Fix xds client active wait time metric also add more logs.

bohhyang:
## Summary
Fixes a few issues in xds client metrics so that:
1. when initial data has not been received (for wildcard subscriber, could be haven't received the last batch), reconnections won't reset subscribe time and the subscriber state (still in INIT_PENDING), so that the wait time and server latency metrics can be tracked with the initial subscribe time. 
2. when IRV enabled, after reconnect, receiving no new update since existing data is up-to-date, the active wait time metric is 0. 
3. when IRV is disabled, receiving the same data after reconnection will set the subscribers to FETCHED state (from PENDING_AFTER_RECONNECT), so the wait time metric is 0.

## Testing Done
QEI deploy indis-canary, verified under both IRV enabled and disabled cases, the logs and metrics are as expected.

### DIFF
--- a/d2/src/test/java/com/linkedin/d2/xds/TestXdsClientImpl.java
+++ b/d2/src/test/java/com/linkedin/d2/xds/TestXdsClientImpl.java
@@ -1017,19 +1017,20 @@ public class TestXdsClientImpl {
     fixture.setAllSubscribersSubscribedAt(System.currentTimeMillis() - TEN_MINS);
 
     // active initial wait time contributed from 8 subscribers for:
-    // individual node, uri, uri map, and wildcard node, uri map, d2 cluster/service name, callees, and callees wildcard subscribers
+    // individual node, uri, uri map, callees, and wildcard node, uri map, d2 cluster/service name and callees subscribers
     fixture.verifySubscribersState(INIT_PENDING);
-    verifyWaitTimeFromSubscribers(xdsClient.getActiveInitialWaitTimeMillis(), 6);
+    verifyWaitTimeFromSubscribers(xdsClient.getActiveInitialWaitTimeMillis(), 8);
 
     if (responseAction == ResponseAction.UPDATE) {
       xdsClient.handleResponse(DISCOVERY_RESPONSE_NODE_DATA1);
       xdsClient.handleResponse(RESPONSE_D2URI1);
       xdsClient.handleResponse(RESPONSE_WITH_SERVICE_NAMES);
-      fixture._xdsClientImpl.handleResponse(DISCOVERY_RESPONSE_CALLEES_DATA1);
+      xdsClient.handleResponse(DISCOVERY_RESPONSE_CALLEES_DATA1);
     } else if (responseAction == ResponseAction.REMOVAL) {
       xdsClient.handleResponse(DISCOVERY_RESPONSE_NODE_DATA_WITH_REMOVAL);
       xdsClient.handleResponse(DISCOVERY_RESPONSE_URI1_DATA_WITH_REMOVAL);
       xdsClient.handleResponse(RESPONSE_WITH_NAME_REMOVAL);
+      xdsClient.handleResponse(DISCOVERY_RESPONSE_CALLEES_DATA_WITH_REMOVAL);
     }
     // All subscribers have received first response, so they won't contribute to active initial
     // wait time anymore.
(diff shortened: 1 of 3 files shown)


d2/src/test/java/com/linkedin/d2/xds/TestXdsClientImpl.java
```diff
@@ -1039,58 +1047,99 @@ private enum ResponseAction
...
+  }
 
-    // 3 subscribers are contributing to active initial wait time now
-    activeWaitTime = fixture._xdsClientImpl.getActiveInitialWaitTimeMillis();
-    Assert.assertTrue(activeWaitTime >= Time.minutes(30) && activeWaitTime < Time.minutes(40));
+  private void verifyWaitTimeFromSubscribers(long waitTime, int numSubscribers)
+  {
+    // each subscriber contributes 10 mins wait time
+    Assert.assertTrue(waitTime >= numSubscribers * TEN_MINS && waitTime < (numSubscribers + 1) * TEN_MINS,
+        "Expected wait time from " + numSubscribers + " subscribers, actual wait time: " + waitTime);
```
ZoabKapoor: Super nit: if we're specifying the actual `waitTime` in the error message, should we also specify the expected wait time (numSubscribers * 10 mins) in addition to the `numSubscribers`?

bohhyang: oh, I think the number of subscribers will give the devs simpler view of what's wrong, since the comment/test states clearly that each subscriber contributes 10 mins of wait time. The expected wait time actually will be a very big number and not descriptive then. I can instead include the actual number of subscribers.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -913,22 +916,24 @@ static class ResourceSubscriber
     @Nullable
     private ResourceUpdate _data;
     private final Clock _clock;
-    private long _subscribedAt;
-    private SubscriberFetchState _fetchState = SubscriberFetchState.INIT_PENDING;
+    private AtomicReference<Long> _subscribedAt = new AtomicReference<>(0L);
```
brycezhongqing:
```suggestion
private AtomicLong _subscribedAt = new AtomicLong(0L);
```

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -1053,23 +1077,34 @@ void onRemoval()
...
-      _subscribedAt = _clock.currentTimeMillis();
-      _fetchState = SubscriberFetchState.PENDING_AFTER_RECONNECT;
+      // only change state to PENDING_AFTER_RECONNECT if the previous state was FETCHED.
+      // if the previous state was still init pending, no change to the state and subscribed time
+      if (_fetchState.compareAndSet(FETCHED, SubscriberFetchState.PENDING_AFTER_RECONNECT))
+      {
+        _subscribedAt.set(_clock.currentTimeMillis());
+        _log.info("Reset fetch state to PENDING_AFTER_RECONNECT for {} {} after reconnect.", _type, _resource);
+      }
     }
```
brycezhongqing, line 1116: There is a corner case:

If a reconnect happens while the subscriber is still in INIT_PENDING state (initial data not yet received), the _subscribedAt timestamp is NOT updated.

Scenario:
- Client subscribes at T0 (state = INIT_PENDING)
- Connection drops and reconnects at T1 (state still INIT_PENDING)
- reset() is called but does nothing
- Data finally arrives at T2
Wait time metric will show (T2 - T0) instead of (T2 - T1)

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
```diff
@@ -955,21 +966,30 @@ void addWatcher(ResourceWatcher watcher)
...
 
     @VisibleForTesting
-    void onData(ResourceUpdate data, XdsServerMetricsProvider metricsProvider, boolean isIrvEnabled)
+    void onData(ResourceUpdate data, XdsServerMetricsProvider metricsProvider)
```
brycezhongqing, line 998: CMIIW: State is set to FETCHED immediately, even if the data is null or invalid. If an invalid update arrives first, the state becomes FETCHED, and subsequent valid updates won't be tracked as "initial data". 

Should we only track the valid data ?

bohhyang: This is expected, since the invalid data is sent from INDIS too, which is the initial data. Later on, if the data becomes valid, it's an update.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: this is intended as the right behavior, please see item 1 in the PR description.

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
brycezhongqing: Got it. Thanks for the heads up!

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
brycezhongqing: > reconnections won't reset subscribe time and the subscriber state (still in INIT_PENDING)

Got it. So we have this

d2/src/main/java/com/linkedin/d2/xds/XdsClientImpl.java
bohhyang: ah, yeah, will update, thx.

PR #1123, opened 2025-10-24
Add Indis based downstream fetcher for d2 warmup

shivamgupta1:
Summary
------
Add a new downstream fetcher that can get the list of usual downstreams from Indis.

Testing Done
------
Added comprehensive UTs. Also tested by deploying locally.

PR #1124, opened 2025-10-29
Indis based d2 warmup for restli client

shivamgupta1:
Description
--------
Adds fetching of downstream services to warmup using Indis. Builds on #1123.

Testing done
--------
Deployed locally.

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/util/D2CalleeInfoRecorder.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/util/D2CalleeInfoRecorder.java
@@ -8,10 +8,6 @@ import javax.annotation.Nonnull;
  */
 public interface D2CalleeInfoRecorder {
   /**
-   * Starts the recorder to send periodic updates.
-   */
-  void start();
-  /**
    * Records a callee service name.
    * @param serviceName the callee service name to record
    */
@@ -25,9 +21,4 @@ public interface D2CalleeInfoRecorder {
 
   @Nonnull
   String getScope();
-
-  /**
-   * Stops periodic updates and shuts down the recorder.
-   */
-  void shutdown();
 }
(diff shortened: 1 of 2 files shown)


d2/src/main/java/com/linkedin/d2/balancer/util/D2CalleeInfoRecorder.java
```diff
@@ -0,0 +1,24 @@
+package com.linkedin.d2.balancer.util;
+
+import javax.annotation.Nonnull;
+
+
+/**
+ * Records the callee service information and periodically sends out the recorded information to a persistence medium.
+ */
+public interface D2CalleeInfoRecorder {
```
shivamgupta1, line 9: This interface should've been added here and re-used in grpc infra. However, now I'll add an adapter internally to re-use that one for its implementation.

d2/src/main/java/com/linkedin/d2/balancer/util/WarmUpLoadBalancer.java
```diff
@@ -446,6 +477,10 @@ public TransportClient getClient(Request request, RequestContext requestContext)
     // the call fails, we still *intend* to use serviceName, so it should be in _usedServices.
     String serviceName = LoadBalancerUtil.getServiceNameFromUri(request.getURI());
     _usedServices.add(serviceName);
+    if (_d2CalleeInfoRecorder != null && D2_SCHEME_NAME.equalsIgnoreCase(request.getURI().getScheme()))
```
TylerHorth, line 480: I'm curious, what is the reason for checking the scheme?

TylerHorth (approved):

d2/src/main/java/com/linkedin/d2/balancer/util/WarmUpLoadBalancer.java
shivamgupta1: I'm not sure if this decorator LB is only ever used for d2:// requests. Currently, the recording of warmup service names doesn't happen here. Those are fetched from the d2 data stored on disk, which happens asynchronously via the property buses. I just wanna be sure that we don't record any unneeded information here.

PR #1125, opened 2025-11-04
Release 29.81.0

shivamgupta1:
As title

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-version=29.80.3
+version=29.81.0
 group=com.linkedin.pegasus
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=29.80.3
+version=29.80.4
```
nickibi: Please bump minor version, if there is  new method/interface/classes introduced https://github.com/linkedin/rest.li/pull/1124. LinkedIn dependency check will only fail build for minor version bump not the patch one of Rest.li.

shivamgupta1: Updated, thanks

PR #1128, opened 2025-11-27
[WIP]: Add OpenTelemetry metrics support for ClusterInfo

aadityaraj7769:
### Summary
This PR adds OpenTelemetry (OTel) metrics instrumentation to the ClusterInfoSensor.

### Changes
- **New Interface**: Added `ClusterInfoOtelMetricsProvider` interface for collecting ClusterInfo metrics via OpenTelemetry
  - CanaryDistributionPolicy

- **No-op Implementation**: Added `NoOpClusterInfoOtelMetricsProvider` as default implementation when metrics are disabled

- **Integration**: 
  - Integrated metrics provider into `JmxManager` constructor which is called from container MP.
  - Metrics provider passed to `ClusterInfoJmx` for metrics collection.

- **Metrics Tracked**:
  - Gauges: CanaryDistributionPolicy

### Testing
- Updated test fixtures to mock the new metrics provider

### Backward Compatibility
- Fully backward compatible - defaults to no-op provider when not configured

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/jmx/ClusterInfoJmx.java
+++ b/d2/src/main/java/com/linkedin/d2/jmx/ClusterInfoJmx.java
@@ -18,14 +18,22 @@ package com.linkedin.d2.jmx;
 
 import com.linkedin.d2.balancer.LoadBalancerStateItem;
 import com.linkedin.d2.balancer.simple.ClusterInfoItem;
+import com.linkedin.d2.jmx.ClusterInfoOtelMetricsProvider;
 
 
 public class ClusterInfoJmx implements ClusterInfoJmxMBean
 {
   private final ClusterInfoItem _clusterInfoItem;
+  private final ClusterInfoOtelMetricsProvider _clusterInfoOtelMetricsProvider;
 
-  public ClusterInfoJmx(ClusterInfoItem clusterInfoItem) {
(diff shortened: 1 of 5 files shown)


d2/src/main/java/com/linkedin/d2/jmx/ClusterInfoJmx.java
```diff
@@ -37,11 +45,19 @@ public ClusterInfoItem getClusterInfoItem()
...
-      case CANARY: return 1;
-      default: return -1;
+      case STABLE: 
+        _clusterInfoOtelMetricsProvider.recordCanaryDistributionPolicy(clusterName, 0);
+        return 0;
+      case CANARY: 
+        _clusterInfoOtelMetricsProvider.recordCanaryDistributionPolicy(clusterName, 1);
+        return 1;
+      default: 
+        _clusterInfoOtelMetricsProvider.recordCanaryDistributionPolicy(clusterName, -1);
```
yashness, line 59: This seems wrong. Why would a getter function try to record metrics? getter can't have a side effect of "doing" something.

Also, I'm wondering, why rest.li needs to be aware of otel vs rrd? all rest.li layer should do is "expose" metrics to be recorded & helper methods to read those. container code should use those methods & accordingly record in otel or not based on config flag.

yashness (changes requested):

ChrisCarini:

PR #1132, opened 2026-01-09
Clear DarkCluster Buffer on Refresh

jcleezer:
When a change is detected to the dark cluster configuration the dark cluster strategies are all recreated. When this happens with ConstantQpsStrategy a reference to its rate limiter is maintained and the rate limiter remains active until the buffer is drained. If this happens frequently a large number of dark requests will be maintained in memory exceeding the buffers limit and the QPS will also remain higher than configured until all orphaned buffers are drained.

This PR does two things. First we fix the dangling reference issue. Second when strategies are removed we clear any buffers that they have.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.81.2] - 2026-01-09
+- Clear DarkCluster Buffer on Refresh
+
 ## [29.81.1] - 2025-11-04
 - Fix xds client active wait time metric also add more logs.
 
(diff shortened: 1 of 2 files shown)


darkcluster/src/main/java/com/linkedin/darkcluster/impl/DarkClusterStrategyFactoryImpl.java
```diff
@@ -278,14 +275,32 @@ public void onSuccess(DarkClusterConfigMap updatedDarkConfigMap)
...
+              _partitionToDarkStrategyMap.compute(partitionId, (key, oldDarkStrategyMap) -> {
+                // Shutdown any old strategies that are being replaced.
+                if (oldDarkStrategyMap != null)
+                {
+                  for (Map.Entry<String, DarkClusterStrategy> oldEntry : oldDarkStrategyMap.entrySet())
+                  {
+                    final DarkClusterStrategy oldStrategy = oldEntry.getValue();
+                    if (oldStrategy != null)
+                    {
+                      oldStrategy.shutdown();
```
prashantpogde, line 299: Is this MT-safe for ongoing operations that are still using the old ring buffer ?

jcleezer: If you look at the "old way" (line 288) we used to just completely overwrite the map and assume the old map/buffers would get garbage collected. Turns out that didnt always happen hence the duplicate buffers. This is explicitly clearing all buffers before overwriting the map.

PR #1133, opened 2026-01-14
Add methods to clear our EvictingCircularBuffer

jcleezer:
We want to be able to clear out EvictingCircularBuffer when new dark cluster strategies are made. This PR adds a clear method.

Added a unit test to verify.


shivamgupta1 (approved):
LGTM

PR #1135, opened 2026-01-23
Support consistent hashing at cluster level

hieusydo:
Copy `D2RelativeStrategyProperties` from D2Service to D2Cluster. Reusing this schema will maintain backward compatability, and the move is to align with long-term vision that load balancing strategy should be configured at the cluster level for all gRPC clients. 

With this change, users will be able to configure Consistent Hashing strategy similar like before, but at the cluster level:
```
<entry key="relativeStrategyProperties">
  <map>
    <entry key="ringProperties">
      <map>
        <entry key="consistentHashAlgorithm" value="pointBased"/>
      </map>
    </entry>
  </map>
</entry>
```

### DIFF
--- a/d2-schemas/src/main/pegasus/com/linkedin/d2/D2Cluster.pdl
+++ b/d2-schemas/src/main/pegasus/com/linkedin/d2/D2Cluster.pdl
@@ -54,4 +54,9 @@ record D2Cluster includes D2ChangeTimeStamps {
   * Options for configuring the outlier detection. Only used by gRPC clients.
   */
   outlierDetectionConfig: optional OutlierDetectionConfig
+
+  /**
+   * Properties used for the relative load balancing strategy. Only used by gRPC clients.
+   */
+  relativeStrategyProperties: optional D2RelativeStrategyProperties
 }


d2-schemas/src/main/pegasus/com/linkedin/d2/D2Cluster.pdl
```diff
@@ -54,4 +54,9 @@ record D2Cluster includes D2ChangeTimeStamps {
   * Options for configuring the outlier detection. Only used by gRPC clients.
   */
   outlierDetectionConfig: optional OutlierDetectionConfig
-}
\ No newline at end of file
+
+  /**
+   * Properties used for the relative load balancing strategy. Only used by gRPC clients.
+   */
+  relativeStrategyProperties: optional D2RelativeStrategyProperties
```
TylerHorth, line 61: I think we have the opportunity to clean things up here. Most of the config in D2RelativeStrategyProperties is irrelevant to both gRPC java and Envoy. We can define a new message type here (possibly inspired by the standard xDS configuration) with just the relevant fields. We should also document that this applies only to non-Java gRPC clients for now, since gRPC Java won't read these properties.

PR #1136, opened 2026-01-27
Support load balancing policy at cluster level and RING_HASH config

hieusydo:
## Change Summary

1. Support configuring load balancing policy at cluster level
2. Add PDL schemas for lb configs ported from Envoy proto
    -  [config.cluster.v3.Cluster.RingHashLbConfig](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#config-cluster-v3-cluster-ringhashlbconfig)
    -  [RoundRobinLbConfig](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#config-cluster-v3-cluster-roundrobinlbconfig)
    - [LeastRequestLbConfig](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#config-cluster-v3-cluster-leastrequestlbconfig)

## Testing
`export JAVA_HOME="/Library/Java/JavaVirtualMachines/jdk1.8.0_282-msft.jdk/Contents/Home" && export PATH="$JAVA_HOME/bin:$PATH" && ./gradlew compileTestJava`

### DIFF
--- a/d2-schemas/src/main/pegasus/com/linkedin/d2/ClientType.pdl
+++ b/d2-schemas/src/main/pegasus/com/linkedin/d2/ClientType.pdl
@@ -5,11 +5,6 @@ namespace com.linkedin.d2
  */
 enum ClientType {
   /**
-   * All client types
-   */
-  ALL
-
-  /**
    * D2 client
    */
   D2
--- a/d2-schemas/src/main/pegasus/com/linkedin/d2/LoadBalancingPolicy.pdl
+++ b/d2-schemas/src/main/pegasus/com/linkedin/d2/LoadBalancingPolicy.pdl
@@ -2,18 +2,18 @@ namespace com.linkedin.d2
 
 /**
  * A load balancing policy configuration with associated client types.
- * The policy must have an explicit list of client types that can support such policy.
- * This ensures that the policy is only used by the supported client types.
+ * This provides a mechanism for introducing new load balancing policies
+ * that are not yet supported by all clients.
  */
 record LoadBalancingPolicy {
-
-  /**
-   * The client types that should use this load balancing policy.
-   */
-  clientTypes: array[ClientType]
 
   /**
    * The load balancing policy configuration.
    */
   config: LoadBalancingPolicyConfig
+
+  /**
+   * The client types that should use this load balancing policy.
+   */
+  clientTypes: optional array[ClientType]
 }


d2-schemas/src/main/pegasus/com/linkedin/d2/LoadBalancingPolicyConfig.pdl
```diff
@@ -0,0 +1,8 @@
+namespace com.linkedin.d2
+
+/**
+ * Union of all supported load balancing policy configurations.
+ * This enforces type safety while maintaining extensibility.
+ * New load balancing policies can be added by extending this union.
+ */
+typeref LoadBalancingPolicyConfig = union[RingHashLbConfig]
```
harrisonlhl123: Does this mean for now, there's only 1 policy in the array (RingHashLbConfig) and in the future we can add more policies?

gu0keno0: Should we at least have two? RoundRobin and RingHash?

Envoy Weighte RR has more features we can consider to use in the future https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/load_balancing_policies/client_side_weighted_round_robin/v3/client_side_weighted_round_robin.proto#envoy-v3-api-msg-extensions-load-balancing-policies-client-side-weighted-round-robin-v3-clientsideweightedroundrobin

gu0keno0: Also, least loaded as well.

It adds to the scope of this PR, yet I think probably easier for us to bundle together all the necessary supports for the near future in a single PR.

gu0keno0: better to define this as a record that contains a union of a list of possible lb configs. Then in the record, we can have metadata such as "name", supported dataplanes (enum), etc

This way all of our data planes can easily consume the array.

hieusydo: yes that's the idea

d2-schemas/src/main/pegasus/com/linkedin/d2/D2Cluster.pdl
```diff
@@ -54,4 +54,14 @@ record D2Cluster includes D2ChangeTimeStamps {
...
+
+  /**
+   * The load balancing policy to use for this cluster. Only used by gRPC clients.
+   * Each client will iterate over the list in order and stop at the first policy that it
+   * supports. This provides a mechanism for starting to use new LB policies that are not yet
+   * supported by all clients.
+   * This is inspired by Envoy's load balancing policy configuration:
+   * https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#config-cluster-v3-loadbalancingpolicy
+   */
+  loadBalancingPolicies: optional array[LoadBalancingPolicyConfig]
```
gu0keno0: maybe make this a map:

map[ClientType]LoadBalancingPolicies

And make LoadBalancingPolicies a record {policies: array[LoadBalancingPolicyConfig]}

This way, different ClientType do not interfere with each other.

hieusydo: Hm IMHO the map will introduce complexities to the service owners. This schema dictates how service owners define their lps-d2 configs and they might not be aware of all client types. They will likely end up with dup configs for all client types
```
{
    GRPC_JAVA: { loadBalancingPolicies: [LEAST_REQUEST, ROUND_ROBIN, RING_HASH] } ,
    GRPC_RUST: { loadBalancingPolicies: [LEAST_REQUEST, ROUND_ROBIN, RING_HASH] } ,
    GRPC_ENVOY: { loadBalancingPolicies: [ROUND_ROBIN, RING_HASH] }
...
}
```
Moreoever, if we happen to introduce a new client type, we need to migrate all service configs.

On the other hand, we can still keep the ClientType separate with a simple array `loadBalancingPolicies: [LEAST_REQUEST, ROUND_ROBIN, RING_HASH]`.Service owners don't have to think about client types. Each client can still iterate over the array to pick the one it supports.

gu0keno0: How about

```
record LbPolicy {
client_types: Array[ClientType],
configs: LbConfig
}
```

if client_types is empty, it means it applies to all client types

LbConfig can be the Union

gu0keno0: Make client_types required, and support wildcard option like "ALL", so that it accomodates to both today's situation and tomorrow's ideal situation.

gu0keno0 (approved):

PR #1137, opened 2026-02-09
Add getClusterName method to D2Client

zhihaodev:
Expose getClusterName(URI, Callback<String>) to resolve D2 service URIs to their cluster names via ServiceProperties.

The implementation is similar to `getMetadata`: https://github.com/linkedin/rest.li/blob/80ef7965f7421ea264e189f39d53cbb848c717c1/d2/src/main/java/com/linkedin/d2/balancer/clients/DynamicClient.java#L170

Bumping minor version as it introduces a new public API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,12 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
-## [29.85.0] - 2026-02-07
-- Add getClusterName method to D2Client
+## [29.85.0] - 2026-02-11
+Add getClusterName method to D2Client
+
+## [29.84.1] - 2026-02-10
+- Add missing ClusterProprties fields load balancing policies and outlier detection.
+- Use aliased union for load balancing policy config.
 
 ## [29.84.0] - 2026-02-03
 - Add load balancing policies to D2Cluster and LB config for RingHash, LeastRequest, and RoundRobin.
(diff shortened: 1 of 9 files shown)


Jnry (approved):

d2/src/main/java/com/linkedin/d2/balancer/clients/DynamicClient.java
```diff
@@ -197,6 +197,37 @@ public void onSuccess(ServiceProperties serviceProperties)
...
+      return;
+    }
+    String serviceName = LoadBalancerUtil.getServiceNameFromUri(uri);
+    _balancer.getLoadBalancedServiceProperties(serviceName, new Callback<ServiceProperties>()
+    {
+      @Override
+      public void onError(Throwable e)
+      {
+        error(_log, e);
+        callback.onSuccess(null);
```
TylerHorth: Doesn't callback have an onError method?

zhihaodev: I was following the existing pattern in `getMetadata` https://github.com/linkedin/rest.li/blob/master/d2/src/main/java/com/linkedin/d2/balancer/clients/DynamicClient.java#L181 :
```
      @Override
      {
        error(_log, e);
        callback.onSuccess(Collections.emptyMap());
      }
```
I can go with either way.

TylerHorth: Let's propagate the error. Who knows why it was done that weird way in the other method. IMO as a user of this API, it would be better if it either returned the cluster name, or an exception. What can I do with "null"? Maybe we should also change the other null cases to call onError with a NPE.

zhihaodev: Sure, let me update this.

PR #1140, opened 2026-02-11
adding logs to detect ZK use.

abhics1002:
adding logs to detect ZK use

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
@@ -273,7 +273,7 @@ public class D2ClientBuilder
     // Adding logs to detect if any non-INDIS load balancer factory is being used.
     if (!loadBalancerFactory.isIndisOnly())
     {
-      LOG.error("[ACTION REQUIRED] Zookeeper-based D2 Client "
+      LOG.warn("[ACTION REQUIRED] Zookeeper-based D2 Client "
           + "is deprecated (unless talking to a locally-deployed ZK, or for testing EI ZK) and must be migrated to INDIS. "
           + "See instructions at go/onboardindis , application identity: ", D2Utils.getAppIdentityName());
     }


d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
```diff
@@ -269,6 +269,16 @@ public D2Client build()
           + "Using in stack: {}", stackTrace);
     }
 
+    // Adding logs to detect if any non-INDIS load balancer factory is being used.
+    if (!loadBalancerFactory.isIndisOnly())
```
singhsanjay12, line 274: should we log the application name here? to easily identify the callers for ZK

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
```diff
@@ -269,6 +269,16 @@ public D2Client build()
...
+    {
+      LOG.error("[ACTION REQUIRED] Zookeeper-based D2 Client "
+          + "is deprecated (unless talking to a locally-deployed ZK, or for testing EI ZK) and must be migrated to INDIS. "
+          + "See instructions at go/onboardindis.\n"
+          + "Failing to do so will block other apps from stopping ZK announcements and will be escalated for site-up "
+          + "stability.");
```
singhsanjay12: minor : we can remove this line, seems redundant, since we already said "deprecated" earlier.

abhics1002: updated.

d2/src/main/java/com/linkedin/d2/balancer/D2ClientBuilder.java
```diff
@@ -269,6 +270,14 @@ public D2Client build()
           + "Using in stack: {}", stackTrace);
     }
 
+    // Adding logs to detect if any non-INDIS load balancer factory is being used.
+    if (!loadBalancerFactory.isIndisOnly())
+    {
+      LOG.error("[ACTION REQUIRED] Zookeeper-based D2 Client "
+          + "is deprecated (unless talking to a locally-deployed ZK, or for testing EI ZK) and must be migrated to INDIS. "
+          + "See instructions at go/onboardindis , application identity: ", D2Utils.getAppIdentityName());
```
shivamgupta1, line 278: Is it possible to log this in some internal library? We usually don't add go links in OSS code. We also don't want this log for the still remaining external users of Restli.

abhics1002: there are some users within Li who uses this without container or offspring, so we would not know which are those?

shivamgupta1: Could we change this to a warn log?

abhics1002: updated.

PR #1144, opened 2026-02-18
Stable Constant Qps

jcleezer:
1. Problem Statement
When setting a target QPS for dark cluster forking, the observed dispatch rate does not match the configured target. The error was be as high as +33% or -25% depending on the target QPS value.

2. Test Setup
Source cluster: IRPS-irps-feedstorage-test17-0 (1 pod)
Dark cluster: IRPS-irps-feedstorage-test9-0 (1 pod)
Incoming traffic: ~350-380 req/s (consistent across all tests)
Buffer: size=2000, TTL=10 seconds for ConstantQpsRateLimiter
Cluster size ratio: 1:1

3. Root Cause Analysis
3.1 How the Rate Limiter Works
The dispatch chain is:

IrpRcService.trafficRecord()
  → ConstantQpsForkingStrategy.handleRequest()     [~350/s incoming]
    → ConstantQPSDarkClusterStrategy.handleUnaryRequest()
      → rateLimiter.submit(callback)                [adds to circular buffer]
        → EventLoop dispatches from buffer at rate   [observed QPS]
          → IrpBaseDarkClusterDispatcher.unaryCall() [actual dark cluster call]
3.2 Key Finding: Buffer Replays Requests
The EvictingCircularBuffer.get() does not remove items from the buffer. Items stay until they expire (TTL=10s) or are overwritten by newer items. The rate limiter's event loop circles through the buffer, re-dispatching the same requests to maintain the target rate even when incoming traffic is lower than the target.
From the SI source code comment:
> "should only be used in cases where the user demands a constant rate of callback execution, and it's not important that all callbacks are executed, or executed only once."
This is by design for dark cluster testing.

4. Proposed Fix for SI Library
4.1 Requirement
The rate limiter must dispatch at rates that are not constrained to integer-ms periods. For a target of 750 QPS, the ideal period is 1.333ms -- the fix must achieve sub-millisecond timing precision without changing the SI library's public API.
5.2 Algorithm: Fractional Permit Accumulation
Instead of refilling a fixed burst of permits every Math.round(period) ms, accumulate fractional permits each millisecond based on the exact rate.
The key insight: run the event loop at a fixed 1ms tick, but track permits as a double and accumulate targetQps / 1000.0 permits per tick.

permitsPerMs = targetQps / 1000.0
Every 1ms tick:
  permitBalance += permitsPerMs
  while permitBalance >= 1.0:
    dispatch one request
    permitBalance -= 1.0

### DIFF
--- a/r2-core/src/main/java/com/linkedin/r2/transport/http/client/SmoothRateLimiter.java
+++ b/r2-core/src/main/java/com/linkedin/r2/transport/http/client/SmoothRateLimiter.java
@@ -211,23 +211,41 @@ public class SmoothRateLimiter implements AsyncRateLimiter
   }
 
   /**
-   * A event loop implementation that dispatches and executes {@link Callback}s from the queue based
-   * on available permits. If permits are exhausted, the event loop will reschedule itself to run
-   * at the next permit issuance time. If the callback queue is exhausted, the event loop will exit
-   * and need to be restarted externally.
+   * An event loop that dispatches {@link Callback}s from the pending queue at the configured rate.
    *
-   * If there are more tasks than the max in the buffer, they'll be immediately executed to align with the limit
-   * <p>
-   * Event loop is meant to be run in a single-threaded setting.
+   * <p>Permits are refreshed every {@link Rate#getPeriodRaw()} milliseconds using fractional
+   * arithmetic so that sub-millisecond periods accumulate correctly. For example, at 750 QPS
+   * with burst=1 the internal period is 1.333ms. The period boundary advances as
+   * 0 -> 1.333 -> 2.666 -> 4.0, producing exactly 3 dispatches per 4ms (= 750/s).</p>
+   *
+   * <p>If permits are exhausted the loop reschedules itself for the next period boundary.
+   * If the callback queue is drained the loop exits and is restarted by the next
+   * {@link #submit(Callback)} call. If the buffer exceeds the configured max, at least one
+   * dispatch is forced to prevent a leak.</p>
+   *
+   * <p>This class is designed to run on a single-threaded {@link ScheduledExecutorService}
+   * and requires no synchronization.</p>
    */
   private class EventLoop
   {
     private final Clock _clock;
 
-    private long _permitTime;
-    private int _permitAvailableCount;
-    private int _permitsInTimeFrame;
+    /**
+     * Fractional period boundary. Advanced by {@link Rate#getPeriodRaw()} on each refresh
+     * so that sub-millisecond remainders carry over instead of being rounded away.
+     */
+    private double _permitTime;
+
+    /** Number of permits available for dispatch in the current period (fractional). */
+    private double _permitAvailableCount;
+
+    /** Total permits issued when the current period started (used to track consumption on rate change). */
+    private double _permitsInTimeFrame;
+
+    /** Absolute time of the next already-scheduled loop tick (prevents duplicate scheduling). */
     private long _nextScheduled;
+
+    /** Earliest wall-clock time at which a dispatch is allowed (supports execution delay). */
     private long _delayUntil;
 
     EventLoop(Clock clock)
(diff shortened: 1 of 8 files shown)


r2-core/src/main/java/com/linkedin/r2/transport/http/client/SmoothRateLimiter.java
```diff
@@ -211,60 +211,85 @@ public int getPendingTasksCount(){
...
+   *
+   * <p>If permits are exhausted the loop reschedules itself for the next period boundary.
+   * If the callback queue is drained the loop exits and is restarted by the next
+   * {@link #submit(Callback)} call. If the buffer exceeds the configured max, at least one
+   * dispatch is forced to prevent a leak.</p>
+   *
+   * <p>This class is designed to run on a single-threaded {@link ScheduledExecutorService}
+   * and requires no synchronization.</p>
    */
   private class EventLoop
```
shivamgupta1, line 279: It would be best to avoid modifying preexisting code and instead create a new class with the different behavior. It would be even better if we can add a config flag to control ramping.

r2-core/src/main/java/com/linkedin/r2/transport/http/client/SmoothRateLimiter.java
```diff
@@ -211,60 +211,85 @@ public int getPendingTasksCount(){
...
-   * A event loop implementation that dispatches and executes {@link Callback}s from the queue based
-   * on available permits. If permits are exhausted, the event loop will reschedule itself to run
-   * at the next permit issuance time. If the callback queue is exhausted, the event loop will exit
-   * and need to be restarted externally.
+   * An event loop that dispatches {@link Callback}s from the pending queue at the configured rate.
    *
-   * If there are more tasks than the max in the buffer, they'll be immediately executed to align with the limit
-   * <p>
-   * Event loop is meant to be run in a single-threaded setting.
+   * <p>Permits are refreshed every {@link Rate#getPeriodRaw()} milliseconds using fractional
```
shivamgupta1: Could you comment on the distributed behavior of this new algorithm? Consider that there are multiple client hosts – for e.g. target qps is 1, and there are 4 hosts on the client side, then we're gonna see 4 calls arrive simultaneously at the end of a 4 second period? Is this behavior same as the current algorithm?

r2-core/src/main/java/com/linkedin/r2/transport/http/client/SmoothRateLimiter.java
```diff
@@ -101,6 +121,28 @@ public SmoothRateLimiter(ScheduledExecutorService scheduler, Executor executor,
...
+   * @param scheduler              Scheduler used to execute the internal non-blocking event loop. MUST be single-threaded
+   * @param executor               Executes the tasks for invoking #onSuccess and #onError (only during #callAll)
+   * @param clock                  Clock implementation that supports getting the current time accurate to milliseconds
+   * @param pendingCallbacks       THREAD SAFE and NON-BLOCKING implementation of callback queue
+   * @param bufferOverflowMode     What to do if the max buffer is reached
+   * @param rateLimiterName        Name assigned for logging purposes
+   * @param executionTracker       Adjusts the behavior of the rate limiter based on policies/state of RateLimiterExecutionTracker
+   * @param enablePrecisePeriodTracking   When {@code true}, uses double-precision period and permit tracking
+   *                               to eliminate millisecond quantization errors. When {@code false}, uses
+   *                               integer-rounded period and permit values. Defaults to {@code false}.
```
TylerHorth, line 141: If the issue is just integer rounding of milliseconds, wouldn't the simplest solution be to change the time source from milliseconds to nanoseconds (and maybe change integers to longs)?

r2-core/src/main/java/com/linkedin/r2/transport/http/client/SmoothRateLimiter.java
```diff
@@ -211,60 +254,98 @@ public int getPendingTasksCount(){
...
+      // Refresh permits when the current period has elapsed.
+      if (now - _permitTime >= period)
       {
-        _permitTime = now;
-        _permitAvailableCount = rate.getEvents();
-        _permitsInTimeFrame = rate.getEvents();
+        if (_enablePrecisePeriodTracking)
+        {
+          // Advance by exact fractional period so sub-millisecond remainders accumulate.
+          _permitTime += period;
```
TylerHorth, line 340: How large is the period? Could this cause drift?

r2-core/src/main/java/com/linkedin/r2/transport/http/client/SmoothRateLimiter.java
```diff
@@ -273,10 +354,17 @@ public void loop()
...
       {
-        // We prefer running above the limit then risking a leak
-        _permitAvailableCount++;
+        if (_enablePrecisePeriodTracking)
+        {
+          _permitAvailableCount = Math.max(_permitAvailableCount, 1.0);
+        }
+        else
+        {
+          _permitAvailableCount++;
```
TylerHorth, line 366: Why is precise branch different to not precise branch?

r2-core/src/main/java/com/linkedin/r2/transport/http/client/SmoothRateLimiter.java
```diff
@@ -319,9 +406,21 @@ public void loop()
...
+            nextRunRelativeTime = _delayUntil - now;
+          }
+          else if (_enablePrecisePeriodTracking)
+          {
+            // Round up so the scheduler never fires before the fractional boundary.
+            nextRunRelativeTime = Math.max(1, (long) Math.ceil(_permitTime + period - now));
+          }
+          else
+          {
+            nextRunRelativeTime = Math.max(0, (long) (_permitTime + period) - now);
```
TylerHorth, line 422: Why is precise branch different?

r2-core/src/main/java/com/linkedin/r2/transport/http/client/SmoothRateLimiter.java
```diff
@@ -211,60 +254,98 @@ public int getPendingTasksCount(){
...
 
     EventLoop(Clock clock)
     {
       _clock = clock;
       _permitTime = _clock.currentTimeMillis();
       Rate rate = _rate;
-      _permitAvailableCount = rate.getEvents();
-      _permitsInTimeFrame = rate.getEvents();
+      _permitAvailableCount = _enablePrecisePeriodTracking ? rate.getEventsRaw() : rate.getEvents();
+      _permitsInTimeFrame = _enablePrecisePeriodTracking ? rate.getEventsRaw() : rate.getEvents();
```
TylerHorth, line 304: What is the difference between the raw and not-raw apis?

ChrisCarini:

PR #1151, opened 2026-02-27
Add potential clients cache to avoid recomputation on every request

TylerHorth:
## Summary

`SimpleLoadBalancer.getPotentialClients()` is a hotspot in production profiles. It runs O(n) loops on every request (n = total hosts in partition) to build a `Map<URI, TrackerClient>` by iterating all URIs, filtering banned hosts, resolving TrackerClients, and applying subsetting. All inputs to this computation change only on state events (host join/leave, config updates), yet the work was repeated on every request.

This PR moves the O(n) computation into the state-change event handlers in `SimpleLoadBalancerState`, precomputing and caching the result in a `ConcurrentHashMap`. The request path becomes a single O(1) map lookup. The feature is gated behind a new config flag `enablePotentialClientsCache` (default: `false`) so it can be rolled out incrementally.

Key changes:
- `SimpleLoadBalancerState` maintains a `_potentialClientsCache` keyed by `(serviceName, scheme, partitionId)`, rebuilt eagerly by subscriber callbacks after URI, service, or cluster property updates
- `SimpleLoadBalancer.getPotentialClients()` checks the cache first, falling back to the existing O(n) logic when the cache is disabled or returns null
- New `enablePotentialClientsCache` flag plumbed through `D2ClientConfig` → `D2ClientBuilder` → `ZKFSTogglingLoadBalancerFactoryImpl` → `SimpleLoadBalancerState`
- Removed expensive `LogUtil.debug()` calls from the hot path (`Logger.isDebugEnabled()` is costly at high call rates)

## Testing Done

- 30 new property-style tests in `PotentialClientsCacheTest` verify equivalence between the cached and uncached code paths across 12 input scenarios varying URI count (1–20), cluster-banned count (0–10), and service-banned count (0–5)
- Lifecycle tests verify cache rebuild on URI changes, ban-list changes, service removal, and multi-service-per-cluster behavior
- All existing d2 tests pass

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerState.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerState.java
@@ -60,7 +60,6 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ScheduledExecutorService;
@@ -153,11 +152,15 @@ public class SimpleLoadBalancerState implements LoadBalancerState, ClientFactory
   private final boolean _enablePotentialClientsCache;
 
   /**
-   * Precomputed map of potential clients for load balancing, keyed by (serviceName, scheme, partitionId).
+   * Precomputed map of potential clients for load balancing.
+   * Structure: serviceName -> scheme -> partitionId -> Map<URI, TrackerClient>
    * Rebuilt eagerly on state change events (URI/service/cluster property updates).
    * Read by SimpleLoadBalancer on the request path for O(1) client lookups.
+   *
+   * On rebuild, the entire inner map for a service is replaced atomically, which
+   * naturally removes stale entries for partitions that no longer exist.
    */
-  private final ConcurrentMap<PotentialClientsKey, Map<URI, TrackerClient>> _potentialClientsCache;
+  private final ConcurrentMap<String, Map<String, Map<Integer, Map<URI, TrackerClient>>>> _potentialClientsCache;
 
   /*
    * Concurrency considerations:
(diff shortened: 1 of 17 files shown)


d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerState.java
```diff
@@ -147,6 +149,18 @@ public class SimpleLoadBalancerState implements LoadBalancerState, ClientFactory
...
+  /**
+   * Precomputed map of potential clients for load balancing.
+   * Structure: serviceName -> scheme -> partitionId -> Map<URI, TrackerClient>
+   * Rebuilt eagerly on state change events (URI/service/cluster property updates).
+   * Read by SimpleLoadBalancer on the request path for O(1) client lookups.
+   *
+   * On rebuild, the entire inner map for a service is replaced atomically, which
+   * naturally removes stale entries for partitions that no longer exist.
+   */
+  private final ConcurrentMap<String, Map<String, Map<Integer, Map<URI, TrackerClient>>>> _potentialClientsCache;
```
golthitarun, line 163: Looks like outermap is only concurrentMap, would this cause any race conditions on inner maps ?

TylerHorth: The inner maps are immutable, so no risk for race conditions

d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerState.java
```diff
@@ -823,6 +860,224 @@ public TrackerClient getClient(String serviceName, URI uri)
...
+      TrackerClient client = trackerClients.get(uri);
+      if (client != null)
+      {
+        result.put(uri, client);
+      }
+    }
+    return result;
+  }
+
+  private Map<URI, TrackerClient> buildPotentialClientsSubsetting(
```
shivamgupta1, line 1005: Is it possible to leverage `getPotentialClientsSubsetting` for re-use here in some capacity?

TylerHorth: The implementation isn't identical, but yes we could likely refactor to reuse some of previous code. However, then we'd need to do dark cluster testing again. I'd rather:

1. Use current solution
2. Ramp globally via global config
3. Once stable, delete the old code

What do you think?

shivamgupta1 (approved):
LGTM

d2/src/main/java/com/linkedin/d2/balancer/simple/SimpleLoadBalancerState.java
shivamgupta1: Sounds good

PR #1157, opened 2026-03-30
Fix millisecond quantization errors in SmoothRateLimiter with precise period tracking using enablePrecisePeriodTracking

albert3010:
Summary
This PR introduces an opt-in feature (enablePrecisePeriodTracking) to eliminate millisecond quantization errors in SmoothRateLimiter that cause significant rate inaccuracies when the internal period is a non-integer millisecond value.

Problem
The current SmoothRateLimiter implementation uses integer-rounded millisecond periods (Rate.getPeriod() and Rate.getEvents()), which causes quantization errors for certain QPS values. For example:

750 QPS with burst=1 → true period is 1.333ms, but rounds to 1ms → actual rate ~1000 QPS (+33% error)
1500 QPS with burst=2 → true period is 1.333ms, but rounds to 1ms → actual rate ~2000 QPS (+33% error)
1729 QPS with burst=2 → true period is 1.157ms, but rounds to 1ms → actual rate ~2000 QPS (+16% error)
This impacts clients requiring precise rate control, especially at mid-range QPS values where fractional millisecond periods are common.

Solution
Introduced a new boolean flag enablePrecisePeriodTracking (defaults to false for backward compatibility) that enables double-precision period tracking:

Classic mode (enablePrecisePeriodTracking=false, default):

Uses Rate.getPeriod() / Rate.getEvents() (integer-rounded)
Period boundary snaps to wall-clock time on each refresh
Preserves existing behavior for all current clients
Precise mode (enablePrecisePeriodTracking=true, opt-in):

Uses Rate.getPeriodRaw() / Rate.getEventsRaw() (double-precision)
Period boundary advances by exact fractional period, accumulating sub-millisecond remainders
Example: 750 QPS → period boundary advances as 0 → 1.333 → 2.666 → 4.0, producing exactly 3 dispatches per 4ms (= 750/s)
Changes
Core Implementation:

SmoothRateLimiter.java: Added _enablePrecisePeriodTracking field and conditional logic in EventLoop to switch between classic and precise modes
ConstantQpsRateLimiter.java: Added new constructor accepting enablePrecisePeriodTracking flag
Key Implementation Details:

EventLoop fields (_permitTime, _permitAvailableCount, _permitsInTimeFrame) use double to support both modes
Four conditional branches in EventLoop (constructor, updateWithNewRate(), two locations in loop()) switch behavior based on the flag
Precise mode uses Math.ceil() for scheduler delays to prevent firing before fractional boundaries
Testing:

Added testFractionalPeriodAccuracy() to TestSmoothRateLimiter (750 QPS, <10% error)
Added testFractionalPeriodRateAccuracy() to TestConstantQpsRateLimiter (1729 QPS, <5% error)
Added testClassicModeHasQuantizationError() contrast test verifying classic mode exhibits expected >15% error at 1500 QPS
All existing tests pass with default false flag, confirming zero regression

Backward Compatibility
✅ Fully backward compatible

Default behavior unchanged (enablePrecisePeriodTracking=false)
All existing constructors preserved with default false value
All 55 existing tests pass without modification
Downstream modules (e.g., darkcluster) compile and work without changes

### DIFF
--- a/r2-core/src/main/java/com/linkedin/r2/transport/http/client/ConstantQpsRateLimiter.java
+++ b/r2-core/src/main/java/com/linkedin/r2/transport/http/client/ConstantQpsRateLimiter.java
@@ -38,11 +38,26 @@ public class ConstantQpsRateLimiter extends SmoothRateLimiter
 {
   private final EvictingCircularBuffer _evictingCircularBuffer;
 
-
   public ConstantQpsRateLimiter(
       ScheduledExecutorService scheduler, Executor executor, Clock clock, EvictingCircularBuffer callbackBuffer)
   {
-    super(scheduler, executor, clock, callbackBuffer, BufferOverflowMode.NONE, "ConstantQpsRateLimiter", new UnboundedRateLimiterExecutionTracker());
+    this(scheduler, executor, clock, callbackBuffer, false);
+  }
+
+  /**
+   * Constructs a new instance with an option to enable precise period tracking.
+   *
+   * @param enablePrecisePeriodTracking When {@code true}, uses double-precision period and permit
+   *                             tracking to eliminate millisecond quantization errors at rates where the
+   *                             internal period is non-integer (e.g., 750 QPS &rarr; 1.333ms period).
+   *                             When {@code false} (default), uses integer-rounded period and permit values.
+   */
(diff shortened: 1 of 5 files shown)


jcleezer (approved):
Looks good

PR #1161, opened 2026-04-08
Fix SmoothRateLimiter precision loss from millisecond rounding

TylerHorth:
## Summary

- `Rate.getPeriod()` rounds to the nearest millisecond via `Math.round()`, causing significant QPS error for rates with fractional-ms periods (e.g. 750 QPS with burst=1 has a 1.333ms period that rounds to 1ms → +33% error)
- Switches `SmoothRateLimiter` to nanosecond time tracking via new additive APIs: `Clock.currentTimeNanos()`, `Rate.getPeriodNanos()`, `Rate.getEventIntervalMillis()/getEventIntervalNanos()`, and `RateLimiterExecutionTracker.getNextExecutionDelayNanos()`
- No existing APIs were changed — all additions are backward-compatible with defaults that delegate to existing methods

## Testing Done

- Added `testFractionalPeriodAccuracy` in `TestSmoothRateLimiter`: sets 750 QPS with burst=1 (1.333ms period), runs for 3s, asserts <5% error
- Verified the test **fails on old code**: dispatches 3000 in 3s instead of 2250 (33.3% error from rounding 1.333ms → 1ms)
- Verified the test **passes on new code**: dispatches ~2250 in 3s (<5% error using 1,333,333ns period)
- All existing rate limiter tests pass (`TestSmoothRateLimiter`, `TestConstantQpsRateLimiter`, `TestRampUpRateLimiter`, `BaseTestSmoothRateLimiter`)

### DIFF
--- a/pegasus-common/src/main/java/com/linkedin/util/clock/Clock.java
+++ b/pegasus-common/src/main/java/com/linkedin/util/clock/Clock.java
@@ -37,6 +37,6 @@ public interface Clock
    */
   default long currentTimeNanos()
   {
-    return System.nanoTime();
+    return currentTimeMillis() * 1_000_000L;
   }
 }
--- a/pegasus-common/src/main/java/com/linkedin/util/clock/Clock.java
+++ b/pegasus-common/src/main/java/com/linkedin/util/clock/Clock.java
@@ -31,4 +31,12 @@ public interface Clock
    * @return the current time of this clock in milliseconds.
    */
   long currentTimeMillis();
+
+  /**
+   * @return the current time of this clock in nanoseconds.
+   */
+  default long currentTimeNanos()
+  {
+    return System.nanoTime();
+  }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 
+## [29.85.7] - 2026-04-08
+- Fix SmoothRateLimiter precision loss by switching from millisecond to nanosecond time tracking, eliminating ~33% QPS error for rates with fractional-ms periods.
+
 ## [29.85.6] - 2026-03-06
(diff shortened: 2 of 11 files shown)


pegasus-common/src/main/java/com/linkedin/util/clock/Clock.java
```diff
@@ -31,4 +31,12 @@ public interface Clock
    * @return the current time of this clock in milliseconds.
    */
   long currentTimeMillis();
+
+  /**
+   * @return the current time of this clock in nanoseconds.
+   */
+  default long currentTimeNanos()
```
shivamgupta1, line 38: Could we rename it or add docs to indicate that this is monotonic nano time (and not wall clock time)?

TylerHorth: The interface makes no claims about the time source. It just says it returns the "time of this clock". I think we can leave the details unspecified / implementation defined.

shivamgupta1: That's okay too, but let me explain my motivation. If the purpose of adding this method is measuring precise time intervals, then it's good for the method to indicate that the time is monotonic (and thus useful for that purpose). It's also good for the method to make it clear that this isn't necessarily wall clock time.

shivamgupta1 (approved):

pegasus-common/src/main/java/com/linkedin/util/clock/Clock.java
TylerHorth: Makes sense. I started adding that documentation, but ended up going down a rabbit hole about whether System.nanoTime is actually monotonic or not. Apparently [this isn't so clear cut](https://stackoverflow.com/a/8854104). I think if this is important to users they should research the underlying time-source themselves 😂

shivamgupta1: TIL, thanks

PR #1166, opened 2026-04-30
Emit request size bytes metric for XDS discovery requests

singhsanjay12:
## Summary
- Adds `recordRequestSizeBytes(clientName, sizeBytes)` to `XdsClientOtelMetricsProvider` interface
- Wires it through `XdsClientJmx.recordRequestSizeBytes(sizeBytes)`
- Emits the serialized proto size on every `sendDiscoveryRequest` call via `request.getSerializedSize()`
- `NoOpXdsClientOtelMetricsProvider` updated with a no-op impl

## Motivation
Lets operators detect XDS clients whose `initial_resource_versions` maps are approaching gRPC's per-message size limit on reconnect. No performance impact — protobuf caches the serialized size, which gRPC reuses during serialization.

## Test plan
- [ ] Verify `recordRequestSizeBytes` is called in `XdsClientImplTest` on `sendDiscoveryRequest`
- [ ] Implement `recordRequestSizeBytes` in `XdsClientOtelTracker` in the `container` repo with a corresponding entry in `XdsClientSemantics`


abhics1002 (approved):
LGTM

abhics1002:
you need to update version and publish a new version to use this changes.

PR #1169, opened 2026-05-08
Fix cluster subsetting silent failure when peer URIs are IPv6 (SI-52492)

jcleezer:
## Summary                                                                                                                                                                                                                                        
  - `ZKDeterministicSubsettingMetadataProvider` compared the host's FQDN identity against `URI.getHost()`-form peer cluster URIs. When the d2 client materializes peer URIs in IPv6 form (typical with `-Djava.net.preferIPv6Addresses=true`,
  default-on across most LinkedIn services), the FQDN never matches the `[ipv6]` form returned by `URI.getHost()` on JDK 17, so `instanceId = -1` and the subsetter fell back to the full host set. The failure was silent because the no-match     
  branch logged at DEBUG.                                                                                                                                                                                                                           
  - Resolve a `Set<String>` of candidate identities (FQDN, IPv4, bracketed and unbracketed IPv6, scope ids preserved) at construction via an injectable `AddressResolver`, and match peer URIs with `Set.contains` instead of `indexOf`.
  - Bump the no-match log to WARN with a once-per-outage `_warnedIdentityNotFound` guard that re-arms on a successful match so flapping clusters don't spam the log.                                                                                
  - Public 3-arg/4-arg constructors are preserved, so consumers (notably container's `D2ClientFactory`) need no changes.                                                                                                                            
  - Companion fix to grpc-infra PR #5819. Tracking ticket: SI-52492.                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           
  ## Testing Done                                                                                                                                                                                                                                   
  - [x] Local code review completed                                                                                                                                                                                                                 
  - [x] Unit tests added: `testCandidateIdentities_includesFqdnIpv4AndIpv6`, `testCandidateIdentities_dnsFailure_fallsBackToFqdn`, `testCandidateIdentities_keepsIpv6ScopeId` (uses `Inet6Address.getByAddress` for cross-OS portability),
  `testGetSubsettingMetadata_matchesViaIpv6`, `testGetSubsettingMetadata_noMatch_warnsOnceThenReArmsOnRecovery`.                                                                                                                                    
  - [x] Pre-existing `testGetSubsettingMetadata` (FQDN match path) still passes.                                                                                                                                                                    
  - [x] All 6 tests pass via `./gradlew :d2:test --tests com.linkedin.d2.balancer.subsetting.ZKDeterministicSubsettingMetadataProviderTest` (JDK 11, Gradle 6.9.4).                                                                                 
  - [ ] Production verification: post-deploy, expect `RelativeLoadBalancerStrategy.<svc>-https-LoadBalancerStrategy.TotalHostsCount` to drop from full cluster size to ~`_minClusterSubsetSize` for clusters with `_enableClusterSubsetting=true`.

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/subsetting/ZKDeterministicSubsettingMetadataProvider.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/subsetting/ZKDeterministicSubsettingMetadataProvider.java
@@ -140,8 +140,14 @@ public class ZKDeterministicSubsettingMetadataProvider implements DeterministicS
           }
           if (addr instanceof Inet6Address)
           {
-            identities.add("[" + addrStr + "]");
-            identities.add(addrStr);
+            // Funnel both forms through the same canonicalizer the peer-side resolver uses
+            // (see canonicalizeIpv6IfApplicable below). For the common case this is a no-op
+            // round-trip — Inet6Address.getHostAddress() already returns the canonical
+            // expanded form — but routing both sides through the same function keeps the
+            // local and peer identity strings in lockstep by construction, e.g. for
+            // IPv4-mapped IPv6 (::ffff:a.b.c.d) which auto-converts to a plain IPv4 address.
+            identities.add(canonicalizeIpv6IfApplicable("[" + addrStr + "]"));
+            identities.add(canonicalizeIpv6IfApplicable(addrStr));
           }
           else
           {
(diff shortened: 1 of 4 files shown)


shivamgupta1 (approved):

d2/src/main/java/com/linkedin/d2/balancer/subsetting/ZKDeterministicSubsettingMetadataProvider.java
shivamgupta1, line 258: The DNS resolution might be expensive but it only runs when there's a change to the cluster's version.

PR #1170, opened 2026-05-11
Avoid per-request HashSet allocation in RelativeLoadBalancerStrategy

idarmans:
JFR identified the HashSet copy constructor call in `RelativeLoadBalancerStrategy.getTrackerClient` to be a hotspot since this is executed once per request. The copied set is only consumed by the executor task on cluster changes, not by the steady-state path, such that the construction is wasted most of the time.

Since the map that backs the TrackerClient values are not mutated and have no duplicates, defer the construction of the set to when it's actually necessary.

Bonus: fix flaky test in StateUpdaterTest

### DIFF
--- a/d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/RelativeLoadBalancerStrategy.java
+++ b/d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/RelativeLoadBalancerStrategy.java
@@ -50,12 +50,21 @@ public class RelativeLoadBalancerStrategy implements LoadBalancerStrategy
 
   private final StateUpdater _stateUpdater;
   private final ClientSelector _clientSelector;
+  private final boolean _enableRelativeStrategyDeferredAllocation;
 
   public RelativeLoadBalancerStrategy(StateUpdater stateUpdater,
                                       ClientSelector clientSelector)
   {
+    this(stateUpdater, clientSelector, false);
+  }
+
+  public RelativeLoadBalancerStrategy(StateUpdater stateUpdater,
+                                      ClientSelector clientSelector,
+                                      boolean enableRelativeStrategyDeferredAllocation)
+  {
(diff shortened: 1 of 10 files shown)


d2/src/main/java/com/linkedin/d2/balancer/strategies/relative/RelativeLoadBalancerStrategy.java
```diff
@@ -90,7 +99,9 @@ public TrackerClient getTrackerClient(Request request,
       return null;
     }
 
-    _stateUpdater.updateState(new HashSet<>(trackerClients.values()), partitionId, clusterGenerationId, shouldForceUpdate);
+    _stateUpdater.updateState(
+        _enableRelativeStrategyDeferredAllocation ? trackerClients.values() : new HashSet<>(trackerClients.values()),
+        partitionId, clusterGenerationId, shouldForceUpdate);
```
shivamgupta1, line 104: After @TylerHorth 's recent changes (#1151), it seems that the map of clients is also mutated from an async thread. In general too, may not be a good idea to assume that the map of tracker clients supplied here is immutable.

shivamgupta1: Actually it appears that this map will be immutable. Is it possible to use `ImmutableMap` for making this clear?

idarmans: In the potential client cache, yes it's immutable as well, and it does a value swap when the lb state is updated, such that it's safe for us to do a .values here. Unfortunately we don't have access to any lib to indicate immutability in the interface (guava is not pulled into the classpath for example)

shivamgupta1 (approved):

PR #1171, opened 2026-05-13
Make recordRequestSizeBytes a default no-op for backward compat

singhsanjay12:
## Summary
- Changes `XdsClientOtelMetricsProvider#recordRequestSizeBytes` from an abstract method to a `default` no-op so older implementations (e.g., container <= 38.31.0) keep linking against pegasus >= 29.85.10.
- Adds a CHANGELOG entry under `[Unreleased]`.

## Background
PR #1166 made `recordRequestSizeBytes` abstract, breaking linkage for consumers that resolve container <= 38.31.0 with pegasus >= 29.85.10 in the same version set. Making it a `default` no-op is source- and binary-compatible; real implementors still override.

## Testing Done
- [x] Local code review completed
- [ ] CI build green on this PR

No behavioral test added: the change replaces an abstract declaration with an empty `default` body, so there is no logic to assert. The existing implementation in container that overrides this method (and the `recordRequestSizeBytes` call sites covered by the tests added in #1166) continue to exercise the real recording path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and what APIs have changed, if applicable.
 
 ## [Unreleased]
 - Avoid per-request HashSet allocation in RelativeLoadBalancerStrategy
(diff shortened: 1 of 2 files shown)


abhics1002 (approved):
LGTM

